PR #19, opened 2018-10-14
Integrating rmp and courseinfo

the-rango:


the-rango:
Manually merged from react-almanac into DbMultiCalendar instead. Pull request no longer needed.

PR #28, opened 2019-02-25
Loop the loading GIF

wytesk133:
The loading GIF should be looped especially when loading takes longer than the length of the GIF/video.


the-rango:
Thanks for the great suggestion! I have included it in the latest branch. And when we finalize things in there, this change will be reflected in master.

Also, the 'p' in 'autoplay' needs to capitalized.

PR #34, opened 2019-04-06
export ics file

JustineKoa:
Added Feature:
- Button to click and export .ics file on schedule currently being viewed
- Includes: courses, custom events, finals
Changed:
- Changed the icon for the screenshot feature to a camera because I wanted to use the icon being previously used for the download .ics feature
- Altered some of the other files like Calendar, App, CalendarPaneToolbar (very small additions) to pass a prop to export component... some of the changes look wild but it's because I was playing around to figure out how things worked but in the end it should be relatively how it started.
Note:
- Currently nothing that will handle the special case of Fall's Week 0
- Hard coded in recent first Mondays in a dictionary in order to skip having to calculate first Monday of the quarter based on the final schedule
- I got ics.js from someone on Github to make actually creating the .ics file easier.
- I left more comments in the code to make things more clear
- Tested it the best I could, but let me know if there's errors or better ways to do things (still learning React and how everything connects). If there's a better way to get information for the finals schedule please let me know, I couldn't really figure it out so I worked with what I had which made the logic a little obscure. :)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1971,6 +1971,7 @@
         "anymatch": "2.0.0",
         "async-each": "1.0.1",
         "braces": "2.3.2",
+        "fsevents": "1.2.7",
         "glob-parent": "3.1.0",
         "inherits": "2.0.3",
         "is-binary-path": "1.0.1",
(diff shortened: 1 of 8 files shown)


the-rango:
Note to refactor: generate finals schedule in calendarize and store as a state in app.js, instead of generating it when displaying finals and when downloading ics, for the sake of efficiency and code reuse.

the-rango:
This was sooo amazing!! Very great job and really clear documentations! It works like a charm, outstanding job for you especially given that you had to work with the funky data structures of react, for the first time!!

gitletH:
Since the start date is hardcoded, it will fail once the hardcoded dates are exhausted. 

It should not fail silently. You can disable the button when you fail to get the start date so that the users will not click on it.

the-rango:
@gitletH This was intentional. I told Justine to hard code this. A unique advantage that we have as an organization is that we will actually have the manpower to update these hard coded information every quarter. Now for good style, we should do it a better way, but we'll do that once we have more time lol. It's about small goals :)

gitletH:
@the-rango I see.

PR #35, opened 2019-04-06
Prerender when build for production

gitletH:
Prerender the static page so that the rendering time on the client side is greatly reduced.
Need reviews for the correctness of the prerendered page.

Outcome: The page shows up almost instantly. It takes a few seconds before this change.

Before
image

After
image


the-rango:
Thanks fam!! Looks good and much faster!

PR #37, opened 2019-04-07
Hover explanation

emmohac:
Hover over enrollment explanation


the-rango:
Really good stuff! Next I suggest:
1) Apply the same thing to TabularView.js
2) Fixing color of course stats td by passing down classname
3) Adding a delay to triggering the hoverover

the-rango:
Already implemented and committed! I made a few changes to implement the delayed hover-over and everything else mentioned in previous comment. Not merging since already committed.

PR #39, opened 2019-04-12
Go Green

gitletH:
Made various changes to improve the performance(First MeaningPaint) of the main page as an extension to #35. The speed index is finally green :)

What changed:
1. Maintain a local copy of the Google Roboto font that we use since it takes pretty long for clients to fetch it from Google.
2. Defer load as many components as I can find.
3. Use `react-snap` instead of `react-snapshot` because it is better maintained. And we also hope to use some of its amazing incoming features like [inlineCss](https://github.com/stereobooster/react-snap#inlinecss) once they are released.

image

That's all I can do for now. I don't have much time to worry about this since there're so much to do. What can be done:
1. Redesign the way we update the components. There're a lot of heavy updates going on that lags the app.
1. Redesign the way we fetch static data(those json files). Even though webpack splits them up for us, the main page is still not rendered before those data are parsed and executed.
1. Or we can just be lazy and have an app shell for users to see before we finish rendering the main page.

This PR should not change the functionality/visual of the app. It's only meant to be a performance improvement. Feel free to reject this PR if it breaks anything.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1416,11 +1416,6 @@
       "resolved": "https://registry.npmjs.org/@xtuc/long/-/long-4.2.1.tgz",
       "integrity": "sha512-FZdkNBDqBRHKQ2MEbSC17xnPFOhZxeJ2YGSfr2BKf3sujG49Qe3bB+rGCwQfIaA7WHnGeGkSijX4FuBCdrzW/g=="
     },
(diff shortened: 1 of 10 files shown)


the-rango:
Thanks fam for helping us figure this out! I have added everything from this PR.

PR #40, opened 2019-04-18
Latest

willsunnn:
Added the functionality to copy schedules from one schedule tab to another. 

This is done through a method in App.js handleCopySchedule that gets a list of the classes from the currentScheduleIndex, copies the classes, changes the class index to the new index, concatenates the new classes to the end of the current class list, and then updates the state

The functionality was added in the UI through the TabularView component. It adds a dropdown menu that has multiple options to choose which schedule to copy the current schedule to. 

There is a change to the package.json and package-lock.json files. I don't think the added packages and dependencies are required, and might have been added when testing the use of a different dropdown menu library. It should be safe to not merge those changes, but I haven't tested the code without it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -151,6 +151,49 @@
         }
       }
     },
+    "@react-bootstrap/react-popper": {
+      "version": "1.2.1",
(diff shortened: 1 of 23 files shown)


the-rango:
Looks great!

PR #54, opened 2019-05-07
Add pre-commit hook for Prettier

wytesk133:
This PR adds a [Git Hook](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) that formats the code every time before committing to ensure consistency in coding style over the entire code base.

This PR involves three additional tools installed as dev dependencies:
- [prettier](https://prettier.io/) is a code formatter. [But why?](https://prettier.io/docs/en/why-prettier.html)
- [husky](https://github.com/typicode/husky) allows configuring Git Hooks from the JavaScript side, such as in package.json.
- [lint-staged](https://github.com/okonet/lint-staged) ensures that the formatter only runs against changed files, leaving the rest of the repo untouched.

The following configuration for Prettier is completely from my observation of the existing code. Feel free to let me know if these need to be updated. Read more about the settings [here](https://prettier.io/docs/en/options.html).

After merging this PR, please also merge #55

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1132,6 +1132,15 @@
       "resolved": "https://registry.npmjs.org/@nodelib/fs.stat/-/fs.stat-1.1.3.tgz",
       "integrity": "sha512-shAmDyaQC4H92APFoIaVDHCx5bStIocgvbwQyxPRrbUY20V1EYTbSDchWbuwlMG3V17cprZhA6+78JfB+3DTPw=="
     },
+    "@samverschueren/stream-to-observable": {
+      "version": "0.3.0",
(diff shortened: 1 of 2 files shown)


wytesk133:
You may want to debate this (once and for all) with the team and get their approvals since styles can be subjective.

You may also want to consider the styles adopted by popular repositories:
- https://github.com/facebook/react/blob/master/.prettierrc.js
- https://github.com/mui-org/material-ui/blob/next/prettier.config.js
- https://github.com/babel/babel/blob/master/.prettierrc

the-rango:
That sounds like a great idea! Thank you!

PR #71, opened 2019-06-29
Latest_Update

PepziC0les:
Removed CourseCodeBar and CourseNumBar options from AdvancedSearch and moved CourseNum next to DeptSearchBar.


PepziC0les:
Commented and added PropTypes for added classes

PR #81, opened 2019-07-12
Graphs V2.0

Indigo-N-I:
The graphs here now use the new api, key is not yet made but can easily be added

One lingering glitches: 
1. Legend at the bottom will be overlayed on the graph instead of under until one of the elements is clicked and updated
2. Some dates are not displayed on the X axis but are displayed when hovered over

graph glitch
graph glitch 1


ChaseC99:
Closing this, since Graphs have been overhauled in the new AntAlmanac rewrite.

PR #88, opened 2019-07-29
Update Dora with latest code

the-rango:
So that I do not have to wait for npm install each time I switch branches


the-rango:
Made a mistake in conflict resolution. Will redo.

PR #92, opened 2019-09-11
Added indexes to marker when a day is selected

falsidge:
Just a little something to make it easier to keep track of the marker for visually impaired users
This sorts the events list by date, however, so a different implementation might be desired.


the-rango:
This is a great idea! Thanks for the PR!

(Apologies for taking so long to get to this, summer was too busy 😓)

the-rango:
This should now be on AntAlmanac.com.

PR #94, opened 2019-11-10
Bug Fix: Department Search Bar

ChaseC99:
Fixed a glitch where the department could not be cleared in the search form.
Also updated setDept to clear the department search label when set to null.


ChaseC99:
Here's the bug that this fixes:

PR #104, opened 2020-10-10
Feature/refactor

emmohac:
@the-rango please review and approve.


ChaseC99:
Hi @emmohac! 

Thanks so much for taking the time to go through and clean up some of this code. 
Unfortunately, last summer we started a massive rewrite that restructured the entire project. `main` has diverged significantly since then. As a result we will be unable to merge these changes.

PR #142, opened 2021-03-02
Add news component

### LINKED ISSUE #110: News display
I want to add a little box/area where we can put in information about changes, news information, etc. We need to implement a backend endpoint that serves this data to the frontend. We can rapidly and easily change the information displayed by editing the mongoDB with the information we'd like to serve. This would be good for instances where we announce new features or changes or even notices if our services are unavailable (like AANTS) etc.

- [x] Create datatabase location and data model
- [x] Create API endpoint
- [x] Implement frontend

devsdevsdevs:
## Summary
The news section basically lets us put up notifications or messages for our users by simply adding documents to the MongoDB. [An endpoint /api/news has been added to antalmanac-backend ](https://github.com/icssc-projects/antalmanac-backend/commit/f0a59d9246d383cba093a9d13d03a91973d1f466). 

image

## Test Plan
- Error message when endpoint responds with non 200 code
- Skeleton shows up during load
- Dates are parsed properly
- Too many news items triggers scroll
- Editing DB results in instantaneous changes on frontend

## Issues
#110 

## Future Followup
Implement notification dot when the user has not viewed the latest news item. PR should be merged after this is complete.

### DIFF
--- a/client/src/components/App/News.js
+++ b/client/src/components/App/News.js
@@ -39,7 +39,7 @@ class News extends PureComponent {
         anchorEl: null,
         newsItems: null,
         loading: true,
-        showDot: true,
+        showDot: false,
     };
 
     componentDidMount = async () => {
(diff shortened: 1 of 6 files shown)


client/src/components/App/News.js
```diff
@@ -0,0 +1,126 @@
...
+                        </ListItem>
+                        {index !== this.state.newsItems.length - 1 ? <Divider /> : null}
+                    </Fragment>
+                );
+            });
+        } else if (this.state.loading === false && this.state.newsItems === null) {
+            return (
+                <ListItem alignItems="flex-start" className={classes.listItem} dense>
+                    <Typography variant="caption" gutterBottom>
+                        {'Error loading news items'}
```
ChaseC99: Instead of displaying an error message, **we should just say that there are no new announcements.** 
Makes us look bad if we display an error. Users wouldn't think twice about that if it just says there are no if new updates.

client/src/components/App/News.js
```diff
@@ -0,0 +1,126 @@
...
+                            </Typography>
+                            <Typography variant="caption" gutterBottom color="textSecondary">
+                                {moment(newsItem.date).tz('America/Los_Angeles').format('MMMM Do YYYY')}
+                            </Typography>
+                        </ListItem>
+                        {index !== this.state.newsItems.length - 1 ? <Divider /> : null}
+                    </Fragment>
+                );
+            });
+        } else if (this.state.loading === false && this.state.newsItems === null) {
```
ChaseC99: Right now if there are no items in the newsItems array, it displays an empty module. We could handle empty lists the same way as errors (see other comment).
```suggestion
        } else if (this.state.loading === false) {
```

ChaseC99 (changes requested):
Really excited about how easy it will be to create News updates moving forward 😄 📰 

Two small things:
- I left a comment about changing how we handle errors. 
- I think you might need to update the submodule to the latest commit. I had to run `git submodule update --recursive --remote` to get the new endpoint. 

Other than that, this is looking awesome!

client/src/components/App/News.js
```diff
@@ -0,0 +1,126 @@
...
+            this.setState({ newsItems: json.news, loading: false });
+        } catch (e) {
+            console.error('Error loading news items:', e);
+            this.setState({ newsItems: null, loading: false });
+        }
+    };
+
+    getNewsItems = () => {
+        const { classes } = this.props;
+        if (this.state.loading === false && this.state.newsItems !== null) {
```
ChaseC99:
```suggestion
        if (this.state.loading === false && this.state.newsItems !== null && this.state.newsItems.length) {
```
Suggestion to fail the check if there are no items in newItems. We don't want to return an empty module.

devsdevsdevs:
Ready for final review + merge

ChaseC99 (approved):
Looks fantastic!

PR #149, opened 2021-03-19
Material UI Time Picker

### LINKED ISSUE #135: Time pickers in advanced search and custom events modal
Implemented with material-ui pickers

braedons:
## Summary
The previous, text entry time picker has been replaced by a material-ui pop-up. AM and PM options are also now supported. The minute hand is restricted to 60-minute increments, but it is still ignored when selecting the time.

time-picker

## Test Plan
1. Try searching without specifying start/end times to make sure it functions normally first.
2. Enter a start time, check if results all come after specified time.
3. Enter an end time, check that all results come between the two times.
4. Click on the start time, then click clear, search, then make sure start time is no longer restricted, but that end time still is.
5. Clear end time and make sure classes span the entire day.

## Issues
#135


ChaseC99 (changes requested):
The code looks good but I'm not sure if this is the right design direction..... 

Personally I think it's less user-friendly to have to click around on a clock. The current implementation of of the time picker seems easier to use and also allows the user to type in their desired time.
image

I'm not quite sure what the point of #135 is... maybe @devsdevsdevs could elaborate?

Sorry to be so critical about the design after you've already put in the time to implement this :(

devsdevsdevs:
> 
> 
> The code looks good but I'm not sure if this is the right design direction.....
> 
> Personally I think it's less user-friendly to have to click around on a clock. The current implementation of of the time picker seems easier to use and also allows the user to type in their desired time.
> image
> 
> I'm not quite sure what the point of #135 is... maybe @devsdevsdevs could elaborate?
> 
> Sorry to be so critical about the design after you've already put in the time to implement this :(

I wanted this change mainly because this experience is not universal. That nice dropdown is native to Chrome and doesn't exist in Firefox.

braedons:
image

I could swap the current implementation for a "KeyboardTimePicker", which allows for keyboard input, as well as providing a button for the pop-up.

braedons:
I've updated it to allow for keyboard input now.

devsdevsdevs (approved):
Excellent work!

ChaseC99 (changes requested):
Uh oh looks like the backend submodule got accidentally updated in this PR! Please revert that change before landing :)

In the future, I'd recommend **_against_** using `git add .` or `git add -a`. Instead, we should be adding each file individually. It's annoying but it makes sure we don't accidentally include unwanted changes.

Also I'm still not sure about the clock :/ It's great that it works across different browsers, but if we could find a material-ui dropdown picker, like the current site's, that would be ideal. Maybe we should sync with Design to see what they think?

braedons:
Ugh, thanks for catching that. I'll keep that in mind next time.
As for the dropdown, if @devsdevsdevs's cool with it, I'm sure swapping it out would be super easy.

devsdevsdevs:
> 
> 
> Ugh, thanks for catching that. I'll keep that in mind next time.
> As for the dropdown, if @devsdevsdevs's cool with it, I'm sure swapping it out would be super easy.

Yeah, I'm cool with it

braedons:
dropdown_time_picker

Ok, so I've reverted the previous commit to undo the submodule change. And I've switched from the MaterialUI time-picker to a standard dropdown.

Any comments on the design of it? Would breaking up the hour and the am/pm be better? This is how WebSOC has it.

client/src/components/SearchForm/AdvancedSearch.js
```diff
@@ -89,6 +81,61 @@ class AdvancedSearchTextFields extends PureComponent {
...
+                    {hour}:00pm
+                </MenuItem>
+            );
+            endsBeforeMenuItems.push(
+                <MenuItem key={`ends-before-${hour}pm-item`} value={`${hour} + :00pm`}>
+                    {hour}:00pm
+                </MenuItem>
+            );
+        }
+
```
ChaseC99: We can significantly condense this block of code by looping through an array of times and doing some fancy js stuff with `.map`.  
```suggestion
        // List of times from 2:00am-11:00pm
        const menuItemTimes = [
            ...[...Array(10).keys()].map(v => v+2+":00am"), 
            "12:00pm",
            ...[...Array(11).keys()].map(v => v+1+":00pm")
        ]

        // Creates a MenuItem for time selection
        const createdMenuItemTime = (time) => (
            <MenuItem value={`${time}`}>
                {time ? time : <em>None</em>}
            </MenuItem>
        );

        // Build arrays of MenuItem elements for time selection
        const startsAfterMenuItems = ["", "1:00am", ...menuItemTimes].map(time => createdMenuItemTime(time));
        const endsBeforeMenuItems = ["", ...menuItemTimes].map(time => createdMenuItemTime(time));
```

I could be wrong but I don't believe we need a `key` attribute on `MenuItem`.

client/src/components/SearchForm/AdvancedSearch.js
```diff
@@ -39,27 +42,16 @@ class AdvancedSearchTextFields extends PureComponent {
...
-                    );
+            if (event !== '') {
+                this.setState(
+                    {
+                        [name]: event.target.value,
+                    },
+                    () => {
+                        updateFormValue('startTime', this.state.startTime);
+                        updateFormValue('endTime', this.state.endTime);
+                    }
```
ChaseC99: This is much cleaner than the previous code - great work!

client/src/components/CoursePane/CourseRenderPane.js
```diff
@@ -131,8 +131,8 @@ class CourseRenderPane extends PureComponent {
                 sectionCodes: formData.sectionCode,
                 instructorName: formData.instructor,
                 units: formData.units,
-                endTime: formData.endTime,
-                startTime: formData.startTime,
+                endTime: formData.endTime, //? format(formData.endTime, 'hh:00aaa') : formData.endTime,
+                startTime: formData.startTime, //? format(formData.startTime, 'hh:00aaa') : formData.startTime,
```
ChaseC99: Did you mean to include these comments?

ChaseC99 (changes requested):

braedons:
Thanks, Chase, for holding my hand through this ;). I implemented your improvement for the time generation, but I added a key attribute since it _does_ throw an error/warning otherwise. Everything still seems to work after testing.

ChaseC99 (approved):
Great work powering through all of the revisions for this PR!! Code looks solid 👌

PR #150, opened 2021-03-29
Custom events box

### LINKED ISSUE #111: "Add Custom Events" box improvements
- [x] Disable "Add Event" button in modal until boxes have been checked to select both the day(s) and the selected schedule(s)
- [x] Add a tool tip message over disabled "Add Event" button indicating that the days and schedule boxes need to be checked.
- [x] Cancel button should clear the event name

kartik-bhagatwala:
## Summary
Three improvements to the custom events box:
- Disable "Add Event" button in modal until boxes have been checked to select both the day(s) and the selected schedule(s)
 
output

- There is a tool tip message when the "Add Event" button is disabled, indicating that the days and schedule boxes need to be checked.

image
- Cancel button now clears the event name

## Test Plan
1. Button is disabled without anything clicked, tooltip active
2. Button is still disabled with only one of each section (day or schedule) selected
3. Button is enabled when one or more boxes from both sections are selected, tooltip not present on hover
4. Event name is cleared when cancel is pressed
## Issues
#111

### DIFF
--- a/client/src/components/CustomEvents/CustomEventDialog.js
+++ b/client/src/components/CustomEvents/CustomEventDialog.js
@@ -41,7 +41,7 @@ class CustomEventDialog extends PureComponent {
     };
 
     handleOpen = () => {
-        this.setState({ open: true })
+        this.setState({ open: true });
         ReactGA.event({
             category: 'antalmanac-rewrite',
             action: 'Click Custom Event button',
@@ -165,7 +165,12 @@ class CustomEventDialog extends PureComponent {
                             Cancel
                         </Button>
 
-                        <Button onClick={() => this.handleClose(false)} variant="contained" color="primary">
+                        <Button
+                            onClick={() => this.handleClose(false)}
+                            variant="contained"
+                            color="primary"
+                            disabled={!(this.state.scheduleIndices.length && this.state.days.some(Boolean))}
+                        >
                             {this.props.customEvent ? 'Save Changes' : 'Add Event'}
                         </Button>
                     </DialogActions>
--- a/client/src/components/CustomEvents/CustomEventDialog.js
+++ b/client/src/components/CustomEvents/CustomEventDialog.js
@@ -164,15 +164,21 @@ class CustomEventDialog extends PureComponent {
                         <Button onClick={() => this.handleClose(true)} color="primary">
                             Cancel
                         </Button>
-
-                        <Button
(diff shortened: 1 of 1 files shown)


client/src/components/CustomEvents/CustomEventDialog.js
```diff
@@ -164,10 +164,21 @@ class CustomEventDialog extends PureComponent {
                         <Button onClick={() => this.handleClose(true)} color="primary">
                             Cancel
                         </Button>
-
-                        <Button onClick={() => this.handleClose(false)} variant="contained" color="primary">
-                            {this.props.customEvent ? 'Save Changes' : 'Add Event'}
-                        </Button>
+                        <Tooltip
+                            title="Schedule and day must be checked"
+                            disableHoverListener={this.state.scheduleIndices.length && this.state.days.some(Boolean)}
```
kartik-bhagatwala: Was wondering if this expression should be made into a variable, since I use it again in Line 176? And if so, where would that be defined?

ChaseC99: Good hutch! It's always good to try and write [DRY code](https://en.wikipedia.org/wiki/Don't_repeat_yourself) :)

My suggestion would be to add a function to the component and then call that function in the render. Something like:
```
isDisabled = () => {
    return this.state.scheduleIndices.length && this.state.days.some(Boolean)
}
```
```suggestion
                            disableHoverListener={this.isDisabled()}
```

client/src/components/CustomEvents/CustomEventDialog.js
```diff
@@ -54,7 +54,7 @@ class CustomEventDialog extends PureComponent {
             this.handleAddToCalendar();
         }
 
-        this.setState({ open: false });
+        this.setState({ open: false, eventName: '' });
     };
```
devsdevsdevs, line 59:
```suggestion
isDisabled = () => {
    return this.state.scheduleIndices.length && this.state.days.some(Boolean)
}
```

devsdevsdevs (changes requested):
Looks good, just commit Chase's suggestions

kartik-bhagatwala:
Added a function to check if the button should be disabled, as well as fixed a bug where pressing the cancel button would not properly clear state, so button would appear enabled even when nothing was checked. Should be ready to merge now!

ChaseC99 (approved):
Fantastic work!

PR #152, opened 2021-03-31
Fix ics export bug for classes with TBA times

### LINKED ISSUE #151: Download ICS: Unable to download schedule that has a TBA class time
User bug report: "Cannot download calendar as an .ics file when there is an online only class added to the schedule"

ChaseC99:
## Summary
Download ics would fail when a class's time was TBA.
This commit now skips classes with TBA times, since it doesn't make sense to include those classes on a calendar.

## Test Plan
- Add classes to a schedule
- Add a class with a "TBA" time
- Verify that download .ics file works

## Issues
Close #151

### DIFF
--- a/client/src/components/Calendar/ExportCalendar.js
+++ b/client/src/components/Calendar/ExportCalendar.js
@@ -68,6 +68,7 @@ const dateToIcs = (date) => {
 // Note(chase): We currently don't use this function because timezones are too annoying.
 //  If you want to use this function, you also need to set the VEvent's
 //  startInputType/endInputType to 'utc'
+// eslint-disable-next-line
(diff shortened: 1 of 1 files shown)


client/src/components/Calendar/ExportCalendar.js
```diff
@@ -68,6 +68,7 @@ const dateToIcs = (date) => {
 // Note(chase): We currently don't use this function because timezones are too annoying.
 //  If you want to use this function, you also need to set the VEvent's
 //  startInputType/endInputType to 'utc'
+// eslint-disable-next-line
```
ChaseC99, line 71: Not related to change, just wanted to suppress a warning

Issue #147, opened 2021-03-07
Download ICS: refactor to use calendarize helper


ChaseC99:
ExportCalendar uses AppStore's getAddedCourse instead of getEventsInCalendar, because getEventsInCalendar was missing critical information.

- [x] Determine what info was missing from getEventsInCalendar (I forgot already 😅)
- [x] Refactor the calendarize function to include the necessary info
- [x] Refactor ExportCalendar to use getEventsInCalendar

JacE070:
I'm going to get this done by Oct. 26 (3 weeks)

Update: Might be done 1 week earlier. By Oct. 19.

Update: Gotta get it back to the original delivery day : )

JacE070:
### Analysis of Current Functions
AppStore.getEventsInCalendar returns both getCurrentCourses and getCurrentCustomEvents.

What we extract and use from Appstore.getCurrentCourses:
`{
term,
deptCode,
courseNumber,
courseTitle,
section: { sectionType, instructors, meetings, finalExam },
}`

`meetings: array(WebsocSectionMeeting) {
timeIsTBA,
bldg,
days,
startTime,
endTime
}`

What we get from AppStore.calendarizeCourseEvents:
`{
color,
term,
title,
courseTitle
locations,
showLocationInfo,
instructors,
sectionCode,
sectionType,
start,
end,
finalExam,
isCustomEvent
}`

What we get from AppStore.calendarizeCustomEvents:
`{
customEventID,
color,
start,
isCustomEvent,
end,
title
}`

### What's Missing?

#### AppStore.calendarizeCourseEvents
- deptCode
- days

#### AppStore.calendarizeCustomEvents
- days
- bldg (feature is under review)
- term

JacE070:
Where to retrieve the **term** for customEvent? Looks like **term** is not defined in RepeatingCustomEvent, which means we don't know the start and end date.

JacE070:
Then comes a problem: do we want users to export their custom event schedule?

PR #157, opened 2021-05-04
Add finals to the .ics export

### LINKED ISSUE #139: Download ICS: Add Finals to .ics download
Currently, only the class schedule is downloaded. We should include finals and custom events.

This will require a small refactor to `client/src/stores/calenderizeHelpers.js` because 2018 is hardcoded as the year for all events. That should pull the year from the course term.

ChaseC99:
## Summary
Add Finals to the .ics export!
image

## Test Plan
- Add an 'am' final, Add a 'pm' final, Add a final that starts 'am' and ends 'pm'
- Add a course that doesn't have a final (should be skipped)
- Add a course that has a 'TBA' final (should be skipped)
- Add a final from each quarter

## Issues
Closes #139 

## Future Followup
Move functions in ExportCalendar over to the calendarize helper (#147)

### DIFF
--- a/client/src/components/Calendar/ExportCalendar.js
+++ b/client/src/components/Calendar/ExportCalendar.js
@@ -101,14 +101,32 @@ const getFirstClass = (date, time) => {
     ];
 };
 
+// getExamTime returns the start and end datetime of an exam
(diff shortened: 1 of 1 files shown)


devsdevsdevs (approved):
Looks good to me!

Issue #156, opened 2021-05-03
Handling schedules with courses from mulitple terms


devsdevsdevs:
- [x] Add snackbar with warning color when schedule is not homogenous
- [x] Add term label for courses in event tooltip

ChaseC99:
Closed by #159

PR #159, opened 2021-05-08
Added warning when multiple terms detected and added term info to tooltip

braedons:
## Summary
Previously, users could add courses from multiple terms, but after being added, there would be no indication of these terms. This change adds a warning snackbar every time the user adds a course and multiple terms are present.
warning

It also adds a row in the event tooltips indicating the term.
tooltip_term

## Test Plan
* Verify that adding courses from two different terms produces a warning snackbar
* Verify that adding courses from the same term does not produce a warning
* Verify that calendar event tooltips display a term and that it's correct
* Make sure the placement of the term in the tooltip is sensible (i.e. should it be listed second?)
* Make sure that displaying the warning after every class is not excessive (i.e. should it only appear once a new term is added?)

## Issues
#156

## Future Followup (Optional)
* Add term for each course in Added Classes page

### DIFF
--- a/client/src/actions/AppStoreActions.js
+++ b/client/src/actions/AppStoreActions.js
@@ -39,9 +39,12 @@ export const addCourse = (section, courseDetails, term, scheduleIndex, color) =>
     const addedCourses = AppStore.getAddedCourses();
 
     let existingCourse;
+    let multipleTerms = false;
 
     for (const course of addedCourses) {
-        if (course.section.sectionCode === section.sectionCode && term === course.term) {
+        if (term !== course.term) {
+            multipleTerms = true;
+        } else if (course.section.sectionCode === section.sectionCode) {
             existingCourse = course;
             if (course.scheduleIndices.includes(scheduleIndex)) {
                 return;
@@ -50,6 +53,8 @@ export const addCourse = (section, courseDetails, term, scheduleIndex, color) =>
             }
         }
     }
+
+    if (multipleTerms) openSnackbar('warning', 'Course added from different term');
 
     if (color === undefined) {
         const setOfUsedColors = new Set(addedCourses.map((course) => course.color));
--- a/client/src/actions/AppStoreActions.js
+++ b/client/src/actions/AppStoreActions.js
@@ -59,7 +59,7 @@ export const addCourse = (section, courseDetails, term, scheduleIndex, color) =>
             'warning',
             `Course added from different term.\nSchedule now contains courses from ${[...multipleTerms]
                 .sort()
-                .join(', ')}`,
+                .join(', ')}.`,
             null,
             null,
             { whiteSpace: 'pre-line' }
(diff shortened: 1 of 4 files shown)


ChaseC99 (approved):
Looks clean, good work! 👍

devsdevsdevs:
The wording might be a little confusing to users, I was thinking to explain it a bit more like "Course added from term "2021 Fall", other courses in this schedule are from different term(s)" but I think it's a little verbose 🤔

braedons:
> The wording might be a little confusing to users, I was thinking to explain it a bit more like "Course added from term "2021 Fall", other courses in this schedule are from different term(s)" but I think it's a little verbose thinking

Adding the term makes sense. Maybe list all terms now present and say something like, "Schedule now contains courses from 2021 Fall, 2021 Spring, 2021 Winter"?

ChaseC99:
I think both would sound good together!
```
Course added from different term.\n
Schedule now contains courses from 2021 Fall, 2021 Spring, 2021 Winter.
```

client/src/components/App/NotificationSnackbar.js
```diff
@@ -1,11 +1,7 @@
 import React, { PureComponent } from 'react';
-import CheckCircleIcon from '@material-ui/icons/CheckCircle';
-import ErrorIcon from '@material-ui/icons/Error';
-import InfoIcon from '@material-ui/icons/Info';
 import CloseIcon from '@material-ui/icons/Close';
 import { amber, green } from '@material-ui/core/colors';
 import IconButton from '@material-ui/core/IconButton';
-import WarningIcon from '@material-ui/icons/Warning';
```
ChaseC99, line 8: Thank you for removing this! The warnings in the terminal were getting annoying lol

client/src/actions/AppStoreActions.js
```diff
@@ -51,6 +54,17 @@ export const addCourse = (section, courseDetails, term, scheduleIndex, color) =>
         }
     }
 
+    if (multipleTerms.size > 1)
+        openSnackbar(
+            'warning',
+            `Course added from different term.\nSchedule now contains courses from ${[...multipleTerms]
```
ChaseC99, line 60: Nit: can you add a period?
```suggestion
            `Course added from different term.\nSchedule now contains courses from ${[...multipleTerms].
```

client/src/actions/AppStoreActions.js
```diff
@@ -51,6 +54,17 @@ export const addCourse = (section, courseDetails, term, scheduleIndex, color) =>
...
+                .sort()
+                .join(', ')}`,
```
ChaseC99: Nit: can you add a period?
```suggestion
            `Course added from different term.\nSchedule now contains courses from ${[...multipleTerms]
                .sort()
                .join(', ')}.`,
```

PR #162, opened 2021-05-12
Change schedule selector to dropdown

ChaseC99:
## Summary
Change schedule switcher to a dropdown menu.
schedule-dropdown

## Test Plan
- Add a class to each schedule, verify they appear correctly
- Add a custom event to each schedule, verify they appear correctly
- Save and reload the schedules, verify the events persist

## Future Followup (Optional)
Later on, we'll probably swap "Schedule {1-4}" out for named schedules associated with an account.

### DIFF
--- a/client/src/actions/AppStoreActions.js
+++ b/client/src/actions/AppStoreActions.js
@@ -302,12 +302,7 @@ export const undoDelete = (event) => {
     }
 };
 
-export const changeCurrentSchedule = (direction) => {
-    let newScheduleIndex;
-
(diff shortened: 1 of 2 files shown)


braedons:
image

This issue pops up when selecting schedule 5. There's also no option to add a custom class or searched-for class to schedule 5. Seems like it's there by mistake?

ChaseC99:
Good catch - thanks. I'll try to reproduce this locally and figure out what's going on.

ChaseC99:
Ah you're right; Schedule 5 was added by mistake! 

Before we only had schedules 1-4. The 5th one caused the index-out-of bounds error.

braedons (approved):
Everything looks good now! No errors found anymore.

PR #168, opened 2021-07-14
Map additions

falsidge:
## Summary
Added indexes to map (was retroactively taken out from the repo after the new merge) 
Added routing to map 
Uses Directions API and not the Static Tile API, which allows for 100,000 requests for free https://www.mapbox.com/pricing/#navigation so it should not be a problem. The fetches are cached so it should not use many API requests.
This is only used to give rough estimates for a schedule. It is not suppose to show exactly where the classes are. The direction API are given coordinates that were already included in AntAlmanac's json file.

https://user-images.githubusercontent.com/8692395/126052532-0535e80c-a9f3-4eb5-95c1-1ead409a44b9.mp4

## Test Plan
Tested when there is only 2 courses are less and added checks for these cases

## Future Followup (Optional)

Add the markers to show up on hover on path? This may cause the user to not know about this feature.
Possibly lookup the exactly location of classes. (Not sure if mapbox has this but it would be nice if we have map.uci.edu access)

### DIFF
--- a/client/src/components/Map/MapMarkerPopup.js
+++ b/client/src/components/Map/MapMarkerPopup.js
@@ -34,7 +34,7 @@ class MapMarkerPopup extends PureComponent {
                                         top: -0.75rem;
                                         text-align: center; 
                                         color: white" >
-                                        ${this.props.index}
+                            ${this.props.index}
                         </div>
                     </div>`,
         });
--- a/client/src/components/Map/MapMarkerPopup.js
+++ b/client/src/components/Map/MapMarkerPopup.js
@@ -15,7 +15,7 @@ class MapMarkerPopup extends PureComponent {
             labelAnchor: [-3.5, 0],
             popupAnchor: [0, -21],
             className: '',
-            html:   `<div style="position:relative;
+            html: `<div style="position:relative;
                                     left: -1rem;
                                     top: -1rem;">
                         <span style="background-color: ${color};
@@ -24,16 +24,19 @@ class MapMarkerPopup extends PureComponent {
                                         display: block;
                                         left: -1rem;
                                         top: -1rem;
-                                        position: relative;
+                                        position: absolute;
                                         border-radius: 1.9rem 1.9rem 0;
                                         transform: rotate(45deg);
-                                        border: 1px solid #FFFFFF" />
+                                        border: 1px solid #FFFFFF">
+                        </span>
                         <div style="position: absolute;
                                         width: 1.75rem;
                                         height: 1.75rem;
-                                        top: 0.25rem
-                                        text-align: center" >
-                                        ${index}
+                                        left: -1rem;
+                                        top: -0.75rem;
+                                        text-align: center" 
+                                        text-color: "white" >
+                                        ${this.props.index}
                         </div>
                     </div>`,
         });
--- a/client/src/components/Map/MapMarkerPopup.js
+++ b/client/src/components/Map/MapMarkerPopup.js
@@ -15,9 +15,7 @@ class MapMarkerPopup extends PureComponent {
             labelAnchor: [-3.5, 0],
             popupAnchor: [0, -21],
             className: '',
-            html: `<div style="position:relative;
-                                    left: -1rem;
-                                    top: -1rem;">
+            html: `<div style="position:relative;">
                         <span style="background-color: ${color};
                                         width: 1.75rem;
                                         height: 1.75rem;
@@ -34,8 +32,8 @@ class MapMarkerPopup extends PureComponent {
                                         height: 1.75rem;
                                         left: -1rem;
                                         top: -0.75rem;
-                                        text-align: center" 
-                                        text-color: "white" >
+                                        text-align: center; 
+                                        color: white" >
                                         ${this.props.index}
                         </div>
                     </div>`,
--- a/client/src/components/Map/MapMarkerPopup.js
+++ b/client/src/components/Map/MapMarkerPopup.js
@@ -15,16 +15,27 @@ class MapMarkerPopup extends PureComponent {
             labelAnchor: [-3.5, 0],
             popupAnchor: [0, -21],
             className: '',
-            html: `<span style="background-color: ${color};
-                            width: 1.75rem;
-                            height: 1.75rem;
-                            display: block;
-                            left: -1rem;
-                            top: -1rem;
-                            position: relative;
-                            border-radius: 1.9rem 1.9rem 0;
-                            transform: rotate(45deg);
-                            border: 1px solid #FFFFFF" />`,
+            html:   `<div style="position:relative;
+                                    left: -1rem;
+                                    top: -1rem;">
+                        <span style="background-color: ${color};
+                                        width: 1.75rem;
+                                        height: 1.75rem;
+                                        display: block;
+                                        left: -1rem;
+                                        top: -1rem;
+                                        position: relative;
+                                        border-radius: 1.9rem 1.9rem 0;
+                                        transform: rotate(45deg);
+                                        border: 1px solid #FFFFFF" />
+                        <div style="position: absolute;
+                                        width: 1.75rem;
+                                        height: 1.75rem;
+                                        top: 0.25rem
+                                        text-align: center" >
+                                        ${index}
+                        </div>
+                    </div>`,
         });
     };
 
--- a/client/src/components/Map/UCIMap.js
+++ b/client/src/components/Map/UCIMap.js
@@ -7,6 +7,7 @@ import DayTabs from './MapTabsAndSearchBar';
 import MapMarkerPopup from './MapMarkerPopup';
 import Locate from 'leaflet.locatecontrol';
 import Leaflet from 'leaflet';
+import { duration } from '@material-ui/core';
 
 class LocateControl extends PureComponent {
     componentDidMount() {
@@ -48,13 +49,15 @@ export default class UCIMap extends PureComponent {
         eventsInCalendar: AppStore.getEventsInCalendar(),
         currentScheduleIndex: AppStore.getCurrentScheduleIndex(),
         poly: [],
+        info_markers: [],
+        info_marker: null,
     };
 
     getRoute = (day) => {
         if (day) {
             let index = 0;
-            let coords = '';
-            let coords_array = [];
+            let coords = ''; // lat and lng of markers to be passed to api
+            let coords_array = []; // lat and lng of markers to be used later
             let colors = [];
 
             // Filter out those in a different schedule or those not on a certain day (mon, tue, etc)
@@ -110,6 +113,8 @@ export default class UCIMap extends PureComponent {
                         ]; // Path is a list of paths for each waypoint. For example, path[0] is the path to waypoint 0, path[1] is the path from 0 to 1... etc.
 
                         let poly = []; // Arrays of polyline to be added to map
+                        let info_markers = [];
+
                         poly.push(
                             <Polyline color={colors[0]} positions={[path[0][0], coords_array[0]]} dashArray="4" />
                         ); // Draw a dashline from waypoint 0 to start of route
@@ -123,7 +128,47 @@ export default class UCIMap extends PureComponent {
                                 path.push([[lng, lat]]);
                                 if (waypointIndex !== 0) {
                                     poly.push(
-                                        <Polyline color={colors[waypointIndex - 1]} positions={path[waypointIndex]} />
+                                        <Polyline
+                                            color={colors[waypointIndex - 1]}
+                                            positions={path[waypointIndex]}
+                                            index={waypointIndex}
+                                            map={this}
+                                            onmouseover={function () {
+                                                let [
+                                                    position,
+                                                    color,
+                                                    duration,
+                                                    miles,
+                                                ] = this.options.map.state.info_markers[this.options.index - 1];
+                                                this.options.map.setState({
+                                                    info_marker: (
+                                                        <Marker
+                                                            position={position}
+                                                            opacity={1.0}
+                                                            icon={Leaflet.divIcon({
+                                                                iconAnchor: [0, 14],
+                                                                labelAnchor: [-3.5, 0],
+                                                                popupAnchor: [0, -21],
+                                                                className: '',
+                                                                iconSize: [1000, 14],
+                                                                html: `<div style="background-color: white;border-left-color: ${color};border-left-style: solid;width: fit-content;border-left-width: 5px;padding-left: 10px;padding-right: 10px;padding-top: 4px;padding-bottom: 4px;">
+                                                                    <span style="color:${color}">
+                                                                    ${duration} 
+                                                                    </span>
+                                                                    <br>
+                                                                    <span style="color:#888888">
+                                                                    ${miles}
+                                                                    </span>
+                                                                </div>`,
+                                                            })}
+                                                        ></Marker>
+                                                    ),
+                                                });
+                                            }}
+                                            onmouseout={function () {
+                                                this.options.map.setState({ info_marker: null });
+                                            }}
+                                        />
                                     ); // Draw path from last waypoint to next waypoint
                                     let duration =
                                         obj['routes'][0]['legs'][waypointIndex - 1]['duration'] > 30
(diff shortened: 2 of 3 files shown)


ChaseC99:
We're super stoked about this feature!!

This weekend, I'll have some time to review the code and I'll also pull the analytics on our map usage to verify that these new requests won't push us over the api limit.

Thanks for opening a PR; looking forward to reviewing it!

client/src/stores/AppStore.js
```diff
@@ -19,6 +19,7 @@ class AppStore extends EventEmitter {
         this.eventsInCalendar = [];
         this.finalsEventsInCalendar = [];
         this.unsavedChanges = false;
+        this.setMaxListeners(100);
```
ChaseC99: Please remove

client/src/stores/AppStore.js
```diff
@@ -128,6 +129,7 @@ class AppStore extends EventEmitter {
                 this.emit('addedCoursesChange');
                 break;
             case 'CHANGE_CURRENT_SCHEDULE':
+                console.log('change current schedule');
```
ChaseC99: Please remove

client/src/components/Map/UCIMap.js
```diff
@@ -45,6 +47,125 @@ export default class UCIMap extends PureComponent {
...
+                const buildingCode = event.bldg.split(' ')[0];
+                const id = locations[buildingCode];
+                const locationData = buildingCatalogue[id];
+
+                if (locationData === undefined) return;
+
+                colors.push(event.color);
+
+                if (day) {
+                    if (coords) coords += ';';
```
ChaseC99: For clarity
```suggestion
                    if (coords) { coords += ';'; }
```

client/src/components/Map/UCIMap.js
```diff
@@ -45,6 +47,125 @@ export default class UCIMap extends PureComponent {
...
+        this.state.eventsInCalendar
+            .sort((event, event2) => event.start - event2.start)
+            .forEach((event) => {
+                // Filter out those in a different schedule or those not on a certain day (mon, tue, etc)
+                if (
+                    event.isCustomEvent ||
+                    !event.scheduleIndices.includes(this.state.currentScheduleIndex) ||
+                    !event.start.toString().includes(DAYS[day])
+                )
+                    return;
```
ChaseC99: Nit: you could `.filter` for this

client/src/components/Map/UCIMap.js
```diff
@@ -45,6 +47,125 @@ export default class UCIMap extends PureComponent {
...
+
+                // Get building code, get id of building code, which will get us the building data from buildingCatalogue
+                const buildingCode = event.bldg.split(' ')[0];
```
ChaseC99: A few buildings have 2 word building codes ("UN CLB", "Student Center", "Mesa Court", "Middle Earth",  "UCI Health",  "Hitachi, PH", "Dean's Office").

It's unlikely that there are classes here, but it's probably better to be safe, especially if new 2-word building codes are added in the future.

```suggestion
                const buildingCode = event.bldg.split(' ').slice(0, -1).join(' ');
```

client/src/components/Map/UCIMap.js
```diff
@@ -45,6 +47,125 @@ export default class UCIMap extends PureComponent {
...
+                access_token: ACCESS_TOKEN,
+            }).toString();
+
+            fetch(url, {
+                method: 'GET',
+                headers: { 'Content-Type': 'application/json' },
+            }).then((response) => {
+                response.json().then((obj) => {
+                    let latlng = obj['routes'][0]['geometry']['coordinates'];
+                    let waypoint = 0;
```
ChaseC99: - Pull `waypoints` out of `obj`
- Rename `waypoint` to `waypointIndex`
```suggestion
                    let waypoints = obj['waypoints'];
                    let waypointIndex = 0;
```

client/src/components/Map/UCIMap.js
```diff
@@ -45,6 +47,125 @@ export default class UCIMap extends PureComponent {
...
+                steps: false,
+                access_token: ACCESS_TOKEN,
+            }).toString();
+
+            fetch(url, {
+                method: 'GET',
+                headers: { 'Content-Type': 'application/json' },
+            }).then((response) => {
+                response.json().then((obj) => {
+                    let latlng = obj['routes'][0]['geometry']['coordinates'];
```
ChaseC99: `latlng` is a little deceiving, since it's actually a list of lats and lngs. 

Rename to `coordinates` and leave a comment explaining that it's a list of latlngs
```suggestion
                    let coordinates = obj['routes'][0]['geometry']['coordinates'];
```

client/src/components/Map/UCIMap.js
```diff
@@ -45,6 +47,125 @@ export default class UCIMap extends PureComponent {
...
+                    let waypoint = 0;
+                    let path = [
+                        [[obj['waypoints'][waypoint]['location'][1], obj['waypoints'][waypoint]['location'][0]]],
+                    ];
+                    let poly = [];
+                    poly.push(<Polyline color={colors[0]} positions={[path[0][0], coords_array[0]]} dashArray="4" />);
+
+                    for (let i of latlng) {
```
ChaseC99: Hard to understand what `i` represents at first glance. We can actually destructure here to pull out the lat and lng. The rest of the code in this for loop should be updated accordingly to use lat/lng instead of i[0/1]

```suggestion
                    for (let [lat, lng] of coordinates) {
```
_Notes: this assumes `latlng` is renamed to `coordinates`, per my previous comment_

client/src/components/Map/UCIMap.js
```diff
@@ -45,6 +47,125 @@ export default class UCIMap extends PureComponent {
...
+                            path.push([[i[1], i[0]]]);
+                            if (waypoint != 0) {
+                                poly.push(<Polyline color={colors[waypoint - 1]} positions={path[waypoint]} />);
+                                poly.push(
+                                    <Polyline
+                                        color={colors[waypoint - 1]}
+                                        positions={[path[waypoint][path[waypoint].length - 1], coords_array[waypoint]]}
+                                        dashArray="4"
+                                    />
+                                );
```
ChaseC99: This code is quite difficult to understand because of the response from the API. It'd be hard for other devs to work modify this code in the future.

To make it more readable, we should pull the waypoints and the coordinates out of `obj`, so we no longer reference `obj` again in the code.

Please also add comments, including an explanation what waypoints and coordinates represent.

client/src/components/Map/UCIMap.js
```diff
@@ -45,6 +47,125 @@ export default class UCIMap extends PureComponent {
...
         eventsInCalendar: AppStore.getEventsInCalendar(),
         currentScheduleIndex: AppStore.getCurrentScheduleIndex(),
+        poly: [],
+    };
+
+    getRoute = (day) => {
+        let index = 0;
+        let coords = '';
+        let coords_array = [];
+        let colors = [];
```
ChaseC99: Add comments explaining what these vars are for

client/src/components/Map/UCIMap.js
```diff
@@ -45,6 +47,125 @@ export default class UCIMap extends PureComponent {
...
+                headers: { 'Content-Type': 'application/json' },
+            }).then((response) => {
+                response.json().then((obj) => {
+                    let latlng = obj['routes'][0]['geometry']['coordinates'];
+                    let waypoint = 0;
+                    let path = [
+                        [[obj['waypoints'][waypoint]['location'][1], obj['waypoints'][waypoint]['location'][0]]],
+                    ];
+                    let poly = [];
+                    poly.push(<Polyline color={colors[0]} positions={[path[0][0], coords_array[0]]} dashArray="4" />);
```
ChaseC99: I'm trying to understand what these two positions items represent.

Is this correct?
- coords_array[0] is the pin of the building on the map
- path[0][0] is the start of the route (where solid line begins)
- the line between two points is the dashed line seen below vvvv
image

client/src/components/Map/UCIMap.js
```diff
@@ -45,6 +47,125 @@ export default class UCIMap extends PureComponent {
...
+                                                        Math.floor(
+                                                            obj['routes'][0]['legs'][waypoint - 1]['distance'] /
+                                                                1.609 /
+                                                                10
+                                                        ) / 100
+                                                    ).toString() + ' mi'
+                                                }
+                                                </span>
+                                            </div>`,
+                                        })}
```
ChaseC99: The math here is brilliant, but it's hard to follow inside of the string interpolation.
Pull all the computations out of the string and set them as variables above line 124 (`poly.push(`)

client/src/components/Map/MapMarkerPopup.js
```diff
@@ -15,16 +15,28 @@ class MapMarkerPopup extends PureComponent {
...
+                        </span>
+                        <div style="position: absolute;
+                                        width: 1.75rem;
+                                        height: 1.75rem;
+                                        left: -1rem;
+                                        top: -0.75rem;
+                                        text-align: center; 
+                                        color: white" >
+                                        ${this.props.index}
+                        </div>
```
ChaseC99: Nit: I think this indentation can come in a bit
```suggestion
                            ${this.props.index}
                        </div>
```

ChaseC99 (changes requested):
This is an awesome PR! 

The code to index the map markers is pretty solid. 
The path routing looks great and I can tell a lot of work went into it! The code is brilliant but there's a lot going on there, so it might be a bit hard for future devs to understand. I've left a few comments on how to refactor it so it'll be easier for others to work with in the future.

Let me know if you have questions or need any help. Can't wait to get this out users!

### Requested Changes

- Left additional comments with some recommended changes
- I know in the "Future Followup" you mentioned potentially making the time/distance markers show up on hover. I think you should try to do that in this PR. For routes that are close to each other, the markers cover a lot of the path. 
image
- Indexing edge case: when a 2+ classes are in the same building, the index skips over it. 
In this scenario, the student's classes go (1) NS1 -> (2) NS1 -> (3) HSLH. 
The markers should show HSLH as (3) instead of (2).
image
_Side note: Don't worry about indicating that there are two classes in NS1. Seems like it will be tricky, so I'll probably ask our design team for help. I've opened issue #169 to address this in the future. For now, just make sure HSLH is properly labeled (3)_ 



### API Usage
Did some math to anticipate our API usage. We should remain in the free tier! 
> The map tab is viewed about 4,000 times/month.
> There are 5 days in a week and about 2-4 classes per day. 
> So, 4000 * 5 * 3 ≈ 60000 requests/month. 

Since this feature could make our map tab much more popular, we'll continue to monitor the usage to make sure it doesn't go over the 100k limit.

client/src/components/Map/UCIMap.js
```diff
@@ -45,6 +47,125 @@ export default class UCIMap extends PureComponent {
...
+                    poly.push(<Polyline color={colors[0]} positions={[path[0][0], coords_array[0]]} dashArray="4" />);
+
+                    for (let i of latlng) {
+                        path[waypoint].push([i[1], i[0]]);
+                        if (
+                            i[0] === obj['waypoints'][waypoint]['location'][0] &&
+                            i[1] === obj['waypoints'][waypoint]['location'][1]
+                        ) {
+                            path.push([[i[1], i[0]]]);
+                            if (waypoint != 0) {
```
ChaseC99:
```suggestion
                            if (waypoint !== 0) {
```

client/src/components/Map/UCIMap.js
falsidge: Yes, this is correct.

client/src/components/Map/UCIMap.js
```diff
@@ -1,11 +1,13 @@
...
-import { Map, TileLayer, withLeaflet } from 'react-leaflet';
+import { Map, TileLayer, withLeaflet, Polyline, Marker } from 'react-leaflet';
 import buildingCatalogue from './static/buildingCatalogue';
 import locations from '../SectionTable/static/locations.json';
 import AppStore from '../../stores/AppStore';
 import DayTabs from './MapTabsAndSearchBar';
 import MapMarkerPopup from './MapMarkerPopup';
 import Locate from 'leaflet.locatecontrol';
+import Leaflet from 'leaflet';
+import { duration } from '@material-ui/core';
```
ChaseC99: Accidental import, please remove
```suggestion

```

client/src/components/Map/UCIMap.js
```diff
@@ -72,51 +239,56 @@ export default class UCIMap extends PureComponent {
...
+                const id = locations[buildingCode];
+                const locationData = buildingCatalogue[id];
+                const courseString = `${event.title} ${event.sectionType} @ ${event.bldg}`;
 
-            // Get building code, get id of building code, which will get us the building data from buildingCatalogue
-            const buildingCode = event.bldg.split(' ')[0];
-            const id = locations[buildingCode];
-            const locationData = buildingCatalogue[id];
-            const courseString = `${event.title} ${event.sectionType} @ ${event.bldg}`;
+                index++; // index temp fix for courses in same building
```
ChaseC99: Is it actually a "temp" fix? Seems like a solid fix to me
```suggestion
                index++; // always increment index to account for courses within the same building
```

client/src/components/Map/UCIMap.js
```diff
@@ -72,51 +239,56 @@ export default class UCIMap extends PureComponent {
...
 
+                markers.push(
+                    <MapMarkerPopup
+                        image={locationData.imageURLs[0]}
+                        markerColor={event.color}
+                        location={locationData.name}
+                        lat={locationData.lat}
+                        lng={locationData.lng}
+                        acronym={acronym}
+                        index={this.state.day ? (index + 1).toString() : ''}
```
ChaseC99: Since you increment before generating the MapMarkerPopup, we can remove the +1. This should fix the bug where all the classes start at 2
```suggestion
                        index={this.state.day ? index.toString() : ''}
```

client/src/components/Map/UCIMap.js
```diff
@@ -45,6 +48,170 @@ export default class UCIMap extends PureComponent {
         selected_acronym: '',
         eventsInCalendar: AppStore.getEventsInCalendar(),
         currentScheduleIndex: AppStore.getCurrentScheduleIndex(),
+        poly: [],
+        info_markers: [],
+        info_marker: null,
+    };
+
+    getRoute = (day) => {
+        if (day) {
```
ChaseC99, line 59: This fixes the bug where the previous route doesn't get cleared when there's only one class on the map
```suggestion
        // Clear any existing route on the map
        this.setState({ poly: [], info_marker: null });
        
        if (day) {
```

client/src/components/Map/UCIMap.js
```diff
@@ -45,6 +48,170 @@ export default class UCIMap extends PureComponent {
...
+                                waypointIndex++;
+                            }
+                        }
+                        this.setState({ poly: poly, info_markers: info_markers, info_marker: null });
+                    });
+                });
+            }
+        } else {
+            this.setState({ poly: [], info_marker: null });
+        }
```
ChaseC99: Per my previous suggestion, if you always clear the route at the top of the function, you no longer need this else statement.
```suggestion
        }
```

client/src/components/Map/UCIMap.js
```diff
@@ -45,6 +48,170 @@ export default class UCIMap extends PureComponent {
...
+                        for (let [lat, lng] of coordinates) {
+                            path[waypointIndex].push([lng, lat]); // Creates a path using lat and lng of coordinates until lat and lng matches one of the waypoint's coordinates
+                            if (
+                                lat === waypoints[waypointIndex]['location'][0] &&
+                                lng === waypoints[waypointIndex]['location'][1]
+                            ) {
+                                path.push([[lng, lat]]);
+                                if (waypointIndex !== 0) {
+                                    poly.push(
+                                        <Polyline
```
ChaseC99, line 161: Add a `zIndexOffset` here, so the info is displayed in front of the pin
| Before | After |
|--|--|
| image | image |

```suggestion
                                        <Polyline
                                            zIndexOffset={100}
```

ChaseC99 (changes requested):
Love the hover functionality, this is looking awesome 🔥 

I found a few small bugs and left some suggestions on how to fix them. Once those are resolved I think this is ready to get merged!

- Indexing is off by one. Looks like everything starts at 2 
image


- When a day with only one class is selected, the previous route doesn't get cleared
single_class_bug

falsidge:
Would it be better for the info marker to hover on the mouse?

ChaseC99:
> Would it be better for the info marker to hover on the mouse?

Sure I think that would be cool. If you want to try it, go for it!

falsidge:
> > Would it be better for the info marker to hover on the mouse?
> 
> Sure I think that would be cool. If you want to try it, go for it!

Ok, added.

client/src/components/Map/UCIMap.js
```diff
@@ -285,7 +277,7 @@ export default class UCIMap extends PureComponent {
                             <hr />
                             Class: {`${event.title} ${event.sectionType}`}
                             <br />
-                            Room: {event.bldg.split(' ')[1]}
+                            Room: {event.bldg.split(' ').slice(-1)}
```
ChaseC99, line 280: Good catch

client/src/components/Map/UCIMap.js
```diff
@@ -45,6 +47,164 @@ export default class UCIMap extends PureComponent {
         selected_acronym: '',
         eventsInCalendar: AppStore.getEventsInCalendar(),
         currentScheduleIndex: AppStore.getCurrentScheduleIndex(),
+        poly: [],
+        info_markers: [],
+        info_marker: null,
+    };
+
+    getRoute = (day) => {
```
ChaseC99: Nit: it might make more sense to name this something like `generateRoute`.   
"get" implies that the function is returning data.

client/src/components/Map/UCIMap.js
```diff
@@ -45,6 +47,164 @@ export default class UCIMap extends PureComponent {
...
+
+                        for (let [lat, lng] of coordinates) {
+                            path[waypointIndex].push([lng, lat]); // Creates a path using lat and lng of coordinates until lat and lng matches one of the waypoint's coordinates
+                            if (
+                                lat === waypoints[waypointIndex]['location'][0] &&
+                                lng === waypoints[waypointIndex]['location'][1]
+                            ) {
+                                path.push([[lng, lat]]);
+                                if (waypointIndex !== 0) {
+                                    function setInfoMarker(event) {
```
ChaseC99, line 131: Nit: this function _should_ probably be defined outside of `getRoute`

ChaseC99 (approved):
Outstanding job on this! Thanks so much for your time developing this feature and responding to feedback. Like I said before, I think users are really going to love this 💯 

I left a few nits but they aren't critical, so feel free to ignore them. I'll merge this PR in on Monday.

ALSO
The ICSSC Projects Committee is planning on launching a Discord this fall for people who want to contribute to projects like AntAlmanac. If you're interested, we'd like to invite you to join early and receive a special "AntAlmanac Contributor" role for this PR.
https://discord.gg/GzF76D7UhY

Thanks again for your hard work 🙌

client/src/components/Map/UCIMap.js
falsidge: `this`  gets overrided if put outside and `this` is needed to get the index of the polyline. i am not sure how to fix this problem.

ChaseC99: You could add a parameter to `setInfoMarker` to pass in the `this` object, so you would still be able to reference it.

However like I said, this isn't super critical ¯\_(ツ)_/¯

ChaseC99:
These changes have been deployed 🚢

PR #173, opened 2021-08-13
Add GDIM to the Dept Search Bar

ChaseC99:
## Summary
GDIM (Game Design and Interactive Media) is a new department, added for Fall 2021. 
Before this PR, users were unable to add any GDIM courses to their schedule.

## Test Plan
Verify that GDIM appears in the dept search bar and that users can add a class to their schedule.


ChaseC99:
Is there anything else that needs to be done when adding a new dept?

PR #187, opened 2021-09-19
Fix loading classes when ad banner can't load 

falsidge:
## Summary
The ad banner was not being checked before being accessed, which caused an error, which prevented classes from being loaded if the ad banners failed to load


ChaseC99 (approved):
LGTM! Thanks for fixing this 🙌

PR #189, opened 2021-09-25
Fix color picker not updating and fix routes not changing when schedule changes 

### LINKED ISSUE #172: Map routes aren't redrawn when schedules change
When a user switches schedules, the routes on a map are not reset
maps_routing_bug

falsidge:
## Summary

Fix color picker not updating due to state not changing when props are updated
Fix routes not changing when schedule changes due to state not updating

## Issues

Color picker was not updating if a course was already selected and another course was selected.
Issue #172

### DIFF
--- a/client/src/components/App/ColorPicker.js
+++ b/client/src/components/App/ColorPicker.js
@@ -70,6 +70,16 @@ class ColorPicker extends PureComponent {
             </span>
         );
     }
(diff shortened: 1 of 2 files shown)


ChaseC99 (approved):
Looks great! Thanks for fixing both of these 🙌

PR #194, opened 2021-10-03
Fix clearing a schedule would clear courses/events that appear only in one schedule

### LINKED ISSUE #190: "Clear Schedule" deleting events on other schedules
This is a weird one and looks like there might be multiple bugs around the "Clear Schedule" functionality.

### Steps to Reproduce
_Bug:_ Deletes unsaved events on other schedules
1. Load a schedule
2. Add events to schedule one and two
3. Clear schedule one
_Result:_ See that the events added to two are also cleared

_Bug:_ Deletes custom events on other schedules
1. Add a custom event to two schedules
2. Delete one schedule
3. Delete the same schedule again
_Result:_ See that the custom event was deleted from the other schedule

falsidge:
## Summary
This bug is caused by an extra check that would cause courses and events to not be included if they only appeared in one schedule.
I am not sure why this is here besides potentially speeding the course deletion up if the course was only in one schedule, but it doesn't check if the one schedule is the current schedule. It seems to be unnecessary anyways.
## Test Plan
1. Add courses into a schedule
2. Go to a different schedule
3. Clear schedule
4. Examine that courses in the original schedule are also deleted
## Issues
Fixes #190

### DIFF
--- a/client/src/actions/AppStoreActions.js
+++ b/client/src/actions/AppStoreActions.js
@@ -256,25 +256,15 @@ export const clearSchedules = (scheduleIndicesToClear) => {
     const addedCourses = AppStore.getAddedCourses();
     const customEvents = AppStore.getCustomEvents();
     const addedCoursesAfterClear = addedCourses.filter((course) => {
-        if (course.scheduleIndices.length === 1) {
-            return false;
-        } else {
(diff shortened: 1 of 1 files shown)


ChaseC99 (approved):
Yay thanks figuring this out! I also have no clue why that check was there LOL

Great work 💯

Issue #196, opened 2021-10-13
Changing Color is Broken


ChaseC99:
Trying to change a course color from the schedule pane or added classes pane causes the site to crash.
I believe this is because `componentDidUpdate` is changing the state and causing infinite recursion.
![color_bug](https://user-images.githubusercontent.com/29494270/137207111-35b37c77-2bbc-4e21-92e8-43971d855fd5.gif)

PR #197, opened 2021-10-14
Color Picker Fixes

falsidge:
## Summary
Color picker would not change color and randomly move to the left.
## Issues
#196
Color picker was not setting the state properly due componentDidUpdate setting the color state back to the props.
## Future Followup (Optional)
I am not sure if this entirely fixes the problem.


ChaseC99:
Unfortunately this just changes the color for everything that gets clicked :/
color_bug

ChaseC99:
Dispatcher was a clever way to fix this but I think I found a simpler way to do this in #198

PR #198, opened 2021-10-17
Fix Course Color Bug

### LINKED ISSUE #196: Changing Color is Broken

ChaseC99:
## Summary
Calendar events show the correct color for the color picker.

## Test Plan
- Load a schedule with courses and custom events
- Change the color of courses and custom events
- Change the color from the "Added Classes" section
- Verify that the courses change color and that the color picker is correctly reset on each new event
color_bug_fix

## Issues
Closes #196

### DIFF
--- a/client/src/components/App/ColorPicker.js
+++ b/client/src/components/App/ColorPicker.js
@@ -32,7 +32,9 @@ class ColorPicker extends PureComponent {
             if (this.props.isCustomEvent) changeCustomEventColor(this.props.customEventID, this.state.color);
             else changeCourseColor(this.props.sectionCode, this.state.color, this.props.term);
         });
+    };
 
+    handleColorChangeComplete = () => {
         ReactGA.event({
             category: 'antalmanac-rewrite',
             action: 'Change Course Color',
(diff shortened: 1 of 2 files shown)


client/src/components/Calendar/ScheduleCalendar.js
```diff
@@ -183,8 +184,13 @@ class ScheduleCalendar extends PureComponent {
         const { currentTarget } = event;
         event.stopPropagation();
 
-        if (courseInMoreInfo.sectionType !== 'Fin')
-            this.setState({ anchorEvent: currentTarget, courseInMoreInfo: courseInMoreInfo });
+        if (courseInMoreInfo.sectionType !== 'Fin') {
+            this.setState({
+                anchorEvent: currentTarget,
+                courseInMoreInfo: courseInMoreInfo,
+                calendarEventKey: Math.random(),
```
ChaseC99, line 191: Update the `calendarEventKey` to a random number onClick to force a re-render for `CourseCalendarEvent`

client/src/components/Calendar/ScheduleCalendar.js
```diff
@@ -242,6 +248,7 @@ class ScheduleCalendar extends PureComponent {
                         open={Boolean(this.state.anchorEvent)}
                     >
                         <CourseCalendarEvent
+                            key={this.state.calendarEventKey}
```
ChaseC99, line 251: `CourseCalendarEvent` wasn't fully re-rendering, which is why the colors didn't update. By changing the key, we guarantee a full re-render.

client/src/components/App/ColorPicker.js
```diff
@@ -32,7 +32,9 @@ class ColorPicker extends PureComponent {
             if (this.props.isCustomEvent) changeCustomEventColor(this.props.customEventID, this.state.color);
             else changeCourseColor(this.props.sectionCode, this.state.color, this.props.term);
         });
+    };
 
+    handleColorChangeComplete = () => {
```
ChaseC99, line 37: Move GA event to new `handleColorChangeComplete` function.  onChange is triggered many times while a user is dragging the color picker around. GA was firing analytics for this, messing up our numbers. By moving it into the onChangeComplete callback, we fix this bug.

PR #211, opened 2022-01-06
Automatic light/dark mode

ryanmohta:
Here’s automatic dark mode for the people like me who use light mode in the day and dark mode in the night. I swapped out the toggle switch for a radio button selection, where “automatic” aligns with the user’s system settings. Automatic will be the default for new users, but their selection will be saved in local storage.

https://user-images.githubusercontent.com/19882060/148345503-f2cf5916-3227-46f3-9d29-74902819f991.mov

I tested this feature by switching between all combinations of light/dark/auto mode and switching my system settings. I also refreshed the page in between to ensure that the local storage and startup worked also. Feel free to test it yourself though in case I missed anything!

### DIFF
--- a/client/src/actions/AppStoreActions.js
+++ b/client/src/actions/AppStoreActions.js
@@ -387,13 +387,13 @@ export const copySchedule = (from, to) => {
     });
 };
 
-export const toggleDarkMode = (switchEvent) => {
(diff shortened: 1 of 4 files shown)


EricPedley (approved):
Looks good and works on Windows

PR #215, opened 2022-01-20
Add CNAME

ChaseC99:
## Summary
This adds the CNAME to our `public` folder. The CNAME is necessary for GitHub Pages to resolve our domain name.

## Test Plan
Monitor the next deploy to verify that the CNAME file is present on the `gh-pages` branch.

### DIFF
--- a/client/CNAME
+++ b/client/CNAME
@@ -1,1 +1,0 @@
-antalmanac.com
--- a/client/public/CNAME
+++ b/client/public/CNAME
@@ -1,0 +1,1 @@
+antalmanac.com


ryanmohta (changes requested):
This looks good, but I did notice that there’s also a `CNAME` file in the `client` folder that was added a few months ago with the exact same content as this one. I’m guessing that one wasn’t carried over to the main site, so should we remove that file to avoid confusion with this?

ChaseC99:
> This looks good, but I did notice that there’s also a `CNAME` file in the `client` folder that was added a few months ago with the exact same content as this one. 

AHH great catch!! This was the issue. I guess I did add the CNAME file before but it was in the wrong spot. Just pushed a new commit to remove that misplaced file.

ryanmohta (approved):
:shipit:

PR #216, opened 2022-01-21
Fixed search bug caused by auto light/dark mode

ryanmohta:
## Summary
In my previous PR (#211), I changed the storage variable `darkMode` to `theme` to better reflect its now three possible states (light, dark, and auto). Accordingly, I changed its getter method `getDarkMode` to `getTheme`, but forgot to change all occurrences of it in the codebase. The crash was caused by attempting to call the old `getDarkMode` function which no longer existed.

Now, I’ve changed all occurrences of `getDarkMode` with a helper function `isDarkMode`, which calls `getTheme` internally and determines from its value — along with the OS’s dark mode state if in automatic mode — if the user is currently in dark or light mode.

## Test Plan
I tried searching for classes and that works as expected now. I also tried switching my OS’s setting while loading the search results to see if the loading GIF would switch between light and dark mode versions, but it loaded too fast for me to test that which makes me think that it is very rare that a user would encounter this issue (and even if they did, it wouldn’t crash the site, they just wouldn’t be able to see the loading GIF)

### DIFF
--- a/client/src/components/App/App.js
+++ b/client/src/components/App/App.js
@@ -10,17 +10,18 @@ import DesktopTabs from '../CoursePane/DesktopTabs';
 import AppStore from '../../stores/AppStore';
 import { MuiPickersUtilsProvider } from '@material-ui/pickers';
 import DateFnsUtils from '@date-io/date-fns';
+import { isDarkMode } from '../../helpers';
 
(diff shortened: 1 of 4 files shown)


ChaseC99 (approved):
Let's get this out asap

PR #218, opened 2022-01-28
Add backend instructions to readme, fix potential recursive install bug

ChaseC99:
## Summary
This PR does 2 things:
1. Adds a section to the README explaining how to get the backend up and running.

2. Fixs a recursive install bug that happens when the backend submodule was not cloned.
If the submodule didn't exist, `npm run server-install` would actually just call `npm install` in the root directory, which calls server-install again, resulting in an infinite loop.

## Test Plan
Run `npm install` with and without the backend submodule. Verify that it works both times.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,3 +13,15 @@
 
 4. The site should load on http://localhost:3000  
 As you make changes to the React application in `client/src`, those changes will be automatically reflected on the site.
+
+### Running the Backend
+The backend server **isn't necessary for frontend development**.
+
+However if you do want to run the backend, you must do the following:
+- Make sure the antalamanac-backened submodule exists.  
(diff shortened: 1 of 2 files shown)


ChaseC99:
Having `npm install` fall into an infinite loop can be really confusing for new contributors who accidentally do `git clone` instead of `git clone --recursive`.

... not that this would ever happen to me though 👀

Voark:
Will not be testing that recursion bug so am assuming you are right on that one. Only suggestion I have is to maybe add that the backend only is verified to work on node 12 until we transition away from serverless.

ChaseC99:
> add that the backend only is verified to work on node 12 until we transition away from serverless.

Good call! Added ✔️

Voark (approved):

Issue #112, opened 2020-12-14
Maps: Refactor and enhancements


devsdevsdevs:
- [ ] Implement a carousel for multiple images
- [x] If multiple courses are at the same location, show number of overlapping courses and list all courses at the location
- [x] Use addedCourses instead of eventsInCalendar to create markers

ChaseC99:
Clarification on the "Implement a carousel for multiple images"  
This is referring to how the Classrooms website has multiple images for each building. On the maps view, we could implement a card that cycles through each image. Not a huge priority issue though.
![image](https://user-images.githubusercontent.com/29494270/151779532-644b3fb8-3d64-4c65-8f75-40e010d151ef.png)

kartik-bhagatwala:
Thanks for the input. I think I'll just submit the merge without the carousel. I was trying a couple third party libraries for the carousel but I ran into various bugs that I couldn't resolve when using those.

ChaseC99:
What branch are you changes on? We'll want to make sure it doesn't conflict with #223

PR #223, opened 2022-01-30
Fix overlapping dashed lines. Overlapping pins now stack.

### LINKED ISSUE #169: Map pins don't indicate when multiple classes are in one building
### Problem
When a student has 2+ classes in the same building, the map does not reflect that. Currently it only shows the color of the first class. 

### Screenshot
image

### Objective
Find a good way to indicate that a user has multiple classes in the building. 

### Ideas
- Show both colors on the pin
- Put 2 pins next to each other
- On hover, show a popup

### Related
PR #168 adds numbers to the pins to represent the order of classes in the schedule. Maybe we could do something with the pin number to show multiple classes are in that building?

falsidge:
## Summary
Fix dashed lines having different colors than the solid lines.

Pins now stack on top of each other.

image

Old
image

## Test Plan
Checked pins on all days. Tested 5 pins stacked on one building.

## Issues
#169 

#112 also exists and may conflict with this

## Future Followup (Optional)
Future ideas

One popup for the entire stack?
Add time of courses to popup

### DIFF
--- a/client/src/components/Map/UCIMap.js
+++ b/client/src/components/Map/UCIMap.js
@@ -118,10 +118,14 @@ export default class UCIMap extends PureComponent {
                         let poly = []; // Arrays of polyline to be added to map
                         let info_markers = [];
 
-                        poly.push(
-                            <Polyline color={colors[0]} positions={[path[0][0], coords_array[0]]} dashArray="4" />
-                        ); // Draw a dashline from waypoint 0 to start of route
-
+                        if (
+                            waypoints[0]['location'][0] !== waypoints[1]['location'][0] ||
+                            waypoints[0]['location'][1] !== waypoints[1]['location'][1]
+                        ) {
+                            poly.push(
+                                <Polyline color={colors[0]} positions={[path[0][0], coords_array[0]]} dashArray="4" />
+                            ); // Draw a dashline from waypoint 0 to start of route
+                        }
                         for (let [lat, lng] of coordinates) {
                             path[waypointIndex].push([lng, lat]); // Creates a path using lat and lng of coordinates until lat and lng matches one of the waypoint's coordinates
                             if (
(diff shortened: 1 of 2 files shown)


ChaseC99:
I think this is a brilliant way to handle this!! Tested the changes locally and it looks awesome 👌 
I'll take some time to actually review the code this week.

ChaseC99:
> #112 also exists and may conflict with this

I don't think they really conflict. To add an image carousel, we would just be updating the existing `MapMarkerPopup` component. But I think this overlapping pins feature is way more important than the carousel anyways.

client/src/components/Map/UCIMap.js
```diff
@@ -232,28 +246,41 @@ export default class UCIMap extends PureComponent {
...
                     !(
-                        event.isCustomEvent ||
-                        !event.scheduleIndices.includes(AppStore.getCurrentScheduleIndex()) ||
-                        !event.start.toString().includes(DAYS[this.state.day])
+                        (
+                            event.isCustomEvent ||
+                            !event.scheduleIndices.includes(AppStore.getCurrentScheduleIndex()) ||
+                            !event.start.toString().includes(DAYS[this.state.day]) ||
+                            courses.has(event.sectionCode) || // Remove duplicate courses that appear in the calendar
+                            !courses.add(event.sectionCode)
```
ChaseC99, line 262: Won't this always return true, since the set isn't empty?

ChaseC99 (approved):
Looks awesome 👍  
Always so impressed with your work on the map component!

client/src/components/Map/UCIMap.js
falsidge: it's a trick and I think it might not be very clear but I couldn't think of an easier way to add filtering of the sectioncode without going through the array again at the time.
if !event.scheduleIndices.includes(AppStore.getCurrentScheduleIndex()) or !event.start.toString().includes(DAYS[this.state.day]) conditions are true then it is negated and returns false to the filter (if the schedule indices does not include the current index or the event does not start in the current day). If those are passed, check if the courses set if the event has already been added, if true it return false after negation and filter skips it else it adds to the set and return false, which actually returns true.

a better way to handle this is to "Use addedCourses instead of eventsInCalendar to create markers"  as #112 mentioned

PR #230, opened 2022-02-06
Saturday and Sunday columns are visible if there's a weekend course/event

### LINKED ISSUE #141: Add Saturday/Sunday Columns iff Schedule has a Sat/Sun Class
We should show columns for Sat and Sun if (and only if) the user's schedule has a Sat or Sun class.
Very few classes are on Sat/Sun but there are some for drama and occasionally writing.

We should also do this for Saturday/Sunday finals. 
_Note: Saturday finals should be before Sunday on the week view_

Douglas-Hong:
## Summary
- If a course is on a Saturday or Sunday, the calendar will expand so that all seven days are visible (Sunday to Saturday).
- If a weekend course is removed, the calendar will revert back to the standard view (Monday to Friday).
- If a course has a final on a Saturday or Sunday, the calendar that shows the finals will also expand to show all seven days; however, it will start on Saturday instead of Sunday because Saturday finals are earlier.
- Just to be consistent, users can have custom events on Saturdays/Sundays.

## Test Plan
- Screenshotting works properly if the calendar has a weekend course, final, and/or custom event.
- Downloading an .ics file still works if there's something on the weekend.
- Adding non-weekend courses still works and are in the correct position on the calendar.
- The calendar properly expands/contracts when you add/remove weekend events.
- Saving/loading a schedule with a weekend event works; the weekend event will still appear.

Demo:

https://user-images.githubusercontent.com/78244965/152706517-49bd1194-32b5-4746-b024-a91afa42ffae.mp4

## Issues
- Since the columns get smaller when the calendar expands, the styling for each event is a little off.
- Closes #141

### DIFF
--- a/client/src/components/Calendar/ScheduleCalendar.js
+++ b/client/src/components/Calendar/ScheduleCalendar.js
@@ -213,6 +213,13 @@ class ScheduleCalendar extends PureComponent {
         const events = this.getEventsForCalendar();
         const hasWeekendCourse = events.some((event) => event.start.getDay() == 0 || event.start.getDay() == 6);
 
+        // If a final is on a Saturday or Sunday, let the calendar start on Saturday
+        moment.locale('es-us', {
+            week: {
+                dow: hasWeekendCourse && this.state.showFinalsSchedule ? 6 : 0,
+            },
+        });
+
         return (
             <div className={classes.container} onClick={this.handleClosePopover}>
                 <CalendarPaneToolbar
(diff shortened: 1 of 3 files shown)


ChaseC99:
I wonder if we can make the Add Custom Event module wider so "Saturday" isn't on a new line 
image

ChaseC99 (approved):
FANTASTIC WORK ON THIS PR 👏 

This feature is so cool and I think lots of students will be pleasantly surprised to see their Sa/Su classes on the calendar! 

Your test plan is also super solid - It's thorough and I believe you thought of every possible edge case ✅ 

Great job overall 🚢

PR #234, opened 2022-02-12
Persist search results

### LINKED ISSUE #124: Keep search results page up when switching tabs
Right now, if you search for classes, then go to the maps or added classes tab, going back to the search tab will show you the form. We need to keep the search results pane mounted somehow, with little cost to performance.

EricPedley:
## Summary
After this change, when you switch from the search results tab to added classes or map and back, it keeps the search results up. I also made queryWebsoc cache its results, so doing a search with the same parameters is faster.
## Test Plan
Make some searches, switch to other tabs and back, check that the search results are still up.
## Issues
Closes #124


EricPedley:
I think a lot of these commits are from me using this branch instead of fix-warnings by accident. Most of the real changes are in the commits with messages "`made queryWebsoc() return json body...`" and "`moved state of RightPane to RightPaneStore`"

client/src/components/CoursePane/CourseRenderPane.js
```diff
@@ -142,36 +142,27 @@ class CourseRenderPane extends PureComponent {
...
+                        courseData: flattenSOCObject(jsonResp),
+                        bannerName: jsonAdInfo.bannerName,
+                        bannerLink: jsonAdInfo.bannerLink,
+                    });
                 } else {
                     this.setState({
                         loading: false,
-                        error: true,
+                        error: false,
+                        courseData: flattenSOCObject(jsonResp),
```
ChaseC99: This will probably be a merge conflict with #239.  
Let's close that one first and then merge the changes into this branch.

ChaseC99 (changes requested):
Great job on the caching! There's actually no load time when we remove ads. 
no_ads


Wait for #239 to close, merge in the changes, and we should be good to go!

ChaseC99:
#239 is completed. Resolve the conflicts and you should be good to go! Check in with @kartik-bhagatwala if you have any questions about the ad banner stuff.

ChaseC99 (changes requested):
Sorry one more edge case that I thought of - when the user presses the back button, we should clear the cache. Otherwise, if they keep searching for the same class the course enrollment numbers won't update.
image

EricPedley:
The cache is in JS memory, so it gets cleared every time they open the page or refresh. I don't see how having new enrollment data on every search is helpful because I can't think of why you would need the data more accurately than the number when you started your session. Can you elaborate on that use case further?

ChaseC99:
> Can you elaborate on that use case further?

### Scenario
Today's the day of your enrollment window. You finished planning your schedule but you leave the site open and keep checking back to see if your classes fill up. Your enrollment window opens up and you go to sign up for classes. `Error: this class is full`. What? AntAlmanac says there are still spots open.

### Explanation
Some classes fill up quick during the enrollment period. It's a really bad user experience if AntAlmanac's numbers and websoc don't match up. We'd lose users' trust. I don't want people to have that thought in the back of their head - "is the data on AntAlmanac reliable?". They shouldn't  feel the need to check WebSoc to verify enrollment numbers. 

Sure they could refresh the page but that's not great ux. Calling the API isn't time expensive, so **each time they search for a class, it should be pulling the latest data from websoc.**

ChaseC99 (approved):
Great work Eric! This is an awesome feature that is really going to improve the user experience. I can't tell you how many times I've been annoyed having to re-enter my search after switching between tabs.

PR #242, opened 2022-02-14
Add mobile tab bar to switch between views

### LINKED ISSUE #125: Mobile site
Instead of adapting the desktop site to be mobile compatible with responsive design, I think we should design a fully mobile experience and deploy on m.antalmanac.com or something like that.

ChaseC99:
## Summary
AntAlmanac now works on mobile! 

When the screen size is less than 750px, a tab bar will appear on the bottom with calendar and search tabs. Users can switch between either the Calendar view or the right pane view.

mobile
_Note: the gif looks like it's off-center, but that's just because I misaligned the screen recording_

In this PR, I also created a new component called `Home` since `App` was starting to get bloated. There were several small css changes too.

## Test Plan
- On mobile, search for a class, add a class, view class on the map, view classes under "Added Classes", view class on the calendar, click on classes in the calendar
- On desktop, verify that the styling for the right pane did not create any regressions. Check class search, added classes, and map.

## Issues
Closes #125

## Future Followup
The more I look at our styling, the sadder I become. There's a lot of refactoring that needs to be done with our CSS. It's pretty inconsistent and hard to work with. 

I think we should look into standardizing the left pane calendar and right pane desktop tabs into a `View` component that had a menu bar and content. This would help us keep margins and paddings are consistent.

### DIFF
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -82,6 +82,6 @@
     </head>
     <body style="overflow: hidden">
         <noscript> You need to enable JavaScript to run this app. </noscript>
-        <div id="root"></div>
+        <div id="root" style="height: 100vh"></div>
     </body>
 </html>
--- a/client/src/components/App/App.js
+++ b/client/src/components/App/App.js
@@ -1,17 +1,12 @@
 import React, { PureComponent } from 'react';
-import { Grid, CssBaseline, createMuiTheme } from '@material-ui/core';
+import { createMuiTheme } from '@material-ui/core';
 import { ThemeProvider } from '@material-ui/core/styles';
-import Calendar from '../Calendar/ScheduleCalendar';
 import ReactGA from 'react-ga';
-import NotificationSnackbar from './NotificationSnackbar';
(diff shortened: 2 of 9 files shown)


Douglas-Hong (commented):
This looks really nice! The mobile version works exactly the same as the desktop version; downloading an .ics file/screenshot and saving/loading schedules still work normally.

For the custom events page, it looks a bit cluttered because the names for each day take a lot of space. Is it possible to just use the abbreviations of those days (Sun, Mon, Tue, etc.) on mobile?

CustomEvent

ChaseC99:
> Is it possible to just use the abbreviations of those days (Sun, Mon, Tue, etc.) on mobile?

Good idea! I've created #244 for that since I think it could be a good task for the fellowship students.

ChaseC99:
Any other changes you'd like to see? Or is this good to go?

Douglas-Hong (approved):
I have no other suggestions, it looks ready to go!

PR #245, opened 2022-02-16
Remove serverless

Voark:
## Summary
Update start script to work with serverless-less backend.

## Test Plan
Manually tested and checked if schedules still load and save.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
         "server-install": "cd ./antalmanac-backend && if test -f package.json; then npm install; fi",
         "dependency-install": "concurrently -n \"BACKEND,FRONTEND\" \"npm run server-install\" \"npm run client-install\"",
         "postinstall": "npm run dependency-install",
-        "server": "cd ./antalmanac-backend && npm start",
+        "server": "npm run start --prefix ./antalmanac-backend",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -7,7 +7,7 @@
         "server-install": "cd ./antalmanac-backend && if test -f package.json; then npm install; fi",
         "dependency-install": "concurrently -n \"BACKEND,FRONTEND\" \"npm run server-install\" \"npm run client-install\"",
         "postinstall": "npm run dependency-install",
-        "server": "cd antalmanac-backend && sls offline --stage development --noPrependStageInUrl",
+        "server": "cd ./antalmanac-backend && npm start",
```
ChaseC99:
```suggestion
        "server": "npm run start --prefix ./antalmanac-backend",
```

For consistently, let's use `--prefix` like we do with the frontend

ChaseC99 (changes requested):

package.json
Voark: Added, makes sense

ChaseC99 (approved):
Thanks for leading this initiative! Now that severless is gone, onboarding new devs will be so much easier 💯

PR #246, opened 2022-02-16
Fixed site-breaking error

EricPedley:
## Summary
The error came from using element.type.name as a key when mapping over lists of components. For some reason this works in dev but not in prod. The fix was using the index from map as the key instead.
## Test Plan
Made sure the client built successfully by running `npm run build` and serving with `serve -s build`
## Issues
None (but resolves the "Emergency Rollback" Discord thread)


ChaseC99 (approved):
Works for me too now! Thanks for the quick fix.

PR #256, opened 2022-02-18
Add Past Enrollment button that links to ZotTracker

### LINKED ISSUE #255: Add ZotTracker Button
Add a `CourseInfoButton` to `client/src/components/SectionTable/SectionTable.js` that redirects to ZotTracker. 

Call this button "Past Enrollment" and use a chart icon.

Douglas-Hong:
## Summary
- For each course, there is a Past Enrollment button that links to that course's enrollment data on ZotTracker. In ZotTracker, the user will be able to see the enrollment data of every lecture, lab, discussion, etc. for the most recent offering of the course. For example, CS 121 was most recently offered in Winter 2022, so the user will be directed to the data for Winter 2022 first.

## Test Plan
- The redirect link works on both desktop and mobile.
- In ZotTracker, the correct course's data is shown, and the user's search history on ZotTracker is updated correctly.
- The redirect link is a proper URL; if a department is more than one word, the redirect is successful because we encoded it correctly.

Demo:

https://user-images.githubusercontent.com/78244965/154623969-0d3decfb-7f2d-43fe-9793-00dfac0620df.mp4

## Issues
Closes #255 

## Future Followup:
- The Past Enrollment button takes up a lot of space on mobile, so we should probably improve the styling.

### DIFF
--- a/client/src/components/SectionTable/SectionTable.js
+++ b/client/src/components/SectionTable/SectionTable.js
@@ -16,6 +16,7 @@ import CourseInfoBar from './CourseInfoBar';
 import SectionTableBody from './SectionTableBody';
 import CourseInfoButton from './CourseInfoButton';
 import { Help, Assessment, Assignment } from '@material-ui/icons';
+import ShowChartIcon from '@material-ui/icons/ShowChart';
 import PropTypes from 'prop-types';
 
 const styles = {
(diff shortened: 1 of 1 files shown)


ChaseC99 (approved):
THIS IS HYPED

so excited to use it!

PR #262, opened 2022-02-28
Show delete button on course search page

### LINKED ISSUE #237: Show delete button in row after adding class
After clicking the plus button on the search results to add a class, the row should change to show the delete and change color buttons like there are on the added classes tab.

EricPedley:
## Summary
When you add a class from the course search page, the add button is replaced by the color and delete buttons.
antalmanac deletebutton
(gif generated with [ScreenToGif](https://www.screentogif.com/); I highly recommend it)
## Test Plan
Actions tested:
- add a class
- delete class from course search
- delete class from calendar
- delete class from added class tab
- change color from course search
- change color from added class tab
- change color from calendar (this doesn't update the color picker immediately in either course search or added classes, but this was a bug beforehand. I'm gonna create a separate issue for that)
- switch tabs, confirm that class stays added or deleted
## Issues
Closes #237 
## Future Followup (Optional)
Identified a new bug, going to create a new issue for that. (#263)


ChaseC99 (changes requested):
This change stacks the icons. Can you get them side-by-side like before?
| Before | After |
| --- | --- |
| image | image |

client/src/components/AddedCourses/CustomEventDetailView.js
```diff
@@ -40,15 +40,19 @@ const CustomEventDetailView = (props) => {
...
+                    return array[1] ? accumulator + 'Tue ' : accumulator + '';
                 case 3:
-                    return array[3] ? accumulator + 'Thu ' : accumulator + '';
+                    return array[2] ? accumulator + 'Wed ' : accumulator + '';
                 case 4:
+                    return array[3] ? accumulator + 'Thu ' : accumulator + '';
+                case 5:
                     return array[4] ? accumulator + 'Fri ' : accumulator + '';
+                case 6:
+                    return array[4] ? accumulator + 'Sat ' : accumulator + '';
```
ChaseC99: Sorry I don't quite understand what's going on here. 

Why is case 1 and 0 bother `array[0]`? And why are 5 and 6 both `array[4]`?
How does this relate to the color change?

client/src/components/App/ColorPicker.js
```diff
@@ -73,7 +74,7 @@ class ColorPicker extends PureComponent {
                         onChangeComplete={this.handleColorChangeComplete}
                     />
                 </Popover>
-            </span>
+            </Fragment>
```
ChaseC99: (see previous comment)
```suggestion
            </>
```

client/src/components/App/ColorPicker.js
```diff
@@ -43,9 +44,9 @@ class ColorPicker extends PureComponent {
 
     render() {
         return (
-            <span style={{ color: this.state.color }}>
+            <Fragment>
```
ChaseC99: Nit: use `<>` instead of fragment
This is just a shorthand for React.Fragment (literally does the same thing), but it is more concise and doesn't require an import.
```suggestion
            <>
```

client/src/components/App/ColorPicker.js
```diff
@@ -5,6 +5,7 @@ import { SketchPicker } from 'react-color';
 import { changeCourseColor, changeCustomEventColor } from '../../actions/AppStoreActions';
 import { ColorLens } from '@material-ui/icons';
 import ReactGA from 'react-ga';
+import { Fragment } from 'react';
```
ChaseC99: Unneeded import when we use `<>` instead
```suggestion

```

client/src/components/SectionTable/SectionTableButtons.js
```diff
@@ -0,0 +1,81 @@
...
+                >
+                    <Delete fontSize="small" />
+                </IconButton>
+                {<ColorPicker color={color} isCustomEvent={false} sectionCode={sectionCode} term={term} />}
+            </div>
+        </TableCell>
+    );
+});
+
+export const ScheduleAddCell = withStyles(styles)((props) => {
```
ChaseC99, line 43: Good call bringing this to separate file 👍

client/src/components/SectionTable/SectionTableButtons.js
```diff
@@ -0,0 +1,81 @@
...
+                        ReactGA.event({
+                            category: 'antalmanac-rewrite',
+                            action: 'Click Delete Course',
+                            label: 'Added Course pane',
+                        });
+                    }}
+                >
+                    <Delete fontSize="small" />
+                </IconButton>
+                {<ColorPicker color={color} isCustomEvent={false} sectionCode={sectionCode} term={term} />}
```
ChaseC99: Why is this wrapped in `{}`?

ChaseC99 (changes requested):

client/src/components/AddedCourses/CustomEventDetailView.js
EricPedley: It's not related to the color change, it's just a bug I found while testing my changes. And yeah I messed up the array indices 🤦‍♂️.

client/src/components/SectionTable/SectionTableButtons.js
EricPedley: don't know. removed the curly braces

client/src/components/App/ColorPicker.js
EricPedley: A bunch of other places in our code use Fragment, so I thought we were using it just to be more explicit. Should we open an issue to refactor all instances of Fragment to <>?

ChaseC99: Yes please!

EricPedley: ok im on it

ChaseC99 (approved):
This looks awesome! Once again, great work powering through all of the revisions. 

Sorry for the delay getting this approved, been busy with the end of the quarter 😅

PR #265, opened 2022-03-01
Fix install script to work on Windows

ChaseC99:
## Summary
The previous `server-install` command did not work on Windows.  

Credit to @falsidge to finding and fixing this: 
> this uses test which only works in linux
> so the project currently won't install the dependencies on windows 
> ```
> "client-install": "npm install ./client",
> "server-install": "npm install ./antalmanac-backend", 
> ```
> should work

## Test Plan
- Run `npm i` on a workspace **without** the backend recursively installed. Verify that the client still installs correctly and runs without the backened
- Run `npm i` on a workspace **with** the backend recursively installed. Verify that both install and run correctly.

## Issues
Update for #218 

## Future Followup 
If this user didn't recursively install the backend, this will (correctly) fail. However the error message doesn't help the user understand why it failed. It should tell them that they need to recursively install the backend.
image


ChaseC99:
Hmm for some reason the backend node_modules aren't getting created with these changes...

ChaseC99:
@falsidge I'm so confused as to why the backend `node_modules` don't get installed... can you try checking out `install-fix` as see if it works for you?

ChaseC99:
OH WAIT
This script actually works totally fine! The `node_modules` for both client and backend just get installed at the root level, under one single `node_modules` folder. 

You might need to delete you existing `node_modules` and `package-lock.json`s to get `npm i` working though

ChaseC99:
So now the question becomes ~ is it better to:
- Implement this PR, with all the `node_modules` in the root directory
OR
- Write a custom node script for installing dependencies that also checks if the file exists similar to how we had before

package.json
```diff
@@ -3,10 +3,11 @@
     "version": "1.0.0",
     "description": "antalmanac.com",
     "scripts": {
-        "client-install": "cd ./client && npm install",
-        "server-install": "cd ./antalmanac-backend && if test -f package.json; then npm install; fi",
+        "client-install": "cd client && npm install ./",
+        "server-install": "cd antalmanac-backend && npm install ./",
         "dependency-install": "concurrently -n \"BACKEND,FRONTEND\" \"npm run server-install\" \"npm run client-install\"",
         "postinstall": "npm run dependency-install",
+        "preinstall": "git clone https://github.com/icssc-projects/antalmanac-backend.git 2>NUL &",
```
ChaseC99: Wait actually I don't think we can do a clone here since it should be added as a submodule.

I'm just going to write a small node script to handle this. Should be simpler than messing around with different bash commands.

ChaseC99:
Turns out this was a super simple fix... we just needed to add a `--prefix` flag 🤦‍♂️ 
```
"client-install": "npm install --prefix ./client",
"server-install": "npm install --prefix ./antalmanac-backend",
```

This simplifies the install script and gracefully exits when the antalmanac-backend submodule wasn't recursively installed.

PR #282, opened 2022-03-11
Unit filter enhancements

### LINKED ISSUE #259: Unit filter doesn't take multiple inputs
The unit filter doesn't support multiple inputs ("4, 0"), it only take a single number.
image

- Change this input field to be a text field, so users can input multiple units. 
- Make sure the query to the api is correctly formatted

ecxyzzy:
## Summary

Add support for "VAR" and multiple inputs to the unit filter; for instance, an input of "4,2" will return all courses that have either 2 or 4 units, while "VAR" returns only the courses that have variable units. Regardless of the query type, these courses are sorted in ascending order of course code, just like any other query.

Note that this does diverge from the behavior of WebSOC, as its Units field seems to discard all non-numeric characters after the last numeric character encountered.

## Test Plan

Actions tested:

1. The input `VAR` on its own returns all courses with variable units.
2. Two numerical inputs `x` and `y` returns all courses with `x` or `y` units.
3. One numerical input `x` and the input `VAR` returns all courses with `x` units or variable units.
4. Test cases 2) and 3) with varying order return the same results.
5. All test cases detailed above, but with the inclusion of varying amounts of whitespace either before, within, or after the query, return the same result.

## Issues

Closes #259.

## Known issues/Caveats

* There are provisions in the backing code to return an empty response if any child responses from splitting the unit input were empty. However, since a full response is returned given gibberish in the `units` field and an otherwise valid request, and an empty response is returned given an invalid response, this failure mode is not exactly meaningful.
* Unrelated to the issue at hand: While running `npm install` the `package[-lock].json` of both the frontend and backend were modified for reasons unknown. I did not commit them because the functionality did not change in their absence; it may be due to the fact that I'm using WSL as my dev environment.

### DIFF
--- a/client/src/components/CoursePane/CourseRenderPane.js
+++ b/client/src/components/CoursePane/CourseRenderPane.js
@@ -10,7 +10,7 @@ import loadingGif from '../SearchForm/Gifs/loading.gif';
 import darkModeLoadingGif from '../SearchForm/Gifs/dark-loading.gif';
 import GeDataFetchProvider from '../SectionTable/GEDataFetchProvider';
 import LazyLoad from 'react-lazyload';
-import { queryWebsoc, isDarkMode } from '../../helpers';
+import { queryWebsoc, queryWebsocMultiple, isDarkMode } from '../../helpers';
 
 const styles = (theme) => ({
     course: {
@@ -148,7 +148,12 @@ class CourseRenderPane extends PureComponent {
             };
 
             try {
-                const jsonResp = await queryWebsoc(params);
+                let jsonResp;
+                if (params.units.includes(',')) {
+                    jsonResp = await queryWebsocMultiple(params, 'units');
+                } else {
+                    jsonResp = await queryWebsoc(params);
+                }
                 this.setState({
                     loading: false,
                     error: false,
--- a/client/src/components/SearchForm/AdvancedSearch.js
+++ b/client/src/components/SearchForm/AdvancedSearch.js
@@ -115,8 +115,8 @@ class AdvancedSearchTextFields extends PureComponent {
                     label="Units"
                     value={this.state.units}
                     onChange={this.handleChange('units')}
-                    type="number"
-                    helperText="ex. 3, 4, 1.7"
+                    type="search"
+                    helperText="ex. 3, 4, or VAR"
                     className={classes.units}
                 />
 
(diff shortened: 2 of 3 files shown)


ChaseC99:
First of all - great PR! 
Summary, Test Plan, etc. are all very well written. 

I really like the new feature of multiple inputs to the unit filter but I'm not sure about implementation.... The `queryWebSoc` function has become too bloated. 

It might be better to move the unit logic to a separate helper function or maybe even out to the [CourseRenderPane component](https://github.com/icssc-projects/AntAlmanac/blob/main/client/src/components/CoursePane/CourseRenderPane.js). I'm gonna take some more time to think about this but I'd love to hear your thoughts as well!

ecxyzzy:
The code for splitting and combining responses should definitely go in a helper function, yeah. That was an oversight on my part.

I didn't understand how the `CourseRenderPane` component was relevant at first but I think I get it now—would it be better to implement the code to check for multiple unit inputs somewhere in [`loadCourses()`](https://github.com/icssc-projects/AntAlmanac/blob/main/client/src/components/CoursePane/CourseRenderPane.js#L131-L164)? That way `queryWebSoc()` remains as it was, with only the logic necessary to do what its name suggests.

ChaseC99:
Yeah exactly. I'm still not sure which approach would be cleaner though. I'm open to either a helper function in queryWebSoc or loadCourses. Play around with both and let me know which way you think would be better!

ChaseC99:
> Unrelated to the issue at hand: While running npm install the package[-lock].json of both the frontend and backend were modified for reasons unknown. I did not commit them because the functionality did not change in their absence; it may be due to the fact that I'm using WSL as my dev environment.

Thanks for bringing this to my attention. I was running an older version of NPM. It should be all good now. Try merging in the latest version of `main`.

ecxyzzy:
I decided to go with something of a hybrid approach, adding helper functions to support making multiple queries in CourseRenderPane. Running through all the test cases I listed in the original PR didn't turn up any issues.

Also as I mentioned in the commit messages this approach could potentially allow for future expansion with comma-separated search options, so let me know what you think :)

ChaseC99 (approved):
I think your hybrid approach is a brilliant way to handle multiple queries! `queryWebsocMultiple` and `combineSOCObjects` have good separation of logic and leaves the door open for future feature expansion.

Great work on this PR 👏 
I'll get this deployed later tonight.

Issue #224, opened 2022-01-31
Bad contrast on some items in dark mode


EricPedley:
It's hard to see the arrow on the back button when looking at the class search results (see screenshot below)
![image](https://user-images.githubusercontent.com/48658337/151758207-a7cceef8-343f-4730-9096-69aa315e890c.png)
My roommate was using it for the first time and couldn't figure out how to get out of that screen cause the back button doesn't look like a back button.
Maybe we might add text saying "Back" next to the button and/or change the button icon for better contrast.

jon-pruessmeier:
Hey Eric,
I would like to work on it. Could you please assign me to it?

jbeanz:
I can do this, can you assign me?

ChaseC99:
Hi @jbeanz, are you a student at UC Irvine? The `good first task` issues are set aside for students who would like to learn more about web development.

jbeanz:
> Hi @jbeanz, are you a student at UC Irvine? The `good first task` issues are set aside for students who would like to learn more about web development.

I am not unfortunately :/ Can I still work on it? I ask because I'm a student preparing for job search.

ChaseC99:
I apologize but like I mentioned, we have these issues set aside for our students. Thank you for your interest in the project though!

EricPedley:
Renamed issue to be more general because I also want to point out the bad contrast on the load/save popups: 
![image](https://user-images.githubusercontent.com/48658337/155476245-076a9f8e-7880-49a4-b3d2-02a67e8009d6.png)
It's dark blue text against a black background, and with a blue light filter on it's basically invisible. Maybe make those into buttons with blue backgrounds and white text?

ChaseC99:
Adding onto this, when you hover over the Back button it goes slightly transparent and looks bad in both dark and light mode. We should change this behavior. 
<img width="103" alt="image" src="https://user-images.githubusercontent.com/29494270/158049578-c2b942a5-cd49-4318-b1ab-a99b13a966fa.png">

akins1:
I'll take this one!

elainelt:
I'll take this one! :)

devsdevsdevs:
Closing since merged

PR #287, opened 2022-03-13
Make class search back button float over results

ChaseC99:
## Summary
The Back Button now floats over the search results. 
back

## Test Plan
- Verify that it floats over content on both desktop and mobile
- Verify that the back button still works 

## Issues
N/A

## Future Followup
When you hover over the button, it goes slightly transparent and doesn't look very good. We should fix this in #224


Douglas-Hong (approved):
Looks good! I agree that the styling of the button should be improved. The button's arrow is barely visible on dark mode, and the button when hovered doesn't look good.

Issue #183, opened 2021-08-25
Improve developer onboarding experience


ChaseC99:
Goals:

- [x] Setting up and running AntAlmanac should be an extremely easy and fast process for new developers. 
- [x] Any developer should be able to run the project and work on the frontend without the backend .env file. 
- [x] Documentation should be publicly available

Know Issues:
- Backend server fails and throws many errors without `AA_MONGODB_URI` .env value
    - We shouldn't provide this key to everyone.
    - Should gracefully fail when missing, so terminal output doesn't become a warzone.
- Open source developers will not be able to access backend data like `news`, `loadUserData`, `users`, or `banners`.
    - These endpoints aren't necessary for most frontend development. However It may be annoying that they cannot save a schedule for local dev though.
    - For now, won't fix
- Open source developers can't access informational endpoints like `enrollmentData` and `websocapi`
    - This data can come from the PeterPortal API (#181 and #182)
- Notion pages are not publicly available

Other potential areas of improvement:

- [x] Simplify npm commands to install and run the project
- [ ] Create a `/scripts` directory so we can have better start and deploy processes
- [ ] Reduce # of npm packages [?]
- [ ] Make some notion docs public

ChaseC99:
#187 fixed an issue that prevented classes from loading due to the `/news` endpoint failing

EricPedley:
All the known issues are fixed now, besides the notion thing. I think by now a lot of the stuff on Notion is outdated or not useful for open-source developers so I'm just gonna close this issue.

PR #289, opened 2022-03-13
Remove backend

ChaseC99:
## Summary
Now that the backend is available at `dev.api.antalmanac.com`, we can use that instead of running it locally! This drastically simplifies the onboarding experience for new developers. Devs no longer need to recursively clone the project or have the .env file for the backend.

In this PR, we do the following:
- remove the backend submodule
- move the contents of `/client` into the root directory
- update package.json in the root directory to be for the website
- update the readme accordingly

## Test Plan
- Delete package-lock.json, delete /node_modules, and re-run `npm i`
- Run `npm start` and verify that that site loads on `localhost:3000`. 
- Verify that I can load and save schedules to the `dev.api.antalmanac.com` backend.
- Run `npm run deploy` and verify that deployment succeeds

## Issues
#183  

## Future Followup
- Maybe make it easier to hook up the backend with the frontend for local development? Right now it requires a code change but it would be nice if it was just a flag or env variable.


ChaseC99:
This is a big PR. In addition to testing the branch locally, the most important files to review are:
- package.json
- README.md

It might be easier to look at the commits inside of all of the changed files at once.

Voark (approved):
LGTM, tested locally and seems to work fine. Agreed on the followup, a quick and dirty solution could be to just maybe leave the old return in the function but commented. However I don't really see it being necessary anytime soon so we can probably just implement a better solution later.

PR #290, opened 2022-03-13
Changed Subject icon to ExpandLess and ExpandMore based on this.state…

### LINKED ISSUE #278: Change SchoolDeptCard component from a Subject icon to a drop down icon
Right now, we use a `Subject` icon for the [SchoolDeptCard component](https://github.com/icssc-projects/AntAlmanac/tree/main/client/src/components/CoursePane/SchoolDeptCard.js). This is ugly. 
image

Instead, we should show a down icon `v` to expand the comment and an up icon `^` to hide the comment.
- [ ] When `this.state.commentsOpen` is false, display an `ExpandMore` icon.
- [ ] When `this.state.commentsOpen` is true, display an `ExpandLess` icon.

_(You can probably use a conditional render statement for this `{ condition ? component if true : component if false }`)_

yialan6:
….commentsOpen.

## Summary
Replaced the Subject icon with ExpandLess/ExpandMore icons based on whether this.state.commentsOpen is T/F.
## Test Plan
Verified that the icons were replaced on antalmanac.com
## Issues
#278

Screen Shot 2022-03-13 at 5 55 55 AM
Screen Shot 2022-03-13 at 5 56 04 AM

### DIFF
--- a/client/src/components/CoursePane/SchoolDeptCard.js
+++ b/client/src/components/CoursePane/SchoolDeptCard.js
@@ -1,6 +1,6 @@
 import React, { PureComponent, Fragment } from 'react';
 import { Grid, Paper, Typography, withStyles, Collapse } from '@material-ui/core';
-import { Subject } from '@material-ui/icons';
+import { ExpandLess, ExpandMore } from '@material-ui/icons';
 
 const styles = (theme) => ({
     school: {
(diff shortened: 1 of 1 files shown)


ChaseC99 (approved):
This looks really good! Thanks for your contribution 🙌

PR #293, opened 2022-03-14
Custom Events header is now hidden if there aren't any events

### LINKED ISSUE #270: When there are no custom events, hide the "Custom Events" text on the Added Classes View
In this schedule I do not have any custom events. However at the bottom of the screen, I still see the "Custom Events" text. The "Custom Events" text should only be displayed if the schedule has custom events in it.

image

ethand605:
image

## Summary
Custom Events header is not hidden if there aren't any events

## Test Plan
Verified that the header is not shown without any events

## Issues
Closes #270

### DIFF
--- a/src/components/AddedCourses/AddedCoursePane.js
+++ b/src/components/AddedCourses/AddedCoursePane.js
@@ -101,6 +101,14 @@ class AddedCoursePane extends PureComponent {
         this.setState({ customEvents: AppStore.getCustomEvents() });
     };
 
+    custEventHeader = (props) => {
+        const hasEvents = props.hasEvents;
+        if (hasEvents) {
+            return <Typography variant="h6">Custom Events</Typography>;
+        }
+        return;
+    };
+
     getGrid = () => {
         return (
             <Fragment>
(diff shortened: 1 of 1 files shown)


src/components/AddedCourses/AddedCoursePane.js
```diff
@@ -173,7 +181,7 @@ class AddedCoursePane extends PureComponent {
                         </Grid>
                     );
                 })}
-                <Typography variant="h6">Custom Events</Typography>
+                <custEventHeader hasEvents={this.state.customEvents.length !== 0} />
```
ChaseC99: Good idea using `.length`! We can actually simplify this down to one line using a [conditional render](https://beta.reactjs.org/learn/conditional-rendering#logical-and-operator-). Here the Typography will only be displayed if the previous statement is true.
```suggestion
                {this.state.customEvents.length > 0 && <Typography variant="h6">Custom Events</Typography>}
```

ChaseC99 (changes requested):

ChaseC99 (approved):
Looks good! Congrats on your first AntAlmanac contribution 🎉

PR #294, opened 2022-03-14
Fix indexing for old custom events

### LINKED ISSUE #292: Old custom events are one day off
If the user created a custom event before [https://github.com/icssc-projects/AntAlmanac/pull/230](https://github.com/icssc-projects/AntAlmanac/pull/230), that custom event is now one day before the day it's supposed to be on. The following screenshot is a good example; the first custom event is supposed to be on Monday but it is now on Sunday. 

BugCustomEvent

Douglas-Hong:
## Summary
- Custom events that were added before #230 are now one day earlier than they're supposed to be. This is because Sunday is now the first day instead of Monday.
- To resolve this issue, we have to check if a custom event's `days` array is of length 5 (instead of 7) in `calenderizeHelpers.js`. If the array is of length 5, `dayIndex` should be incremented by 1 so that all the old custom events are shifted to the right day.
- Additionally, we have to make sure the custom events have the correct day on the Added Classes page. The Custom Events dialog for old custom events should also have the correct days checked/selected.

## Test Plan
- Added custom events using the code before #230. These custom events are now in the correct place after this bug fix.
- The Added Classes page correctly displays custom events.
- The Custom Events dialog is correct for old custom events.
- Adding/deleting custom events on the weekdays and weekend still works.
- Loading, saving, and screenshotting the schedule all work normally.

## Issues
Closes #292

### DIFF
--- a/src/components/AddedCourses/CustomEventDetailView.js
+++ b/src/components/AddedCourses/CustomEventDetailView.js
@@ -38,7 +38,9 @@ const CustomEventDetailView = (props) => {
         });
 
         const dayAbbreviations = ['Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat'];
-        const daysString = days.map((includeDate, index) => (includeDate ? dayAbbreviations[index] : '')).join(' ');
+        const daysString = days
+            .map((includeDate, index) => (includeDate ? dayAbbreviations[days.length === 7 ? index : index + 1] : ''))
+            .join(' ');
 
         return `${startTime.format('h:mm A')} — ${endTime.format('h:mm A')} • ${daysString}`;
     };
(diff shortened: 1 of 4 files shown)


ChaseC99 (approved):
LGTM! Good work fixing this.

ChaseC99:
Follow up question on this - 
If I understand the code correctly, for someone who had `days.length == 5` if they load and resave their schedule using this change, the next time they load, `days.length == 7` right?

Since future users won't have this problem and this fix will eventually resolve the issue for everyone from this quarter, does that mean next year we would be able to remove the extra logic for this fix to simplify the code?

Douglas-Hong:
Currently, if a custom event had `days.length == 5`, the user would have to edit the custom event and save that edit in order to get `days.length == 7`. I can definitely try to make it so that saving the schedule again will make it change to `days.length == 7`; that sounds much better.

We can definitely remove all the conditional logic after this quarter ends because all future custom events will have an array that supports the 7-day calendar.

ChaseC99:
> I can definitely try to make it so that saving the schedule again will make it change to days.length == 7; that sounds much better.

Since we can delete the conditional logic after this quarter, this is probably not worth the effort. I'll add an issue for this so we don't forgot!

PR #296, opened 2022-03-14
Fix ics download

### LINKED ISSUE #295: Download Button unresponsive
When clicked, download button does not prompt user to save .ics file.

image

ChaseC99:
## Summary
ics `createEvent` expects a list of ints.
Having year and day as strings caused it to break.

It's unclear why this broke now.
I'm assuming it's because package-lock was updated to a newer version of ics.

## Test Plan
- Download a schedule
- Verify that download succeeds and that the finals dates are correct.

## Issues
Closes #295

## Future Followup
We should add automated tests for ics downloads. This feature breaks fairly often.

### DIFF
--- a/src/components/Calendar/ExportCalendar.js
+++ b/src/components/Calendar/ExportCalendar.js
@@ -113,8 +113,8 @@ const getExamTime = (exam, year) => {
     const [examStartTime, examEndTime] = parseTimes(time);
 
     return [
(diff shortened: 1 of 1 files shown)


Douglas-Hong (approved):
Looks good! I downloaded some schedules and all the courses and finals are in the correct place!

PR #298, opened 2022-03-15
Simplify SOC updates

### LINKED ISSUE #227: Simplify process to update course term
Right now you have to update the current term in a few different places in the code.
Ideally, we update one spot and then everything else references that single source of truth.
Having one spot to update would also make this easier to update in the future.

ecxyzzy:
## Summary
Collect all term data used by AntAlmanac into `termData`, which exports an `Array` of `Term` objects and the index of the default `Term` (as per the default term in WebSOC's term dropdown). This way, updates for future quarters simply involves adding a new `Term` object to the front of the `termData.termData` array literal, and updating the `termData.defaultTerm` constant to the index of the new default term.

The `defaultTerm` constant is useful for cases where the first `Term` in the array is not necessarily the current or immediate next term. For instance, WebSOC's default term as of the time of writing is 2022 Spring, while the Schedules of Classes for all Summer Sessions have already been released and are ahead of 2022 Spring in the dropdown.

## Test Plan
1. The Term dropdown in the Class Search tab still works as it did prior to this change.
2. Downloading a schedule (with classes from non-Summer Session terms) as an `.ics` file still works as it did prior to this change.

## Issues
Closes #227.

## Future Followup (Optional)
There have been proposals regarding scraping in other projects (e.g. icssc-projects/peterportal-public-api#16), so I'm wondering whether it would be possible and/or useful to scrape the registrar page for future terms and relevant dates. This might be out of scope for AntAlmanac, but if implemented properly the terms wouldn't be needed to be added manually even to `termData`.

### DIFF
--- a/src/components/Calendar/ExportCalendar.js
+++ b/src/components/Calendar/ExportCalendar.js
@@ -6,22 +6,15 @@ import Today from '@material-ui/icons/Today';
 import { saveAs } from 'file-saver';
 import { createEvents } from 'ics';
 import AppStore from '../../stores/AppStore';
+import { termData } from '../../termData';
 
-// Hardcoded first mondays
-// You can find the start dates here: https://www.reg.uci.edu/calendars/academic/tenyr-19-29.html
-// Note: months are 0-indexed
 // TODO(chase): support summer sessions
-const quarterStartDates = {
-    '2019 Fall': [2019, 8, 26],
-    '2020 Winter': [2020, 0, 6],
-    '2020 Spring': [2020, 2, 30],
-    '2020 Fall': [2020, 9, 1],
-    '2021 Winter': [2021, 0, 4],
-    '2021 Spring': [2021, 2, 29],
-    '2021 Fall': [2021, 8, 23],
-    '2022 Winter': [2022, 0, 3],
-    '2022 Spring': [2022, 2, 28],
-};
+let quarterStartDates = {};
+for (const q of termData) {
+    if (!q.shortName.includes('Summer') && q.startDate) {
+        quarterStartDates[q.shortName] = q.startDate;
+    }
+}
 
 const daysOfWeek = ['Su', 'M', 'Tu', 'W', 'Th', 'F', 'Sa'];
 const daysOffset = { SU: -1, MO: 0, TU: 1, WE: 2, TH: 3, FR: 4, SA: 5 };
--- a/src/components/Calendar/ExportCalendar.js
+++ b/src/components/Calendar/ExportCalendar.js
@@ -9,12 +9,9 @@ import AppStore from '../../stores/AppStore';
 import { termData } from '../../termData';
 
 // TODO(chase): support summer sessions
-let quarterStartDates = {};
-for (const q of termData) {
-    if (!q.shortName.includes('Summer') && q.startDate) {
-        quarterStartDates[q.shortName] = q.startDate;
-    }
-}
+const quarterStartDates = termData
+    .filter((term) => !term.longName.includes('Summer') && term.startDate.length)
+    .reduce((prev, curr) => ({ ...prev, [curr.shortName]: curr.startDate }), {}); // https://stackoverflow.com/q/36388401
 
 const daysOfWeek = ['Su', 'M', 'Tu', 'W', 'Th', 'F', 'Sa'];
 const daysOffset = { SU: -1, MO: 0, TU: 1, WE: 2, TH: 3, FR: 4, SA: 5 };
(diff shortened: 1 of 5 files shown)


src/components/Calendar/ExportCalendar.js
```diff
@@ -6,22 +6,15 @@ import Today from '@material-ui/icons/Today';
...
-    '2021 Fall': [2021, 8, 23],
-    '2022 Winter': [2022, 0, 3],
-    '2022 Spring': [2022, 2, 28],
-};
+let quarterStartDates = {};
+for (const q of termData) {
+    if (!q.shortName.includes('Summer') && q.startDate) {
+        quarterStartDates[q.shortName] = q.startDate;
+    }
+}
```
ChaseC99: We can use filter and reduce here. 
(I wrote this code pretty quick, so idk if it compiles lol)
```suggestion
let quarterStartDates = termdata
    .filter(term => !q.shortName.includes('Summer') && q.startDate)
    .reduce((quarters, term) => quarters[term.shortName] = term.startDate, {});
```

src/components/SearchForm/TermSelector.js
```diff
@@ -34,45 +35,9 @@ class TermSelector extends PureComponent {
...
-                    <MenuItem value={'2015 Fall'}>2015 Fall Quarter</MenuItem>
-                    <MenuItem value={'2015 Summer2'}>2015 Summer Session 2</MenuItem>
-                    <MenuItem value={'2015 Summer10wk'}>2015 10-wk Summer</MenuItem>
-                    <MenuItem value={'2015 Summer1'}>2015 Summer Session 1</MenuItem>
-                    <MenuItem value={'2015 Spring'}>2015 Spring Quarter</MenuItem>
-                    <MenuItem value={'2015 Winter'}>2015 Winter Quarter</MenuItem>
-                    <MenuItem value={'2014 Fall'}>2014 Fall Quarter</MenuItem>
+                    {termData.map((q) => (
+                        <MenuItem value={q.shortName}>{q.longName}</MenuItem>
+                    ))}
```
ChaseC99: Nit: use `term` instead of `q`. I know `q` is just shorthand for quarter, but it's a little more clear to just write out the whole word `term`
```suggestion
                    {termData.map((term) => (
                        <MenuItem value={term.shortName}>{term.longName}</MenuItem>
                    ))}
```

src/termData.js
```diff
@@ -0,0 +1,64 @@
...
+    new Term('2015 Summer2', '2015 Summer Session 2'),
+    new Term('2015 Summer10wk', '2015 10-wk Summer'),
+    new Term('2015 Summer1', '2015 Summer Session 1'),
+    new Term('2015 Spring', '2015 Spring Quarter'),
+    new Term('2015 Winter', '2015 Winter Quarter'),
+    new Term('2014 Fall', '2014 Fall Quarter'),
+];
+
+// The index of the default term in termData, as per WebSOC
+const defaultTerm = 3;
```
ChaseC99: Nit: let's move this to the top of the file.

src/termData.js
```diff
@@ -0,0 +1,64 @@
+function Term(shortName, longName, startDate = []) {
+    this.shortName = shortName;
+    this.longName = longName;
+    this.startDate = startDate;
+    Object.seal(this);
```
ChaseC99: Is it necessary to seal?
I have no clue what the best practice is for stuff like this. Although we won't be modifying this object elsewhere, I tend to avoid added extra code if it's not needed.

src/components/SearchForm/TermSelector.js
ChaseC99: Side note: I'm REALLY stoked to not have to write out these long `<MenuItem>` elements every again haha

src/components/EnrollmentGraph/GraphModalContent.js
```diff
@@ -94,9 +95,13 @@ class GraphModalContent extends PureComponent {
...
-                        <MenuItem value={'2021 Winter'}>2021 Winter Quarter</MenuItem>
-                        <MenuItem value={'2021 Spring'}>2021 Spring Quarter</MenuItem>
-                        <MenuItem value={'2021 Fall'}>2021 Fall Quarter</MenuItem>
+                        {termData
+                            .slice(defaultTerm + 2, defaultTerm + 8)
+                            .filter((q) => !q.shortName.includes('Summer'))
+                            .reverse()
+                            .map((q) => (
+                                <MenuItem value={q.shortName}>{q.longName}</MenuItem>
+                            ))}
```
ChaseC99, line 104: Just so you know, the GraphModals are currently hidden from users. We are completely changing how we handle course enrollment graphs, so that this data and the available terms will be fetched from the PeterPortal API. Don't worry about `GraphModalContent.js` in your test plan.

ChaseC99 (changes requested):
I'm SO excited for this PR! Changing the term was always a major pain each quarter. 

I've just left a few nits. Once those are resolved, this will be good to go!

src/termData.js
ecxyzzy: StackOverflow is pretty scarce on the topic but it seems like `Object.seal` (and `Object.freeze`, which _should_ have been the method used here…) is really only used to implement "proper" immutability as found in other programming languages. Since we won't be modifying these objects elsewhere, it's probably fine to remove it.

ChaseC99 (approved):
PR looks great. Thanks so much for doing this! It will be really helpful for future quarters :D

PR #300, opened 2022-03-15
Replaced instances of Fragment with its shorthand version

### LINKED ISSUE #273: Replace instances of Fragment with <>
<> is a shorthand for Fragment and doesn't require an import. See https://reactjs.org/docs/fragments.html#short-syntax. Note that if a fragment is part of an array map, it needs to be explicitly a Fragment so it can get a key prop (https://reactjs.org/docs/fragments.html#keyed-fragments).

RandyHuynh5815:
## Summary
Replaced all instances of <Fragment></Fragment> with its shorthand syntax (<></>)

## Test Plan
- Verified that the syntax changes do not affect the overall performance of AntAlmanac

## Issues
Closes #273

### DIFF
--- a/src/components/App/NotificationSnackbar.js
+++ b/src/components/App/NotificationSnackbar.js
@@ -50,17 +50,15 @@ class NotificationSnackbar extends PureComponent {
     snackbarAction = (key) => {
         const { classes } = this.props;
         return (
-            <>
-                <IconButton
-                    key="close"
-                    color="inherit"
-                    onClick={() => {
-                        this.props.closeSnackbar(key);
-                    }}
-                >
-                    <CloseIcon className={classes.icon} />
-                </IconButton>
-            </>
+            <IconButton
+                key="close"
+                color="inherit"
+                onClick={() => {
+                    this.props.closeSnackbar(key);
+                }}
+            >
+                <CloseIcon className={classes.icon} />
+            </IconButton>
         );
     };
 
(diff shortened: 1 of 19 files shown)


src/components/App/NotificationSnackbar.js
```diff
@@ -51,7 +50,7 @@ class NotificationSnackbar extends PureComponent {
     snackbarAction = (key) => {
         const { classes } = this.props;
         return (
-            <Fragment>
+            <>
```
ChaseC99: We only need to use a `<>` when there is more than 1 element being returned. In this case, since it's just `<IconButton>`, we can remove the `<>`

This is because the return statement can only return one element. So if there multiple elements **next to each other**, we wrap them in `<>`. Otherwise it's not needed.
| 1 Element | Multiple Elements |
| --- | --- |
| Doesn't need `<>` | Needs `<>` |
| image | image |

src/components/Map/UCIMap.js
```diff
@@ -348,7 +348,7 @@ export default class UCIMap extends PureComponent {
 
     render() {
         return (
-            <Fragment>
+            <>
```
ChaseC99: Looks like you can remove it from here too since it's just one `<Map>` element
```suggestion
```

src/components/Notification.js
```diff
@@ -102,7 +101,7 @@ class CustomizedSnackbars extends React.Component {
 
     render() {
         return (
-            <Fragment>
+            <>
```
ChaseC99: Not needed here.
```suggestion
```

ChaseC99 (changes requested):
I highlighted a few areas where we actually didn't need Fragments/`<>`. Check if you can find any other spots where you can remove it. See my comment for more info.

ChaseC99 (approved):
Looks good to me, thanks for cleaning this up!

PR #301, opened 2022-03-16
Add refresh button to CourseRenderPane

### LINKED ISSUE #116: Implement refresh button

ChaseC99:
## Summary
Adds a refresh button to the Class Search page, so users can reload the latest data from WebSoc.
refresh

## Test Plan
- Search for a class
- Press the refresh button
    - verify that the component reloads
- Switch to "Added Class" tab and then back to "Class Search"
    - verify that the course data is still cached

## Issues
Close #116

### DIFF
--- a/src/components/CoursePane/CoursePaneButtonRow.js
+++ b/src/components/CoursePane/CoursePaneButtonRow.js
@@ -1,7 +1,7 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import { IconButton, Tooltip } from '@material-ui/core';
-import { ArrowBack } from '@material-ui/icons';
+import { ArrowBack, Refresh } from '@material-ui/icons';
 import { withStyles } from '@material-ui/core/styles';
 
 const styles = {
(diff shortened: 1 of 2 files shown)


EricPedley (changes requested):
Changes look mostly good. I feel like having an integer prop passed to CourseRenderPane to force re-render is a bit hack-ish, but probably good enough for now. I think the better way would be making the RightPaneStore emit an event when the cache is cleared, and for CourseRenderPane to listen and re-render itself in that case. I can create an issue for that after we merge these changes.

Since we're adding a refresh button, I think we should not clear the cache automatically when you go back to the search menu and do a different search. We just need to remove lines 3 and 58 from `RightPaneStore.js`. It makes the experience much nicer when you have to search a department again and it just loads instantly, plus I think people constantly looking for enrollment updates would be using the new refresh button instead of repeatedly doing the same search.

ChaseC99:
Hmm I still think we run into the same issue though... someone leaves a tab open overnight, comes back to search in the morning, and gets results that are different from WebSOC.

I would be open to caching if we also included a time limit on it. Maybe we add a time field in the cache so if a cached result is older than 30min, we ignore it and fetch from WebSOC again.

We can create a separate issue for this if you'd like to work on it!

EricPedley:
> Hmm I still think we run into the same issue though... someone leaves a tab open overnight, comes back to search in the morning, and gets results that are different from WebSOC.
> 
> I would be open to caching if we also included a time limit on it. Maybe we add a time field in the cache so if a cached result is older than 30min, we ignore it and fetch from WebSOC again.
> 
> We can create a separate issue for this if you'd like to work on it!

Yeah this makes a lot of sense. I'll make the issue.

EricPedley (approved):

Issue #279, opened 2022-03-10
Migrate to mui v5


ChaseC99:
We're currently on v4
https://mui.com/guides/migration-v4/

This will mess up the colors so whoever tackles this should also fix the theme after migrating in the same PR. The new colors don't necessarily need to be the same as it was before.

EricPedley:
While on the topic of MUI, whoever makes this change should also fix these MUI-related console warnings:
![image](https://user-images.githubusercontent.com/48658337/158688264-9809b70c-c680-43e7-accc-6e2c0af09fcd.png)
(index.js and App.js)
![image](https://user-images.githubusercontent.com/48658337/158688290-123ab672-aa2c-486a-a1ff-03d9132aa573.png)
(News.js line 145)

ecxyzzy:
Per the migration guide we'll also want to upgrade `react-scripts` to latest, which will resolve almost all of the deprecation warnings when running `npm i`. The stragglers get their own issues but are also relatively minor fixes.

PR #302, opened 2022-03-16
Import from Study List feature

### LINKED ISSUE #225: Add "import from study list" feature
For people who are migrating to antalmanac for the first time, it would be convenient to be able to copy/paste your entire study list into a form somewhere on the site and it automatically add all your classes. Like for example, you could highlight your study list like the following: 
screenshot of study list being highlighted
and paste it in. The pasted text would be in this form:
```
  33226 WRITING   39B SEM   Z  4.0  GR      T T   11:00-12:20   SSL  117       
   35670 I&C SCI    51 LEC   A  6.0  GR      T T   08:00-09:20   HSLH 100A      
   35671 I&C SCI    51 DIS 1A   0.0  GR mc    W    07:00-07:50pm HSLH 100A      
   35684 I&C SCI    51 LAB A4   0.0  GR      T T   02:00-03:20   TBA  TBA       
   36010 IN4MATX    43 LEC   A  4.0  GR      T T   12:30-01:50   HIB  100       
   36012 IN4MATX    43 DIS 2    0.0  GR         F  10:00-10:50   DBH  1300      
   37180 STATS      67 LEC B    4.0  GR mc   T T   05:00-06:20pm ICS  174       
   37182 STATS      67 DIS B2   0.0  GR mc      F  04:00-04:50   ICS  180    

```

Parsing out the course codes can just be done by running a regex on 5 digit numbers in the user input.

ecxyzzy:
## Summary

Allows new users to AntAlmanac to import their WebReg study list, which they can then save to the backend if they so choose.

Demo (you may need to open in new tab to view it in full resolution):
antalmanac-import-studylist
<sub>Incidentally, if anyone knows of a dark theme for WebReg please let me know…</sub>

## Test Plan
1. Leaving the textbox blank results in an error `Snackbar` telling the user that they cannot import an empty/invalid Study List.
2. Filling the textbox with gibberish results in the same `Snackbar` as above.
3. Pasting a properly formatted Study List as copy-pasted from WebReg, with the term selector in the right-hand pane set to the corresponding term, correctly imports the Study List into the currently focused Schedule.
4. Same as 3) but with the Study List from StudentAccess, which has a drastically different format compared to the Study List from WebReg, also correctly imports the Study List into the currently focused Schedule.

~~5. Any errors due to API calls failing (e.g. due to a term mismatch causing certain section codes to not exist) will result in a `Snackbar` notifying the user of an error , and the currently focused Schedule will be cleared. Any classes that were successfully added will remain added to the currently focused Schedule (possible caveat?).~~ Deprecated as of 577f946.

5. If a section cannot be added (e.g. due to a term mismatch causing that section code to not exist), the import process will continue silently. Upon conclusion of the process, a snackbar with the appropriate variant will be displayed, along with relevant information (success => all sections imported, warning => some sections imported, error => no sections imported).
6. Typing in any number of valid section codes for the current term, as long as they are delimited by a non-numeric character, correctly imports the courses as in 3) and 4). This is not really an expected usecase, but is useful if you just have a list of course codes for whatever reason, and works properly.

## Issues
Closes #225.

## Possible Enhancements before Merging

* Theming
  * Location: ~~Is the top bar appropriate for a feature like this?~~ Still provisional but has been moved into the overflow menu on mobile as of ad5efc2.
  * Icon: ~~Are there better icons than "download?" It might be a bit confusing.~~ Changed to `PostAdd` in a208b82 with the intent to update it to `ContentPasteGo` after migrating to mui v5.
* Error Handling: ~~The import can fail because of a multitude of reasons (API failure, mismatched term, empty/invalid input, etc). Do we want to tell the user what specifically went wrong, or is it okay to just say that it failed? If it comes down to user error, do we want to further delineate what exactly they did wrong?~~ Added logging to console in 3ee4ff9.
* Clearing Schedules: ~~This might not be such a good idea if (for example) the user tries to overwrite an existing schedule with a malformed Study List import, which would cause them to lose everything in that schedule. Maybe making a copy of the existing schedule and restoring it if something goes wrong would be better?~~ Removed in 3ee4ff9.
* Term Selection: ~~Add a drop-down in the dialog so users can select the term their Study List is from.~~ Implemented in 5857ce5.
* User Experience: Add a blurb to the dialog that tells the user where to actually find their Study List, as well as what and what not to copy.

Any thoughts and/or possible additional features are greatly appreciated.

### DIFF
--- a/src/components/App/CustomAppBar.js
+++ b/src/components/App/CustomAppBar.js
@@ -12,6 +12,7 @@ import { ReactComponent as MobileLogo } from './mobile-logo.svg';
 import News from './News';
 import AboutPage from './AboutPage';
 import ConditionalWrapper from './ConditionalWrapper';
+import ImportStudyList from './ImportStudyList';
 
 const styles = {
     appBar: {
@@ -46,6 +47,8 @@ const CustomAppBar = (props) => {
                 {isMobileScreen ? <MobileLogo height={32} /> : <Logo height={32} />}
 
                 <div style={{ flexGrow: '1' }} />
+
+                <ImportStudyList />
 
                 <LoadSaveScheduleFunctionality />
 
--- a/src/components/App/CustomAppBar.js
+++ b/src/components/App/CustomAppBar.js
@@ -48,8 +48,6 @@ const CustomAppBar = (props) => {
 
                 <div style={{ flexGrow: '1' }} />
 
-                <ImportStudyList />
-
                 <LoadSaveScheduleFunctionality />
 
                 <ConditionalWrapper
@@ -66,6 +64,7 @@ const CustomAppBar = (props) => {
                     {[
                         <SettingsMenu />,
                         <NotificationHub />,
+                        <ImportStudyList />,
                         <Tooltip title="Give Us Feedback!">
                             <Button
                                 onClick={() => {
--- a/src/components/App/ImportStudyList.js
+++ b/src/components/App/ImportStudyList.js
@@ -17,6 +17,13 @@ import { termData } from '../../termData';
 import MenuItem from '@material-ui/core/MenuItem';
 import Select from '@material-ui/core/Select';
 import InputLabel from '@material-ui/core/InputLabel';
+import { withStyles } from '@material-ui/core/styles';
+
+const styles = {
+    input: {
+        'margin-top': '10px',
+    },
+};
 
 class ImportStudyList extends PureComponent {
     state = {
@@ -119,6 +126,7 @@ class ImportStudyList extends PureComponent {
     }
 
     render() {
+        const { classes } = this.props;
         return (
             <>
                 {/* TODO after mui v5 migration: change icon to ContentPasteGo */}
@@ -133,26 +141,30 @@ class ImportStudyList extends PureComponent {
                             the right term selected below.
                         </DialogContentText>
                         {/* TODO refactor to use a modified TermSelector */}
-                        <InputLabel>Term</InputLabel>
-                        <Select value={this.state.selectedTerm} onChange={this.handleChange}>
-                            {termData.map((term, index) => (
-                                <MenuItem key={index} value={term.shortName}>
-                                    {term.longName}
-                                </MenuItem>
-                            ))}
-                        </Select>
+                        <div className={classes.input}>
+                            <InputLabel>Term</InputLabel>
+                            <Select value={this.state.selectedTerm} onChange={this.handleChange}>
+                                {termData.map((term, index) => (
+                                    <MenuItem key={index} value={term.shortName}>
+                                        {term.longName}
+                                    </MenuItem>
+                                ))}
+                            </Select>
+                        </div>
                         <br />
-                        <InputLabel>Study List</InputLabel>
-                        <TextField
-                            autoFocus
-                            fullWidth
-                            multiline
-                            margin="dense"
-                            type="text"
-                            placeholder="Paste here"
-                            value={this.state.studyListText}
-                            onChange={(event) => this.setState({ studyListText: event.target.value })}
-                        />
+                        <div className={classes.input}>
+                            <InputLabel>Study List</InputLabel>
+                            <TextField
+                                autoFocus
+                                fullWidth
+                                multiline
+                                margin="dense"
+                                type="text"
+                                placeholder="Paste here"
+                                value={this.state.studyListText}
+                                onChange={(event) => this.setState({ studyListText: event.target.value })}
+                            />
+                        </div>
                     </DialogContent>
                     <DialogActions>
                         <Button onClick={() => this.handleClose(false)} color="primary">
@@ -168,4 +180,4 @@ class ImportStudyList extends PureComponent {
     }
 }
 
-export default ImportStudyList;
+export default withStyles(styles)(ImportStudyList);
--- a/src/components/App/ImportStudyList.js
+++ b/src/components/App/ImportStudyList.js
@@ -49,7 +49,6 @@ class ImportStudyList extends PureComponent {
         this.setState({ isOpen: false }, async () => {
             document.removeEventListener('keydown', this.enterEvent, false);
             if (doImport) {
-                document.removeEventListener('keydown', this.enterEvent, false);
                 const sectionCodes = this.state.studyListText.match(/\d{5}/g);
                 if (!sectionCodes) {
                     openSnackbar('error', 'Cannot import an empty/invalid Study List.');
--- a/src/components/App/ImportStudyList.js
+++ b/src/components/App/ImportStudyList.js
@@ -1,0 +1,122 @@
+import React, { PureComponent } from 'react';
+import {
+    Button,
+    Dialog,
+    DialogActions,
+    DialogContent,
+    DialogContentText,
+    DialogTitle,
+    TextField,
+} from '@material-ui/core';
+import { queryWebsoc } from '../../helpers';
+import store from '../../stores/RightPaneStore';
+import { addCourse, clearSchedules, openSnackbar } from '../../actions/AppStoreActions';
+import AppStore from '../../stores/AppStore';
+import { GetApp } from '@material-ui/icons';
+
+class ImportStudyList extends PureComponent {
+    state = {
+        isOpen: false,
+        selectedTerm: null,
+        studyListText: '',
+    };
+
+    handleError = (index) => {
+        clearSchedules([index]);
+        openSnackbar('error', 'An error occurred while trying to import the Study List.');
+    };
+
+    handleOpen = () => {
+        this.setState({ isOpen: true, selectedTerm: store.getFormData().term });
+    };
+
+    handleClose = (doImport) => {
+        this.setState({ isOpen: false }, () => {
+            document.removeEventListener('keydown', this.enterEvent, false);
+            this.setState({ studyListText: '' });
+        });
+        if (doImport) {
+            document.removeEventListener('keydown', this.enterEvent, false);
+            if (!(this.state.studyListText && this.state.studyListText.match(/\d{5}/g))) {
+                openSnackbar('error', 'Cannot import an empty/invalid Study List.');
+                return;
+            }
+            const currIndex = AppStore.getCurrentScheduleIndex();
+            try {
+                Promise.all(
+                    this.state.studyListText
+                        .split('\n')
+                        .map((line) => line.match(/\d{5}/g))
+                        .filter((id) => id)
+                        .flat()
+                        .map((sectionCode) => queryWebsoc({ term: this.state.selectedTerm, sectionCodes: sectionCode }))
+                )
+                    .then((r) => {
+                        r.forEach((response) => {
+                            const courseData = response.schools[0].departments[0].courses[0];
+                            addCourse(courseData.sections[0], courseData, this.state.selectedTerm, currIndex);
+                        });
+
+                        openSnackbar('success', 'Study List successfully imported!');
+                    })
+                    .catch(() => this.handleError(currIndex));
+            } catch (e) {
+                this.handleError(currIndex);
+            }
+        }
+    };
+
+    enterEvent = (event) => {
+        const charCode = event.which ? event.which : event.keyCode;
+        if (charCode === 13 || charCode === 10) {
+            event.preventDefault();
+            this.handleClose(true);
+        }
+    };
+
+    componentDidUpdate(prevProps, prevState, snapshot) {
+        if (!prevState.isOpen && this.state.isOpen) {
+            document.addEventListener('keydown', this.enterEvent, false);
+        } else if (prevState.isOpen && !this.state.isOpen) {
+            document.removeEventListener('keydown', this.enterEvent, false);
+        }
+    }
+
+    render() {
+        return (
+            <>
+                <Button onClick={this.handleOpen} color="inherit" startIcon={<GetApp />}>
+                    Import
+                </Button>
+                <Dialog open={this.state.isOpen}>
+                    <DialogTitle>Import Schedule</DialogTitle>
+                    <DialogContent>
+                        <DialogContentText>
+                            Paste the contents of your Study List below to import it into AntAlmanac.
+                        </DialogContentText>
+                        <TextField
+                            autoFocus
+                            fullWidth
+                            multiline
+                            margin="dense"
+                            type="text"
+                            placeholder="Paste here"
+                            value={this.state.studyListText}
+                            onChange={(event) => this.setState({ studyListText: event.target.value })}
+                        />
+                    </DialogContent>
+                    <DialogActions>
+                        <Button onClick={() => this.handleClose(false)} color="primary">
+                            Cancel
+                        </Button>
+                        <Button onClick={() => this.handleClose(true)} color="primary">
+                            Import
+                        </Button>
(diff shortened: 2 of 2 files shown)


src/components/App/CustomAppBar.js
```diff
@@ -47,6 +48,8 @@ const CustomAppBar = (props) => {
 
                 <div style={{ flexGrow: '1' }} />
 
+                <ImportStudyList />
+
```
ChaseC99: Still not sure if the AppBar is the best place for this, but for now I think it should probably be the 3rd item in the list after `Load`

ChaseC99 (commented):
This is a sweet new feature that I am 100% going to use!

I left a few comments but I'm going to defer the final review to @EricPedley since he initially created the issue.

My only other concern is that users might not know what to copy from WebReg. You have a good description for what to do, but users need a LOT of hand-holding LOL. 

Potential future follow: maybe we should make a "Tips" box above the Google Analytics privacy policy that tells users about features they might not know about (adding custom events, signing up for notifications, exporting to Google Calendar, etc., and, of course, importing from WebReg) #303

ChaseC99:
> Error Handling: The import can fail because of a multitude of reasons (API failure, mismatched term, empty/invalid input, etc).

I think the way you handle it right now is good! You should also **log the error to the console** though, so more advanced users (and us) can debug.

For mismatched terms, I think the way that it displays a snackbar warning to the user is good enough. 
But maybe we should add a dropdown box inside the dialog (preset to the defaultTerm) to let them choose their term. 

> Clearing Schedules: This might not be such a good idea

I don't think we should clear the schedule. The way it adds the classes to the existing schedule right now is good.


Side note ~ this code also works when you copy the study list from student access: https://www.reg.uci.edu/access/student/studylist/?seg=U <- NICE!

ChaseC99:
> Incidentally, if anyone knows of a dark theme for WebReg please let me know…

@ryanmohta ? 👀

src/components/App/ImportStudyList.js
```diff
@@ -0,0 +1,122 @@
...
+            }
+            const currIndex = AppStore.getCurrentScheduleIndex();
+            try {
+                Promise.all(
+                    this.state.studyListText
+                        .split('\n')
+                        .map((line) => line.match(/\d{5}/g))
+                        .filter((id) => id)
+                        .flat()
+                        .map((sectionCode) => queryWebsoc({ term: this.state.selectedTerm, sectionCodes: sectionCode }))
```
devsdevsdevs: Instead of querying websoc/PP API for each section code, since a studylist will have all courses in the same term, you can build a list of section codes and submit them in a single request. Keep in mind though, this will limit you to ~8 section codes (last time that I checked), so chunk your codes into groups of 8 and submit them in more requests if you need to.

src/components/App/ImportStudyList.js
```diff
@@ -0,0 +1,122 @@
...
+    handleError = (index) => {
+        clearSchedules([index]);
+        openSnackbar('error', 'An error occurred while trying to import the Study List.');
+    };
+
+    handleOpen = () => {
+        this.setState({ isOpen: true, selectedTerm: store.getFormData().term });
+    };
+
+    handleClose = (doImport) => {
```
devsdevsdevs, line 48: Would be good to make this function async to avoid .then etc.

devsdevsdevs (changes requested):

devsdevsdevs:
Really neat feature 👍

ecxyzzy:
> My only other concern is that users might not know what to copy from WebReg. You have a good description for what to do, but users need a LOT of hand-holding LOL.

Yeah I figured that was the case. I'll take a crack at writing up a short guide that can fit in the pop-up, but that will have to wait until WebReg comes back up since I don't remember off the top of my head how to get there exactly.

src/components/App/ImportStudyList.js
ecxyzzy: Not exactly sure I follow—the reason I used `.then` and `.catch` on `Promise.all` was to run multiple queries in parallel, and catch any errors from them. Is there a cleaner way to write this if the function was made `async`? Even with the section code chunking you suggested (great idea by the way) there are still cases where multiple queries will have to be made.

devsdevsdevs: It'd just be a simple refactor,  so you'd do await on the Promise.all instead of chaining it with .then. I think we should use just promote using async instead of promise chaining because it looks cleaner, functionally it's identical.

ecxyzzy: Gotcha.

Yeah, async-await definitely looks cleaner now that I've figured out how it works. Thanks for the suggestions :D

ChaseC99:
Something really cool I realized about this - you can literally just type in a few random course codes on one line and they still load into the schedule! Doesn't need to match the format of the study list. Good job on that regex.

src/components/App/ImportStudyList.js
```diff
@@ -0,0 +1,168 @@
...
+                }
+            } catch (e) {
+                this.handleError(e);
+            }
+        }
+    };
+
+    enterEvent = (event) => {
+        const charCode = event.which ? event.which : event.keyCode;
+        if (charCode === 13 || charCode === 10) {
```
EricPedley, line 108: Can you add a comment for what these charCodes are? I figured out that 13 is enter but in the first two references I looked at I couldn't find what charCode 10 is and had to use String.fromCharCode to figure out it was newline.

src/components/App/ImportStudyList.js
```diff
@@ -0,0 +1,168 @@
...
+        if (doImport) {
+            document.removeEventListener('keydown', this.enterEvent, false);
+            if (!this.state.studyListText.match(/\d{5}/g)) {
+                openSnackbar('error', 'Cannot import an empty/invalid Study List.');
+                return;
+            }
+            const currSchedule = AppStore.getCurrentScheduleIndex();
+            const sectionCodes = this.state.studyListText
+                .split('\n')
+                .map((line) => line.match(/\d{5}/g))
```
EricPedley, line 55: You can do this matching without splitting the string and mapping each line. Running match on the entire study list string should do the trick. See this screenshot:
(running a regex match in the console to illustrate the point)
Also I think the filter is unnecessary, and also if you run the match on the entire string isntead of each line the flatten shouldn't be needed either.

src/components/App/ImportStudyList.js
```diff
@@ -0,0 +1,168 @@
...
+                .filter((id) => id)
+                .flat();
+            let sectionsAdded = 0;
+            try {
+                (
+                    await Promise.all(
+                        sectionCodes
+                            .reduce((result, item, index) => {
+                                // WebSOC queries can have a maximum of 8 course codes in tandem
```
EricPedley: Where did you find this info or how did you figure this out? I tried changing the chunk size to higher numbers when testing your code and it still seems to work.

src/components/App/ImportStudyList.js
```diff
@@ -0,0 +1,168 @@
...
+                                result[chunkIndex] ? result[chunkIndex].push(item) : (result[chunkIndex] = [item]);
+                                return result;
+                            }, []) // https://stackoverflow.com/a/37826698
+                            .map((sectionCode) =>
+                                queryWebsoc({ term: this.state.selectedTerm, sectionCodes: sectionCode.join(',') })
+                            )
+                    )
+                ).forEach((response) => {
+                    response.schools
+                        .map((school) => school.departments)
```
EricPedley: Since you aren't returning anything here, using repeated map/flatten is more confusing than just using normal loops. I recommend refactoring it for readability. See this example of going through the websoc data from `helpers.js`:
```
for (const school of jsonResp.schools) {
      for (const department of school.departments) {
          for (const course of department.courses) {
                  addedCourses.push({
                      ...sectionCodeToInfoMapping[section.sectionCode],
                      deptCode: department.deptCode,
                      courseNumber: course.courseNumber,
                      courseComment: course.courseComment,
                      prerequisiteLink: course.prerequisiteLink,
                      section: section,
                  });
              }
          }
      }
  }

```
Also since we're having to do this twice, it would be nice to write a function to encapsulate extracting the individual course info from a websoc query. You can do this yourself or create an issue for it if you want.

src/components/App/ImportStudyList.js
EricPedley: You can do this matching without splitting the string and mapping each line. Running match on the entire study list string should do the trick. See this screenshot:
(running a regex match in the console to illustrate the point)
Also I think the filter is unnecessary, and also if you run the match on the entire string isntead of each line the flatten shouldn't be needed either.

src/components/App/ImportStudyList.js
EricPedley, line 113: Can you add a comment for what these charCodes are? I figured out that 13 is enter but in the first two references I looked at I couldn't find what charCode 10 is and had to use String.fromCharCode to figure out it was newline.

EricPedley (changes requested):
Nice job! Looks really good functionally but there are some refactors that would make the code cleaner / more readable. Feel free to push back on any of these suggestions if you think the way you did it was better (but explain why).

EricPedley:
Also, if you want a dark theme for WebReg check out the [Dark Reader](https://darkreader.org/) browser extension.

EricPedley:
> Icon: Are there better icons than "download?" It might be a bit confusing.

The list of all material icons is here: https://mui.com/components/material-icons/
Some that might be more appropriate than "download":
- ContentPaste
- ContentPasteGo
- ImportExport
- Upload
- CloudUpload
- PostAdd
- AddBox

ChaseC99:
> The list of all material icons is here: https://mui.com/components/material-icons/

Note: we actually use v4 of mui, so the list of available icons is slightly smaller: https://v4.mui.com/components/material-icons/

src/components/App/ImportStudyList.js
```diff
@@ -0,0 +1,168 @@
...
+                <Dialog open={this.state.isOpen}>
+                    <DialogTitle>Import Schedule</DialogTitle>
+                    <DialogContent>
+                        <DialogContentText>
+                            Paste the contents of your Study List below to import it into AntAlmanac. Make sure you have
+                            the right term selected below.
+                        </DialogContentText>
+                        <InputLabel>Term</InputLabel>
+                        <Select value={this.state.selectedTerm} onChange={this.handleChange}>
+                            {termData.map((term) => (
```
EricPedley: This causes a console warning because each MenuItem has no unique key prop. Get the index param from map and pass it to each menu item as a key prop (See `CustomAppBar.js` lines 81-83 for a good example of this).

EricPedley: The same problem also exists at TermSelector.js line 38. And now that I see that file, I'm thinking it might be appropriate to use that here instead of a new Select component? I can see that just dropping it in wouldn't work, but you could maybe modify TermSelector to take an onchange function as a prop and use that. (i.e. replace updateFormValue with the function passed as a prop).

src/components/App/ImportStudyList.js
ecxyzzy: devsdevsdevs mentioned it [here](https://github.com/icssc-projects/AntAlmanac/pull/302#discussion_r827536026) and it seemed to be correct based on initial testing; I did some more testing with the REST API and it appears the limit is actually 10. Thanks for pointing that out—I'll address it.

src/components/App/ImportStudyList.js
ecxyzzy: Will do. Is it best practice to make a separate branch/PR for new functionality separate from this feature (e.g. this and refactoring `TermSelector` like you mentioned [here](https://github.com/icssc-projects/AntAlmanac/pull/302#discussion_r828435583)), or is it fine to just commit to this branch?

EricPedley: Normally I just add it to the same branch. Not sure if that is best practice though. @ChaseC99 what are your thoughts on when to make an entirely new branch vs just lumping a change in with an existing branch that isn't entirely related?

ecxyzzy:
Of the icons mentioned [here](https://github.com/icssc-projects/AntAlmanac/pull/302#issuecomment-1069563348) only the last five seem to be available in mui v4, which is a shame because I think `ContentPasteGo` fits the bill perfectly. `PostAdd` seems like a decent replacement though—thoughts?

src/components/App/ImportStudyList.js
ChaseC99: I usually try to keep my PRs as small as possible, so it's easier for others to review.

In this case, let's create a new issue, leave a todo comment in the code, and immediately open a new PR after merging this one.

That helper function will affect core functionality, so it'd probably be best to test it thoroughly in a separate PR.

ChaseC99:
> . . . I think `ContentPasteGo` fits the bill perfectly. `PostAdd` seems like a decent replacement though—thoughts?

I agree :( Let's go with PostAdd for now. In 
#279, leave a comment about switching it to ContentPasteGo after we migrate to v5

src/components/App/ImportStudyList.js
ecxyzzy: Done. I've also created issues #308 and #309 for this and the TermSelector issue Eric mentioned below.

ChaseC99 (changes requested):
Sorry for all the revision request but could you add some top/bottom margin in between term and the input field? They're a little too close right now.
image

ecxyzzy:
Also, I did some more testing while adding the margins, and I think this PR might need a bit longer to simmer. Here were the main issues I encountered:

* ~~If the user hits enter/clicks import immediately after pasting, then `setState` might not have fired yet, so the importing process will naturally fail because it's operating on an empty text buffer. Is there a clean way to ensure that `setState` has actually fired?~~ Fixed in c2d3943 by moving it into the callback for the `setState` in `handleClose`, not really sure why I wrote all the code outside in the first place.
* Since we're calling `addCourse` once for each course, if the user imports a study list from one quarter but already has course(s) on their schedule from another, then the warning snackbar will fire once per course they imported, which is a lot of spam/visual clutter. Not really sure what the cleanest way around this is short of touching that function.
* ~~There's also the blurb to point users to where the study list actually is, which I can't write yet because WebReg is "down"—it shows the study list as soon as you log in, which is not the case when students can register for classes.~~ Done in 80d8d91.

ecxyzzy:
I think this feature is just about done so I'm requesting a review now; the wrapper for `addCourse` to prevent multiple snackbars firing will likely go in another PR since it's not directly relevant, unless that's something that y'all want to see in this one.

Apologies Eric if that notified you twice—not really sure why GitHub decided to cancel my review request the first time.

ChaseC99:
Functionality wise, this is awesome! I just used it on my own schedule and it worked great :D 
I'm going to wait for @devsdevsdevs or @EricPedley to approve though since they previously requested code changes.

>  the wrapper for addCourse to prevent multiple snackbars firing will likely go in another PR since it's not directly relevant, unless that's something that y'all want to see in this one.

Sounds good to me. I want to rework how we handle that warning anyways (for example it shouldn't fire on empty or cleared schedules #319)

EricPedley (approved):

devsdevsdevs (approved):

ChaseC99:
Deployed! I've also added an announcement under the News icon.

Fantastic work here 💯 The PR had a ton of comments but you did a great job responding to them and resolving all concerns.
Really looking forward to this feature :D

PR #307, opened 2022-03-17
Made enter button submit search form in class search

### LINKED ISSUE #261: Enter button should submit search form
When the user is on the class search page, pressing the `enter` or `return` button should trigger the search
image

sravtall:
## Summary
I made the search button work with the enter key. The file edited is SearchForm.js.
Note: In the demo, there is no panel or anything on the side. Just emphasizing that my mouse is away.

https://user-images.githubusercontent.com/97429892/158724441-0042f3ad-57e9-427f-aff1-d6ee762ceed4.mp4

## Test Plan
Verified it by running it locally through localhost (npm)

## Issues
Closes "Enter button should submit search form" #261

### DIFF
--- a/src/components/SearchForm/DeptSearchBar/DeptSearchBar.js
+++ b/src/components/SearchForm/DeptSearchBar/DeptSearchBar.js
@@ -98,7 +98,6 @@ class DeptSearchBar extends PureComponent {
                     openOnFocus={true}
                     getOptionSelected={this.compareValues}
                     getOptionLabel={(option) => option.deptLabel}
-                    // blurOnSelect
                     onChange={this.handleSetDept}
                     includeInputInList={true}
                     noOptionsText="No departments match the search"
--- a/src/components/SearchForm/DeptSearchBar/DeptSearchBar.js
+++ b/src/components/SearchForm/DeptSearchBar/DeptSearchBar.js
@@ -98,7 +98,7 @@ class DeptSearchBar extends PureComponent {
                     openOnFocus={true}
                     getOptionSelected={this.compareValues}
                     getOptionLabel={(option) => option.deptLabel}
-                    blurOnSelect
+                    // blurOnSelect
                     onChange={this.handleSetDept}
                     includeInputInList={true}
                     noOptionsText="No departments match the search"
--- a/src/components/SearchForm/SearchForm.js
+++ b/src/components/SearchForm/SearchForm.js
@@ -42,23 +42,6 @@ const styles = {
 };
 
 class SearchForm extends PureComponent {
-    /*
-    componentDidMount = () => {
-        this.form.addEventListener('keydown', this.enterEvent);
-    };
-
-    componentWillUnmount = () => {
-        this.form.removeEventListener('keydown', this.enterEvent);
-    };
-
-    enterEvent = (event) => {
-        if (event.code === 'Enter' || event.key === 'Enter') {
-            console.log("Enter is clicked")
-            event.preventDefault();
-            this.props.toggleSearch();
-        }
-    };
-    */
     onFormSubmit = (event) => {
         event.preventDefault();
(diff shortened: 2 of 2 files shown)


ChaseC99 (changes requested):
Right now if you press "enter" at any other time, like loading a schedule, it triggers a search. 

Search should only happen when the form is in focus. I'm not sure what's the best way to handle this is... 

My initially thought is that you could get rid of `enterEvent`, wrap everything in a `<form>`, and then add a `type="submit"` prop to the Search button. But this will mess up the formatting and might require a larger refactor.

sravtall:
I just pushed some changes, but there are still some bugs involved in it. I have sort of been stuck for a while, so I put these changes out there to see if you possibly have an idea of resolving it.

I ended up making two possible solutions to resolve the issue at hand. One is using `eventListener` and the other is wrapping the code in a `<form>`. Each has there own unique issue.

#### `eventListener` approach through `enterEvent` and `ref` (this code should be commented)
I was able to address this by changing the scope from the entire document to a singular component. The component ended up being a `div` that contained all of the parts of the `SearchForm`.  (Used a `ref`)

After input is given and the enter is pressed, the website proceeds to look for the classes as expected. (All this only happens when the focus is on the form).

The issue comes with the `<Autocomplete>` component in the DeptSearchBar.js file. This solution seems to override any input with the search bar and proceed with the search without autocompleting the prompt. For instance, if I type "comp", I would hopefully expect "COMPSCI" to fill in. But, it proceeds to look for classes with "COMP", ultimately leading to no classes being found. The `<Autocomplete>` works if I select the option and then return.

#### Wrapping the code in `<form>` (this code is not commented, so active)
Surprisingly, wrapping the code this way does not mess up any of the formattings. I  had to add the `<form>` tags and the relevant `onSubmit` code.

After input is given and enter is pressed, the website only looks for classes if the element altered or inputted is not a  `<Select>`. This would include components like `GESelector` and `TermSelector`.

For example, I select "GE 1a" and proceed to click enter, the form does not submit as the options in the `<Select>` tab pops back up. In this case, at least the `<Autocomplete>` component works as intended.

---
I'm not really sure which method would be preferred here. I still keep working on it, but if you have any ideas please let me know. Sorry for the inconvenience. 

Also, side note, I commented out the `blurOnSelect` prop in the `<Autocomplete>` component as only then does clicking enter start a search for the "Department" section. Not sure if this is supposed to work like this or not.

ChaseC99 (changes requested):
Dang nice working getting it working with `<form>`!! I think this is def the best approach. 

> the website only looks for classes if the element altered or inputted is not a `<Select>`. This would include components like GESelector and TermSelector.  

Honestly I think it's actually **better** this way. When I click enter on a dropdown, I expect a list of options, not a search to be triggered. Department and the input fields were the important ones that should have enter-to-search functionality. 

So just clean up the comments/dead code and I feel like this will probably be good to go!

sravtall:
Sorry for responding late.

I cleaned up the code as per the suggestions and the enter button should be working through the form method.

The only thing that is a bit weird is the `<Select>` interaction where after a selection is made and the user clicks out, the enter functionality doesn't work. I think it's because when you click out of the `<Select>`, the form goes out of focus. 

If this is okay, then I think everything is good to go! If there are any more doubts or grievances regarding this, please feel free to reach out. Once again, apologies for the late response.

ChaseC99 (approved):
Looks and works awesome! Fantastic job on this pr 💯 
This is a great quality-of-life feature that will make searching for classes feel much smoother :)

> The <Select> interaction where after a selection is made and the user clicks out, the enter functionality doesn't work. I think it's because when you click out of the <Select>, the form goes out of focus.

Makes sense since the form is out of focus. I don't think it's a big deal.

> apologies for the late response.

No worries dude! Thanks for taking the time to work on this.

Issue #304, opened 2022-03-16
Clicking location pins on the map spawns popup in wrong location


devsdevsdevs:
Clicking the location pin doesn't accurately place the popup on the pin. This used to work fine, might want to look into changes on map component or its parent components.

![image](https://user-images.githubusercontent.com/30539874/158498137-69da417b-d2a5-466e-ab4c-bd4967a1b73c.png)

ChaseC99:
Hmm I don't even get a popup when I click on it.  
@falsidge can you look into this after finals?

PR #311, opened 2022-03-17
Fix search pins

falsidge:
## Summary
Search pins were not showing due to undefined offsets. Setting the offset to 0 will make it show up in the proper location. Unfortunately the pin will not stack with class pins.

image

## Test Plan

Searched for place and confirmed popup showed above pin.

## Issues

#304


package.json
```diff
@@ -41,7 +41,7 @@
         "react-lazyload": "^3.1.0",
         "react-leaflet": "^2.7.0",
         "react-router-dom": "^6.2.1",
-        "react-scripts": "^3.4.1",
```
ChaseC99: Do we want to update react-scripts right now?

falsidge: Oh I just realized I have a node-modules folder in a folder above the projects folder, which causes react-scripts to warn about dependencies. We don't need to update it right now/

ChaseC99 (changes requested):
Can you revert package-lock.json too plz?

Douglas-Hong (changes requested):
The popups for locations that were searched look pretty good to me! However, the maps for specific days (Mon, Tue, etc.) seem to be broken because of this change. It looks like the pins for each class are one day off.


https://user-images.githubusercontent.com/78244965/160179973-2b40ae9f-db85-4712-9653-b45f163fa0b5.mp4

Douglas-Hong (approved):
Everything looks good to me!

PR #312, opened 2022-03-18
Handle custom events with 6 days

### LINKED ISSUE #310: Custom events with 6 days are off by one day
Custom events that have a `days` array of length 6 instead of 5 or 7 are now one day later than they're supposed to be. This is due to the conditional logic that was implemented in #292. The following is an example of a custom event with 6 values in its `days` array:

```
{
  "days": [
    false,
    true,
    false,
    true,
    false,
    false
  ],
  "scheduleIndices": [
    {
      "$numberInt": "0"
    }
  ],
  "_id": {
    "$oid": "620d3bd77156990009db557f"
  },
  "color": "#551a8b",
  "title": "",
  "start": "09:55",
  "end": "11:55",
  "customEventID": "1638679414951"
},
```

The event is supposed to be on Monday and Wednesday, but it is actually on Tuesday and Thursday.

Douglas-Hong:
## Summary
- Custom events with 6 days are now in the correct place. This was done by deleting the conditional logic from #292 and adding some code in `src\stores\AppStore.js`. This should be more concise and less prone to bugs.
- All custom events will have a `days` array of length 7 after the user saves, even if the user didn't directly edit the event.

## Test Plan
- Try adding custom events with a `days` array of length 5 or 6. This can be done by editing the `days` array in MongoDB. These custom events should still be in the correct place.
- Adding and editing custom events still work properly.
- Saving will make all custom events have a `days` array of length 7.

## Issues
Closes #310 

## Future Followup
- We should find out why some custom events have 6 values (instead of 5 or 7) in their `days` array.

### DIFF
--- a/src/components/AddedCourses/CustomEventDetailView.js
+++ b/src/components/AddedCourses/CustomEventDetailView.js
@@ -38,9 +38,7 @@ const CustomEventDetailView = (props) => {
         });
 
         const dayAbbreviations = ['Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat'];
-        const daysString = days
-            .map((includeDate, index) => (includeDate ? dayAbbreviations[days.length === 7 ? index : index + 1] : ''))
-            .join(' ');
+        const daysString = days.map((includeDate, index) => (includeDate ? dayAbbreviations[index] : '')).join(' ');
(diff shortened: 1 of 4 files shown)


ChaseC99 (approved):
LGTM!

Issue #260, opened 2022-02-24
Visual Bug on Course Results Page


EricPedley:
When the window is narrow, the icons on the course results page render outside of their buttons. We should make them stack on top of the text, or disappear, or truncate the text.
![image](https://user-images.githubusercontent.com/48658337/155464054-cc841f7d-51d0-4d6a-b958-644578cf6dbd.png)

ChaseC99:
I feel like if we remove the icon on smaller screen sizes, that should be good.

Maybe we should also rename "Prerequisites" to "Prereqs"?

heyitsmichellelin:
I'll take this one for the fellowship task!

devsdevsdevs:
Closing since fix merged

PR #314, opened 2022-03-20
Modified course icons for smaller screens

heyitsmichellelin:
## Summary
Turned Course Info Button from class component to function component so the icon can disappear with smaller screen sizes. This corresponds to issue #260.

## Test Plan
Run the program on local host and resize the window to make sure the button functions properly.

### DIFF
--- a/src/components/SectionTable/CourseInfoButton.js
+++ b/src/components/SectionTable/CourseInfoButton.js
@@ -12,9 +12,8 @@ const styles = {
     },
 };
 
-function CourseInfoButton({ text, icon, redirectLink }) {
+function CourseInfoButton({ classes, text, icon, redirectLink }) {
     const isMobileScreen = useMediaQuery('max-width: 750px)');
-    const { classes } = { text, icon, redirectLink };
 
     return (
         <Button
--- a/src/components/SectionTable/CourseInfoButton.js
+++ b/src/components/SectionTable/CourseInfoButton.js
@@ -1,25 +1,44 @@
-import React, { PureComponent } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import { Button } from '@material-ui/core';
+import { useMediaQuery } from '@material-ui/core';
 
-class CourseInfoButton extends PureComponent {
-    render() {
-        const { text, icon, redirectLink } = this.props;
-        return (
-            <Button
-                startIcon={icon}
-                variant="contained"
-                size="small"
-                style={{ marginRight: '4px', backgroundColor: '#385EB1', color: '#fff' }}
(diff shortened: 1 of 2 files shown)


src/components/SectionTable/CourseInfoButton.js
```diff
@@ -1,26 +1,45 @@
...
+//                onClick={(event) => {
+//                    window.open(redirectLink);
+//                }}
+//            >
+//                {text}
+//            </Button>
+//        );
+//    }
+//}
+
```
ChaseC99: We can delete this now :) 
```suggestion

```

src/components/SectionTable/CourseInfoButton.js
```diff
@@ -1,26 +1,45 @@
...
-    }
+function CourseInfoButton({ text, icon, redirectLink }) {
+    const isMobileScreen = useMediaQuery('max-width: 750px)');
+
+    return (
+        <Button
+            startIcon={!isMobileScreen && icon}
+            variant="contained"
+            size="small"
+            style={{ marginRight: '4px', backgroundColor: '#385EB1', color: '#fff' }}
```
ChaseC99: While you're working on this, could you make this style a separate class? You can look at other files to see how we do styles

src/components/SectionTable/CourseInfoButton.js
```diff
@@ -1,26 +1,45 @@
...
+//            >
+//                {text}
+//            </Button>
+//        );
+//    }
+//}
+
 CourseInfoButton.propTypes = {
     text: PropTypes.string.isRequired,
     icon: PropTypes.object.isRequired,
```
ChaseC99: I don't think these prototypes work right now. See this link for how to do it with function components: https://reactjs.org/docs/typechecking-with-proptypes.html#function-components

ChaseC99 (changes requested):
This is looking great! Just have a few suggestions and then it should be good to go.

src/components/SectionTable/CourseInfoButton.js
heyitsmichellelin: I tried to make style into a class but it keeps giving me this error in the console and a blank white page after I do a search.

ChaseC99: We can try to debug more later, but for now try this:

- Add `classes` as prop. 
function CourseInfoButton ( { <code>classes</code>, text, icon, redirectLink }) {

- Delete line 17.
```diff
- const { classes } = { text, icon, redirectLink };
```

ChaseC99 (approved):
Looks great!

PR #315, opened 2022-03-21
Added GE fulfillment to course info bar

### LINKED ISSUE #248: Add GE fulfillments to course info bar
When a user clicks on a course for more info, list any GEs that it might satisfy below the prereqs.
image

We already have access to this information in the REST request. You should be able to get this information off of the `ge_list` option.
image

dxiuu:
## Summary
- Added the General Education categories that are fulfilled into the course info bar.

## Test Plan
- Verified that if a course satisfied GE categories, the respective categories showed up.

## Issues
- closes #248


ChaseC99 (changes requested):
I don't think you fetched and merged like I mentioned here: https://github.com/icssc-projects/AntAlmanac/issues/248#issuecomment-1066225870

The new file structure is causing a merge conflict. You'll need to fetch and merge before we can accept this.

Issue #323, opened 2022-04-01
Calendar Color picker closes when clicking inside the Color picker popup


falsidge:
The calendar color picker popup should stay open when selecting a color and close when clicking anywhere outside the popup.

This is due to changeCourseColor sending a addedCoursesChange action causing a updateCurrentScheduleIndex, which closes all popups in ScheduleCalendar.js. A simple solution is to remove the handleClosePopover, but popups will not close when classes are added or updated. Another is to add a different action for colorChange, which could also fix the color not updating in course panel

ecxyzzy:
Closing since #326 was merged.

Issue #263, opened 2022-02-28
Color picker on right side doesn't change when modified from calendar view


EricPedley:
Steps to reproduce:
1. add a class to schedule
2. go to added classes page
3. click on the class in the calendar view, and change its color using the color picker
4. observe that the color of the color picker on added classes doesn't change

![(alt text) the problem visualized as a gif](https://user-images.githubusercontent.com/48658337/155907541-8d0f7f52-864c-4338-ab14-651fe60793a8.gif)

EricPedley:
Tagged for someone else to do. My thoughts for this one were to look at the calendar view (ScheduleCalendar.js) to see how it listens for color changes, and just copy that to the section table code, which is where the color picker is in (SectionTable.js, SectionTableBody.js, SectionTableButtons.js). Pro tip is to ctrl+p search for these files in VSCode instead of digging through the file structure.

devsdevsdevs:
>  Pro tip is to ctrl+p search for these files in VSCode instead of digging through the file structure.

Or ctrl + shift + f for the Webstorm/JetBrains enjoyers

alisa-lu:
I'll take this one

ecxyzzy:
Closing since #326 was merged.

PR #326, opened 2022-04-05
Colorpicker fix for update and closing

falsidge:
## Summary
This adds a mapping between course code/ event id to the event emitters of color pickers, which should allow consistent and faster updating of the color pickers. This also fixes the colorpicker closing when click on it in the calendar.

## Test Plan
Tested all 3 color pickers in added courses, calendar, and search and they all update now.

## Issues
#263
#323 

## Future Followup (Optional)
I am not sure the best implementation for updating only the color pickers  as updating everything creates a lot of lag in the color picker. This was the most efficient I could think of, but it adds new emitters to AppStore.

### DIFF
--- a/src/actions/AppStoreActions.js
+++ b/src/actions/AppStoreActions.js
@@ -326,6 +326,8 @@ export const changeCustomEventColor = (customEventID, newColor) => {
     dispatcher.dispatch({
         type: 'CUSTOM_EVENT_COLOR_CHANGE',
         customEventsAfterColorChange,
+        customEventID,
+        newColor,
     });
 };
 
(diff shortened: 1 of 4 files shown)


src/actions/AppStoreActions.js
```diff
@@ -343,6 +345,8 @@ export const changeCourseColor = (sectionCode, newColor, term) => {
     dispatcher.dispatch({
         type: 'COURSE_COLOR_CHANGE',
         addedCoursesAfterColorChange,
+        sectionCode,
+        newColor,
```
ChaseC99, line 349: Alternatively, in [AppStore](https://github.com/icssc/AntAlmanac/blob/color-fix/src/stores/AppStore.js#L189-L190), we could compare `this.addedCourses` to `action.addedCoursesAfterColorChange` to figure out what the changed color was/section code was. 

This doesn't require another emitter but it also would take longer to parse. I'm undecided on whether or not it's better...

I'm gonna think about this longer.

TBH we might want to rethink how we handle state management. Our current AppStore model is pretty confusing and clunky.

ChaseC99 (approved):
Tested locally and everything looks good. Great work!

PR #327, opened 2022-04-06
Genericize TermSelector

### LINKED ISSUE #309: TermSelector improvements
Currently `TermSelector` causes a console warning because each `MenuItem` does not have a unique `key` prop. (Granted, I am responsible for this since it was introduced in #298 when I slimmed down the Massive Term List of Doom. Oops.)

Additionally, it would be nice if along with a fix for this, `TermSelector` was improved so it can take some sort of `onChange` function so that it's not restricted to being used in the search form (e.g. in the #302 dialog box). [Originally mentioned [here](https://github.com/icssc-projects/AntAlmanac/pull/302#discussion_r828435583).]

ecxyzzy:
## Summary
Refactor `TermSelector` such that it takes two props, changeState and `fieldName`. `changeState` is a function that takes two arguments, a field and a value, and changes the state of the other component so that it matches its internal state. `fieldName` specifies the name of the field in the other component's state to change.

`ImportStudyList` has additionally been refactored to use this, instead of bringing in MUI props to roll our own `TermSelector`-alike component locally.
## Test Plan
The term dropdowns in the right pane and the Import window still work as they did prior to this change.

## Issues
Closes #309.

### DIFF
--- a/src/components/App/ImportStudyList.js
+++ b/src/components/App/ImportStudyList.js
@@ -13,15 +13,13 @@ import RightPaneStore from '../../stores/RightPaneStore';
 import { addCourse, openSnackbar } from '../../actions/AppStoreActions';
 import AppStore from '../../stores/AppStore';
 import { PostAdd } from '@material-ui/icons';
-import { termData } from '../../termData';
-import MenuItem from '@material-ui/core/MenuItem';
-import Select from '@material-ui/core/Select';
 import InputLabel from '@material-ui/core/InputLabel';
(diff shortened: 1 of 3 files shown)


EricPedley (approved):
Looks good. I would give extra feedback but you got it all right first try. Excellent job.

EricPedley:
Adding Chase so he can merge it. I'm not authorized to merge into main.

ChaseC99:
Ah sorry I'll fix those permissions tonight

PR #328, opened 2022-04-06
Improved "different terms" warning triggers

### LINKED ISSUE #319: Improve "different terms" warning trigger
We have a warning that shows up when you add classes from different schedules
image

This warning erroneously appears if:
- you have a different term on another schedule (e.g. "Schedule 1" is 2022 Spring and "Schedule 2" is 2022 Winter)
- you clear an old schedule (e.g. press the delete button to clear "Schedule 1" which was 2022 Spring and then try to add a 2022 Winter class)
- you import a schedule for one term into another term (#302), it sends multiple warning banners instead of one

ecxyzzy:
## Summary
Refactored `addCourse` so that it would trigger the warning for different terms in the same schedule only when appropriate. Additionally, implemented a wrapper `addCoursesMultiple` that is used by `ImportStudyList`, which would suppress all warnings and trigger the warning once only if the user tried to import a Study List from a different term.

## Test Plan
1. Adding a class from one term to an empty schedule, when another schedule has classes from another term, no longer triggers the warning.
2. Clearing a schedule that had classes from one term, and then adding classes from another term, no longer triggers the warning. (NB: I could not reproduce this on `main`, but tested it here regardless to make sure I didn't accidentally introduce a regression.)
3. Importing a Study List from a different term into the current schedule triggers the warning only once.

## Issues
Closes #319.

### DIFF
--- a/src/actions/AppStoreActions.js
+++ b/src/actions/AppStoreActions.js
@@ -35,7 +35,7 @@ const arrayOfColors = [
     blueGrey[500],
 ];
 
-export const addCourse = (section, courseDetails, term, scheduleIndex, color) => {
+export const addCourse = (section, courseDetails, term, scheduleIndex, color, quiet) => {
     const addedCourses = AppStore.getAddedCourses();
     const terms = termsInSchedule(addedCourses, term, scheduleIndex);
     let existingCourse;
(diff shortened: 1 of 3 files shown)


ChaseC99 (approved):
Followed your test plan and everything works perfectly! 
Great job 💯

PR #338, opened 2022-04-10
Add expiration to websoc cache

### LINKED ISSUE #318: Cache results between different searches
Right now if you go from the search results page to the search form page, it clears the websoc cache. It would make a better user experience to not clear the cache, but if we're doing this we need to add timestamps to the items in the cache and discard them if they're stale when we get a cache hit.

EricPedley:
## Summary
Before, the cache was cleared on every new search and when pressing refresh. Now, it isn't cleared on search and still is cleared by the fresh button, as well as after 30 minutes.
## Test Plan
I changed the expiration time to 5 seconds to see whether the expiration worked and it does, then changed it back to 30 minutes.
## Issues
Closes #318

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -85,7 +85,8 @@ export async function queryWebsoc(params) {
     // Construct a request to PeterPortal with the params as a query string
     const url = new URL(PETERPORTAL_WEBSOC_ENDPOINT);
     const searchString = new URLSearchParams(params).toString();
-    if (websocCache[searchString]) {
+    if (websocCache[searchString]?.timestamp > Date.now() - 30 * 60 * 1000) {
+        //if cache hit and less than 30 minutes old
         return websocCache[searchString];
     }
     url.search = searchString;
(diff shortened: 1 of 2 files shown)


src/helpers.js
```diff
@@ -101,6 +102,8 @@ export async function queryWebsoc(params) {
         }).then((res) => res.json());
         websocCache[searchString] = backupResponse;
         return backupResponse;
+    } finally {
+        websocCache[searchString].timestamp = Date.now();
```
ChaseC99, line 106: Let's avoid using `finally` here by doing something like this:
```suggestion
websocCache[searchString] = { ...backupResponse, timestamp: Date.now() };
return backupReponse;
```

ChaseC99 (changes requested):
Small nit and it should be good to go!

src/helpers.js
EricPedley: I used `finally` there because it should be doing the same thing (adding timestamp) regardless of whether the try or catch runs, and doing it without `finally` would make for duplicate code (same thing for lines 95 and 103). What is the reason we should be avoiding `finally`?

ChaseC99: Ah I didn't see what was in the try block, since I was on my phone. Makes sense not to duplicate code by using finally 👍

ChaseC99 (approved):
Good stuff! Our search is going to feel so much faster now :racing_car:

PR #344, opened 2022-04-15
Hide LegacySearch 

ChaseC99:
## Summary
- Move old search to LegacySearch component
- Move term to top
- Move fuzzy inside of form

| Mobile | Desktop |
| --- | --- |
| image | image |
| image | image |

## Test Plan
- Tested locally for mobile and desktop

## Future Followup
"Legacy Search" dropdown looks kinda ugly. Maybe there's more we can do there.

### DIFF
--- a/src/components/App/PrivacyPolicyBanner.js
+++ b/src/components/App/PrivacyPolicyBanner.js
@@ -5,7 +5,8 @@ import { withStyles } from '@material-ui/core';
 const styles = {
     container: {
         padding: 12,
-        marginTop: '10%',
+        marginBottom: '10px',
+        marginRight: '5px',
     },
 };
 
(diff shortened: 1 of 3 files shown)


ecxyzzy (changes requested):
This is awesome! Basically how I intended to do it, file name and all, except _much_ cleaner UI/UX-wise. Good stuff.

In terms of specific feedback: The large gap between the legacy search and the privacy policy banner looks kind of awkward. If that's what you meant by the legacy search dropdown looking kinda ugly, I'd be inclined to agree.

Maybe we could move the spacing to within the LegacySearch component, or just put the privacy policy banner on the bottom of the RightPane? Would love to hear your thoughts, since this is a bit out of my wheelhouse.

ChaseC99:
How's this? I moved privacy banner to the bottom but I wonder if we should have a larger gap between "Search" and "Legacy Search"

privacy-banner

ecxyzzy:
Pushing it to the bottom make it look better for sure!

I don't have any strong feelings either way about making the gap larger, so go for it if you think it'd look good.

ChaseC99:
I'll leave it for now and we'll update it accordingly after some user testing. 

Unless there are any other code changes, can you approve?

ecxyzzy (approved):
LGTM!

PR #347, opened 2022-04-17
Dummy PR to test Github Action

Voark:
This PR is not meant to be merged

PR #343, opened 2022-04-14
Fuzzy Search

ecxyzzy:
## Summary
Implement fuzzy searching with the `websoc-fuzzy-search` module.

## Testing
https://staging-343.antalmanac.com

## To-do
- [x] #345
- [x] Currently, queries manipulate the `RightPane` directly, which means the legacy search form will reflect any autocompleted fields after the user returns from the search view. Not really sure if there's a cleaner way to go about this. (It's fine.)
## Issues
Closes #266.

PR #348, opened 2022-04-17
Staging actions

Voark:
## Summary
Add Github Action that deploys a copy of the pull request to AWS. The action is triggered by pull requests with the `deploy` label and should update on changes to the PR. Heavily references https://github.com/jgoux/preview-environments-per-pull-request-using-aws-cdk-and-github-actions .

This action deploys:
- S3 bucket to store site
- Cloudfront distribution and corresponding record to redirect URL (`staging-pr_num.antalmanac.com`) to S3 bucket

The dev backend is used in the deployment.

The URL can be easily accessed by clicking `pr-xxx-staging-actions`
image

## Test Plan
https://github.com/icssc/AntAlmanac/pull/347 for the full struggle, otherwise just check this PR.

## Future Followup (Optional)

### DIFF
--- a/.github/workflows/clean_up_pr.yaml
+++ b/.github/workflows/clean_up_pr.yaml
@@ -1,32 +1,53 @@
 name: "Pull Request Clean Up"
 
 on:
-  workflow_dispatch:
   pull_request:
     branches: main
     types: [unlabeled, closed]
 
+permissions:
+  id-token: write
+  contents: read
+  deployments: write
+
 jobs:
   clean-up:
     if: |
-      (github.event.action == 'unlabeled' && github.event.label.name == ':rocket: deploy') ||
-      (github.event.action == 'closed' && contains(github.event.pull_request.labels.*.name, ':rocket: deploy'))
+      (github.event.action == 'unlabeled' && github.event.label.name == 'deploy') ||
+      (github.event.action == 'closed' && contains(github.event.pull_request.labels.*.name, 'deploy'))
     runs-on: ubuntu-latest
+    env: # env variables for Node.js
+      HOSTED_ZONE_ID: ${{ secrets.HOSTED_ZONE_ID }}
+      CERTIFICATE_ARN: ${{ secrets.CERTIFICATE_ARN }}
+      PR_NUM: ${{ github.event.pull_request.number }}
+
     steps:
       - name: inject slug/short variables
         uses: rlespinasse/github-slug-action@v3.x
 
       - name: set STAGE variable in environment for next steps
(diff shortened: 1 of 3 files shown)


ecxyzzy:
> Currently, the deployed URL is `pr_num.antalmanac.com` because of the current certificate. However, maybe another URL such as `pr_num.staging.antalmanac.com` would be better, but it would require changing the certificate.

How about `staging-pr_num.antalmanac.com`? That shouldn't require a certificate change since it's still a third-level domain.

ecxyzzy (approved):
Tested on #343 by squash-merging your branch and it works perfectly, good stuff! Need someone with write access to `main` to merge it though.

src/api/endpoints.js
```diff
@@ -1,6 +1,7 @@
 function endpointTransform(path) {
-    if (process.env.NODE_ENV === 'development') return `https://dev.api.antalmanac.com${path}`;
-    else return `https://api.antalmanac.com${path}`;
+    if (process.env.NODE_ENV === 'development' || process.env.REACT_APP_STAGING === 'true')
+        return `https://dev.api.antalmanac.com${path}`;
+    else return `https://dev.api.antalmanac.com${path}`;
```
ChaseC99: One of these should be https://api.antalmanac.com right?

.github/workflows/clean_up_pr.yaml
```diff
@@ -0,0 +1,66 @@
...
+permissions:
+  id-token: write
+  contents: read
+  deployments: write
+
+jobs:
+  clean-up:
+    if: |
+      (github.event.action == 'unlabeled' && github.event.label.name == 'deploy') ||
+      (github.event.action == 'closed' && contains(github.event.pull_request.labels.*.name, 'deploy'))
```
ChaseC99, line 17: Have you tested the on close event yet? Maybe we can with #347?

ChaseC99 (changes requested):

.github/workflows/clean_up_pr.yaml
Voark: Yup, already tested. We can also test here too.

ChaseC99 (approved):
This is INCREDIBLE. Going to be a game changer for testing PRs. Deployment process looks good and so does the clean up process. 

The time to deploy is pretty long (around 7mins 40secs). It looks like most of the time is spent on deploying the stack to AWS. I'm not sure what we can do to reduce that time though.

image

But it shouldn't matter because since we're a public repo I believe we have [unlimited minutes](https://resources.github.com/devops/tools/automation/actions/#:~:text=Use%20it%20on%20your%20public,their%20own%20GitHub%20Action%20server) right?

PR #350, opened 2022-04-19
Add schedules with their own names

### LINKED ISSUE #332: Add ability to create additional schedules
Users should be able to generate additional schedules. 

If possible, it'd be nice for them to reorder schedules in the dropdown list too.

### LINKED ISSUE #333: Start with one schedule, instead of 4 by default
Currently every user starts with 4 schedules by default. Change this behavior to start with just one and add additional schedules later.

Douglas-Hong:
## Summary
- Users can create additional schedules by clicking the + button next to the current schedule's name.
- Once the + button is clicked, a dialog will pop up and prompt the user for the name of the new schedule.
- Once a schedule is added, the user will be redirected to the new schedule and can add courses and custom events.
- If a user has no schedules, they will just be shown one schedule called `Schedule 1`.

## Test Plan
- Add a schedule and see if its name appears in the list of schedules.
- Add courses to the new schedule and see if the Added Courses pane updates.
- Add custom events to the new schedule and see if the Added Courses pane updates.
- Make sure the new schedule does not affect any other schedules.
- Test various schedule functions (finals calendar, undo last deleted course, download as an ics file, screenshot, copy schedule, clear schedule, viewing maps).
- Test if a user with no schedules starts with only `Schedule 1`.

Demo:

https://user-images.githubusercontent.com/78244965/163896748-213fc040-8ddf-4144-b77e-68c1ac5e823d.mp4

## Issues
- Closes #332 
- Closes #333 

## Next Steps
- Allow user to rename schedule
- Allow user to delete schedule
- Redesign the calendar pane toolbar; instead of the new + button, maybe create a central menu for adding a schedule, deleting the current schedule, adding custom events, and renaming the current schedule.

### DIFF
--- a/src/actions/AppStoreActions.js
+++ b/src/actions/AppStoreActions.js
@@ -37,8 +37,7 @@ const arrayOfColors = [
 
 export const addCourse = (section, courseDetails, term, scheduleIndex, color) => {
     const addedCourses = AppStore.getAddedCourses();
-    const scheduleNames = AppStore.getScheduleNames();
-
+    const terms = termsInSchedule(addedCourses, term, scheduleIndex);
     let existingCourse;
     let multipleTerms = new Set([term]);
 
@@ -77,6 +76,7 @@ export const addCourse = (section, courseDetails, term, scheduleIndex, color) =>
         if (color === undefined) color = '#5ec8e0';
     }
 
+    const scheduleNames = AppStore.getScheduleNames();
     if (existingCourse === undefined) {
         const newCourse = {
             color: color,
--- a/src/actions/AppStoreActions.js
+++ b/src/actions/AppStoreActions.js
@@ -37,7 +37,7 @@ const arrayOfColors = [
 
 export const addCourse = (section, courseDetails, term, scheduleIndex, color) => {
     const addedCourses = AppStore.getAddedCourses();
-    const terms = termsInSchedule(addedCourses, term, scheduleIndex);
+
     let existingCourse;
     let multipleTerms = new Set([term]);
 
--- a/src/actions/AppStoreActions.js
+++ b/src/actions/AppStoreActions.js
@@ -349,10 +349,12 @@ export const changeCourseColor = (sectionCode, newColor, term) => {
 export const copySchedule = (from, to) => {
     const addedCourses = AppStore.getAddedCourses();
     const customEvents = AppStore.getCustomEvents();
+    const scheduleNames = AppStore.getScheduleNames();
 
     const addedCoursesAfterCopy = addedCourses.map((addedCourse) => {
         if (addedCourse.scheduleIndices.includes(from) && !addedCourse.scheduleIndices.includes(to)) {
-            if (to === 4) return { ...addedCourse, scheduleIndices: [0, 1, 2, 3] };
+            if (to === scheduleNames.length)
+                return { ...addedCourse, scheduleIndices: scheduleNames.map((_, index) => index) };
             else
                 return {
                     ...addedCourse,
@@ -365,7 +367,8 @@ export const copySchedule = (from, to) => {
 
     const customEventsAfterCopy = customEvents.map((customEvent) => {
         if (customEvent.scheduleIndices.includes(from) && !customEvent.scheduleIndices.includes(to)) {
-            if (to === 4) return { ...customEvent, scheduleIndices: [0, 1, 2, 3] };
+            if (to === scheduleNames.length)
+                return { ...customEvent, scheduleIndices: scheduleNames.map((_, index) => index) };
             else
                 return {
                     ...customEvent,
--- a/src/actions/AppStoreActions.js
+++ b/src/actions/AppStoreActions.js
@@ -398,3 +398,10 @@ export const toggleTheme = (radioGroupEvent) => {
         action: 'toggle theme',
     });
 };
+
+export const addSchedule = (scheduleName) => {
+    dispatcher.dispatch({
+        type: 'ADD_SCHEDULE',
+        scheduleName,
+    });
+};
--- a/src/actions/AppStoreActions.js
+++ b/src/actions/AppStoreActions.js
@@ -400,8 +400,10 @@ export const toggleTheme = (radioGroupEvent) => {
 };
 
 export const addSchedule = (scheduleName) => {
+    const newScheduleNames = [...AppStore.getScheduleNames(), scheduleName];
+
     dispatcher.dispatch({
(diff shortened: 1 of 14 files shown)


Voark:
Oh I forgot this branch has not merged the previous changes yet so the action won't trigger 😄

src/components/Calendar/AddScheduleDialog.js
```diff
@@ -0,0 +1,74 @@
...
+        padding: '3px 7px',
+        minWidth: 0,
+        minHeight: 0,
+    },
+    textField: {
+        marginBottom: '25px',
+    },
+});
+
+class AddScheduleDialog extends PureComponent {
```
ChaseC99: Trying to move away from Class Components. Can you make this a [functional component](https://reactjs.org/docs/components-and-props.html) instead?

src/components/Calendar/AddScheduleDialog.js
```diff
@@ -0,0 +1,74 @@
...
+    };
+
+    handleNameChange = (event) => {
+        this.setState({ scheduleName: event.target.value });
+    };
+
+    handleAdd = () => {
+        addSchedule(this.state.scheduleName);
+        this.setState({ isOpen: false });
+    };
```
ChaseC99: Right now you can add a schedule with an empty string as the name. Do we want to allow this or force them to provide a non-empty string?

src/components/Calendar/CalendarPaneToolbar.js
```diff
@@ -61,15 +63,16 @@ const CalendarPaneToolbar = (props) => {
...
                 onChange={handleScheduleChange}
             >
-                <MenuItem value={0}>Schedule 1</MenuItem>
-                <MenuItem value={1}>Schedule 2</MenuItem>
-                <MenuItem value={2}>Schedule 3</MenuItem>
-                <MenuItem value={3}>Schedule 4</MenuItem>
+                {props.scheduleNames.map((name, index) => (
+                    <MenuItem value={index}>{name}</MenuItem>
+                ))}
             </Select>
```
ChaseC99: We should add some padding-left inside so that the text isn't right up against the left side. 
image

ChaseC99: Not sure how the best way to do this is though. I think we might need to apply the css to `MenuItem`

src/actions/AppStoreActions.js
```diff
@@ -349,10 +349,12 @@ export const changeCourseColor = (sectionCode, newColor, term) => {
 export const copySchedule = (from, to) => {
     const addedCourses = AppStore.getAddedCourses();
     const customEvents = AppStore.getCustomEvents();
+    const scheduleNames = AppStore.getScheduleNames();
 
     const addedCoursesAfterCopy = addedCourses.map((addedCourse) => {
         if (addedCourse.scheduleIndices.includes(from) && !addedCourse.scheduleIndices.includes(to)) {
-            if (to === 4) return { ...addedCourse, scheduleIndices: [0, 1, 2, 3] };
+            if (to === scheduleNames.length)
+                return { ...addedCourse, scheduleIndices: scheduleNames.map((_, index) => index) };
```
ChaseC99, line 365: Could you add a comment here explaining that when `to === scheduleNames.length`, the "Add to all" option was selected.

I was a little confused initially.

src/components/AddedCourses/AddedCoursePane.js
```diff
@@ -101,12 +104,18 @@ class AddedCoursePane extends PureComponent {
         this.setState({ customEvents: AppStore.getCustomEvents() });
     };
 
+    loadScheduleNames = () => {
+        this.setState({ scheduleNames: AppStore.getScheduleNames() });
+    };
+
     getGrid = () => {
         return (
```
ChaseC99: Let's pull this out to a variable outside of the render, so the jsx is cleaner.
```suggestion
    getGrid = () => {
        const scheduleName = this.state.scheduleNames[AppStore.getCurrentScheduleIndex()]
        const scheduleUnits = this.state.totalUnits
        return (
```

src/components/AddedCourses/AddedCoursePane.js
```diff
@@ -101,12 +104,18 @@ class AddedCoursePane extends PureComponent {
...
             <>
                 <div className={this.props.classes.titleRow}>
                     <Typography variant="h6">
-                        {`Schedule ${AppStore.getCurrentScheduleIndex() + 1} (${this.state.totalUnits} Units)`}
+                        {`${this.state.scheduleNames[AppStore.getCurrentScheduleIndex()]} (${
+                            this.state.totalUnits
+                        } Units)`}
```
ChaseC99: Once the values are pulled out to vars, we can simplify this.
```suggestion
                        {`${scheduleName} (${scheduleUnits} Units)`}
```

src/stores/AppStore.js
```diff
@@ -184,6 +195,8 @@ class AppStore extends EventEmitter {
                 this.emit('customEventsChange');
                 break;
             case 'LOAD_SCHEDULE':
+                // If the user already had schedules, load up four schedules
+                this.scheduleNames = ['Schedule 1', 'Schedule 2', 'Schedule 3', 'Schedule 4'];
```
ChaseC99, line 199: I know the backend isn't ready yet, but how will this work when a user has saved named schedules?

ChaseC99 (changes requested):
This is looking really sweet! I played around with it locally and it feels great 💯  

I've left a handful of comments. PR should be good to go once those are resolved. 

Also these were a few other things I found:
- On Class Search, we also need to update the dropdown to add to other schedules 
image

- The "Cancel" button should be white in Dark Mode (looks like "Add Custom Event" has a similar accessibility issue) image

src/components/Calendar/AddScheduleDialog.js
Douglas-Hong: I think it's a good idea to not allow them to submit an empty string. Should we also not allow the user to submit a name that is entirely whitespace?

ChaseC99: Yeah sounds like a good idea

src/stores/AppStore.js
Douglas-Hong: If we use the current schema and slightly adjust it so that it handles schedule names, we can just set `this.scheduleNames` equal to the user's saved `scheduleNames` array.

If we use the other schema, we can iterate through the user's schedules and put all the names into an array.

ChaseC99 (approved):
Great work here - this feature is coming along really nicely!!

PR #355, opened 2022-04-21
Fix selected term getting cleared on fuzzy search

ecxyzzy:
## Summary
Fixes a regression introduced in #343 where the `RightPaneStore` was reset to default when triggering a fuzzy search, in order to make sure that all relevant fields would be clear. This had the unintended side effect of also resetting the selected term to the current default term, which meant that fuzzy searches for classes in other terms would not actually be possible.

## Test Plan
Ensure that searches for terms other than the current default actually return matching courses from that term.


ChaseC99 (approved):
Good catch. Will get this deployed asap!

Issue #324, opened 2022-04-02
Google Calendar Timezone


EricPedley:
Importing the downloaded ICS file to Google Calendar puts the events 7 hours off because it interprets them as UTC time instead of PST.  To fix this, we need to add a `VTIMEZONE` section to the ICS file, like this:
```
BEGIN:VTIMEZONE
TZID:America/Los_Angeles
X-LIC-LOCATION:America/Los_Angeles
BEGIN:DAYLIGHT
TZOFFSETFROM:-0800
TZOFFSETTO:-0700
TZNAME:PDT
DTSTART:19700308T020000
RRULE:FREQ=YEARLY;BYMONTH=3;BYDAY=2SU
END:DAYLIGHT
BEGIN:STANDARD
TZOFFSETFROM:-0700
TZOFFSETTO:-0800
TZNAME:PST
DTSTART:19701101T020000
RRULE:FREQ=YEARLY;BYMONTH=11;BYDAY=1SU
END:STANDARD
END:VTIMEZONE
```
it goes between `X-WR-TIMEZONE` and `BEGIN:VEVENT`

EricPedley:
I'm seeing now that we're using a library to manage the ics file downloads so just adding that section to the ICS file might not be the solution we're looking for.

ChaseC99:
For now, let's add a tooltip reminding user to put their calendar in PST

PR #356, opened 2022-04-22
ExportCalendar improvements

### LINKED ISSUE #143: Download ICS: Add Summer Sessions
Currently unable to download summer sessions. 
- Add summer session start dates.
- Account for summer's different quarter length.

### LINKED ISSUE #329: Add Snackbar Warning for Summer .ics Download
.ics download don't work for summer yet. It will be added in #143 

In the meantime, we should provide a snackbar warning instead of failing silently.

ecxyzzy:
## Summary
Add support for downloading .ics calendars for Summer Sessions, as well as support for calendars which are not set to PST/PDT thanks to Eric's suggestion in #324 to inject a `VTIMEZONE` section into the .ics file.

Also converted `ExportCalendarButton` into a functional component and added the starting dates for 2022 Summer Sessions so users will be able to export their Summer Session schedules.

## Test Plan
1. Test that summer sessions can be exported successfully and imported into Google Calendar with no issue, with start/end dates in accordance with the Quarterly Academic Calendar. (NB: Since all Summer Session courses have TBA finals dates, even as far back as 2017, this aspect of the code could not be tested. It should function as intended, given that exporting finals works fine for regular quarters.)
2. Test that calendars imported into a timezone that does not observe Daylight Savings Time reflect the correct local time after transitioning between PDT and PST.

## Issues
Closes #143; closes #329.

## Future Followup
* Something to do with #147. Will involve significant refactoring though, considering it has quite a bit of technical debt (calendar items are hacked together using `Date` objects with an internal date of January 2018).
* Revert or modify the `VTIMEZONE` section if/when DST gets patched out by the government.

### DIFF
--- a/src/components/Calendar/ExportCalendar.js
+++ b/src/components/Calendar/ExportCalendar.js
@@ -189,6 +189,8 @@ const getRRule = (bydays, quarter) => {
         case 'Summer10wk':
             if (bydays.includes('FR')) count -= 1; // instruction ends Thursday of Week 10
             break;
+        default:
+            break;
     }
     return `FREQ=WEEKLY;BYDAY=${bydays};INTERVAL=1;COUNT=${count}`;
 };
--- a/src/components/Calendar/ExportCalendar.js
+++ b/src/components/Calendar/ExportCalendar.js
@@ -168,22 +168,28 @@ const getTermLength = (quarter) => (quarter.startsWith('Summer') && quarter !== 
 // getRRule returns a string representing the recurring rule for the VEvent
 //  Ex: ["TU", "TH"] -> "FREQ=WEEKLY;BYDAY=TU,TH;INTERVAL=1;COUNT=20"
 const getRRule = (bydays, quarter) => {
-    let count = getTermLength(quarter) * bydays.length; // Number of occurances in the quarter
-
-    if (quarter === 'Fall') {
-        for (const byday of bydays) {
-            switch (byday) {
-                case 'TH':
-                case 'FR':
-                case 'SA':
-                    count += 1;
-                    break;
-                default:
-                    break;
+    let count = getTermLength(quarter) * bydays.length; // Number of occurences in the quarter
+    switch (quarter) {
+        case 'Fall':
+            for (const byday of bydays) {
+                switch (byday) {
+                    case 'TH':
+                    case 'FR':
+                    case 'SA':
+                        count += 1; // account for Week 0 course meetings
+                        break;
+                    default:
+                        break;
+                }
             }
-        }
+            break;
+        case 'Summer1':
+            if (bydays.includes('MO')) count += 1; // instruction ends Monday of Week 6
+            break;
+        case 'Summer10wk':
+            if (bydays.includes('FR')) count -= 1; // instruction ends Thursday of Week 10
+            break;
     }
-
     return `FREQ=WEEKLY;BYDAY=${bydays};INTERVAL=1;COUNT=${count}`;
 };
 
@@ -251,6 +257,7 @@ const exportCalendar = () => {
         if (!err) {
             // Download the .ics file
             saveAs(
+                // inject the VTIMEZONE section into the .ics file
                 new Blob([val.replace('BEGIN:VEVENT', vTimeZoneSection)], { type: 'text/plain;charset=utf-8' }),
                 'schedule.ics'
             );
--- a/src/components/Calendar/ExportCalendar.js
+++ b/src/components/Calendar/ExportCalendar.js
@@ -10,7 +10,7 @@ import { openSnackbar } from '../../actions/AppStoreActions';
 import { termData } from '../../termData';
 
 const quarterStartDates = termData
-    .filter((term) => term.startDate.length)
+    .filter((term) => term.startDate !== undefined)
     .reduce((prev, curr) => ({ ...prev, [curr.shortName]: curr.startDate }), {});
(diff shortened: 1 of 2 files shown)


src/components/Calendar/ExportCalendar.js
```diff
@@ -9,15 +9,33 @@ import AppStore from '../../stores/AppStore';
 import { openSnackbar } from '../../actions/AppStoreActions';
 import { termData } from '../../termData';
 
-// TODO(chase): support summer sessions
 const quarterStartDates = termData
-    .filter((term) => !term.longName.includes('Summer') && term.startDate.length)
-    .reduce((prev, curr) => ({ ...prev, [curr.shortName]: curr.startDate }), {}); // https://stackoverflow.com/q/36388401
-
+    .filter((term) => term.startDate.length)
```
EricPedley: Why exactly does `term.startDate` have a `length` (i.e. why is it a list instead of an object)? I feel like this is confusing enough that it warrants a comment here or considering a refactor.

src/components/Calendar/ExportCalendar.js
```diff
@@ -98,7 +95,7 @@ const getFirstClass = (date, time) => {
 
 // getExamTime returns the start and end datetime of an exam
 //  Ex: ("Mon Jun 7 10:30-12:30pm", "2019") -> [[2019, 6, 7, 10, 30], [2019, 6, 7, 12, 30]]
-const months = { Mar: 3, Apr: 4, Jun: 6, Jul: 7, Aug: 8, Sep: 9, Nov: 11, Dec: 12 };
+const months = { Mar: 3, Jun: 6, Aug: 8, Sep: 9, Dec: 12 };
```
EricPedley: Why are we removing these specific months from this object?

src/components/Calendar/ExportCalendar.js
```diff
@@ -185,99 +187,92 @@ const getRRule = (bydays, quarter) => {
...
-                    endOutputType: 'local',
-                    title: `${deptCode} ${courseNumber} ${sectionType}`,
-                    description: `${courseTitle}\nTaught by ${instructors.join('/')}`,
-                    location: `${meeting.bldg}`,
-                    start: firstClassStart,
-                    end: firstClassEnd,
-                    recurrenceRule: rrule,
-                });
+    // Construct an array of VEvents for each event
+    let events = [];
```
EricPedley: Pedantic, but I think `events` can be const.

EricPedley (changes requested):
Looks mostly good but I found a bug after testing:

Try adding 36550 and 36556 to a schedule and download it then import to a Google Calendar. The first class is displayed as Monday June 20, but instruction starts on June 21st according to the [quarterly calendar](https://www.reg.uci.edu/calendars/quarterly/2021-2022/quarterly21-22.html#:~:text=%C2%A0%C2%A0%C2%A0%C2%A0%C2%A0%C2%A0%C2%A0%C2%A0%C2%A0%C2%A0%C2%A0%C2%A0%C2%A0%C2%A0%C2%A0%C2%A0Jun%2020-,Instruction%20begins,-Jun%2021)

The timezone fix appears to be working, however. Good job on that. Also, after further testing, I can confirm that all other aspects of adding summer classes work besides this issue.

Also I have a few minor code quality feedback items.

src/components/Calendar/ExportCalendar.js
ecxyzzy: I went through the Quarterly Academic Calendars on reg.uci.edu and found that final exams always took place in March (Winter), June (Spring), July (Summer I), August (Summer 10wk), September (Summer II), and December (Fall), so April and November aren't needed. Good catch, though, since I also erroneously removed the entry for July.

src/components/Calendar/ExportCalendar.js
ecxyzzy: Prior to #298 the list of terms for which calendar exports were supported was stored as a mapping of the term code to a 3-array containing the year, 0-indexed month, and day of the first day of instruction of said term; when I implemented that change I decided to keep the data structure as it was. Checking for its `length` here just ensures that we're not pulling in any data for terms for which we don't support calendar exports (i.e. their `startDate` field is not filled out in `termData`).

Commenting is definitely a good idea here, but if you have an idea for how to refactor it I'd love to hear it.

ecxyzzy:
Regarding the inconsistency between the first calendar event and actual start dates for SS1 and SS10wk this year: it's because Juneteenth happens to be observed on the Monday of Week 1 for those terms. If you consider the "actual" start date to be June 20th, then these terms all have the same number of instruction days as the same terms from prior years.

So I think it would be fine if the first class meeting was reported as Monday; the only way I could think of actually making this work for future years is to encode whether a holiday observed prior to the start of instruction shortens the instruction period.

src/components/Calendar/ExportCalendar.js
EricPedley: I removed the default value for startDate, and made filter check if it's undefined explicitly. I think checking if it's defined makes way more sense than checking if its length is nonzero.
https://github.com/icssc/AntAlmanac/pull/356/commits/30b30d4149f1729c6af29051d152b3035d0f1b2b

EricPedley (approved):

src/components/Calendar/ExportCalendar.js
ecxyzzy: Yeah that makes more sense to me as well—thanks for making this change.

PR #358, opened 2022-04-23
Search Keyboard Shortcuts

### LINKED ISSUE #357: Search Keyboard Shortcuts and Improved Parsing
Some things that would be nice to add to our course search form:
- Pressing enter (without having to press down arrow first) selects the first autofill result and submits the search
- adding more common shortenings like we do for `cs` being short for `compsci`, or` ics` -> `i&c sci`
  - `bio` -> `biosci` or `biochem`
  - `ess` -> `earthss`
  - `wr` -> `writing`
  - There are probably more. Feel free to add more to this issue and it might also be a good idea to ask around.
- Pressing backspace or esc on the results page goes back to the search bar.
- Parse lists of course codes separated by commas like `cs 171, 178` or `ics32, stats67,`

ecxyzzy:
## Summary
Add keyboard shortcuts pertinent to search; pressing Enter now automatically submits the first result returned by fuzzy search, and pressing Backspace or Escape in the results view goes back to the search view without the user needing to click on "back."

## Test Plan
1. Pressing Enter when focused on the search bar submits the first result in the dropdown and triggers a search, as if the user had selected the first result themselves.
2. Pressing Backspace or Escape when in the course listing pane returns the user to the search pane.
3. Any keyboard input outside of these scenarios does not do anything.

NB: I didn't include the parsing improvements as part of this test plan, since it's not on npmjs yet, and I've already tested it with the driver before pushing. If you really wanted to test it with AntAlmanac you could pull the latest version from [websoc-fuzzy-search](https://github.com/icssc/websoc-fuzzy-search) and get npm to use that instead.

## Issues
Closes #357.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,7 +36,7 @@
                 "react-router-dom": "^6.2.1",
                 "react-scripts": "^3.4.1",
                 "recharts": "^1.8.5",
-                "websoc-fuzzy-search": "file:../websoc-fuzzy-search"
+                "websoc-fuzzy-search": "^0.6.3"
             },
             "devDependencies": {
                 "gh-pages": "^3.2.3",
@@ -46,10 +46,12 @@
             }
         },
         "../websoc-fuzzy-search": {
-            "version": "0.6.3",
+            "version": "0.7.0",
+            "extraneous": true,
             "license": "MIT",
             "devDependencies": {
-                "prettier": "^2.6.2"
+                "prettier": "^2.6.2",
+                "typescript": "^4.6.3"
             }
         },
         "node_modules/@ampproject/remapping": {
@@ -21083,8 +21085,9 @@
             }
         },
         "node_modules/websoc-fuzzy-search": {
-            "resolved": "../websoc-fuzzy-search",
-            "link": true
+            "version": "0.6.3",
+            "resolved": "https://registry.npmjs.org/websoc-fuzzy-search/-/websoc-fuzzy-search-0.6.3.tgz",
+            "integrity": "sha512-YBXvT+uNCIMviJu/N7Tw6fL/pGn9DKMfVTBE6B5Bnvd3f3Afgcve8CDUteIv05dw6Nfk5FBPcVPGOo0pwmZGUQ=="
         },
         "node_modules/websocket-driver": {
             "version": "0.6.5",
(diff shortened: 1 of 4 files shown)


package-lock.json
```diff
@@ -46,11 +46,12 @@
...
         },
         "../websoc-fuzzy-search": {
-            "version": "0.6.3",
+            "version": "0.7.0",
             "extraneous": true,
             "license": "MIT",
             "devDependencies": {
-                "prettier": "^2.6.2"
+                "prettier": "^2.6.2",
+                "typescript": "^4.6.3"
```
EricPedley: What's with typescript on the dev dependencies?

src/components/SearchForm/FuzzySearch.js
```diff
@@ -70,56 +137,7 @@ class FuzzySearch extends PureComponent {
...
-                        updateFormValue(
-                            'instructor',
-                            Object.keys(this.state.results).filter((x) => this.state.results[x].name === ident)[0]
-                        );
-                        break;
-                    default:
-                        break;
-                }
-                this.props.toggleSearch();
+                this.doSearch(value);
```
EricPedley, line 140: Making the refactor into a function and adding `enterEvent` separate commits would've made this a bit easier to review cause it's hard to tell if you changed anything after moving it to its own function or just left it the same.

EricPedley (approved):
This is all gonna work but look at my comments (nothing major that would make me block the approval).

EricPedley:
Not sure if this is intentional but I still don't have merge permissions so we need Chase to look at this one.

package-lock.json
ecxyzzy: ~~This shouldn't have happened—not sure how that got in there, probably WebStorm trying to be "helpful." I'll go ahead and revert that.~~

src/components/SearchForm/FuzzySearch.js
ecxyzzy: Will do in the future, thanks for the advice.

package-lock.json
ecxyzzy: Wait no that's a fuzzy search dependency.

The latest version on the repo does have TypeScript as a devDependency because it has TS type declarations, but it looks like switching back to v0.6.3 didn't get rid of the entry for the local package in the `package-lock.json`. I'll see if I can remove it without breaking everything.

ChaseC99:
I've updated permissions. Can you try again @EricPedley?

PR #359, opened 2022-04-24
Add Google Analytics v4 to Calendar Pane and Navbar

### LINKED ISSUE #337: Google Analytics 4 Migration Part 1
We need to migrate to GA 4 by July 1, 2023.

Let's setup a new GA account under the ICSSC email. Other projects will also be using this account.

EricPedley:
## Summary
Added the `react-ga4` library and added events according to this doc (view with UCI account): https://docs.google.com/document/d/1ik9cjG3RqQZuwZGgXtKnFsrXzr6BHm7vZ6vsvnyC9ds/edit?usp=sharing

## Test Plan
Opening a draft PR so I can deploy this and see if the dashboard actually receives events.
## Issues
Closes #337
## Future Followup (Optional)
- Add Right Pane events
- Eventually remove the GA v3 library, `react-ga`

### DIFF
--- a/src/actions/AppStoreActions.js
+++ b/src/actions/AppStoreActions.js
@@ -104,6 +104,12 @@ export const openSnackbar = (variant, message, duration, position, style) => {
 };
 
 export const saveSchedule = async (userID, rememberMe) => {
+    ReactGA4.event({
+        category: 'Navbar',
+        action: 'Save Schedule',
+        label: userID,
+        value: rememberMe,
+    });
     if (userID != null) {
         userID = userID.replace(/\s+/g, '');
 
@@ -153,6 +159,12 @@ export const saveSchedule = async (userID, rememberMe) => {
 };
 
 export const loadSchedule = async (userID, rememberMe) => {
+    ReactGA4.event({
+        category: 'Navbar',
+        action: 'Load Schedule',
+        label: userID,
+        value: rememberMe,
+    });
     if (
         userID != null &&
         (!AppStore.hasUnsavedChanges() ||
--- a/src/actions/AppStoreActions.js
+++ b/src/actions/AppStoreActions.js
@@ -1,6 +1,7 @@
 import dispatcher from '../dispatcher';
 import AppStore from '../stores/AppStore';
 import ReactGA from 'react-ga';
+import ReactGA4 from 'react-ga4';
 import {
     amber,
     blue,
@@ -392,5 +393,10 @@ export const toggleTheme = (radioGroupEvent) => {
     ReactGA.event({
         category: 'antalmanac-rewrite',
         action: 'toggle theme',
+    });
+    ReactGA4.event({
+        category: 'Navbar',
+        action: 'Toggle Theme',
+        label: radioGroupEvent.target.value,
     });
 };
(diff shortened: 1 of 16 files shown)


src/analyticsEnum.js
```diff
@@ -0,0 +1,38 @@
+/**
+ * This is an enum that stores all the
+ * possible category names and associated actions
+ * for Google Analytics
+ */
+const analyticsEnum = {
```
ChaseC99: Let's rename this file and this object to just `analytics`. Sounds cleaner.
```suggestion
const analytics = {
```

src/analyticsEnum.js
```diff
@@ -0,0 +1,38 @@
...
+            DOWNLOAD: 'Download Schedule',
+        },
+    },
+    nav: {
+        title: 'Navbar',
+        actions: {
+            CLICK_NOTIFICATIONS: 'Click Notifications',
+            CLICK_ABOUT: 'Click About Page',
+            CHANGE_THEME: 'Change Theme', // Label is the theme changed to
+            IMPORT_STUDY_LIST: 'Import Study List', // Value is percentage course imported
```
ChaseC99: > // Value is **percentage course** imported

What does "percentage course" mean?

src/actions/AppStoreActions.js
```diff
@@ -103,6 +105,12 @@ export const openSnackbar = (variant, message, duration, position, style) => {
 };
 
 export const saveSchedule = async (userID, rememberMe) => {
+    ReactGA4.event({
+        category: analyticsEnum.nav.title,
+        action: analyticsEnum.nav.actions.SAVE_SCHEDULE,
+        label: userID,
+        value: rememberMe,
+    });
```
ChaseC99: What do you think about exporting a function from `analyticsEnum` that is something like 
```js
function logAnalytics(category, action, label?, value?) {
    ReactGA4.event({
        category: category,
        action: action,
        label: label,
        value: value,
    });
}
```

So that we can just import the function + enum and don't have to import ReactGA4 all over the place?

EricPedley: Yeah I agree. implementing this.

src/analyticsEnum.js
EricPedley: It means the percentage successfully imported. For example, if they try to import 6 courses and only 3 get imported it'll be 0.5. I'll update the comment so it's more clear.

src/analyticsEnum.js
EricPedley: Agreed about the file name, although I think the object name makes it clear it only has string constants and not anything else in it.

EricPedley:
I think the second deploy failed because it was too soon after the first one. From the details:
```
Error [ValidationError]: (long-ass name) is in UPDATE_IN_PROGRESS state and can not be updated.
```
So its nothing to worry about.

src/analytics.js
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * This is just a wrapper around ReactGA4.event so we don't have to import it everywhere
+ * @param {string} category required
+ * @param {string} action required
+ * @param {string} label optional
+ * @param {number} value optional
+ */
+export function logAnalytics({ category, action, label, value }) {
+    ReactGA4.event({ category, action, label, value });
```
ChaseC99, line 49: In your testing, having label and value as null didn't cause any issues right?

EricPedley: Haven't tested for null, but them being undefined doesn't break anything.

ChaseC99 (approved):
Great work on part 1!

PR #363, opened 2022-04-30
Add Fall 2022 Class Schedule

ecxyzzy:
## Summary
Add dropdown option for Fall 2022.

## Test Plan
1. Verify that Fall 2022 is the default term option (NB: this diverges from WebSOC, which is stuck on Spring 2022 for now)
2. Search for classes with Fall 2022 as the selected term and verify that classes for that term show up
3. Export a schedule with Fall 2022 classes as a .ics and verify that it includes the Week 0 classes

### DIFF
--- a/src/termData.js
+++ b/src/termData.js
@@ -1,5 +1,5 @@
 // The index of the default term in termData, as per WebSOC
-const defaultTerm = 3;
+const defaultTerm = 0;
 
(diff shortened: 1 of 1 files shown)


ChaseC99 (approved):
Looks good! Let's get this out 🚀

PR #364, opened 2022-04-30
Google login passport

### LINKED ISSUE #229: Sign in with Google
Add a login button to the navbar.
Also when users click the "Load" button, provide an "Or sign in with Google" option.

Sign in with Google: https://developers.google.com/identity/gsi/web

This should pair nicely with Amazon Cognito:
- https://aws.amazon.com/cognito/ 
- https://docs.aws.amazon.com/cognito/latest/developerguide/google.html

falsidge:
## Summary
Adds login to google, which allows loading and save of schedules to google account. Saving of schedule is hidden when logged in. Email is showed in popup and name and profile pic is displayed in AppBar

## Test Plan
Sign in and schedule loads with name shown.

## Issues
Closes #229 

## Future Followup (Optional)
User profile pic may not display when logged in.


ChaseC99:
Hmm I'm having a hard time testing this locally.. Once I login, no schedules load, it still shows the "Login" button, and then clicking "Login" again does nothing.

src/components/App/AccountBase.js
```diff
@@ -0,0 +1,37 @@
...
+        const user = this.props.user;
+        return (
+            <>
+                {user ? (
+                    <>
+                        <Avatar
+                            alt={user.passport.user.name}
+                            src={user.passport.user.picture}
+                            style={{ marginRight: 5, width: 25, height: 25 }}
+                        ></Avatar>
```
ChaseC99: I think this can be self closing
```suggestion
                        />
```

src/components/App/AccountBase.js
```diff
@@ -0,0 +1,37 @@
...
+    render() {
+        const user = this.props.user;
+        return (
+            <>
+                {user ? (
+                    <>
+                        <Avatar
+                            alt={user.passport.user.name}
+                            src={user.passport.user.picture}
+                            style={{ marginRight: 5, width: 25, height: 25 }}
```
ChaseC99: Let's change these styles to classNames

src/components/App/CustomAppBar.js
```diff
@@ -48,7 +48,7 @@ const CustomAppBar = (props) => {
 
                 <div style={{ flexGrow: '1' }} />
 
-                <LoadSaveScheduleFunctionality />
+                <ScheduleLoginManager></ScheduleLoginManager>
```
ChaseC99: Can this be self closing too?
```suggestion
                <ScheduleLoginManager />
```

src/components/App/AccountBase.js
```diff
@@ -0,0 +1,37 @@
+import React, { PureComponent } from 'react';
+import { AssignmentReturn, AssignmentReturned, SaveAlt } from '@material-ui/icons';
+import { Button, Avatar, Box } from '@material-ui/core';
+import { login, logout, saveUser } from '../../actions/AppStoreActions';
+
+class AccountBase extends PureComponent {
+    render() {
+        const user = this.props.user;
+        return (
```
ChaseC99: Moving forward, everything should be functional components.

```suggestion
const AccountBase = (props) => {
    const { users } = props;
    
    return (
```

src/components/App/ScheduleLoginManager.js
```diff
@@ -0,0 +1,38 @@
+import React, { PureComponent } from 'react';
+import { checkUser, loadUser, loadSchedule } from '../../actions/AppStoreActions';
+import { LoadSchedule, SaveSchedule } from './LoadSaveFunctionality';
+import AccountBase from './AccountBase';
+
+class ScheduleLoginManager extends PureComponent {
+    state = {
+        user: null,
+    };
```
ChaseC99: Convert to functional component
```suggestion
const ScheduleLoginManager = (props) => {
    const [user, setUser] = useState(null);
```

ChaseC99 (changes requested):
Since I haven't been able to test locally yet, here's some feedback on the javascript. 
This is a big PR, so great work on it so far!

ChaseC99:
Note: works on `localhost:3001`

EricPedley:
Updated the PR description to link it to #229.

ChaseC99:
So I've resolved the merge conflicts and the requested changes but I feel like there's a lot going on in this navbar... It's a little overwhelming.
image

Potential solution:
- [x] Make the user's name a dropdown menu
- [x] Put "Logout" and "Load" inside the dropdown
- [ ] Rename "Load" to "Import old schedule"
- [x] Change "Save User" to just the classic "Save" with floppy disk icon

ChaseC99:
Cleaned it up a little bit. I think there is more than can be done to tiddy up the navbar but that should probably be a separate PR
image

ChaseC99:
In addition to another round of code review, we should think about how to introduce this to users. We want to make the transition as smooth as possible. One idea: the first time the login with Google, we can check to see if they have a legacy schedule saved and them prompt them to load it.

Also I wonder if we should show a "[Sign in with Google](https://developers.google.com/identity/branding-guidelines)" button instead of "Login"?

ChaseC99:
Adding "no deploy" since the auth endpoint doesn't work on staging sites anyways

ChaseC99:
I've thought more about 3 user flows that we need to consider.

1. User's first time on AntAlmanac.
This is easy to handle, just go through the normal login flow.

2. User already has a schedule on AntAlmanac.
On login, we should prompt them to import their old schedule.

3. User started to plan a schedule and then logs in.
Right now, logging in will wipe the current schedule. Should we try to maintain that state?

ChaseC99:
> 3. User started to plan a schedule and then logs in.
> Right now, logging in will wipe the current schedule. Should we try to maintain that state?

They already get a warning that their schedule will be wiped, so I think that's probably good enough.

ChaseC99:
Ok update:

Now if there is an existing schedule or unsaved changes, the user will be prompted to import it when they press the login button. 
image

This handles scenarios 2 and 3 with one solution!

I'm worried that the code might be getting complicated and I'd also like to test on some other people to test the ux flow to make sure that it's smooth enough.

Remaining todo tasks:
- [ ] Add a field on the save post request indicating it's a google schedule
Side note: I kind of want to make the field the user's email. That way we also have a way to tie a user (via email) to a uuid.
- [x] When signed in, the load (import legacy) button should append the schedules onto existing schedules instead of overwriting theme

src/components/AppBar/ScheduleLoginManager.js
```diff
@@ -1,31 +1,122 @@
...
+                        return {
+                            ...event,
+                            scheduleIndices: event.scheduleIndices.map(
+                                (index) => (index += userData.scheduleNames.length)
+                            ),
+                        };
+                    });
+
+                    // Add the cached data onto userData
+                    userData.addedCourses = userData.addedCourses.concat(oldData.addedCourses);
```
Voark: I don't think this works if you already have a course in your Google schedule and you try to add the same course from the old schedule. Instead we would have to add the schedule indices if the same course is in both the old and new schedules.

falsidge: ~~This works because of the lines above 50-70. It creates a new schedule rather than overwrite the schedules from the google login.~~

falsidge: Oh I see the problem

ap0nia:
I think it's ok to close because we're gonna be working on a revised implementation of auth within the monorepo setup, but I wrote down additional comments in the issue to keep track of features and whatnot.

PR #365, opened 2022-04-30
Manual Search mobile fixes

ecxyzzy:
## Summary
Currently manual search does not work on mobile because the fuzzy search bar doesn't lose focus properly—this PR (hopefully) resolves that issue.

## Test Plan
1. Test on mobile that manual search works again (requires deploy)

## Issues
image


ecxyzzy:
Looks like it works 🎉

ChaseC99 (approved):
Thanks for the quick fix!

PR #367, opened 2022-05-01
Add space between times

Voark:
## Summary
image
Ever since I noticed this it has bothered me. I only changed how the text is displayed, the underlying data structure is untouched.

## Test Plan
image
Shows up in both class search and added classes section. Couldn't find a class with a TBA time but I don't see how what I did would mess that up.

## Notes
The string before for meeting time before was padded to two characters (e. g. `_9:00` assuming `_` is the space) which adds a space only for one digit hours. This means there are technically two spaces in between `${meeting.days}` and `${meeting.time}` if the time was 1 digit, however for some reason HTML(?) turns multiple spaces into one? This is why there was no issue with double spaces between the weekdays and time.

### DIFF
--- a/src/components/SectionTable/SectionTableBody.js
+++ b/src/components/SectionTable/SectionTableBody.js
@@ -246,9 +246,10 @@ const DayAndTimeCell = withStyles(styles)((props) => {
 
     return (
         <NoPaddingTableCell className={classes.cell}>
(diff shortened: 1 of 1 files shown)


ecxyzzy (approved):
LGTM! Can also confirm that classes with TBA times are not affected by this change.

PR #368, opened 2022-05-01
Rename and delete schedules

### LINKED ISSUE #331: Add ability to rename schedules

Douglas-Hong:
## Summary
- There is a menu on the left of the calendar toolbar to edit your schedule. Inside the menu, you can rename or delete your current schedule.
- To add a schedule, you now have to go to the select menu with all your schedule names and then click the last menu item.

## Test Plan
- If a user renames a schedule, its name is changed in all appropriate places, such as the custom events dialog, the select menu that shows schedule names, and the right pane.
- After a user deletes a schedule, all schedules after the deleted schedule should be shifted back. Make sure the courses, custom events, and finals are correct for each schedule after the deletion occurs.
- If a user deletes a schedule, the current schedule should now be the schedule that was after the deleted schedule; if the user deletes their last schedule, the current schedule should be the schedule that was before the deleted schedule.
- If a user only has one schedule, the "Delete Schedules" button should be disabled.

Demo:

https://user-images.githubusercontent.com/78244965/166159887-21fe6bfc-219a-4ee6-af9f-cb934d308191.mp4

## Issues
- Closes #331


ChaseC99:
Thoughts on adding an "Add" icon to the "Add Schedule" option? I think it would help it stand out a little more.
image

Any of these might look good.
image

src/components/Calendar/DeleteScheduleDialog.js
```diff
@@ -0,0 +1,57 @@
...
+            <MenuItem onClick={handleOpen} disabled={props.scheduleNames.length === 1}>
+                Delete Schedule
+            </MenuItem>
+            <Dialog open={isOpen}>
+                <DialogTitle>Delete Schedule</DialogTitle>
+                <DialogContent>
+                    <DialogContentText>
+                        Are you sure you want to delete {`"${props.scheduleNames[props.scheduleIndex]}"`}? You cannot
+                        undo this action.
+                    </DialogContentText>
```
ChaseC99: Can we put "You cannot undo this action." on a new line?

src/components/Calendar/EditSchedule.js
```diff
@@ -0,0 +1,63 @@
...
+const styles = () => ({
+    editButton: {
+        padding: '3px 7px',
+        minWidth: 0,
+        minHeight: 0,
+    },
+});
+
+const EditSchedule = (props) => {
+    const [anchorEl, setAnchorEl] = useState(null);
```
ChaseC99: Deference classes from props up here so that later on, you can just do `classes.<style>` instead of `props.classes.<style> over and over again.
```suggestion
const EditSchedule = (props) => {
    const { classes } = props;

    const [anchorEl, setAnchorEl] = useState(null);
```

src/components/Calendar/EditSchedule.js
```diff
@@ -0,0 +1,63 @@
...
+                anchorOrigin={{
+                    vertical: 'bottom',
+                    horizontal: 'left',
+                }}
+                transformOrigin={{
+                    vertical: 'top',
+                    horizontal: 'left',
+                }}
+            >
+                <ScheduleNameDialog
```
ChaseC99, line 48: Good work reusing the ScheduleNameDialog component here 👍

src/actions/AppStoreActions.js
```diff
@@ -418,3 +418,61 @@ export const addSchedule = (scheduleName) => {
...
+        });
+
+        if (newScheduleIndices.length > 0) {
+            event.scheduleIndices = newScheduleIndices;
+            newEvents.push(event);
+        }
+    });
+
+    return newEvents;
+};
```
ChaseC99, line 457: Wow adjusting all the indexes seems like a pain.... Once we move to new schema, we can remove this right?

ChaseC99: Also - good comment explaining what's going on here 👍

src/components/Calendar/CalendarPaneToolbar.js
```diff
@@ -55,6 +56,7 @@ const CalendarPaneToolbar = (props) => {
     const isMobileScreen = useMediaQuery('(max-width:630px)');
 
     const [anchorEl, setAnchorEl] = React.useState(null);
+    const [openSchedules, setOpenSchedules] = React.useState(false);
```
ChaseC99: Nit: we should just import useState at the top
```jsx
import React, { useState } from 'react';
```

So that this will simply become:
```suggestion
    const [anchorEl, setAnchorEl] = useState(null);
    const [openSchedules, setOpenSchedules] = useState(false);
```

src/components/Calendar/CalendarPaneToolbar.js
```diff
@@ -64,19 +66,26 @@ const CalendarPaneToolbar = (props) => {
...
     };
 
+    const handleScheduleClick = () => {
+        setOpenSchedules((prev) => !prev);
+    };
+
     return (
         <Paper elevation={0} variant="outlined" square className={classes.toolbar}>
-            <AddScheduleDialog />
+            <EditSchedule scheduleNames={props.scheduleNames} scheduleIndex={props.currentScheduleIndex} />
```
ChaseC99: Deference these from props at the top of the function.

```jsx
const { classes, scheduleNames, currentScheduleIndex } = props;
```

```suggestion
            <EditSchedule scheduleNames={scheduleNames} scheduleIndex={currentScheduleIndex} />
```

src/components/Calendar/ScheduleNameDialog.js
```diff
@@ -0,0 +1,95 @@
...
+import { addSchedule, renameSchedule } from '../../actions/AppStoreActions';
+import { isDarkMode } from '../../helpers';
+
+const styles = () => ({
+    textField: {
+        marginBottom: '25px',
+    },
+});
+
+const ScheduleNameDialog = (props) => {
```
ChaseC99, line 17: Let's also de-structure here too
```suggestion
const ScheduleNameDialog = (props) => {
    const { onOpen, onClose, scheduleIndex, scheduleNames, rename } = props; // there's probably more 
```

ChaseC99 (changes requested):
Functionality looks awesome! A few code clean ups and it should be good to go.

src/actions/AppStoreActions.js
Douglas-Hong: Yes, once we move to the new schema, we don't need this function anymore! In the new schema, each schedule is independent, so deleting a schedule will not affect the courses or custom events of other schedules.

ChaseC99 (approved):
LGTM 🎉

PR #371, opened 2022-05-03
Highlight first entry in autocomplete

Voark:
## Summary
Automatically select the first entry in the autocomplete drop down as it (in my opinion) conveys the functionality more accurately and also removes the need to double down arrow to effectively move only one entry down. Prop from https://mui.com/material-ui/api/autocomplete/#props. 

## Test Plan
image

### DIFF
--- a/src/components/SearchForm/FuzzySearch.js
+++ b/src/components/SearchForm/FuzzySearch.js
@@ -22,14 +22,6 @@ class FuzzySearch extends PureComponent {
         value: '',
     };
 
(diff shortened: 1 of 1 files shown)


ecxyzzy:
This looks great! I did some testing, and it looks like we can get rid of the listener for the Enter key altogether now that the first element is automatically highlighted. Do you mind if I make a commit to this branch for that?

ecxyzzy (approved):
LGTM!

PR #372, opened 2022-05-04
Update websoc-fuzzy-search to 0.7.4

ecxyzzy:
## Summary
Update to the latest version of websoc-fuzzy-search, which fixes issues with searching suffixed course numbers without departments and hyphenated instructor names.

## Test Plan
1. Verify that searching for course numbers with a suffix and no department name returns all courses with that number (e.g. "143a" returns CS 143A etc.)
2. Verify that searching for instructors with hyphenated names works regardless of whether the hyphen is present (e.g. "wongma", "wong-ma", "wong ma" all return Prof. Wong-Ma)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,7 +37,7 @@
                 "react-router-dom": "^6.2.1",
                 "react-scripts": "^3.4.1",
                 "recharts": "^1.8.5",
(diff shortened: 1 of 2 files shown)


Voark (approved):
LGTM, tested the searches again in query and they all work.

PR #375, opened 2022-05-09
Add saving and loading of named schedules from backend

Voark:
## Summary
Load schedule names from backend instead of creating them by default from frontend. In conjunction with https://github.com/icssc/antalmanac-backend/pull/8

## Test Plan
Tested renaming, then saving and loading schedules which works.

## Issues
The backend is deployed on the dev stack for testing, but not yet on main. This branch shouldn't be merged into main  until I deploy to the changes to main.


Douglas-Hong (approved):
Looks good to me! I tested it with existing and new users, and everything works as expected.

PR #377, opened 2022-05-14
Added Analytics Events to Right Pane

### LINKED ISSUE #360: Google Analytics 4 Migration Part 2 (right pane)
Need to add the right pane events according to this doc (shared with UCI accounts): https://docs.google.com/document/d/1ik9cjG3RqQZuwZGgXtKnFsrXzr6BHm7vZ6vsvnyC9ds/edit
Part 2 of #337

EricPedley:
## Summary
Added an event category for each of the tabs on the right pane, and added the events as follows: 
image
https://docs.google.com/document/d/1ik9cjG3RqQZuwZGgXtKnFsrXzr6BHm7vZ6vsvnyC9ds/edit
Also, I changed the color change event so it only fires once when the popup opens instead of every time the user picks a new color.
## Test Plan
I used this chrome extension to make sure the events were firing appropriately: https://chrome.google.com/webstore/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna
## Issues
Closes #360 
## Future Followup (Optional)
Once we get enough data to verify that the v4 analytics are working, we should remove the v3 analytics.

### DIFF
--- a/src/components/SectionTable/CourseInfoButton.js
+++ b/src/components/SectionTable/CourseInfoButton.js
@@ -6,7 +6,6 @@ import { withStyles } from '@material-ui/core/styles';
 
 const styles = {
     button: {
-        marginRight: '4px',
         backgroundColor: '#385EB1',
         color: '#fff',
     },
--- a/src/components/SectionTable/CourseInfoButton.js
+++ b/src/components/SectionTable/CourseInfoButton.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { Button } from '@material-ui/core';
 import { useMediaQuery } from '@material-ui/core';
 import { withStyles } from '@material-ui/core/styles';
+import analyticsEnum, { logAnalytics } from '../../analytics';
 
 const styles = {
     button: {
@@ -11,7 +12,7 @@ const styles = {
     },
 };
 
-function CourseInfoButton({ classes, text, icon, redirectLink }) {
+function CourseInfoButton({ classes, text, icon, redirectLink, anlyticsAction }) {
     const isMobileScreen = useMediaQuery('(max-width: 750px)');
     return (
         <Button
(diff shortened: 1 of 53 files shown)


src/components/SectionTable/CourseInfoButton.js
```diff
@@ -11,7 +12,7 @@ const styles = {
     },
 };
 
-function CourseInfoButton({ classes, text, icon, redirectLink }) {
+function CourseInfoButton({ classes, text, icon, redirectLink, anlyticsAction, analyticsCategory }) {
```
ChaseC99, line 15: For components like this, I wonder if it would be better to pass in a function to log the analytics instead of the action and category?

src/components/SectionTable/SectionTableButtons.js
```diff
@@ -58,6 +69,12 @@ export const ScheduleAddCell = withStyles(styles)((props) => {
         }
 
         if (scheduleIndex !== -1) {
+            if (specificSchedule) {
+                logAnalytics({
+                    category: analyticsEnum.classSearch.title,
+                    action: analyticsEnum.classSearch.actions.ADD_SPECIFIC,
+                });
+            }
```
ChaseC99, line 77: What is the point of the specificSchedule check?

EricPedley: It's to see how often people are using the dropdown to add to a specific schedule, versus just adding to the current schedule with the plus
image
blue - specific schedule.
red - current schedule

src/components/SectionTable/CourseInfoButton.js
EricPedley: I don't think it makes enough difference to warrant changing it. If you have some good reasons then I can do it, but I just don't see why a function would be better.

src/components/SectionTable/SectionTableButtons.js
ChaseC99: Ah nice, this is a good analytic to capture! I'd love to see if people are actually using it.

ChaseC99 (approved):
Lgtm! Great work 💯

PR #380, opened 2022-05-16
Set notification term to default term

ChaseC99:
## Summary
We only allow notification signups for the current term. Previously, we hardcoded the term. This PR takes advantage of defaultTerm/termData, so that the notification signups automatically get updated. 

## Test Plan
- Search for ics classes under Fall 2022.
- Verify that user can sign up for notifications.
- Search for ics classes under Spring 2022.
- Verify that users can **not** sign up for notifications. 

## Future Follow Up
Thoughts on creating a `getDefaultTerm()` function that simply returns `termData[defaultTerm]`? Seems like a common use case and it feels more clear/explicit than keying termData. [RightPaneStore](https://github.com/icssc/AntAlmanac/blob/d0672434018b548e00f2c70b20d460b966f212ee/src/stores/RightPaneStore.js#L9) also does this. Might be a good fellowship task!

### DIFF
--- a/src/components/SectionTable/SectionTableBody.js
+++ b/src/components/SectionTable/SectionTableBody.js
@@ -11,6 +11,7 @@ import AppStore from '../../stores/AppStore';
 import { ColorAndDelete, ScheduleAddCell } from './SectionTableButtons';
 import classNames from 'classnames';
 import { clickToCopy, isDarkMode } from '../../helpers';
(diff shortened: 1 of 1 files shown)


ecxyzzy (approved):
Test plan checks out, LGTM!
I like the idea of the future follow-up, and have created #382 to track it.

PR #384, opened 2022-05-16
Fix esc/backspace conflict w/ notification signup

### LINKED ISSUE #379: Pressing delete key during notification sign up sends user back to search form
When typing out a number in the notification signup modal, if you press the delete/back button it will send the user back to the search form. 

image

Caused by PR https://github.com/icssc/AntAlmanac/pull/358

ecxyzzy:
## Summary
Ensure that the keyboard shortcuts to send the user back to the search form triggers if and only if the search form is open and there are no notification signup popups open. This is done by making the `keydown` event listener persistent, with an additional condition in the event function itself determining whether to trigger the event.

Needs deployment first for testing on mobile.

## Test Plan
1. Verify that escape/backspace works in the course view, with no notification signup popups open
2. Verify that with a notification signup popup open, backspace allows the user to delete their phone number without sending them back to the search form
3. Verify that with a notification signup popup open, escape sends the user back to the search form

## Issues
Closes #379.

### DIFF
--- a/src/components/CoursePane/RightPane.js
+++ b/src/components/CoursePane/RightPane.js
@@ -24,7 +24,10 @@ class RightPane extends PureComponent {
     };
 
     returnToSearchBarEvent = (event) => {
-        if (event.key === 'Backspace' || event.key === 'Escape') {
+        if (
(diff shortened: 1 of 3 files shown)


ChaseC99 (approved):
🚀

PR #385, opened 2022-05-17
Calendar export: added PST timezone information to events

ryanmohta:
## Summary
Currently our calendar events don’t have any timezone-specific information associated with them (otherwise known as “floating datetimes”). This doesn’t seem to be an issue when importing the ICS file into Google Calendar (since you can choose a timezone for the calendar itself), but when importing into iCloud calendar, the events begin and end at the same time of day no matter what timezone you’re in (I actually noticed this issue when sharing my calendar with my brother, who’s in a different timezone).

Now events will include the `America/Los_Angeles` timezone in their `DTSTART` and `DTEND` fields. I chose to go this route instead of adjusting the timezone to UTC and processing it through the ics library because converting it to UTC would remove any Daylight Savings related info from it.

## Test Plan
I tested out adding exported schedules to both my iCloud and Google calendars before and after the change, and simulating multiple timezones. The new changes properly adjust for different timezones and account for Daylight Savings time too.

### DIFF
--- a/src/components/Calendar/ExportCalendar.js
+++ b/src/components/Calendar/ExportCalendar.js
@@ -262,10 +262,14 @@ const exportCalendar = () => {
             action: analyticsEnum.calendar.actions.DOWNLOAD,
         });
         if (!err) {
+            // Add timezone information to start and end times for events
+            const icsString = val
+                .replaceAll('DTSTART', 'DTSTART;TZID=America/Los_Angeles')
+                .replaceAll('DTEND', 'DTEND;TZID=America/Los_Angeles');
             // Download the .ics file
             saveAs(
                 // inject the VTIMEZONE section into the .ics file
-                new Blob([val.replace('BEGIN:VEVENT', vTimeZoneSection)], { type: 'text/plain;charset=utf-8' }),
+                new Blob([icsString.replace('BEGIN:VEVENT', vTimeZoneSection)], { type: 'text/plain;charset=utf-8' }),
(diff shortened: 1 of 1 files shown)


ChaseC99:
Currently we avoid adding a timezone so we don't have to deal with the scenario when daylight savings flips mid-quarter (which would make all of the classes off by 1 hour).

> The new changes [...] account for Daylight Savings time too.

How does the code handle this scenario?

ryanmohta:
I was worried about that too, but found that situation only occurs when converting the events to UTC timestamps and displaying them in local time from your calendar client (since an event specified for midnight UTC would be 4pm Pacific without daylight savings and 5pm with). Turns out, specifying the timezone as `America/Los_Angeles` for each event in the ICS file accounts for this, as it automatically switches from PST (Pacific Standard Time) to PDT (Pacific Daylight Time) when daylight savings begins (and vice versa). I tested this out for a Fall 2022 schedule in iCloud and Google calendars, but feel free to test it out yourself if you like too!

ChaseC99:
@EricPedley I think you were working in this area before. Thoughts?

ecxyzzy:
I tested this in Outlook and it does account for DST properly in timezones that do observe DST and timezones that don't, so I'm fine with merging this personally, but I'm going to defer to Eric since he has more experience with the calendar stuff.

EricPedley:
Tested on Google Calendar and looks good to me.

EricPedley (approved):

PR #386, opened 2022-05-19
Update websoc-fuzzy-search to 0.8.1

ecxyzzy:
## Summary
Update to the latest version of websoc-fuzzy-search, which implements better searching for GEs and actually returns the requested number of entries (why 0.8.0 got skipped over).

This version also has the new writing course numbers in the index thanks to Eric (see icssc/websoc-fuzzy-search#15 for more details), which does break any queries with the old pre-Fall 2022 course numbers, but should be fine.

## Test Plan
1. Verify that searches for a GE category returns that category and all courses that fulfill it.
2. Verify that the fuzzy search dropdown has a maximum of 10 entries, instead of 9.
3. Verify that all other functionality still works as intended.

## Future Followup
Put out a proper release over summer (or whenever the catalogue updates) so the manual changes aren't just a stopgap.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,22 +37,13 @@
                 "react-router-dom": "^6.2.1",
                 "react-scripts": "^3.4.1",
                 "recharts": "^1.8.5",
-                "websoc-fuzzy-search": "^0.7.4"
+                "websoc-fuzzy-search": "^0.8.1"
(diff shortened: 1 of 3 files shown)


EricPedley (approved):
LGTM 🥳

PR #399, opened 2022-06-12
Change cursor hover color in darkmode

SenghoungLim:
## Summary
Change the color of cursor hover in table section body from blue violent to gold for better visibility. 

## Test Plan
- Verified that the color changes work in localhost. 

## Issues
Closes #354

PR #400, opened 2022-06-13
Represent alphanumeric course numbers as decimals

### LINKED ISSUE #378: Improve course number analytics
Right now, the analytics event for course codes only captures the whole number part of the course number, so classes like writing 39 a,b,c and cs 122a,b,c,d will all just appear as "writing 39" or "cs 122" on the analytics page. Since GA can track decimals in the values, it would be nice to represent the alphanumeric course codes as decimals, like 39a -> 39.1, 122c -> 122.3
The part you would need to change to do this is in `AppStoreActions.js`
image
`courseDetails.courseNumber` is a string such as `'39a'`, `'122d'`, or `'121'`

rlmech:
## Summary
Created courseNumAsDecimal function in helpers.js, imported and used courseNumAsDecimal in AppStoreActions.js

## Test Plan
 - Tested function outputs in personal workspace
 - Not sure how to view the effects of my changes on AntAlmanac page, hence why this is a Draft pull request

## Issues
Closes #378

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -225,7 +225,7 @@ export function courseNumAsDecimal(courseNumber) {
     const lastElement = filtered[filtered.length - 1].toUpperCase(); // .toUpperCase() won't affect numeric characters
     const lastElementCharCode = lastElement.charCodeAt(0); // Just checks the first character of the last element in the array
     // Return the last element of the filtered array as an integer if it represents an integer
-    if ('0'.charCodeAt(0) < lastElementCharCode && lastElementCharCode <= '9'.charCodeAt(0)) {
+    if ('0'.charCodeAt(0) <= lastElementCharCode && lastElementCharCode <= '9'.charCodeAt(0)) {
         return parseInt(lastElement);
     }
 
@@ -235,16 +235,15 @@ export function courseNumAsDecimal(courseNumber) {
         return 0;
     }
 
-    // This element is the second to last element of the array, supposedly a string of letters
+    // This element is the second to last element of the array, supposedly a string of numeric characters
     const secondToLastElement = filtered[filtered.length - 2];
-    // The character codes for [A-J] or [a-j] will turn into 0-9, respectively
-    const charCode = lastElement.toUpperCase().charCodeAt(0) - 'A'.charCodeAt(0);
-    if (0 <= charCode && charCode <= 9) {
-        // Add 1 to the end because A will originally represent 0, so now it represents 1. Same applies to the rest.
-        return parseFloat(`${secondToLastElement}.${charCode + 1}`);
+    // The characters within [A-I] or [a-i] will be converted to 1-9, respectively
+    const letterAsNumber = lastElement.toUpperCase().charCodeAt(0) - 'A'.charCodeAt(0) + 1;
+    if (1 <= letterAsNumber && letterAsNumber <= 9) {
+        return parseFloat(`${secondToLastElement}.${letterAsNumber}`);
     } else {
         console.error(
-            `The first character type at the end of the string was not within [A-J] or [a-j], returning last numbers found in string, Violating Character: ${
+            `The first character type at the end of the string was not within [A-I] or [a-i], returning last numbers found in string, Violating Character: ${
                 filtered[filtered.length - 1][0]
             }, Input: ${courseNumber}`
         );
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -205,3 +205,51 @@ export function isDarkMode() {
             return window.matchMedia('(prefers-color-scheme: dark)').matches;
     }
 }
+
+/**
+ * @param {string} courseNumber A string that represents the course number of a course (eg. '122A', '121')
+ * @returns {number} This function returns a number with a decimal representation of the passed in string (eg. courseNumAsDecimal('122A') returns 122.1, courseNumAsDecimal('121') returns 121)
+ */
+export function courseNumAsDecimal(courseNumber) {
+    // I wanted to split the course detail number into letters and digits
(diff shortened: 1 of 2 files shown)


EricPedley:
@Voark The preview deployment isn't working, and I think it's because the PR is coming from a fork instead of a branch. I tried running a deploy on another fellowship deploy and it failed with the same error message,  Any ideas to fix this?
screenshot of github actions logs
Also I think the clean up automation failing for #399 is related:
image

src/helpers.js
```diff
@@ -205,3 +205,51 @@ export function isDarkMode() {
...
+        console.error(`The string did not have numbers, returning 0, Input: ${courseNumber}`);
+        return 0;
+    }
+
+    // This element is the second to last element of the array, supposedly a string of letters
+    const secondToLastElement = filtered[filtered.length - 2];
+    // The character codes for [A-J] or [a-j] will turn into the character codes for 0-9, respectively
+    const charCode = lastElement.toUpperCase().charCodeAt(0) - 'A'.charCodeAt(0) + '0'.charCodeAt(0);
+    // Character codes 48 - 57 are the numbers 0 - 9
+    if ('0'.charCodeAt(0) - 1 < charCode && charCode < '9'.charCodeAt(0)) {
```
EricPedley: Instead of subtracting 1 at the start you can just use <=. Also, it shouldn't exclude 9 from the range. Try testing it with ICS45J

src/helpers.js
```diff
@@ -205,3 +205,51 @@ export function isDarkMode() {
...
+    // If the string does not have any numeric characters
+    if (filtered.length === 1) {
+        console.error(`The string did not have numbers, returning 0, Input: ${courseNumber}`);
+        return 0;
+    }
+
+    // This element is the second to last element of the array, supposedly a string of letters
+    const secondToLastElement = filtered[filtered.length - 2];
+    // The character codes for [A-J] or [a-j] will turn into the character codes for 0-9, respectively
+    const charCode = lastElement.toUpperCase().charCodeAt(0) - 'A'.charCodeAt(0) + '0'.charCodeAt(0);
```
EricPedley: Instead of adding the character code for 0, you can just leave it as the difference between the character and A, then change the below range check to use 0 and 9 instead of their character codes.

src/helpers.js
```diff
@@ -205,3 +205,51 @@ export function isDarkMode() {
...
+    // Character codes 48 - 57 are the numbers 0 - 9
+    if ('0'.charCodeAt(0) - 1 < charCode && charCode < '9'.charCodeAt(0)) {
+        // Add 1 to the end because A will originally represent 0, so now it represents 1. Same applies to the rest.
+        //
+        return parseFloat(`${secondToLastElement}.${String.fromCharCode(charCode + 1)}`);
```
EricPedley: If you implement the suggestion about not adding the character code for 0, you can drop the fromCharCode part of this.

src/helpers.js
```diff
@@ -205,3 +205,51 @@ export function isDarkMode() {
...
+    } else {
+        console.error(
```
EricPedley, line 245: Might be useful to print out which character was outside the range and its position in the string.

EricPedley (changes requested):

src/helpers.js
rlmech: I don't know how to get the position of the character in the string without summing the length of each array element, so I just added the offending character to the error message.

src/helpers.js
rlmech: Suggestion completed

src/helpers.js
EricPedley: Ok yeah good enough

src/helpers.js
```diff
@@ -237,16 +237,16 @@ export function courseNumAsDecimal(courseNumber) {
 
     // This element is the second to last element of the array, supposedly a string of letters
     const secondToLastElement = filtered[filtered.length - 2];
-    // The character codes for [A-J] or [a-j] will turn into the character codes for 0-9, respectively
-    const charCode = lastElement.toUpperCase().charCodeAt(0) - 'A'.charCodeAt(0) + '0'.charCodeAt(0);
-    // Character codes 48 - 57 are the numbers 0 - 9
-    if ('0'.charCodeAt(0) - 1 < charCode && charCode < '9'.charCodeAt(0)) {
+    // The character codes for [A-J] or [a-j] will turn into 0-9, respectively
+    const charCode = lastElement.toUpperCase().charCodeAt(0) - 'A'.charCodeAt(0);
+    if (0 <= charCode && charCode <= 9) {
```
EricPedley: reading this again, I realize that 9 actually should be excluded from the range since we're adding 1 afterward. Also, the variable `charCode` is now an inaccurate name since the number it represents isn't a character code anymore (that is to say, putting it into fromCharCode doesn't give you a sensical result).

EricPedley (changes requested):
Looks good except for one thing. Fix it real quick and then I'll merge this.

src/helpers.js
rlmech: Since 39a -> 39.1, I assumed that ICS45J -> 45.10
However, for courseNumAsDecimal('ICS45J'), my code returns parseFloat('45.10'), which results in 45.1 being returned.
How should I address this problem? Should I return a string instead of numerical values?

rlmech: And would the name "letterAsNumber" work instead of "charCode"?

EricPedley: 45j should just return 45, since 10 is out of the range of one-digit numbers. And yeah that name works.

rlmech: I believe I have done it

EricPedley (approved):
looks good 🚀

PR #406, opened 2022-07-12
Split main bundle

### LINKED ISSUE #403: Move some assets outside main bundle
Right now our page doesn't load until the entire bundle loads. The bundle currently includes stuff that isn't necessary for the initial load like the logo image and other static assets. See my comment on #118 for more details. Anyone that tackles this should research if we can easily move at least the logo, building catalog, and department list outside the main bundle, as well as identify other ways to reduce the initial page load. Success criteria is at minimum moving the logo outisde the bundle, and ideally getting our lighthouse performance score from yellow to green.

ecxyzzy:
## Summary
Reduce the size of the main bundle by chunking static assets, if possible.
* [x] Logo images
* [x] Building catalogue
* [x] Department list

## Test Plan
1. Navigate to the deployed production version and ensure that performance has improved.

## Issues
Closes #403.


ecxyzzy:
Moving the logo images out of the main bundle increased our Lighthouse performance score by 2 points, so it's a start at least.

Currently looking at solutions to move the non-React component static assets (e.g. building catalogue) out of the main bundle since Webpack v4 isn't playing nice.

src/components/RightPane/CoursePane/SearchForm/SearchForm.js
```diff
@@ -64,16 +75,27 @@ const SearchForm = (props) => {
...
-                            onSubmit={() => {
-                                logAnalytics({
-                                    category: analyticsEnum.classSearch.title,
-                                    action: analyticsEnum.classSearch.actions.MANUAL_SEARCH,
-                                });
-                                toggleSearch();
-                            }}
-                            onReset={resetFormValues}
-                        />
+                        <Suspense
```
ecxyzzy: Also implemented lazy loading of legacy search in this commit, derp :upside_down_face:

ecxyzzy:
Our performance score should now be in the green at 90 :tada: We could likely increase it further by considering changes to our tech stack (e.g. the use of `moment` and libraries that fetch remote CSS), but I think this is a good first effort.

ChaseC99:
Impressive speed boost! 

| Old | New |
| --- | --- |
| image | image |

src/components/RightPane/CoursePane/SearchForm/SearchForm.js
ChaseC99: Why does this need a suspense? Are there any network calls?

It does look a bit strange having the course search loading gif displayed here temporarily

src/components/AppBar/CustomAppBar.js
```diff
@@ -44,7 +52,9 @@ const CustomAppBar = (props) => {
     return (
         <AppBar position="static" className={classes.appBar}>
             <Toolbar variant="dense">
-                {isMobileScreen ? <MobileLogo height={32} /> : <Logo height={32} />}
+                <Suspense fallback={<div className={classes.fallback}>Loading...</div>}>
+                    {isMobileScreen ? <MobileLogo height={32} /> : <Logo height={32} />}
+                </Suspense>
```
ChaseC99: The jump here is a bit jarring... 
The width of "Loading..." causes the buttons to compress and just looks a little strange in the bar.

appbar_load

The logo svg is pretty small, do we even need to lazy load it?

src/components/AppBar/Logo.js
```diff
@@ -0,0 +1,3 @@
+import { ReactComponent as Logo } from './logo.svg';
+
+export default Logo;
```
ChaseC99: If we decide not to lazy load this, we can get rid of Logo.js and MobileLogo.js

ChaseC99:
If you really want to speed up the initial page load, I think a better solution would be SSG on build, so we serve html instead of waiting for javascript to render. Might require a bit of work to handle network calls for search though.

src/components/AppBar/CustomAppBar.js
ecxyzzy: Ooh yeah that's something that needs fixing for sure, I didn't even notice it since my internet is so fast :sweat_smile: Something about living in darkness by going faster than light, et cetera.

As for lazy loading logos: that was one of the goals Eric mentioned in #403, so I'll defer to him on that.

src/components/RightPane/CoursePane/SearchForm/SearchForm.js
ecxyzzy: I believe it's good practice to wrap lazy-loaded components in Suspenses, but it might also work without (haven't tested it).

I just decided to use the course search loading gif as a placeholder for the fallback, but we could just leave it blank as well.

ChaseC99: Oh yeah no you should really only lazy load something if there is a blocker in the component (e.g. an awaited network call). So a suspense should not be used here.

src/components/AppBar/CustomAppBar.js
ChaseC99: > I didn't even notice it since my internet is so fast 😅  

In your chrome dev tools, you can go to the network tab and throttle your internet speed to slow things down
throttle

src/components/RightPane/CoursePane/SearchForm/SearchForm.js
ecxyzzy: Oh I remember why I wanted to lazy load legacy search now. It's because it uses the department search bar, which in turn includes the static department list, which was one of the stretch goals for #403.

Removing the suspense breaks things, but even an empty div in the fallback works, and should be less abrupt than the gif.
image

src/components/AppBar/CustomAppBar.js
EricPedley: > As for lazy loading logos: that was one of the goals Eric mentioned in #403, so I'll defer to him on that.

I put it on the issue because it looked like a good place to start. If it's going to be difficult to fix that rendering problem then we should remove the lazy loading, because it's such a small asset it only contributes a few kilobytes to the bundle size.

EricPedley:
> If you really want to speed up the initial page load, I think a better solution would be SSG on build, so we serve html instead of waiting for javascript to render. Might require a bit of work to handle network calls for search though.

Worth investigating, although I think static site generation warrants its own issue. And if we do that, then I don't think putting effort into splitting up the bundle any further is worth it because with SSG the JS is loaded after the page anyway.

src/components/RightPane/Map/LazyUCIMap.js
```diff
@@ -0,0 +1,12 @@
+import React, { Suspense } from 'react';
+
+const LazyUCIMap = () => {
```
EricPedley: What's the thought behind making this its own component? In my opinion, putting this inline in `RightPaneRoot` makes more sense since it's only a few lines and not reused anywhere.

src/components/RightPane/CoursePane/SearchForm/SearchForm.js
EricPedley: With this change, bundle looks like this, with the split chunk for the legacy search highlighted:
screenshot
Without lazy loading legacy search:
screenshot
So we're just shaving off ~7 kb from the initial load, which is less than 1%
screenshot
I'm guessing the largest performance gain we're seeing from this PR here is lazy loading the UCIMap component and the other stuff is just marginal gains. I'll test that and reply in another comment.

EricPedley: This is our bundle without lazy loading the map.
image
Looks like the bundle is roughly 60 kb bigger without the lazy loading, so I think this is where most of the performance is coming from.

EricPedley: Also, reading the earlier conversation it seems like Chase really meant to ask, "why is this being lazy-loaded", because the response Eddy provided was basically that it needs a suspense is _because_ it is being lazy loaded, but I think that got cleared up. For extra clarification, the Suspense is needed to specify content to display while the lazy-loaded component is being loaded (in my understanding).

EricPedley: I think we should just remove the lazy loading for the legacy search, given the negligible performance boost now that I actually did some research. Having our codebase remain as simple as possible is a bigger priority than a 1% performance boost. This probably means removing the lazy loading on the images too, but keeping it on the map since that does provide a significant boost.

EricPedley: Now that I'm realizing we should prioratize based on file sizes, here are some other candidates for lazy loading and their uncompressed sizes:
- RightPane/CoursePane/static folder (the 'no results' images) - 70 kb
- SectionTable folder - 70 kb
- RightPane\CoursePane\SearchForm\Gifs - 2 MB
  - Maybe this gets relatively smaller on the bundle because image compression is better than source code compression, but I think this may be a major reason why our bundle is large. For reference, the entire source folder is 2.5 MB, so lazy loading this gets us potentially 80% size reduction at best.

src/components/AppBar/CustomAppBar.js
EricPedley: After thinking about it more, we should definitely not lazy load these images. Doing so adds extra complexity for almost no benefit.

EricPedley (changes requested):
We should roll back the lazy loading on the legacy search and logo images, and look into lazy loading these components (same list as in my comment on the LegacySearch thread):
- RightPane/CoursePane/static folder (the 'no results' images) - 70 kb
- SectionTable folder - 70 kb
- RightPane\CoursePane\SearchForm\Gifs - 2 MB
  - Maybe this gets relatively smaller on the bundle because image compression is better than source code compression, but I think this may be a major reason why our bundle is large. For reference, the entire source folder is 2.5 MB, so lazy loading this gets us potentially 80% size reduction at best.
 
The other action item is replying to me about why LazyUCIMap needs to be its own component. Other than that you can skip my comments unless you're interested in seeing my thought process and rationale behind decisions.

src/components/RightPane/Map/LazyUCIMap.js
ecxyzzy: This should no longer be in the PR as of daae514, as it's been consolidated into `RightPaneRoot`. I think GitHub might have given you an outdated version of the source tree to review, since my requesting of your review predated that commit.

ecxyzzy:
> RightPane/CoursePane/static folder (the 'no results' images) - 70 kb
> RightPane/CoursePane/SearchForm/Gifs - 2 MB

I think these static assets might already be lazy loaded by design, since they only show up in my Network tab after performing a search. So I'm not sure using `React.lazy()` or similar would have any tangible benefits.

EricPedley (approved):
image
Very big reduction in bundle size from this one. I think we can call this PR good to go.
For reference, here was the original bundle size (before any lazy loading):
image
So it's been reduced by roughly 30%

EricPedley:
Gonna merge this tomorrow. Seems like we're hitting limits on deployments for AWS and I don't want it to screw up the production build when we merge. Also, I created a new branch for exploring bundle optimizations: https://github.com/icssc/AntAlmanac/tree/manual-webpack
If you just run `npx webpack` from that branch, you get output that looks like this from a plugin I installed.
image

ChaseC99:
> Gonna merge this tomorrow. Seems like we're hitting limits on deployments for AWS and I don't want it to screw up the production build when we merge

Jsyk, prod is deployed to GitHub pages and doesn't use AWS

EricPedley:
Good to know. Also, seems like just retrying the build worked, merging now

PR #410, opened 2022-07-17
Replace dispatcher events with AppStore functions

### LINKED ISSUE #334: Migrating away from Flux
The version of `flux` we currently ship is out of date and depends on a deprecated version of `core-js`. In addition, the `flux` project as a whole is in maintenance mode, so it would likely be better in the long run to migrate to another application architecture.

EricPedley:
## Summary
Get rid of `flux` dependency by just skipping the dispatcher for all events. This should work because the dispatcher only has one listener per event and doesn't add any dynamically. This will simplify our architecture without necessitating a rewrite of the state management system with a new framework like Redux or React context API, although it might be worth investigating that at some point. I think simplifying first will make it easier to figure out what kind of state management system we need to switch to, if any.

Currently, I've refactored RightPaneStore to completely remove dependency on flux and everything works fine. Next up is AppStore

I wrote a wiki page about our current state management system as well: https://github.com/icssc/AntAlmanac/wiki/How-our-state-management-system-works
## Test Plan
After everything's done, manually navigate the website to make sure everything still works. Too bad we don't write unit tests 😢, might put that on the agenda at some point because manually testing every PR is getting to be a pain.
## Issues
Closes #334


ecxyzzy (approved):
LGTM, everything works fine from staging.

I like this approach a lot—kudos! The codebase is definitely a lot more readable with all the dispatcher boilerplate gone.

ecxyzzy:
> Switch files again and look at that function. It's one line, `return this.eventtsInCalendar;`, and you notice there was a typo, `eventts` instead of `events`, which wasn't caught because JS objects just let you get any property you want even if it doesn't exist.

TypeScript time? :thinking:

PR #411, opened 2022-07-19
Convert codebase to TypeScript

EricPedley:
spent too much time making this. Used Google Docs, inserted gif and recorded on ScreenToGif

Implementation Plan: rename all js files to ts or tsx depending on whether it contains components or not, then fix all the compiler errors.

This change is too big for one PR so I'm going to split it up into child PRs.
- [x] Convert AppBar: #412
- [x] Convert Calendar: #420 
- [x] Convert Map: #421 
- [x] Convert AddedCourses and SectionTable: #423
- [x] Convert CoursePane and SearchForm: #439
- [x] Convert AppStore and helpers: #444 
- [x] Backport and convert existing changes from `main`: #450 
- [x] Fix all `@ts-ignore` workarounds (e.g. instances of `openSnackbar`) and TODOS related to TS or open new issues for anything to procrastinate on
- [x] Update code style tooling per [here](https://gist.github.com/ecxyzzy/dd7989bdf546942a6b7655602d4e8f79)
- [x] Lint and ship! :rocket:

All of these sub-tasks can also include changes to files outside of their respective folders if necessary but should try to limit the scope to just one section.

Associated dev log: https://github.com/icssc/AntAlmanac/wiki/The-Great-AntAlmanac-TypeScript-Rewritening%E2%84%A2

## Test Plan
Verify successful compilation on preview deployments and click around testing website features

## Issues
Closes #120 and #126

PR #412, opened 2022-07-20
Convert Appbar to TS

EricPedley:
## Summary
Child PR of #411
This just converts all the components in the AppBar folder and the ConditionalWrapper component to TypeScript. No effect on the functionality of the site.
## Test Plan
Make sure the preview deployment works.

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -53,6 +53,9 @@ class App extends PureComponent {
                 primary: {
                     main: '#305db7',
                 },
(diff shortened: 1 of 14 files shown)


ecxyzzy (approved):
LGTM content wise, but some of the files seem to have been missed by Prettier—not entirely sure why given that we have a pre-commit hook.

I think it's fine as long as we remember to do a linting pass before merging #411 into `main`.

PR #415, opened 2022-07-22
add file to remove commits from git blame.

EricPedley:
## Summary
I use the GitLens extension to see who authored certain lines or functions and read the commit messages to make it easier to understand parts of our codebase. Under the hood, the extension uses `git blame`, which has a feature that lets you ignore commits. This is useful for ignoring commits like formatting or renaming files. I'll update and merge this once #411 (TypeScript rewrite) is done and merged to include the squashed commit converting all the code to TS, but I just wanted to make the team aware of this neat feature and see if there are any commits you guys want to add to the exclusion list.

## Future Followup  
If we do any large changes that aren't useful when inspecting the history, put their commit hashes in this file and write a comment describing the commit.


EricPedley:
I don't think we're ever gonna get around to this so closing it.

PR #417, opened 2022-07-23
Fix rightpanestore

falsidge:
## Summary
Fixes termselector, change tabs
## Test Plan
Select term in termselector

## Issues

Reset in legacy search seems to not work. 
Separate EventEmitter as a static variable instead of extending the class? maybe with pass through methods


EricPedley:
Not sure if this is a good solution long-term but I'll merge it right now and look into this later. If this problem exists in RightPaneStore it might also mean that AppStore is broken in some way, because this was all from the removal of flux. Will need to do more testing and research.

EricPedley (approved):

PR #418, opened 2022-07-23
Fix RightPaneStore bugs without making everything static

EricPedley:
## Summary
Fixes the bugs about the reset button not working in the manual search form and the term selector not working. 

The problem was as you said on Discord, that the functions being called by SearchForm were the static versions, not the instance versions, but I fixed that without making the attributes of RightPaneStore static.

For some reason, using the arrow syntax to define the functions makes them attached to the instance while using normal function syntax makes them static, although I think JS only makes this distinction when they're being passed as a first-class object, like `{RightPaneStore.someFunction}` being passed as a prop. For calls like `RightPaneStore.someFunction()` there doesn't seem to be a problem.

This also makes it clear why AppStore doesn't have any of these problems. It's functions are all put in AppStoreActions, so they always operate on the singleton instance of AppStore and don't run the risk of making a static function call where `this` is undefined.

This undoes #417 and keeps all the RightPaneStore attributes instance level. On a higher level, I think this is the right choice because we're adhering more completely to a singleton pattern instead of some weird hybrid of having a singleton object for the emitter but then having static variables. Do let me know if you see any downsides compared to having static attributes.

## Test Plan
Verify that all the right pane actions work:

- [ ] Toggle between search and results
- [ ] Change tabs
- [ ] Update form values
- [ ] Reset the form values

## Future Followup  
Make sure to define all functions in stores as arrow functions so they aren't accidentally static.

### DIFF
--- a/src/components/RightPane/CoursePane/SearchForm/TermSelector.js
+++ b/src/components/RightPane/CoursePane/SearchForm/TermSelector.js
@@ -26,6 +26,7 @@ class TermSelector extends PureComponent {
 
     handleChange = (event) => {
         this.setState({ term: event.target.value });
+        console.log(this.props.changeState);
         this.props.changeState(this.props.fieldName, event.target.value);
     };
 
--- a/src/components/RightPane/CoursePane/SearchForm/TermSelector.js
+++ b/src/components/RightPane/CoursePane/SearchForm/TermSelector.js
@@ -26,7 +26,6 @@ class TermSelector extends PureComponent {
 
     handleChange = (event) => {
         this.setState({ term: event.target.value });
-        console.log(this.props.changeState);
         this.props.changeState(this.props.fieldName, event.target.value);
     };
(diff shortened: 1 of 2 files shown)


src/components/RightPane/CoursePane/SearchForm/TermSelector.js
```diff
@@ -26,6 +26,7 @@ class TermSelector extends PureComponent {
 
     handleChange = (event) => {
         this.setState({ term: event.target.value });
+        console.log(this.props.changeState);
```
ChaseC99: Rogue console.log?

EricPedley: Oopsies

ChaseC99 (approved):
LGTM!

falsidge (approved):
LGTM too

PR #421, opened 2022-07-29
Convert Map folder to TypeScript

EricPedley:
## Summary
Child of #411 
## Test Plan
Make sure map still works


Voark (commented):
Just looked over, didn't run. Some random non blocking thoughts
- `as keyof typeof` seems unnecessarily confusing for what it really means imo, I feel like maybe string would just be fine? 
- Should we type variables as well? I know sometimes the type is obvious and as long as all function returns are typed it is fine, but what is your opinion about just making it consistent and type everything? This also means when looking at a variable after it is declared you wouldn't have to find the initialization to figure out the type.

EricPedley:
> as keyof typeof seems unnecessarily confusing for what it really means imo, I feel like maybe string would just be fine?

It's supposed to indicate that the string will only be one of the specific values that work as a key for that object, and we get a compiler error leaving it as just a string. Here's a good answer explaining it: https://stackoverflow.com/a/62764510/14587004
image

> Should we type variables as well?

Not necessary IMO. At least for VSCode users, the type is inferred automatically by IntelliSense, and you can hover the variable at any time to view the type. It does the same thing for library functions, which is why I think the whole team should get on it or some IDE with the same feature because even if we type every variable, you need the IDE to quickly see signatures for library functions. I think it would be too messy to explicitly type all the variables when you can just let your IDE do it for you. 

That being said, there are still some variables that have `any` as their type so I'll go and fix a few cases I've spotted. Feel free to comment if you see any more.
image

Voark:
> It's supposed to indicate that the string will only be one of the specific values that work as a key for that object, and we get a compiler error leaving it as just a string. 

Fair enough if it errors otherwise. I just thought it's kinda of an unintuitive type that doesn't really contribute much extra information. 

> At least for VSCode users, the type is inferred automatically by IntelliSense,

I'm ngl I thought the default type assigned would be `any` for a variable with a declaration but no type, that makes sense.

ecxyzzy (approved):
GIF of the Space Shuttle in flight

PR #422, opened 2022-08-05
Changes for #421

ecxyzzy:
## Summary
Some changes I would have liked to submit through GitHub's suggestion feature, which I couldn't because of completely arbitrary limitations on its functionality. Oh well.

* Renaming `*.d.ts` to `*.ts`: technically `d.ts` files are used to specify TypeScript types for modules written in JavaScript. We're just declaring types within our own codebase, so `.ts` is fine.
* Switching over to default export: since `building.ts` only exports one interface we can just use a default export and save time not writing the brackets around `Building` every time we need to import it.
* Getting rid of the ts-ignore for the `Locate` constructor: (long rant incoming)
All the relevant `leaflet.locatecontrol` stuff is declared as a namespace under the `leaflet` module in the bindings (which is probably the source of all of our problems), which is all fun and games until the browser runs the transpiled JavaScript and dies because the canonical `leaflet` module doesn't actually export anything named `Control`, which is what the namespace is called. Apparently, [only one (1) npm package](https://github.com/runette/ngx-leaflet-locate) even _depends_ on the TypeScript bindings, and it straight up imports the _browser_ version of the module in `node_modules`. The worst part is that it somehow works and I have no idea how.

## Test Plan
The usual deal (make sure map still works)

## Issues
#421


src/components/RightPane/Map/UCIMap.tsx
```diff
@@ -1,30 +1,32 @@
...
-        const lc = new Locate({
+        const lc = new Control.Locate({
             position: 'topleft',
             strings: {
                 title: 'Look for your lost soul',
             },
             flyTo: true,
         });
-        lc.addTo(map);
+        if (map instanceof Leaflet.Map) {
```
EricPedley: Can we just type cast this when we declare map to avoid the conditional?

src/components/RightPane/Map/UCIMap.tsx
```diff
@@ -1,30 +1,32 @@
...
 import AppStore from '../../../stores/AppStore';
 import MapMenu from './MapMenu';
 import MapMarker from './MapMarker';
-import Locate from 'leaflet.locatecontrol';
-import Leaflet, { LeafletMouseEvent } from 'leaflet';
+import Leaflet, { Control, LeafletMouseEvent } from 'leaflet';
 import analyticsEnum, { logAnalytics } from '../../../analytics';
 import { CalendarEvent, CourseEvent } from '../../Calendar/CourseCalendarEvent';
 import { Coord, MapBoxResponse } from './static/mapbox'
+import '../../../../node_modules/leaflet.locatecontrol/dist/L.Control.Locate.min.js';
```
EricPedley, line 13: Yeah let's just leave it hacky for now. If it breaks later we can just revert to the TS-ignore fix.

EricPedley (commented):
Looking for an answer on my question about the type casting. Not sure if it needs to be changed.

src/components/RightPane/Map/UCIMap.tsx
ecxyzzy: We could just do `lc.addTo(map as Leaflet.Map)`. The conditional guards against `map` being undefined, but if it were _actually_ undefined at runtime that would be a problem of its own. Nice catch.

EricPedley (approved):
ship it
```
         /|\
       /__| )
     /____| ))
   /______| )))
 /________|  )))
         _|____))
 \======| o o /
~~~~~~~~~~~~~~~~~~~
```

PR #353, opened 2022-04-20
Improve section table on mobile

ChaseC99:
## Summary
- Refactor SectionTable and CourseInfoBar to functional components
- Hide Course Info button icon in mobile view
- Change 'Prerequisites' to 'Prereqs' in mobile view

| Before | After |
| --- | --- |
| image | image |

## Test Plan
- View "Added Classes" and "Class Search" in mobile and desktop
- Verify that icons and "Prerequisites" show on desktop
- Verify that no icons and "Prereqs" show on mobile

PR #423, opened 2022-08-09
Convert AddedCourses and SectionTable folders

EricPedley:
## Summary
Convert all the files in said folders to TypeScript
I also deleted the EnrollmentGraph folder since we aren't using that component anymore. I left a link to the last version of the folder in the commented-out line that imports it.
## Test Plan
Try out the class search results and added classes tabs a bit.
## Issues
Child item of #411


ecxyzzy (changes requested):
Looks like the enrollment column might be broken. The first screenshot was taken from staging-423, while the second one is from a local instance running from the latest commit on the `convert-typescript` branch.
image
image

Voark:
> Looks like the enrollment column might be broken...

Besides the int parsing issues, I took a look at the code and what is the point of [this](https://github.com/icssc/AntAlmanac/pull/423/files#diff-3a3f5dd8cc30e4e2dcb089e5f994c07b0e343368896b7a68c160fcc3a77d99caR229)? If we really need conditional rendering based on the value shouldn't we be using a more concrete definition of true and false? 

Edit: refer to comment below for actual link to code, no idea why Github won't let me copy a permalink despite the fact that the URL changes when I highlight the line...

EricPedley:
> Besides the int parsing issues, I took a look at the code and what is the point of [this](https://github.com/icssc/AntAlmanac/pull/423/files#diff-3a3f5dd8cc30e4e2dcb089e5f994c07b0e343368896b7a68c160fcc3a77d99caR229)? If we really need conditional rendering based on the value shouldn't we be using a more concrete definition of true and false?

The link just takes me to the diff page without anything highlighted. Can you add a comment on the line you're pointing out?

src/components/RightPane/SectionTable/SectionTableBody.tsx
```diff
@@ -184,7 +207,15 @@ const LocationsCell = withStyles(styles)((props) => {
...
+    numCurrentlyEnrolled: EnrollmentCount
+    maxCapacity: number
+    numOnWaitlist: number
+    numNewOnlyReserved: number
+}
+
+const SectionEnrollmentCell = withStyles(styles)((props: SectionEnrollmentCellProps) => {
     const { classes, numCurrentlyEnrolled, maxCapacity, numOnWaitlist, numNewOnlyReserved } = props;
 
     return (
```
Voark, line 223: > Besides the int parsing issues, I took a look at the code and what is the point of [this](https://github.com/icssc/AntAlmanac/pull/423/files#diff-3a3f5dd8cc30e4e2dcb089e5f994c07b0e343368896b7a68c160fcc3a77d99caR229)? If we really need conditional rendering based on the value shouldn't we be using a more concrete definition of true and false?

> Edit: refer to comment below for actual link to code, no idea why Github won't let me copy a permalink despite the fact that the URL changes when I highlight the line...

Adding this in reference to my previous comment, this is the closest line I am able to add a comment on. The code I am referring to is lines 229 and 230.

src/components/RightPane/SectionTable/SectionTableBody.tsx
```diff
@@ -75,7 +76,12 @@ const NoPaddingTableCell = withStyles({
     sizeSmall: { padding: '0px 0px 0px 0px' },
 })(TableCell);
 
-const CourseCodeCell = withStyles(styles)((props) => {
+interface CourseCodeCellProps {
+    classes: ClassNameMap
+    sectionCode: string
+}
+
+const CourseCodeCell = withStyles(styles)((props: CourseCodeCellProps) => {
```
Voark, line 84: Also in reference to this entire file, would it be better to split the individual cell components into multiple files? This file is starting to get quite long along with there being many different prop types. Maybe this PR isn't the right place to do these changes but just a thought.

src/components/RightPane/SectionTable/SectionTableBody.tsx
EricPedley: Based on the original PR it was introduced in, I think the conditional is to prevent it displaying an empty string: https://github.com/icssc/AntAlmanac/pull/353#issuecomment-1104408864

EricPedley: https://github.com/icssc/AntAlmanac/pull/423/commits/82fbf5440a7c458543c980093e00e0bb3d51942a

src/components/RightPane/SectionTable/SectionTableBody.tsx
EricPedley: I don't think it's worth splitting because each of the cells is pretty small on its own so I don't think it would help. I do see your point about the file being big but I don't see any way of splitting it that would increase the readability.

ecxyzzy (approved):
```
L G T M
G
T
M
```

src/components/RightPane/SectionTable/SectionTableBody.tsx
Voark: While I don't know for React components I feel like in general it's obviously best practice to compartmentalize code as much as possible. There is the downside that instead of having one long file we'll have a ton of small files but I feel like that isn't that much of a problem. I personally think it increases code readability because it would be much easier to figure out the main point of the file without having to read the entire thing.

EricPedley: I'm going to leave it as-is for now. I started refactoring it and then realized that some of the styles are shared between the cells and the table body itself, and didn't want to abstract away the styles into different files because it splits up the component definitions, which I think we should avoid. If you can think of a way to get around that lmk and I'll throw it in the next PR in the TS conversion.

Voark: Fair enough, I didn't look that deep into it. Considering that's how it works right now then leaving it as is is probably the better option.

PR #427, opened 2022-08-27
Add recruitment banner

### LINKED ISSUE #424: Add ICSSC Advertisement
When someone searches for a class in the school of ICS, we should display a small banner advertising projects committee and the fellowship. Something like this:
mockup screenshot

EricPedley:
## Summary
Add a banner above the search results for classes in the school of ICS advertising our club. I'm not too set on the specific design here so feel free to make suggestions.
image
I also filmed a 2-hour video of myself working on this that we can use as a training video. I'm gonna edit and post it later so people don't have to sit through me fumbling with CSS for so long.
## Test Plan
Make sure the banner displays on ICS, CS, INF, and STATS classes but not other ones.
## Issues
Closes #424 

## Future Followup  
We might want to remove it at some point or comment it out after recruiting season is over for us.

### DIFF
--- a/src/components/RightPane/CoursePane/CourseRenderPane.js
+++ b/src/components/RightPane/CoursePane/CourseRenderPane.js
@@ -97,7 +97,7 @@ const RecruitmentBanner = ({ className }) => (
             Join ICSSC and work on AntAlmanac and other projects!
         </a>
         <br />
-        We have positions for experienced devs and those with zero experience!
+        We have opportunities for experienced devs and those with zero experience!
     </div>
 );
 
--- a/src/components/RightPane/CoursePane/CourseRenderPane.js
+++ b/src/components/RightPane/CoursePane/CourseRenderPane.js
@@ -42,7 +42,6 @@ const styles = (theme) => ({
         height: '100%',
         overflowY: 'scroll',
         position: 'relative',
-        paddingTop: '50px',
     },
     noResultsDiv: {
         height: '100%',
@@ -57,6 +56,12 @@ const styles = (theme) => ({
         display: 'flex',
         justifyContent: 'center',
         alignItems: 'center',
+    },
+    bannerContainer: {
+        height: '50px',
+        paddingLeft: '110px',
+        display: 'flex',
+        justifyContent: 'space-around',
     },
 });
 
@@ -82,6 +87,16 @@ const flattenSOCObject = (SOCObject) => {
         return accumulator;
     }, []);
 };
+
+const RecruitmentBanner = ({ className }) => (
+    <div className={className}>
+        Have web development experience or are interested in learning?
+        <br />
+        <a href="https://forms.gle/v32Cx65vwhnmxGPv8" target="__blank" rel="noopener noreferrer">
+            Join ICSSC and work on AntAlamanc and other projects!
+        </a>
+    </div>
+);
 
 const SectionTableWrapped = (index, data) => {
     const { courseData, scheduleNames } = data;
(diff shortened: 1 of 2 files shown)


Voark:
OK I think it's actually pretty good. I think maybe we could top and/or right align and make the font smaller for design?

src/components/RightPane/CoursePane/CourseRenderPane.js
```diff
@@ -83,6 +88,16 @@ const flattenSOCObject = (SOCObject) => {
     }, []);
 };
 
+const RecruitmentBanner = ({ className }) => (
+    <div className={className}>
+        Have web development experience or are interested in learning?
+        <br />
+        <a href="https://forms.gle/v32Cx65vwhnmxGPv8" target="__blank" rel="noopener noreferrer">
+            Join ICSSC and work on AntAlamanc and other projects!
```
ecxyzzy:
```suggestion
            Join ICSSC and work on AntAlmanac and other projects!
```

EricPedley:
Updated: right aligned, updated text.
screenshot
I left the font the same size as it was before. I think it's fine as-is. If someone else wants to vote for a smaller font I'll change it. I think a vote here is appropriate so we avoid [bikeshedding](https://en.wikipedia.org/wiki/Law_of_triviality)

Voark:
It is just my opinion that it should discrete in order to be easy to overlook when using the website normally as I don't like the idea of having a pseudo-ad. That being said, at the end of the day this is just my opinion and should be taken as such. An alternative suggestion I had is to have it as a banner that can be closed and will never reopen again on its on.

On an unrelated note, what if instead of linking to the Google form, we link to the up and coming ICSSC projects page on the ICSSC website? That way we can reduce the amount of text into a small blurb and have more information for people that click through. Also it is much more actionable to a user as they can then join our Discord server or browse our projects among other things besides just seeing a blank Google form.

EricPedley:
I like the idea of adding a button to close it and saving that decision, but I think it would be better as a fellowship task (AKA I'm too lazy to do it right now). We could even add another line: "Tired of this banner? Implement the close button yourself! (link to that issue)"

My plan with the Google Form was to email the people who signed up there a link to the actual application form when we figure it out. If we get some sort of application link on the website then I think linking to the website first instead of straight to the form is a good idea. Do we have that yet?

Voark:
Sounds good. I was under the impression that the site was almost done when making that comment but it looks like it isn't yet, let's just move forward with the form until it is finished.

src/components/RightPane/CoursePane/CourseRenderPane.js
```diff
@@ -83,6 +89,18 @@ const flattenSOCObject = (SOCObject) => {
     }, []);
 };
 
+const RecruitmentBanner = ({ className }) => (
+    <div className={className}>
+        Interested in web development?
+        <br />
+        <a href="https://forms.gle/v32Cx65vwhnmxGPv8" target="__blank" rel="noopener noreferrer">
+            Join ICSSC and work on AntAlmanac and other projects!
```
ChaseC99, line 97: Should this be "to" instead of "and"? Idk which sounds better.
```suggestion
            Join ICSSC to work on AntAlmanac and other projects!
```

src/components/RightPane/CoursePane/CourseRenderPane.js
```diff
@@ -83,6 +89,18 @@ const flattenSOCObject = (SOCObject) => {
...
+        </a>
+        <br />
+        We have positions for experienced devs and those with zero experience!
```
ChaseC99: Are there positions for zero experience?

EricPedley: I was referring to the fellowship, but now I'm realizing it would be better worded as "opportunities" instead of "positions". Will change.
https://github.com/icssc/AntAlmanac/pull/427/commits/c990837b717cd0c14db4c81de424c652a1a3ce52

src/components/RightPane/CoursePane/CourseRenderPane.js
EricPedley: I see your point with "to" since it's showing how that is the purpose of joining, but I still think "and" sounds better, although "sounds better" is kinda subjective, so if more people think "and" is better we can change it, but for now, I'll leave it as-is.

ChaseC99 (approved):

PR #433, opened 2022-09-11
Staging Action Improvements

Voark:
## Summary
This PR addresses some of the annoyances that have popped up while using this action. The problems and their fixes are:

- Github environments are created by the deployments are not deleted when the deployment is deleted, which means they need to be manually cleaned out. While it is possible to have the action delete environments it needs a custom Github token (which is gotten from a user and cannot be from the ICCSC org), and results in the action saying "{token's user} deployed to some environment" which just feels wrong.
    - Get rid of deployments action
    - To replace the ability to get the staging URL easily I added a comment action that will comment the URL, along with recreating every time a new commit is added so it stays near the bottom of the PR.
- Github deployments are not able to be created when pull requests come from outside the repo because their Github token does not have the correct permissions.
    - Get rid of deployments action 
- AWS permissions are not able to be assumed when pull requests come from outside the repo due to how OIDC federation works.
    - Added AWS credentials to the repo secrets and use those to assume a role in the action instead of getting credentials dynamically through OIDC
- Adding the "no deploy" label doesn't delete the staging stack if it was created, and removing the label doesn't deploy the staging stack
    - Add triggers for the actions on labeled and unlabeled

There are also some other minor changes, main one being forcing only one workflow to be run at a time along with stack status waiting to make the action more resistant to failing.

## Test Plan
Tested the action in this PR by adding and removing the no deploy label to trigger deployment and deletion. 
**Warning:** I have not tested with external PRs, but I am relatively confident it will work. If it doesn't there is another measure that we can try (using [pull request target](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) as a trigger), but that gives actions run from outside repos much more permissions so I am currently trying to avoid that. 

## Issues
Closes #

### DIFF
--- a/.github/workflows/clean_up_pr.yaml
+++ b/.github/workflows/clean_up_pr.yaml
@@ -29,9 +29,11 @@ jobs:
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@master
         with:
-          role-to-assume: ${{ secrets.ACTIONS_IAM_ROLE }}
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: us-east-1
-
+          role-duration-seconds: 2400
+            
       - name: check if stack exists
         id: check-stack-exists
 
--- a/.github/workflows/clean_up_pr.yaml
+++ b/.github/workflows/clean_up_pr.yaml
@@ -5,9 +5,14 @@ on:
     types: [closed, labeled]
     
 
+concurrency:
+  group: ${{ github.head_ref }}
+  cancel-in-progress: true
+
 permissions:
   id-token: write
   contents: read
+  pull-requests: write
   deployments: write
   repository-projects: write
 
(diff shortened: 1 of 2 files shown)


Voark:
Testing

EricPedley (approved):
Sounds good. I'm not really familiar enough with GitHub actions to sight-read the changes here and understand exactly what's going on, so I'll just trust that it works and if it breaks it should be no big deal since this isn't visible to our end users.

EricPedley:
One thought I just had after merging this is that for users outside the icssc org, does it just automatically create the deployment from a PR? I was thinking we might want to have it wait for a contributor to approve the workflow being run before it creates a preview deployment based on a new PR from an outsider, cause otherwise, they might be able to execute arbitrary code by opening a PR that changes the GitHub workflows. Correct me if I'm wrong here.

Voark:
> One thought I just had after merging this is that for users outside the icssc org, does it just automatically create the deployment from a PR? I was thinking we might want to have it wait for a contributor to approve the workflow being run before it creates a preview deployment based on a new PR from an outsider, cause otherwise, they might be able to execute arbitrary code by opening a PR that changes the GitHub workflows. Correct me if I'm wrong here.

There is a setting enabled that doesn't allow first time contributors to run workflows without approval from someone with permissions. I also attempted to scope the AWS role to be less permissive but of course nothing is bulletproof. That being said there is nothing stopping someone who already has permissions to attempt malicious code but of course if they already have permissions they could have done that to begin with without running the action.

PR #434, opened 2022-09-12
Move prerequisites link into course info

### LINKED ISSUE #393: Move Prerequisites Link
The Prerequisites button is redundant since we also display the prereqs inside of the course info.

- [ ] Remove the Prerequisites button
- [ ] Make the "Prerequisites" text (highlighted in the image below), link to the prereqs page like the button does.

image

MinhxNguyen7:
## Summary
- [x] Removed the Prerequisites button
- [x] Made the "Prerequisites" text link to the prereqs page like the button does.
image
The link acts the same as the late prerequisites button, except that it opens the page in a new tab (which I think is standard since it leads to a different site).

## Test Plan
Check that the link leads to the prerequisites page in a new tab.
Check that the button has disappeared for good.

## Issues
Closes #393


src/components/RightPane/SectionTable/SectionTable.js
```diff
@@ -82,21 +82,16 @@ const SectionTable = (props) => {
                     deptCode={courseDetails.deptCode}
                     courseTitle={courseDetails.courseTitle}
                     courseNumber={courseDetails.courseNumber}
+                    prerequisiteAnchorInfo={{
+                        link: courseDetails.prerequisiteLink,
+                        analyticsAction: analyticsEnum.classSearch.actions.CLICK_PREREQUISITES,
```
EricPedley: is this value ever not `analyticsEnum.classSearch.actions.CLICK_PREREQUISITES`? If not, I don't think it needs to be a field in this object, it can just be hard-coded into CouseInfoBar.

EricPedley (changes requested):
1 comment to address

EricPedley (changes requested):
Try it out with this extension (https://chrome.google.com/webstore/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna) on and check the console. I don't see the analytics event firing for some reason. It fires a generic "click" event but not the custom "Click Prerequisites" event.

MinhxNguyen7:
@EricPedley Fyi, you said that this looked good to you in meeting. Can you merge it?

EricPedley (approved):
LGTM gonna merge blind without the preview 😎

PR #438, opened 2022-09-13
Class location link opens maps tab

### LINKED ISSUE #394: Class location links should open maps tab
Right now, when you click a class link in the calendar or section table, it opens up https://map.uci.edu/.

Instead, we should show some love to the Maps tab. It would be cool if clicking it opened the maps tab and selected that building.
image

MinhxNguyen7:
## Summary
As described, clicking on the link of a building in the calendar selects the building in the maps tab.

## Test Plan
I'm not sure, other than that clicking on the link opens the map and focuses on a correctly-placed pin.

## Issues
#394


MinhxNguyen7:
It'd be nice if someone could give me some pointers as to why my implementation of AppStore.emit() and AppStore.on() doesn't work (emitter at `CourseCalendarEvent.js` line `77` and listener at `UCIMap.js` line `299`).

EricPedley:
Looks like it isn't working because the event listener is only registered when the map component is opened. If you have the map open and click the button, then it works. I think you might have to add the listener in `RightPaneRoot`, switch the current tab to UCIMap programmatically, and the pass the event down.

MinhxNguyen7:
> Looks like it isn't working because the event listener is only registered when the map component is opened. If you have the map open and click the button, then it works. I think you might have to add the listener in `RightPaneRoot`, switch the current tab to UCIMap programmatically, and the pass the event down.

Did that. It worked. Thanks!
Problem now is, `MobileHome` doesn't have `DesktopTabs` and therefore `RightPaneRoot` loaded. Time for more cascading listeners.

src/components/Calendar/CourseCalendarEvent.js
```diff
@@ -133,14 +160,13 @@ const CourseCalendarEvent = (props) => {
...
-                                {bldg !== 'TBA' ? (
-                                    <a href={genMapLink(bldg)} target="_blank" rel="noopener noreferrer">
-                                        {bldg}
-                                    </a>
-                                ) : (
-                                    bldg
-                                )}
+                            <td
+                                onClick={() => selectBuilding(bldg)}
+                                className={`${classes.multiline} ${classes.rightCells} ${
```
EricPedley: Kinda poor contrast in dark mode:
image
You can use `helpers.isDarkMode` and do a ternary to pick a different color when it's dark mode.

src/components/RightPane/Map/UCIMap.js
```diff
@@ -241,6 +253,40 @@ export default class UCIMap extends PureComponent {
         this.generateRoute(this.state.day);
     };
 
+    focusOnBuilding = ({ lat, lng, name, acronym, imgURL } = {}) => {
```
EricPedley: Can you make this function also display the popup for the building that normally only opens when you click on the pin? Like this:
image

src/components/Calendar/CourseCalendarEvent.js
```diff
@@ -63,15 +63,42 @@ const styles = {
...
+    },
 };
 
-const genMapLink = (location) => {
-    try {
-        const location_id = locations[location.split(' ')[0]];
-        return 'https://map.uci.edu/?id=463#!m/' + location_id;
-    } catch (err) {
-        return 'https://map.uci.edu/';
+// const genMapLink = (location) => {
```
EricPedley: I think instead of commenting this out, we should move into the popups on UCIMap.
Kind of like this except with some modifications (see below)
image
- margin between the buttons and the text on one line per button, which is going to be a balance between a wider popup and maybe smaller font. 
- different icons per button
- the title should no longer be a link, and maybe make it bold and centered.

src/components/MobileHome.js
```diff
@@ -1,13 +1,36 @@
...
 import DesktopTabs from './RightPane/RightPaneRoot';
 import Calendar from './Calendar/ScheduleCalendar';
+import RightPaneStore from './RightPane/RightPaneStore';
 
 const MobileHome = () => {
     const [selectedTab, setSelectedTab] = useState(0);
 
     const components = [<Calendar isMobile={true} />, <DesktopTabs style={{ height: 'calc(100% - 50px' }} />];
 
+    const focusOnBuilding = (args) => {
```
EricPedley: "args" is pretty vague. Can you think of a better parameter name?

src/components/MobileHome.js
```diff
@@ -1,13 +1,36 @@
...
+        // Since MobileHome doesn't have DesktopTabs permamnently loaded,
```
EricPedley: typo in this comment on "permamnently"

src/components/MobileHome.js
```diff
@@ -1,13 +1,36 @@
...
+        // we need to switch over to it, get a confirmation that it's loaded,
+        // then re-emit 'focusOnBuilding'
+
+        if (selectedTab !== 1) {
+            const reEmitFocus = () => {
```
EricPedley: How does this *not* cause an infinite loop? We're emitting the event that triggers `focusOnBuilding` from within `focusOnBuilding`. I can see that it works since I tested it locally, but how? Adding an explanation in code as a comment might help.

src/components/RightPane/Map/UCIMap.js
```diff
@@ -36,6 +37,20 @@ const ATTRIBUTION_MARKUP =
     '&copy; <a href="http://osm.org/copyright">OpenStreetMap</a> contributors | Images from <a href="https://map.uci.edu/?id=463">UCI Map</a>';
 const DIRECTIONS_ENDPOINT = 'https://api.mapbox.com/directions/v5/mapbox/walking/';
 
+// NOTE: Maybe there's a better place to put this function
```
EricPedley: Yeah, move it inside UCIMap.

src/components/Calendar/CourseCalendarEvent.js
```diff
@@ -63,15 +63,42 @@ const styles = {
...
+//         return 'https://map.uci.edu/';
+//     }
+// };
+
+const selectBuilding = (buildingName) => {
+    if (buildingName !== 'TBA') {
+        RightPaneStore.emit('focusOnBuilding', { name: buildingName });
     }
+
+    /** Explanation of what happens when 'focusOnBuilding' is emitted:
```
EricPedley: Thanks for this comment 👍

EricPedley (changes requested):
7 comments need addressing

src/components/Calendar/CourseCalendarEvent.js
```diff
@@ -63,15 +63,42 @@ const styles = {
...
+     *
+     *  If desktop:
+     *  1) RightPaneRoot recieves 'focusOnBuilding'.
+     *  2a) If the Map tab is selected already, it passes the args down to UCIMap with 'selectBuilding'.
+     *  2b) If the Map tab is not selected, it switches to the Map tab, waits for it to load, and does 1a.
```
MinhxNguyen7: 2b leads to 2a, not 1a

src/components/MobileHome.js
MinhxNguyen7: It doesn't cause an infinite loop because, after the first time it runs, it sets `selectedTab` to 1 (SEARCH), which causes the condition to be false. The function does execute twice, but not infinitely

EricPedley:
Didn't realize this was ready for review. I'll probably get around to reviewing some time next week.

MinhxNguyen7:
Between the TS rewrite and the fact that I haven't looked at this in three months, I would basically have to rewrite everything. I'm going to close this and move it back into the backlog.

MinhxNguyen7:
@Voark Can you see why the deploy is failing? Is it because it's not in the original repo?

Voark:
> @Voark Can you see why the deploy is failing? Is it because it's not in the original repo?

Sorry for the late response, yes if you do a PR from a branch outside the repo it doesn't have the AWS credentials. If I don't respond on GitHub in a reasonable time feel free to ping me on Discord, sometimes I miss the tags in PRs.

MinhxNguyen7:
I'm going to merge this into map-link so we can deploy it. I'll make another PR to merge that into main.

PR #439, opened 2022-09-15
Convert CoursePane and SearchForm

ecxyzzy:
## Summary
Convert the components to TypeScript

## Test Plan
Play around with the right pane

## Issues
Child item of #411


src/components/AppBar/ImportStudyList.tsx
```diff
@@ -42,14 +42,17 @@ class ImportStudyList extends PureComponent<ImportStudyListProps, ImportStudyLis
         studyListText: '',
     };
 
-    onTermSelectorChange = (field: 'selectedTerm', value: any) => {
-        this.setState({ [field]: value });
-    };
+    // The type hack is cool, but string literal as type is not compatible with the `string` type, so I've changed the
```
EricPedley: Good catch. I don't know why this built successfully before or else I would've caught it, but now uncommenting it throws an error, so I agree with this change. You can go ahead and remove the comment and commented-out code now. That would remove a distraction for future devs reading this file.

src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
```diff
@@ -1,8 +1,9 @@
 import React, { PureComponent } from 'react';
-import PropTypes from 'prop-types';
-import { IconButton, Tooltip } from '@material-ui/core';
+import {IconButton, Theme, Tooltip} from '@material-ui/core';
 import { ArrowBack, Refresh } from '@material-ui/icons';
 import { withStyles } from '@material-ui/core/styles';
+import {ClassNameMap} from "notistack";
```
EricPedley: nit: I've been using the ClassNameMap from `@material-ui/core/styles/withStyles` so far. Weird that `notistack` also has one, and it's a type alias with the exact same definition, but I would rather stay consistent from where we're importing it from. Or maybe they're both copying it from a different common place? IDK.

src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
```diff
@@ -44,9 +52,4 @@ class CoursePaneButtonRow extends PureComponent {
     }
 }
 
-CoursePaneButtonRow.propTypes = {
-    showSearch: PropTypes.bool.isRequired,
-    onDismissSearchResults: PropTypes.func.isRequired,
-};
-
-export default withStyles(styles)(CoursePaneButtonRow);
+export default withStyles(styles as unknown as Styles<Theme, {}>)(CoursePaneButtonRow);
```
EricPedley: `as unknown as Styles<Theme, {}>` seems kinda hacky. Was there anything wrong with doing the other approach here? (grep the project for `const styles: Styles<Theme, object>` to see examples)

src/components/RightPane/CoursePane/CourseRenderPane.tsx
```diff
@@ -60,67 +64,75 @@ const styles = (theme) => ({
...
 });
 
-const flattenSOCObject = (SOCObject) => {
+const flattenSOCObject = (SOCObject: WebsocResponse): AACourse[] => {
     const courseColors = AppStore.getAddedCourses().reduce((accumulator, { color, section }) => {
         accumulator[section.sectionCode] = color;
         return accumulator;
     }, {});
-    return SOCObject.schools.reduce((accumulator, school) => {
+    return SOCObject.schools.reduce((accumulator: unknown[], school) => {
```
EricPedley: Is this `unknown[]` just temporary until we convert AppStore to get type inference here? If so we should add a TODO comment, and in the comment also mention trying to remove `as AASection` and `as AACourse[]` because those might get automatically inferred once AppStore.getAddedCourses gets an explicit return type.

src/components/RightPane/CoursePane/SchoolDeptCard.tsx
```diff
@@ -1,8 +1,10 @@
 import React, { PureComponent } from 'react';
-import { Grid, Paper, Typography, withStyles, Collapse } from '@material-ui/core';
+import {Grid, Paper, Typography, withStyles, Collapse, Theme} from '@material-ui/core';
 import { ExpandLess, ExpandMore } from '@material-ui/icons';
+import {ClassNameMap} from "notistack";
```
EricPedley: See my last comment about these ClassNameMap imports

src/components/RightPane/CoursePane/SchoolDeptCard.tsx
```diff
@@ -1,8 +1,10 @@
...
+import {Styles} from "@material-ui/core/styles/withStyles";
 
-const styles = (theme) => ({
+const styles = (theme: Theme) => ({
```
EricPedley: Why does this work here without type parameters? It might be worth dropping the type parameters from other style declarations.

src/components/RightPane/CoursePane/SchoolDeptCard.tsx
```diff
@@ -34,11 +36,18 @@ const styles = (theme) => ({
...
+    name: string;
+    type: string;
+}
+
+class SchoolDeptCard extends PureComponent<SchoolDeptCardProps> {
     state = { commentsOpen: false };
 
     render() {
-        const html = { __html: [this.props.comment] };
+        const html = { __html: this.props.comment };
```
EricPedley, line 50: Why remove the list here? Not saying we shouldn't, but this is a change that isn't just adding types which is why I'm curious.

src/components/RightPane/CoursePane/SchoolDeptCard.tsx
```diff
@@ -73,4 +82,4 @@ class SchoolDeptCard extends PureComponent {
     }
 }
 
-export default withStyles(styles)(SchoolDeptCard);
+export default withStyles(styles as unknown as Styles<Theme, {}>)(SchoolDeptCard);
```
EricPedley: Same thing as my last comment about `as unknown as Styles<Theme, {}>`

src/components/RightPane/CoursePane/SearchForm/AdvancedSearch.tsx
```diff
@@ -28,7 +30,29 @@ const styles = {
...
+
+interface AdvancedSearchProps {
+    classes: ClassNameMap
+}
+
+interface AdvancedSearchState {
+    expandAdvanced: boolean;
+}
+
+class ASTF extends PureComponent<AdvancedSearchTextFieldsProps, AdvancedSearchTextFieldsState> {
```
EricPedley: maybe call this UnstyledAdvancedSearchTextFields or something else verbose but more descriptive. I kinda dislike acronyms. You could also just add a comment explaining it.

src/components/RightPane/CoursePane/SearchForm/AdvancedSearch.tsx
```diff
@@ -59,7 +83,7 @@ class AdvancedSearchTextFields extends PureComponent {
         });
     };
 
-    handleChange = (name) => (event) => {
+    handleChange = (name: string) => (event: ChangeEvent<{ checked?: boolean, name?: string | undefined; value: unknown; }>) => {
```
EricPedley: pretty sure `string|undefined` is redundant if you have the `?` on the parameter to make it optional.

src/components/RightPane/CoursePane/SearchForm/DeptSearchBar/DeptSearchBar.tsx
```diff
@@ -20,8 +21,19 @@ const options = depts.map((dept) => {
     };
 });
 
-class DeptSearchBar extends PureComponent {
-    constructor(props) {
+interface DeptSearchBarProps {
+    classes: ClassNameMap
+}
+
+interface Department { deptLabel: string, deptValue: string, isFavorite: boolean }
```
EricPedley: nit: this is too big for a one-liner

src/components/RightPane/CoursePane/SearchForm/DeptSearchBar/MobileDeptSelector.tsx
```diff
@@ -33,7 +41,7 @@ class MobileDeptSelector extends PureComponent {
             <FormControl className={classes.formControl}>
                 <InputLabel>Department</InputLabel>
                 <Select value={this.state.deptLabel} onChange={this.handleChange} fullWidth>
-                    {depts.map((dept) => {
+                    {(depts as unknown as {value: string, label: string}[]).map((dept) => {
```
EricPedley: maybe convert `depts.js` to TS to avoid having to cast here?

src/components/RightPane/CoursePane/SearchForm/FuzzySearch.tsx
```diff
@@ -1,28 +1,40 @@
...
-const emojiMap = {
+const emojiMap: Record<string, string> = {
     GE_CATEGORY: '🏫', // U+1F3EB :school:
     DEPARTMENT: '🏢', // U+1F3E2 :office:
     COURSE: '📚', // U+1F4DA :books:
     INSTRUCTOR: '🍎', // U+1F34E :apple:
 };
 
-const romanArr = ['I', 'II', 'III', 'IV', 'V', 'VI', 'VII', 'VIII'];
+const romanArr: string[] = ['I', 'II', 'III', 'IV', 'V', 'VI', 'VII', 'VIII'];
```
EricPedley: nit: the explicit type is redundant since this is a literal.

src/components/RightPane/CoursePane/SearchForm/FuzzySearch.tsx
```diff
@@ -38,7 +50,7 @@ class FuzzySearch extends PureComponent {
                 break;
             case emojiMap.DEPARTMENT:
                 RightPaneStore.updateFormValue('deptValue', ident[0]);
-                RightPaneStore.updateFormValue('deptLabel', ident.join(':'));
+                RightPaneStore.updateFormValue('deptLabel', (ident as string[]).join(':'));
```
EricPedley: nit: maybe do the `string[]` cast in the declaration instead?

src/components/RightPane/CoursePane/SearchForm/TermSelector.tsx
```diff
@@ -16,17 +21,17 @@ class TermSelector extends PureComponent {
     };
 
     componentDidMount = () => {
-        document.addEventListener('keydown', this.enterEvent, false);
+        // document.addEventListener('keydown', this.enterEvent, false);
```
EricPedley: Why is this being commented out?

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
     "compilerOptions": {
-        "target": "es5",
+        "target": "es2015",
```
EricPedley, line 3: What's the reason behind changing this?

EricPedley (changes requested):
Lots of comments to address, but a lot of them are small stuff. Overall good job 👍

tsconfig.json
ecxyzzy: Not using ES2015 yields this error:
```
src/components/RightPane/CoursePane/SearchForm/AdvancedSearch.tsx:112:20 - error TS2802: Type 'IterableIterator<number>' can only be iterated through when using the '--downlevelIteration' flag or with a '--target' of 'es2015' or higher.

112             ...[...Array(10).keys()].map((v) => v + 2 + ':00am'),
                       ~~~~~~~~~~~~~~~~
```

If we need to stick with ES5 (e.g. for browser compatibility reasons) then we could probably include the `downlevelIteration` flag instead, but either way the tsconfig needs to be changed.

src/components/RightPane/CoursePane/SearchForm/TermSelector.tsx
ecxyzzy: `TermSelector` never had a `enterEvent` listener. I suspect this was probably copypasta'd from another component in the before-times, and nobody caught it because JavaScript lets you do anything with things that don't exist. If it would make more sense to simply remove that line, I'm fine with it.

src/components/RightPane/CoursePane/SchoolDeptCard.tsx
ecxyzzy: The `dangerouslySetInnerHTML` prop takes an object of type `{__html: string}`. I have no idea how it worked when passed an object with a `string[]` instead (probably more JavaScript spaghetti).

src/components/RightPane/CoursePane/SearchForm/TermSelector.tsx
EricPedley: Yeah, just remove it.

src/components/RightPane/CoursePane/SchoolDeptCard.tsx
ecxyzzy: Not sure what you mean because TypeScript balks about `theme` being implicitly typed `any` when I remove the type parameter.

EricPedley: Maybe I got the terminology wrong, but by type parameters, I meant the stuff that goes in angle brackets, like `Theme` and `object` in `Styles<Theme, object>`. Now I'm realizing that this is Theme instead of Style and it's on the parameter of `styles` instead of `styles` itself, which is why it didn't need any parameters. This way seems simpler, so can you just quickly replace all instances of `const styles: Styles<Theme, object> = (theme)` with `const styles = (theme: Theme)` and remove any unused `Styles` imports?

ecxyzzy: Now that I've looked at it, this is actually an error on my part since the return value of the following statement:

```ts
const styles = (theme: Theme) => ({ […] });
```

is implicitly `any`, and therefore in order for things to work properly I need to add the ugly `as unknown as Styles<Theme, {}>` hack that we want to avoid.

Whereas with your approach:

```ts
const styles: Styles<Theme, object> = (theme) => ({ […] });
```

the type of the `theme` parameter is correctly inferred to be `Theme`, probably because of the `withStyles()` call at the very end of the file.

So what I think is the right thing to do instead is to replace all instances of the former with the latter, and also remove the double typecast that we wanted to get rid of anyways.

src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
ecxyzzy: Issue resolved [here](https://github.com/icssc/AntAlmanac/pull/439#discussion_r977992030).

EricPedley (approved):
🚀

PR #447, opened 2022-11-15
Fix Github Actions not checking status properly

Voark:
## Summary
The wait for stack status steps were not working due to my command to query the stack status being wrong. Also, instead of modifying the step status itself I am inserting the stack status into an environmental variable as per Github's newest recommendations.

## Test Plan
As can be seen, the step isn't skipped now (this workflow was run while a stack was currently creating)
image

## Issues
Closes #

### DIFF
--- a/.github/actions/stabilize_stack/action.yaml
+++ b/.github/actions/stabilize_stack/action.yaml
@@ -1,0 +1,51 @@
+name: "Wait for Stack Stabilize"
+
+inputs:
+  stack_name:
+    required: true
+  aws_access_key_id:
+    required: true
+  aws_secret_access_key:
+    required: true
+  
+outputs:
+  stack-exists: 
+    value: ${{ steps.check-stack-exists.outcome == 'success' && !steps.wait-for-stack-delete.outcome == 'success'}}
+
+runs:
+  using: "composite"
+  steps:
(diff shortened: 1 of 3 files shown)


EricPedley (changes requested):
Is there any way we can encapsulate these steps so we don't have to have the code duplicated between these two workflows? Or if we can't just add a comment or note or something that indicates the code is in two places if anyone in the future needs to update it

Voark:
> Is there any way we can encapsulate these steps so we don't have to have the code duplicated between these two workflows? Or if we can't just add a comment or note or something that indicates the code is in two places if anyone in the future needs to update it

There is a way with reusable workflows, I didn't do it before because the environments aren't shared between the normal workflow and the reusable one. However looking at it now I don't remember exactly why I didn't do it... Will try and see what happens.

EricPedley:
What does "stabilize stack" mean?

Voark:
> What does "stabilize stack" mean?

Wait for the stack status to stabilize, i.e. finish anything in progress.

EricPedley (approved):

EricPedley:
Actually maybe we should add that to the description of the action. I'll open another PR for it.

EricPedley:
I'll open another PR for it

PR #450, opened 2022-11-28
Backport ts changes

EricPedley:
## Summary
Merged main, then converted HelpBox to TS and fixed a bug in the recruitment banner. In resolving merge conflicts I changed a line
`flexDirection: 'left'` to `flexDirection:'row'`, which controls the layout of the recruitment banner. If you change it to column it puts the close button in a weird place so I think I've captured the original intent there. The TS compiler was complaining that 'left' wasn't a valid value for that CSS property.

This is a child PR of #411 
## Test Plan
Just make sure it builds correctly in staging and look at the recruitment banner functionality and the helpbox.


Voark:
Somewhere along the line I messed something up and apparently the staging instance tries to hit the prod API. I fixed it in the newest commit in https://github.com/icssc/AntAlmanac/pull/447, so can you approve that PR and merge it into main and this as well?

MinhxNguyen7 (commented):
You'd probably want the images to scale down instead of scroll right?

image

EricPedley:
> You'd probably want the images to scale down instead of scroll right?


Good catch. Since that problem wasn't introduced by the TS rewrite I'll just open an issue for it and leave it for fixing later: #452

Voark:
Don't think you merged the newest action changes in if I'm not mistaken

EricPedley:
Right now loading a schedule doesn't work on this build because when loading an existing schedule, none of the added classes have a `section` property, whereas when adding a class from the AA search menu, it does have a `section` property. Working on a fix rn.
With section:
image
Without section:
image

Based on playing with breakpoints on the `main` build, what's supposed to happen is that somewhere between being loaded from the API and being passed to AppStore, the added courses get the `section` property appended to them.

EricPedley:
@Voark @MinhxNguyen7 @Douglas-Hong This PR is ready for review again now.

src/helpers.ts
```diff
@@ -37,32 +38,51 @@ export async function queryGraphQL(queryString: string): Promise<GradesGraphQLRe
...
+    addedCourses: {
+        section: AASection;
+        deptCode: string;
+        courseNumber: string;
+        courseTitle: string;
+        courseComment: string;
+        prerequisiteLink: string;
+        color: string;
+        term: string;
+        scheduleIndices: number[];
```
Voark: Is there a reason this isn't just `addedCourses: AppStoreCourse[]`?

.github/workflows/stage_pr.yaml
```diff
@@ -2,38 +2,31 @@ name: "Deploy Pull Request"
 
 on: 
   pull_request:
```
Voark, line 4: Looking back at this and Avent's recent PR I remember changing this line (at least locally) to [`pull_request_target`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) which should allow outside contributors to use the action as if they made a pull request from within the repo. However this comes with the security concern of giving secret access to anyone that is able to run an action through pull requests. While the AWS role that is currently stored in the secrets should have pretty limited account access it still is a concern. Wondering if the cons are worth the benefit.

EricPedley: I think we should defer making this decision to another PR since it's not directly relevant to the TS changes. This PR is just for squashing merge conflicts with main.

Voark: Agreed, just remembered again when I read the code

src/helpers.ts
EricPedley: Not really. Resolved in https://github.com/icssc/AntAlmanac/pull/450/commits/05da393a3f56fb39b4060b7dc22085d80e6c51dc

Voark (approved):

PR #453, opened 2022-12-01
Fix duplicate section code bug

EricPedley:
## Summary
- Fixed this user reported bug: 
image
  The bug was caused by courses with duplicate section codes but different terms not being loaded separately.
- Added a separate error case for 404s as opposed to other errors loading schedules.
- Removed the proxy URL from our `package.json` because it was giving me errors, see screenshot below. I don't think we need it anymore since we don't run the backend locally alongside the client.
image

## Test Plan
To reproduce the bug, load username `dupe-test` on the [prod site](https://antalmanac.com). Try deleting one of the classes on schedule 2. It deletes both. To get there from scratch, create two different schedules, add ENGR 1A fall 22 on one, and ENGR 7B winter 23 on the other, then save the schedule and load it.

Try the same process on the staging deployment or locally and verify the bug doesn't happen. Also, try loading a username that doesn't exist to verify the error handling for the 404s works correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,6 @@
             "git add"
         ]
     },
-    "proxy": "http://localhost:8080",
     "browserslist": {
         "production": [
             ">0.2%",
--- a/src/actions/AppStoreActions.js
+++ b/src/actions/AppStoreActions.js
@@ -182,19 +182,23 @@ export const loadSchedule = async (userID, rememberMe) => {
             }
 
             try {
-                const data = await fetch(LOAD_DATA_ENDPOINT, {
+                const response_data = await fetch(LOAD_DATA_ENDPOINT, {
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -62,7 +62,6 @@
             "git add"
         ]
     },
-    "proxy": "http://localhost:8080",
```
Voark, line 65: I think this was here if someone decided to use the backend locally, maybe it's worth leaving in?

Voark: I didn't fully read the pull request description... I didn't seem to be having that issue but it's probably fine removing.

Voark (approved):
Didn't try and specifically reproduce the bug but otherwise LGTM.

PR #454, opened 2022-12-10
Expand calendar earliest time to 6 AM

EricPedley:
## Summary
- Addresses a user feedback form request to expand the calendar view to include 6 AM
TODO:
- Since this is only useful for a handful of users, we should make the default view start at 8 AM, and show earlier times only if you scroll up.
## Test Plan
Add custom events at 6 AM and verify that it works. Also try to add course and custom events at other times to verify it didn't break any typical use cases.


EricPedley:
Closing because it's a strictly worse version of #570

PR #456, opened 2022-12-16
Update all packages to the latest version

falsidge:
## Summary
This updates all the packages.
## Test Plan
Ensure every UI is properly styled
## Issues
Closes #


.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
```
EricPedley, line 2: Can you leave a comment that either
- explains what this does
or
- links to wherever you learned that this file was necessary and what to put in it?
Thanks

package.json
```diff
@@ -64,14 +61,15 @@
     },
     "browserslist": {
         "production": [
-            ">0.2%",
+            ">0.5%",
```
EricPedley, line 64: What does this browserlist thing do and why are we changing it?

package.json
```diff
@@ -64,14 +61,15 @@
...
             "not dead",
-            "not op_mini all"
+            "not op_mini all",
+            "not IE 11"
         ],
         "development": [
-            "last 1 chrome version",
-            "last 1 firefox version",
-            "last 1 safari version"
+            "last 2 chrome version",
```
EricPedley, line 70: Why are `development` and `production` different here and why are we changing it from last version to last 2 versions?

src/components/AppBar/NotificationSnackbar.js
```diff
@@ -54,8 +54,9 @@ class NotificationSnackbar extends PureComponent {
                 key="close"
                 color="inherit"
                 onClick={() => {
-                    this.props.closeSnackbar(key);
+                    size = 'large' > key;
```
EricPedley: What's going on with this string comparison and assignment? Why does it no longer call `closeSnackbar`?

src/components/Home.js
```diff
@@ -1,18 +1,19 @@
...
+import { Grid, CssBaseline, useMediaQuery } from '@mui/material';
 import Calendar from './Calendar/ScheduleCalendar';
 import Bar from './AppBar/CustomAppBar';
 import DesktopTabs from './RightPane/RightPaneRoot';
 import NotificationSnackbar from './AppBar/NotificationSnackbar';
-import { MuiPickersUtilsProvider } from '@material-ui/pickers';
+import { LocalizationProvider } from '@mui/x-date-pickers/LocalizationProvider';
 import MobileHome from './MobileHome';
-import DateFnsUtils from '@date-io/date-fns';
+import AdapterDateFns from '@mui/lab/AdapterDateFns';
```
EricPedley, line 9: Is this being used anywhere? Also, on a higher-level note why do we need to wrap the entire site in this provider component and why are we replacing it?

src/components/RightPane/CoursePane/CourseRenderPane.js
```diff
@@ -17,7 +17,12 @@ import analyticsEnum from '../../../analytics';
 
 const styles = (theme) => ({
     course: {
-        ...theme.mixins.gutters(),
+        paddingLeft: theme.spacing(2),
+        paddingRight: theme.spacing(2),
+        [theme.breakpoints.up('sm')]: {
```
EricPedley, line 22: Is this like a media query or something?

EricPedley (changes requested):
Great work here!
A few UI changes I'd like to see:
- Can you get rid of the transparent white linear gradient overlayed on top of the app bar?
- Can you make the borders of the term selector and search box only be at the bottom? Right now the term selector border overlaps with the word "term".
- Can you make the border of the calendar toolbar less bright so it matches the main branch?

I can see some other places where this version looks different but IMO they look similar enough or better so don't sweat it about making it look exactly like the original for the most part, minus those comments above.

.husky/pre-commit
falsidge: https://typicode.github.io/husky/#/?id=migrate-from-v4-to-v8

This file was created by `husky install` and `husky add`

This file is for husky to lint the project before you commit

Previously this was in `package.json.`

https://blog.typicode.com/husky-git-hooks-javascript-config/

package.json
falsidge: It's for compatibility. MUI supports a newer browserlist. 
https://mui.com/material-ui/migration/migration-v4/#supported-browsers-and-node-versions

It tells React, which browsers we are targeting, A newer target would mean less polyfills or legacy code and faster and smaller bundles.

package.json
falsidge: Development builds are run on the latest browser for quicker testing. We don't need to change it to last 2 browser version but I guess I changed it because it was in the MUI link

src/components/Home.js
falsidge: I asked in the discord. I am not sure we need this anymore as I don't think we use the MUI datepicker anymore for some reason.

src/components/AppBar/NotificationSnackbar.js
falsidge: I am not sure what's happening here. I ran the script to upgrade MUI so I guess it did something weird and mixed up the MUI snackbar with the notistack one.

EricPedley:
We let this one get stale and now there's too many merge conflicts, and @bevm0 is going to work on porting some of this stuff anyways, so gonna close this one to keep the open PR queue clean.

PR #457, opened 2022-12-21
Refactor schedules data structure

Voark:
Thanks to @EricPedley for the TS rewrite otherwise this would be close to impossible. 

This change should not be merged until after the infrastructure changes are merged in first. 

## Related PRs
- Backend: https://github.com/icssc/antalmanac-backend/pull/11
- CDK: https://github.com/icssc/antalmanac-cdk/pull/4

## Why?
The main reason for this change is to make the schedules data structure more intuitive. The previous way of storing schedules looks like (greatly simplified to):
```
{
	addedCourses: [
		{
			courseTitle: "Course 1",
			scheduleIndicies: [0, 1]
		},
		{
			courseTitle: "Course 2",
			scheduleIndicies: [1]
		}
	],
	scheduleNames: ["Schedule 1", "Schedule 2"]
}
```
Basically, schedules are stored as a list of courses that then stores the shedules that contain it. In this case, "Schedule 1" has both "Course 1" and "Course 2" while "Schedule 2" only has "Course 2". 

The biggest drawback of storing schedule data this way is that the most common use case, modifying a single schedule, becomes quite complex. To view a single schedules we have to iterate through all the courses to find which ones are actually in the schedule, and modifying any schedule requires doing more iteration to filter courses and what not. In an attempt to make these tasks easier there are other data structures in AppStore, such as `addedSectionCodes` and `eventsInCalendar` but now data is duplicated and has to be updated every time a change happens. 

### Undo
- Instead of undoing the last delete, the undo button now undoes the last action (ex. add schedules, clear schedules)
- Uses a primitive method where you save the entire schedule data structure up to 50 states, after some slight stress testing it seems to use 5mb out of ~50mb total used by AntAlmanac, which isn't too bad. 

### Save and load schedule index
- When saving and loading schedules, the schedule that is selected is the one that was last open instead of the first schedule.

## Open Issues
- There are now 3 different ways to interact with the schedule data/AppStore. We have `AppStoreActions`, `AppStore` and `Schedules`. I thought about consolidating it but then `AppStore` would be an extremely long file plus there would be many methods that would just be `someMethod(args): return this.schedule.someMethod(args)`.


src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -134,8 +134,9 @@ class CustomEventDialog extends PureComponent<CustomEventDialogProps, CustomEven
     render() {
         return (
             <>
-                {this.props.customEvent ? (
-                    <IconButton onClick={() => this.setState({ open: true })}>
+                {this.props.customEvent !== undefined ? (
+                    // Dumb ternary added to get rid of TypeScript possibly undefined compile error
```
EricPedley: Have you tried optional chaining here to avoid the ternary?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

src/stores/Schedules.ts
```diff
@@ -0,0 +1,488 @@
...
+     */
+    addCourse(newCourse: ScheduleCourse, scheduleIndex: number = this.getCurrentScheduleIndex(), addUndoState = true) {
+        if (addUndoState) {
+            this.addUndoState();
+        }
+        let courseToAdd = this.getExistingCourse(newCourse.section.sectionCode, newCourse.term);
+        if (courseToAdd === undefined) {
+            const setOfUsedColors = new Set(this.getAllCourses().map((course) => course.section.color));
+
+            const color: string =
```
bevm0, line 191:
```suggestion
const color: string = arrayOfColors.find((materialColor) => !setOfUsedColors.has(materialColor)) || '#5ec8e0';
courseToAdd = {
  ...newCourse,
  section: {
    ...newCourse.section,
    color
  }       
};
```

src/stores/Schedules.ts
```diff
@@ -0,0 +1,488 @@
...
+                arrayOfColors.find((materialColor) => {
+                    if (!setOfUsedColors.has(materialColor)) return materialColor;
+                    else return undefined;
+                }) ?? '#5ec8e0';
+            courseToAdd = newCourse;
+            courseToAdd.section.color = color;
```
ap0nia:
```suggestion
const color: string = arrayOfColors.find((materialColor) => !setOfUsedColors.has(materialColor)) || '#5ec8e0';
courseToAdd = {
  ...newCourse,
  section: {
    ...newCourse.section,
    color
  }       
};
```

src/stores/Schedules.ts
```diff
@@ -0,0 +1,488 @@
...
+        return false;
+    }
+
+    // --- Other methods ---
+    /**
+     * Appends a copy of the current state to previous states to revert to.
+     * Previous states are capped to 50 to prevent memory usage from going infinite.
+     */
+    addUndoState() {
+        const clonedSchedules = JSON.parse(JSON.stringify(this.schedules)) as Schedule[]; // Create deep copy of Schedules object
```
ap0nia:
```suggestion
        const clonedSchedules = structuredClone(this.schedules);
```

Voark: Damn I swear when I looked up deep copying I didn't see this.

ap0nia: Bro same, I learned about it from  [YouTube shorts](https://youtube.com/shorts/XK0V0E3bA-M?feature=share) lol. I swear StackOverflow never mentioned anything about this wizardry

src/stores/calenderizeHelpers.ts
```diff
@@ -2,10 +2,10 @@ import { CourseEvent, CustomEvent } from '../components/Calendar/CourseCalendarE
 import AppStore from './AppStore';
 
 export const calendarizeCourseEvents = () => {
```
ap0nia:
```suggestion
export const calendarizeCourseEvents = (currentCourses: Course[]) => {
```
I personally think this would make the helper functions perfectly decoupled; there's this weird circular dependency between helpers and the AppStore: helpers ask the AppStore for data and the AppStore asks the helpers for data;

By just passing in `currentCourses` as a param, the `AppStore.schedule.getCurrentCourses` can easily be called within the AppStore itself in order to provide the data to the helper. Helpers don't need to retrieve this data from the AppStore themselves, so they would become independent of the AppStore's implementation, and the circular dependency is resolved.

This does require that you fix the function signature of each helper function to ask for any data it needs from AppStore as params, and then adjust the calls from AppStore itself. I think it would be a good improvement, but it may take a bit of work, and I'll also address it during my own re-factoring work if it's not a big concern at the moment.

src/stores/calenderizeHelpers.ts
```diff
@@ -64,10 +63,10 @@ export const calendarizeCourseEvents = () => {
 };
```
ap0nia, line 63: Also, you don't need to type assert the array if you use functional programming, e.g. map, filter, etc. (those are more TypeScript friendly). Here's an example implementation that seems to get me pretty close (there's issues with the final typing because the CourseEvent extends from the Event class, which the returned object lacks properties of unless you assert it:

```ts
export function calendarizeCourseEvents(currentCourses: Course[]) {
  return currentCourses.map((course) => (
    course.section.meetings
      .filter((meeting) => meeting.time.replace(/\s/g, '') !== 'TBA')
      .map((meeting) => {
        const timeString = meeting.time.replace(/\s/g, '');
        const [, startHrStr, startMinStr, endHrStr, endMinStr, ampm] = timeString.match(
          /(\d{1,2}):(\d{2})-(\d{1,2}):(\d{2})(p?)/
        ) as RegExpMatchArray;

        let startHr = parseInt(startHrStr, 10);
        const startMin = parseInt(startMinStr, 10);
        let endHr = parseInt(endHrStr, 10);
        const endMin = parseInt(endMinStr, 10);

        /**
         * map each day to a boolean indicating whether the course meets on that day
         */
        const dates = days.map(day => meeting.days.includes(day))

        if (ampm === 'p' && endHr !== 12) {
          startHr += 12;
          endHr += 12;
          if (startHr > endHr) startHr -= 12;
        }

        return dates
          .filter((shouldBeInCal) => shouldBeInCal)
          .map((_date, index) => {
            const newEvent = {
              color: course.section.color,
              term: course.term,
              bldg: meeting.bldg,
              instructors: course.section.instructors,
              sectionCode: course.section.sectionCode,
              sectionType: course.section.sectionType,
              start: new Date(2018, 0, index, startHr, startMin),
              finalExam: course.section.finalExam,
              end: new Date(2018, 0, index, endHr, endMin),
              isCustomEvent: false as const,
            };
            return newEvent;
          });
      }).flat()
  )).flat();
}
```

src/stores/calenderizeHelpers.ts
Voark: At some point I was going to move the calenderizing functions into `Schedule.ts` but I never got around to it. Will probably add it in an upcoming commit.

src/stores/calenderizeHelpers.ts
Voark: Addressed offline, in summary we can consider this in another PR.

src/stores/calenderizeHelpers.ts
Voark: Decided `Schedules.ts` becomes a little too bloated, changes were implemented.

Voark:
Linked backend and CDK prs in description.

src/stores/Schedules.ts
```diff
@@ -0,0 +1,509 @@
...
+    /**
+     * @return current schedule index
+     */
+    getCurrentScheduleIndex() {
+        return this.currentScheduleIndex;
+    }
+
+    // --- Schedule related methods ---
+    /**
+     * @return the number of schedules
```
EricPedley: Can you remove the autogenerated docstrings if the method name is the same thing as the description?

EricPedley (changes requested):
2 small changes to look at then you're good

ap0nia (approved):
I think it looks good! I've been using it as a guideline for state management in my `aponia` branch and it felt solid 🚀

src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
Voark: The problem is optional chaining evaluates to undefined and the function signature only takes numbers. Also I was referring to L139, IDK if that was clear by the comment.

EricPedley (changes requested):
Client code looks fine rn but we're getting fucked by CORS on the staging deployment. 
image
Even if it's gonna work in prod because there isn't a subdomain preventing the CORS match, I think this'll break future staging deployments right? Can you fix this before we merge?

EricPedley (approved):
Discussed outside of the PR - this won't break prod so we'll resolve the cors problem later.

PR #458, opened 2023-01-02
Adjusted min-width to include tablet screens

### LINKED ISSUE #250: Support Tablet Screen Sizes
> Another potential change we can make in the future is support for tablets that are oriented vertically. If I switch to the dimensions of an iPad Air/Mini, the page is stuck on the calendar and doesn't show the Calendar and Search tabs on the top.
> 
> IPadDimensions
> 
> Otherwise, this looks good to go!

_Originally posted by @Douglas-Hong in https://github.com/icssc-projects/AntAlmanac/pull/247#pullrequestreview-885483583_

sirAvent:
## Summary
Between window widths of 750px - 960px, the page is stuck on the calendar and doesn't show the Calendar and Search tabs on the top. I didn't touch the AppBar because it seems fine to me at that width, although the elements are slightly clipping each other.

## Test Plan
I checked various screen widths, to see when the bug occurs.

## Issues
Closes #250 

## Future Followup  
I adjusted the breakpoint for mobile view to include the buggy widths. A future change could be a different design on tablet widths.

### DIFF
--- a/src/components/Home.js
+++ b/src/components/Home.js
@@ -9,7 +9,7 @@ import MobileHome from './MobileHome';
 import DateFnsUtils from '@date-io/date-fns';
 
 const Home = () => {
-    const isMobileScreen = useMediaQuery('(max-width: 750px)');
+    const isMobileScreen = useMediaQuery('(max-width: 960px)');
 
     return (
         <MuiPickersUtilsProvider utils={DateFnsUtils}>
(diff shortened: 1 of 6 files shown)


Voark:
Not your fault of course, but should probably put all those constants in a global variable to make it easier to change stuff like this in the future. No idea where it would go though.

EricPedley:
> Not your fault of course, but should probably put all those constants in a global variable to make it easier to change stuff like this in the future. No idea where it would go though.

The specific place doesn't matter because it'll be easy to refactor later. Maybe just add a file called `globals.js` under `src` and import it wherever we need magic number constants.

EricPedley:
Completed in #538

PR #460, opened 2023-01-16
Fixed GE-III Bug -- Need Future Improvement

### LINKED ISSUE #448: GE III Bug
Searching for GE III classes is broken. It only shows BIO SCI 48 and the department of education but no classes for it.

image
The schedule of classes successfully displays all of them.
image

JacE070:
## Summary
I fixed the GE III bug by deleting the empty sectionCodes parameterwhen being sent to PeterPortal API. In my experiments, the API sometimes returns empty results when the request is with empty sectionCodes parameter.
## Test Plan
I searched for only GE III, and it worked fine.
## Issues
Closes https://github.com/icssc/AntAlmanac/issues/448
## Future Followup  
Deleting the empty param may not be sustainable. The issue needs further inspection on the behavior of the API.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -100,6 +100,10 @@ export function clearCache() {
 export async function queryWebsoc(params) {
     // Construct a request to PeterPortal with the params as a query string
     const url = new URL(PETERPORTAL_WEBSOC_ENDPOINT);
+    // TODO: API may return empty result if params has empty sectionCodes
+    if (params.sectionCodes === '') {
+        delete params.sectionCodes;
+    }
     const searchString = new URLSearchParams(params).toString();
     if (websocCache[searchString]?.timestamp > Date.now() - 30 * 60 * 1000) {
         //if cache hit and less than 30 minutes old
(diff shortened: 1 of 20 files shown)


JacE070:
Pardon me. Forgot to update my fork. Already done it now.

src/helpers.js
```diff
@@ -100,6 +100,10 @@ export function clearCache() {
 export async function queryWebsoc(params) {
     // Construct a request to PeterPortal with the params as a query string
     const url = new URL(PETERPORTAL_WEBSOC_ENDPOINT);
+    // TODO: API may return empty result if params has empty sectionCodes
```
EricPedley, line 103: Can you investigate this further? See if the API docs mention anything about this, and if not try looking through the PeterPortal API source code for the keyword `sectionCodes` to find where it handles that parameter. If that seems too hard we can discuss more offline about how to dive deep into stuff like this.

I'm asking because knowing more about this will help make sure this code is bug-free.

EricPedley (changes requested):

src/helpers.js
JacE070: Got it

JacE070: Could you try to reproduce the GE-III bug? I tried using multiple devices, but the correct results showed up.


-----


I also tested the other GEs and they also work well.

JacE070:
Re-open PR in TypeScript

PR #461, opened 2023-01-16
Dark Mode Screenshots Bug

JacE070:
## Summary
I fixed the Dark Mode Screenshots Bug by 
1) Setting the backgroundColor parameter to dark color in the html2canvas options. 
2) Creating dark mode CSS style for rbc-headers and rbc-timeslot labels, and using them if it is in dark mode.
## Test Plan
I set the theme to light and dark and took screenshots separately.
## Issues
Closes https://github.com/icssc/AntAlmanac/issues/442
## Future Followup  
~~When the auto mode is turned on, the screenshot will, by default, apply a light theme.
This might not be good for users using auto mode (they may be using the dark theme on their devices).~~ Solved

PR #462, opened 2023-01-28
Dark Mode Screenshots Bug

### LINKED ISSUE #442: Dark Mode Screenshots
If the theme is set to dark, make the screenshot render with a dark background.

JacE070:
## Summary
I fixed the Dark Mode Screenshots Bug by 
1) Setting the backgroundColor parameter to dark color in the html2canvas options. 
2) Creating dark mode CSS style for rbc-headers and rbc-timeslot labels, and using them if it is in dark mode.
3) Fixing bug in my previous PR https://github.com/icssc/AntAlmanac/pull/461
## Test Plan
I set the theme to light and dark and took screenshots separately.
## Issues
Closes https://github.com/icssc/AntAlmanac/issues/442
## Note
This PR is a re-open version of my previous PR https://github.com/icssc/AntAlmanac/pull/461 because of the TS codebase conversion.

### DIFF
--- a/src/components/Calendar/CalendarRoot.tsx
+++ b/src/components/Calendar/CalendarRoot.tsx
@@ -9,7 +9,6 @@ import React, { PureComponent, SyntheticEvent } from 'react';
 import { Calendar, DateLocalizer, momentLocalizer, Views } from 'react-big-calendar';
 import ReactDOM from 'react-dom';
 
-import { isDarkMode } from '../../helpers';
 import AppStore from '../../stores/AppStore';
 import CalendarToolbar from './CalendarToolbar';
 import CourseCalendarEvent, { CalendarEvent } from './CourseCalendarEvent';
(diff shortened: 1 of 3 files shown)


src/components/Calendar/CalendarRoot.tsx
```diff
@@ -223,6 +224,18 @@ class ScheduleCalendar extends PureComponent<ScheduleCalendarProps, ScheduleCale
         canvas.style.color = 'black';
         calendarHeader.style.marginRight = '0px';
 
+        // Set each rbc header and time slot text color to white if the theme is dark
```
EricPedley: I think you should change the contents of this comment. The code is pretty self-explanatory so you don't need to explain _what_ it's doing. However, I have no clue _why_ we're changing these properties of rbc_headers and rbc_time_slots, or even where in the DOM they are. The comment should explain the _why_ of this block, not the _what_.

src/components/Calendar/CalendarRoot.tsx
```diff
@@ -223,6 +224,18 @@ class ScheduleCalendar extends PureComponent<ScheduleCalendarProps, ScheduleCale
...
+        const rbc_headers = document.getElementsByClassName('rbc-header');
+        const rbc_time_slots = document.getElementsByClassName('rbc-label');
+
+        if (isDarkMode()) {
+            for (let i = 0; i < rbc_headers.length; i++) {
```
EricPedley: Nitpick: can you change these to `for of` loops?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

src/components/Calendar/CalendarRoot.tsx
```diff
@@ -223,6 +224,18 @@ class ScheduleCalendar extends PureComponent<ScheduleCalendarProps, ScheduleCale
         canvas.style.color = 'black';
         calendarHeader.style.marginRight = '0px';
 
+        // Set each rbc header and time slot text color to white if the theme is dark
+        const rbc_headers = document.getElementsByClassName('rbc-header');
```
EricPedley: I'd like to avoid using any native `document.getElement...` functions because it's hard to tell what exactly they're changing. Can you find the JSX component that needs these styles and apply them there? If that isn't possible, just comment what element this is changing. It can be in the same code comment that resolves my above code review comment.

src/components/Calendar/CalendarRoot.tsx
```diff
@@ -231,6 +244,17 @@ class ScheduleCalendar extends PureComponent<ScheduleCalendarProps, ScheduleCale
             canvas.style.color = oldColor;
             calendarHeader.style.marginRight = oldMargin;
 
+            // Set text color back
```
EricPedley: Again, this code block is self-explanatory with regards to _what_ it does, but I can't tell _why_ we need to reset the styles. Can you change the comment to explain the _why_, not the _what_?

src/components/Calendar/CalendarRoot.tsx
```diff
@@ -231,6 +244,17 @@ class ScheduleCalendar extends PureComponent<ScheduleCalendarProps, ScheduleCale
...
+            if (isDarkMode()) {
+                for (let i = 0; i < rbc_headers.length; i++) {
+                    // delete rbc_headers[i].style;
```
EricPedley: If you don't need this line anymore can you delete it instead of commenting it out?

src/components/Calendar/calendar.css
```diff
@@ -6,6 +6,14 @@
     justify-content: center;
 }
 
+.rbc-header.dark-mode {
+    border-bottom: none;
+    color: white;
+    display: inline-flex;
+    align-items: center;
```
EricPedley: Why does it need these `display`, `border-bottom`, `align-items`, and `justify-content` properties if it's just a modifier for dark mode? My intuition says that all you'd need to put here is the `color` property, so can you simplify this by removing the unnecessary properties or explain why they actually are necessary?

src/components/Calendar/calendar.css
```diff
@@ -36,6 +44,11 @@
     font-size: 0.9rem;
 }
 
+.rbc-label.dark-mode {
+    font-size: 0.9rem;
```
EricPedley: Same idea as the previous comment, why would the font size need to change for dark mode?

EricPedley (changes requested):
Good job on coming up with a working fix. I'm leaving a bunch of code quality and readability comments for you to consider before I approve this. Since we pass this codebase on to an entirely new group every couple of years it's important that everything is as clear as possible. Feel free to push back on any of my comments you don't think would actually help that goal.

JacE070:
So weird. Now it's working after I added the background color props for _html2canvas_ **AND** removed all the codes that handle the color in handleTakeScreenshot().

When **ONLY** removing the legacy code: the font color changes corresponding to the theme. The background color keeps white.
Thus I only added one line of code to change the background color to fix this.

I would guess the _html2canvas_ only captures the style of column and time slots headers (of the calendar) in dark mode but not the style of background.

EricPedley (approved):
Ship it!
I'm happy that the solution was to _remove_ lines of code.
```
                              __    __    __
                             |==|  |==|  |==|
                           __|__|__|__|__|__|_
                        __|___________________|___
                     __|__[]__[]__[]__[]__[]__[]__|___
                    |............................o.../
                    \.............................../
               ,~')_,~')_,~')_,~')_,~')_,~')_,~')/,~')_
```

ap0nia:
You can also grab a reference to the MUI theme and ask for its current background color; e.g I use `useTheme` and then reference the paper background; this could be more versatile if you mount the calendar inside of a MUI Paper, then you'll always get the exact matching background dynamically

[sample here](https://github.com/icssc/AntAlmanac/blob/aponia/src/components/Buttons/Screenshot.tsx)

JacE070:
> You can also grab a reference to the MUI theme and ask for its current background color; e.g I use `useTheme` and then reference the paper background; this could be more versatile if you mount the calendar inside of a MUI Paper, then you'll always get the exact matching background dynamically
> 
> [sample here](https://github.com/icssc/AntAlmanac/blob/aponia/src/components/Buttons/Screenshot.tsx)

Bro that's clever. Thanks!

PR #463, opened 2023-01-28
GE-III Bug

### LINKED ISSUE #448: GE III Bug

JacE070:
## Summary
I fixed the GE III bug by setting the height props from 300 to 200. This allows for rendering more content in the first round of render.

Can find more documentation in [react-lazyload](https://www.npmjs.com/package/react-lazyload).
## Test Plan
I searched for multiple GEs, and it worked fine.
## Issues
Closes https://github.com/icssc/AntAlmanac/issues/448

### DIFF
--- a/src/components/RightPane/CoursePane/CourseRenderPane.tsx
+++ b/src/components/RightPane/CoursePane/CourseRenderPane.tsx
@@ -1,12 +1,12 @@
-import { Button, Grid, Paper, Theme  } from '@material-ui/core';
+import { Button, Grid, Paper, Theme } from '@material-ui/core';
 import { withStyles } from '@material-ui/core/styles';
-import { ClassNameMap , Styles } from '@material-ui/core/styles/withStyles';
+import { ClassNameMap, Styles } from '@material-ui/core/styles/withStyles';
 import CloseIcon from '@material-ui/icons/Close';
 import React, { PureComponent } from 'react';
 import LazyLoad from 'react-lazyload';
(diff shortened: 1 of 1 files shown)


EricPedley (approved):
I'm merging this rn because it fixes the bug immediately and the diff is tiny, but can you explain exactly _why_ 300 pixels didn't work for some specific screen sizes and why 200 fixes it? I can tell empirically that it works on all the search queries and screen sizes I've tried on the staging deployment so far but I can't exhaustively search every configuration, so it would be nice to know that this solution actually theoretically makes it work for any screen size.

JacE070:
@EricPedley In the documentation, the height props *sets the height of the placeholder that the Lazyloader uses to measure if the component is visible in the first round of render.*
Therefore, theoretically, the smaller value of height more easily let the LazyLoader recognize there are something to render.

image
As you can see here, under the School of Education there are two lazyload-placeholder, both of which has height of 120 pixels.
If we are using Height = 300, these two lazyload-placeholders will not be rendered since 120 * 2 = 240 < 300. 
So we should use at most height = 240 to render the two rows.

After all, I can say the safest way is to set the estimateHeight to 120. I didn't do this since the offset props in the configuration can also contribute to the render.

p.s. increase offset to 700 can also solve this

PR #465, opened 2023-01-30
Duplicate section meetings from different locations

### LINKED ISSUE #257: Classes with multiple locations show up multiple times on calendar
User Report:
image

When a class has 2 locations, it creates 2 events on the calendar. Ideally it should only create one event with both locations. In this case, it seems like it would be a pretty easy `if statement` to check if the class times are the same.

This formatting can mean two different things when there are multiple days (e.g. MW)
- the class is offered at the same time in both rooms (usually the room would be conjoined)
OR
- On Monday the class is in one room and on Wednesday it’s in a different room

When fixing this we just need to make sure we don’t break .ics downloads or squash classes that are locations for different times.

MinhxNguyen7:
## Summary

The meeting duplication was on the API end. I just looked through the response and merged duplicate meetings (separating them with an ampersand). I think this should be fixed on the API side, but this works for now. If we stick to this fix, we should probably refactor meetings.bldg: string -> meetings.locations: string[].

Also, it seems that the map link in the event popup still works despite the mangled name.

## Test Plan

Make sure that everything renders properly in different screen sizes. 

## Issues

Closes #257

### DIFF
--- a/src/components/RightPane/SectionTable/SectionTableLazyWrapper.tsx
+++ b/src/components/RightPane/SectionTable/SectionTableLazyWrapper.tsx
@@ -2,6 +2,7 @@ import React, { Suspense } from 'react';
 
 import { SectionTableProps } from './SectionTable.types';
 
+// This should be in SectionTable.tsx, IMO
 const SectionTable = React.lazy(() => import('./SectionTable'));
 
 export default function SectionTableLazyWrapper(props: SectionTableProps) {
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -211,12 +211,11 @@ function removeDuplicateMeetings(websocResp: WebsocResponse): WebsocResponse {
 
                             // Add the building to existing meeting instead of creating a new one
                             if (sameDayAndTime && !sameBuilding) {
-                                const newMeeting: Meeting = {
+                                existingMeetings[i] = {
                                     days: existingMeetings[i].days,
                                     time: existingMeetings[i].time,
                                     bldg: existingMeetings[i].bldg + ' & ' + meeting.bldg,
                                 };
-                                existingMeetings[i] = newMeeting;
                                 isNewMeeting = false;
                             }
                         }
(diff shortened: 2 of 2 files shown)


src/components/RightPane/SectionTable/SectionTableLazyWrapper.tsx
```diff
@@ -2,6 +2,7 @@ import React, { Suspense } from 'react';
 
 import { SectionTableProps } from './SectionTable.types';
 
+// This should be in SectionTable.tsx, IMO
```
EricPedley, line 5: yes. open another PR for it

EricPedley (approved):
🚀

EricPedley:
@Voark The preview deployment is failing with the error below. Can you look into this before we merge this PR?
```
github-actions-stack-465 failed: Error: The stack named github-actions-stack-465 failed creation, 
it may need to be manually deleted from the AWS console: ROLLBACK_COMPLETE
    at waitForStackDeploy (/home/runner/work/AntAlmanac/AntAlmanac/.github/workflows/actions_stack/node_modules/aws-cdk/lib/api/util/cloudformation.ts:311:11)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at prepareAndExecuteChangeSet (/home/runner/work/AntAlmanac/AntAlmanac/.github/workflows/actions_stack/node_modules/aws-cdk/lib/api/deploy-stack.ts:376:26)
    at CdkToolkit.deploy (/home/runner/work/AntAlmanac/AntAlmanac/.github/workflows/actions_stack/node_modules/aws-cdk/lib/cdk-toolkit.ts:209:24)
    at initCommandLine (/home/runner/work/AntAlmanac/AntAlmanac/.github/workflows/actions_stack/node_modules/aws-cdk/lib/cli.ts:342:12)

```

Voark:
Sorry missed the ping somehow, looking into it. That being said it shouldn't be an issue because the prod site is deployed to Github Pages not AWS.

Voark:
Looks like AWS had some random issue, just restarting the deploy fixed it.

PR #467, opened 2023-02-02
initial commit for monorepo re-write

ap0nia:
## Summary
- Re-wrote the project to accommodate TypeScript monorepo architecture

## Test Plan
- The project structure itself is very minimal
- it can be made to accomodate Turborepo, lint-staged, husky, etc. as needed for project management

## Issues
- A lot of the code in Antalmanac is using older APIs that's now broken
- I haven't migrated anything besides the Antalmanc website; CI/CD integrations will need to be patched to be compatible

Closes #


EricPedley:
For documentation's sake, here's a summary of the convo we had on Discord:
- The diff for this PR is massive and there's no way I can look at everything at once
  - It's also all in one commit, so it's hard to figure out the purpose of different changes.
- The PR does multiple things at once: moving files, updating old libraries, and making styling changes, which also makes it hard to approve all at once.
- However, I do agree with the sentiment of this PR: having a monorepo would make development easier so we can share type definitions between the frontend and  backend, and we should update libraries we're using old versions of like Material UI.

The plan moving forward is to put the updating libraries in its own PR, which I can probably approve pretty easily, and to put the monorepo migration in another PR where we can discuss with @Voark about the feasibility of such a change, since he knows way more about the backend and deployments than I do, and is in the process of rewriting the way we store schedules.

PR #471, opened 2023-02-06
Missing waypoints for course section with fake locations

### LINKED ISSUE #390: Missing waypoints for course section with fake locations
If a section with a "fake" location (e.g. `TBA`, `VRTL REMOTE`, `ON LINE`) is present in a schedule, waypoints do not render properly on the map. The "all" view shows no waypoints or paths at all, and views for individual days show waypoints only if all sections with meetings that day have a real location. Any paths between sections with real locations are still rendered, though without waypoints.

JacE070:
## Summary
Fixed the issue of missing waypoints with the presence of course sections with fake locations by filtering out them.
## Test Plan
I tested the cases that it only has sections with real locations, it only has sections with fake locations, and mixed.
The map functions well.
## Issues

Closes https://github.com/icssc/AntAlmanac/issues/390

### DIFF
--- a/src/components/RightPane/Map/UCIMap.tsx
+++ b/src/components/RightPane/Map/UCIMap.tsx
@@ -2,7 +2,7 @@ import '../../../../node_modules/leaflet.locatecontrol/dist/L.Control.Locate.min
 
 import Leaflet, { Control, LeafletMouseEvent } from 'leaflet';
 import React, { PureComponent } from 'react';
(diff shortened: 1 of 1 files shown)


EricPedley (changes requested):
Can you remove the submodule you accidentally added? Should be as simple as deleting the folder.
image

EricPedley (approved):
🚀

PR #475, opened 2023-02-07
Schedule Help Box Removed

### LINKED ISSUE #449: Remove schedule help box
Now that it isn't the start of the year anymore we should comment out the scheduling tips box. Maybe as a stretch goal, make it only visible during certain weeks of the year so we don't need to manually add and remove it periodically.
screenshot with the box circled

mina03333:
## Summary
Removed schedule help box and made it visible only during September and October.
Screenshot 2023-02-06 161948
## Test Plan
Make sure that you no longer see this help box unless it's the beginning of the year.
## Issues
Closes #449


EricPedley (changes requested):
Can you move the conditional from HelpBox to its parent component, SearchForm? It makes more sense for the logic to be "if it's the beginning of the year, render the helpbox, otherwise, don't", rather than the current way you're implementing it, "if it's the beginning of the year, the helpbox contains its normal contents, otherwise, it's an empty component". Should be a quick fix then we can approve and merge this.

EricPedley (approved):
Ok looks good. Two comments that don't need addressing:
- I renamed the variable you were using to store the current month and added a comment to clarify that it starts at 0=January, not 1=January
- I resolved the merge conflicts with main. In this case, it was a one-line conflict with an import, and I just took the version from main.

Normally I'd put these as comments asking for you to do them yourself but they're really small and this PR is already two weeks old.
Also, make sure you press "re-request review" when you address the comments so I know to look at your PR again.

PR #470, opened 2023-02-03
Update packages, switch to Vite and Yarn

EricPedley:
This is just duplicated from #469 but reopened so that we get the preview deployments
#

> # Main Changes
> - mostly changed only the import statements e.g. 
>   `import ... from '@material-ui/core'` -> `import ... from '@mui/material'`
>   `import ... from '@material-ui/styles'` -> `import ... from '@mui/styles'`
> 
> ## Fixes
> - *all* technical debt (aside from `react-leaflet`, since I don't know how to migrate it from v2 to v4)
> - closes #267 
> - closes #279 
> - closes #335 
> 
> ## Summary
> - no files moved
> - aside from import statements, files changed very little, e.g. to fix some TypeScript/ESLint warnings
> - `yarn` is the package manager because it handles peer dependencies better
> 
> commands:
> - `npm run dev` will start a dev server at [http://localhost:5173](http://localhost:5173)
> - `npm run build` will build the app into the `dist` folder
> - `npm run start` will start the built app locally at [http://localhost:4173](http://localhost:4173)
> 
> ## Files with extra changes
> - [src/components/App.tsx](https://github.com/bevm0/AntAlmanac/blob/main/apps/antalmanac/src/components/App.tsx#L57) has a `GlobalStyles` that overrides the link colors instead of `createTheme`
> - [src/components/AppBar/CustomAppBar](https://github.com/bevm0/AntAlmanac/blob/main/apps/antalmanac/src/components/AppBar/CustomAppBar.tsx#L77) has an array of uninitialized components around line 77;
>    they used to be initialized and passed as children; now they're created during the iteration
> - [src/components/Home](https://github.com/bevm0/AntAlmanac/blob/main/apps/antalmanac/src/components/Home.tsx#L3) uses the `LocalizationProvider` from `@mui/x-date-pickers` and provides the `DateFnsUtils` as the `dateAdapter`
> - [src/components/RightPane/Map/UCIMap](https://github.com/bevm0/AntAlmanac/blob/main/apps/antalmanac/src/components/RightPane/Map/UCIMap.tsx#L22) uses `Control.Locate` from the `leaflet.locatecontrol` library instead of `leaflet` now
> 
> ## TODO
> - MUI migrated to the [`sx prop`](https://mui.com/material-ui/customization/how-to-customize/#1-one-off-customization)  as their preferred styling method instead of `withStyles`
> - I presume this is the cause of styling issues
> 
> ## DX Improvements
> - starting the development server and building the app should be ***blazingly fast***
>

PR #477, opened 2023-02-07
Fix techdebt theme

ap0nia:
## Summary
Replicated #456 's changes for the TypeScript codebase


index.html
```diff
@@ -44,11 +44,9 @@
...
         <meta property="twitter:image" content="https://antalmanac.com/logo.png" />
 
-        <!--
-        <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
-        <link rel="shortcut icon" href="%PUBLIC_URL%/favicon.ico" />
-        <link rel="stylesheet" href="%PUBLIC_URL%/google_font_roboto_300_400_500_700.css" />
-        -->
+        <link rel="manifest" href="/manifest.json" />
+        <link rel="shortcut icon" href="/favicon.ico" />
+        <link rel="stylesheet" href="/google_font_roboto_300_400_500_700.css" />
```
EricPedley, line 49: Just noticed it on this PR, but why is index.html in the root directory instead of in `public` now? Also, why does the font look different on the preview deployment?

src/components/App.tsx
```diff
@@ -35,14 +35,84 @@ class App extends PureComponent {
...
+                //     main: '#305db7',
+                // },
+                // white: {
+                //     main: '#d5d5d5',
+                //     contrastText: '#000',
+                // },
+                // clearButton: {
+                //     main: '#f50057',
+                // },
+                ...(this.state.darkMode
```
EricPedley, line 59: Can we simplify this by just putting the ternary as the value of `pallette` instead of doing this destructuring, since the other properties are commented out? Also, if we're not going to use those commented-out values I think we should just delete them.

src/components/AppBar/ImportStudyList.tsx
```diff
@@ -1,15 +1,15 @@
-import { PostAdd } from '@mui/icons-material';
+import { ContentPasteGo } from '@mui/icons-material';
```
EricPedley, line 1: Good change 👍

src/components/AppBar/ImportStudyList.tsx
```diff
@@ -137,11 +137,10 @@ class ImportStudyList extends PureComponent<ImportStudyListProps, ImportStudyLis
     }
 
     render() {
```
EricPedley, line 139: Can you get this component to look more like it did originally?
Original:
image

Yours:
image

EricPedley (changes requested):
Code looks mostly fine with a few comments. Besides those, here's some other things I'd like fixed:
- The favicon doesn't show up
- The font is different
- A lot of the components got bright white borders with these changes, and I think they looked better before
- In the search results page, the copy schedule and clear schedule buttons look different
- The different tabs (search, added classes, map) get blue text color when they're selected.
- The course info button on the search results is blue when it was previously light grey
- The class statuses don't have color anymore
- In the maps tab, if you click on a pin, the contrast on the "directions" button is pretty bad
- The "Add Custom" popup elements overlap each other
Whenever I say something is different here, the implication is that I think it should look like it did originally. There are some style changes that I thought looked better like the new icon for import study list that you should keep, but for everything else can you either fix them or let me know why that change should stay when we merge this?

After resolving those we can merge this one and #470.

ap0nia:
Good demo of what the project looks like in mui5; gonna work on function components first

PR #478, opened 2023-02-08
added generic function param to delete action

ap0nia:
## Summary
fixed the function definition of `getEventsAfterDeleteSchedule` to not require type asserting

## Fixes
- return type is correctly inferred by the param at call
- better than type asserting


EricPedley:
Closing because this function will get removed when we merge #457 (according to @bevm0)

PR #479, opened 2023-02-09
simplified copy to clipboard

ap0nia:
## Summary
- use navigator API instead of document

- why would you make a new DOM element to write a raw value to the clipboard????????????????????????????????????????
- caveat: navigator API only works in "secure contexts", e.g. localhost and https in production
- https://stackoverflow.com/questions/51805395/navigator-clipboard-is-undefined

### DIFF
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -353,13 +353,7 @@ export const warnMultipleTerms = (terms: Set<string>) => {
 
 export function clickToCopy(event: React.MouseEvent<HTMLElement, MouseEvent>, sectionCode: string) {
     event.stopPropagation();
-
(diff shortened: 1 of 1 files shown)


EricPedley:
You accidentally made this PR against the unmerged refactor schedules branch, so I did an interactive rebase locally to pick out all the commits except where you actually changed the clipboard code, and squashed it into one commit.

EricPedley (approved):
```
    __4___
 _  \ \ \ \
<'\ /_/_/_/
 ((____!___/)
  \0\0\0\0\/
 ~~~~~~~~~~~
```
Ship it

PR #476, opened 2023-02-07
Default class colors

teresa-liang:
## Summary
Added logic that would make the sections (Lec, Dis, Lab, etc.) of the same course the same color on the schedule 
## Test Plan
Can manually check if sections of the same course are the same color
## Issues

Closes #236

PR #481, opened 2023-02-11
Patchnotes modal

### LINKED ISSUE #373: Create Developer Log
- [ ] Create a dev log, documenting changes/new feature every quarter
- [ ] Create a popup page for the new Spring features.

### LINKED ISSUE #425: Patch Notes Modal
Create a modal to let people know what we accomplished every month. We can copy a lot from the About page. Something like this: 
image
To keep low overhead on this, I think it's fine to just hard-code the notes for the first few months until we think of an elegant way to load them dynamically or just settle for the hard-coding.

johnlorenzini:
## Summary
Created a modal to let people know what we accomplished every month. Dialog is presented only on first load using localStorage.
image

## Test Plan
Ensure that dialog appears once, and can be closed by clicking outside the box, using "escape" on the keyboard, or clicking on the "close" button.

## Issues
Closes #425
Closes #373 

## Future Followup
Content will need to be updated by either hard-coding every month, or partially re-written to load dynamically, and localStorage 'visitedcount' property will need to be reset prior to deploying. Also, this seems to have a lot in common with content put in the "News" tab on the AppBar, maybe open a separate issue to allow the dialog to be brought up again through News? As of now it only appears once and never again.

### DIFF
--- a/src/components/PatchNotes.tsx
+++ b/src/components/PatchNotes.tsx
@@ -7,10 +7,10 @@ const PatchNotes = () => {
 
     // show modal only on first visit    
     useEffect(() => {
-        if (localStorage.getItem('visitedcount')) {
+        if (localStorage.getItem('visitedCount') == 'y') {
             setIsOpen(false);
         } else {
-            localStorage.setItem('visitedcount', '1');
+            localStorage.setItem('visitedCount', 'y');
         }
     }, []);
         
@@ -23,7 +23,7 @@ const PatchNotes = () => {
                 or by clicking outside the dialog 
                 */
                 if(reason == 'backdropClick' || reason == 'escapeKeyDown') {
-                setIsOpen(false);
+                    setIsOpen(false);
                 }
             }} open={isOpen}>
                 <DialogTitle>{"What's New - February 2023"}</DialogTitle>
--- a/src/components/PatchNotes.tsx
+++ b/src/components/PatchNotes.tsx
@@ -40,9 +40,6 @@ const PatchNotes = () => {
                             <li>Fixed issues with displaying GE-III courses</li>
                             <li>Fixed courses with multiple locations appearing as independent events</li>
                         </ul>
-                        Other
-                        <ul>
-                        </ul>
                     </DialogContentText>
                 </DialogContent>
                 <DialogActions>
--- a/src/components/PatchNotes.tsx
+++ b/src/components/PatchNotes.tsx
@@ -31,14 +31,17 @@ const PatchNotes = () => {
                     <DialogContentText>
                         Features
                         <ul>
-                            <li>Added a new feature</li>
-                            <li>Added a second new feature</li>
+                            <li>Added 2023 Spring Quarter courses</li>
+                            <li>Courses will now share colors when added to Calendar</li>
+                            <li>Added this changelog!</li>
                         </ul>
                         Bug Fixes
                         <ul>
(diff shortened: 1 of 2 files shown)


src/components/PatchNotes.tsx
```diff
@@ -0,0 +1,60 @@
...
+    
+    return(
+        <>
+            <Dialog fullWidth={true} onClose={(event, reason) => {
+                /* 
+                allow the user to exit the modal using their keyboard 
+                or by clicking outside the dialog 
+                */
+                if(reason == 'backdropClick' || reason == 'escapeKeyDown') {
+                setIsOpen(false);
```
Voark: Super nit but is this missing a tab

src/components/PatchNotes.tsx
```diff
@@ -0,0 +1,60 @@
+import { Button, Dialog, DialogActions, DialogContent, DialogContentText, DialogTitle } from '@material-ui/core';
+import React, { useEffect, useState } from 'react';
+
+// PatchNotes follows structure/layout of AboutPage.tsx
+const PatchNotes = () => {
+    const [isOpen, setIsOpen] = useState(true);
+
+    // show modal only on first visit    
+    useEffect(() => {
+        if (localStorage.getItem('visitedcount')) {
```
Voark: Another nit but I think having the explicit conditional is better for code maintainability even though this is such a simple property.

src/components/PatchNotes.tsx
```diff
@@ -0,0 +1,60 @@
...
+                        <ul>
+                            <li>Added 2023 Spring Quarter courses</li>
+                            <li>Courses will now share colors when added to Calendar</li>
+                            <li>Added this changelog!</li>
+                        </ul>
+                        Bug Fixes
+                        <ul>
+                            <li>Fixed issues with displaying GE-III courses</li>
+                            <li>Fixed courses with multiple locations appearing as independent events</li>
+                        </ul>
```
Voark: I wonder if this can be done by loading a markdown file and converting it to basic HTML instead of having it hard coded in here, would make changing the notes much easier. Seems like [this package](https://www.npmjs.com/package/react-markdown) could work.

johnlorenzini: I thought the plan for the content was to hard code it initially (since it was the beginning of the month) and update it next sprint with a dynamic solution in time for March. Agree that react-markdown is the perfect tool for the job though.

EricPedley: Yeah, hard-coding it initially was the plan and we can stick to it and decide whether or not we want markdown rendering later.

Voark: My bad I must have missed that. I do think though that a dynamic solution is kinda weird because patch notes are coordinated with our commits anyways so might as well just do a commit to update the patch notes. This is assuming by dynamic solution we pull from a DB somewhere for the patch notes.

EricPedley:
This looks good. I'm holding my approval until #476 is merged.

EricPedley (approved):

ap0nia:
Possible enhancement: write patch notes in a `.md` file and import it into the component using [mdx](https://mdxjs.com/)

Patch notes themselves might be useful when they're stored as markdown, e.g. for general reference. This could be valid scenario for mdx, which lets you import markdown files as React components, e.g. with all of the necessary HTML tags, JSX, etc. handled for you.

PR #483, opened 2023-02-12
Add schedule notes

### LINKED ISSUE #413: Add description/notes to schedule
Add a feature to leave a short description on each named schedule to remind yourself why you made it. For example, alternative schedules for if you don't get a class when a major restriction drops.

This might be a good use for the blank space beneath the search bar.

This idea came from a user on the feedback form. Low priority for the moment but I think it would be nice to have unless we think of a better use for all that blank space.

Douglas-Hong:
## Summary
- Users can add a note for each of their schedules. They can also edit, save, and load their notes.
- Users can write notes while they are adding/editing a schedule.
- Schedule notes will be displayed under their added classes; the user can also edit the note directly in the box.

Task list:
- [x] Allow user to write/edit a note while they create or edit a schedule.
- [x] Add a box under the added classes that displays the current schedule's notes.
- [x] Allow schedule notes to be saved and loaded (waiting for schedule structure rewrite). For old schedules, make sure each schedule's note is just an empty string.
- [x] Add a way to edit a schedule note by clicking somewhere in the note box.

## Test Plan
- For a schedule with no notes, make sure the note box displays a message that says there is no note.
- Create a new schedule with notes. Make sure the note is displayed properly in the box.
- Edit a schedule that already has a note, and make sure the note box changes.
- Create multiple schedules with notes. Switch between these schedules and make sure the displayed notes are correct.
- When you delete a schedule, make sure the notes are still displayed correctly for the remaining schedules.
- Test various notes. Make sure newlines are processes and the max character limit is enforced.
- Save the schedules and see if the notes are still there when loaded back.
- Load older schedules that were saved with no schedule notes, then try editing notes for these schedules.

## Issues

Closes #413


src/actions/AppStoreActions.ts
```diff
@@ -229,14 +229,18 @@ export const toggleTheme = (radioGroupEvent: React.ChangeEvent<HTMLInputElement>
...
 
-export const addSchedule = (scheduleName: string) => {
-    AppStore.addSchedule(scheduleName);
+export const addSchedule = (scheduleName: string, scheduleNote: string) => {
+    AppStore.addSchedule(scheduleName, scheduleNote);
 };
 
-export const renameSchedule = (scheduleName: string, scheduleIndex: number) => {
-    AppStore.renameSchedule(scheduleName, scheduleIndex);
+export const editSchedule = (scheduleName: string, scheduleIndex: number, scheduleNote: string) => {
```
Voark: nit: I feel like `scheduleIndex` should be last

src/stores/Schedules.ts
```diff
@@ -79,19 +79,25 @@ export class Schedules {
...
         // Setting schedule index manually otherwise 2 undo states are added
         this.currentScheduleIndex = this.getNumberOfSchedules() - 1;
     }
 
     /**
-     * Renames schedule at index
+     * Edits schedule name and schedule note at index
      */
-    renameSchedule(newScheduleName: string, scheduleIndex: number) {
+    editSchedule(newScheduleName: string, scheduleIndex: number, newScheduleNote: string) {
```
Voark: Nit: same `scheduleIndex` thing as before

Voark (changes requested):
Overall code that you have written seems good to me, but there are a few issues related to undoing and schedule notes. When editing the schedule through the edit schedule button, we add an undo state so the schedule note change their can be undone. However it cannot be undone when updating from the note container because you didn't add an undo state save in `Schedules.ts`. I think this is probably the intended behavior (undo button does not undo schedule note change) but because of how we save schedules the undo button will undo it anyways. To fix this I think we need to add some more complicated undo saving/loading where we save/load all the properties except the notes. @EricPedley what are your opinions overall on undo state tracking for schedule notes?

On a slightly related note I think it would also be OK to remove the ability to edit schedule note from the edit schedule button because of the extra complexity of handling undoes (we still display the note in the pop up just not editable). 

Also small feature request: a "Clear Schedule Notes" button or trash can in the notes area in added classes.

Douglas-Hong:
From standup:
- If the schedule note was edited through a modal, then maybe we can do undo schedule notes.
- If the schedule note was edited through the box itself, it should not be undone.
- Clear button shouldn't be necessary since the user can just highlight all the text and delete it. 
- We should still be able to add a note through a modal.

Voark (commented):
Code changes look good to me, only change I think might help is just using a random number generator instead of a counter every time you add a new schedule so you don't need to keep track and increment the class variable.

Waiting on merge conflict resolution so the changes will be deployed to approve.

Voark (approved):
LGTM!!

PR #487, opened 2023-02-14
added aliases for all directories in src

ap0nia:
aliases: 
```ts
      '$actions': resolve(__dirname, './src/actions'),
      '$api': resolve(__dirname, './src/api'),
      '$components': resolve(__dirname, './src/components'),
      '$lib': resolve(__dirname, './src/lib'),
      '$stores': resolve(__dirname, './src/stores'),
```

example:
`import analytics from '../../../analytics'` -> `import analytics from $lib/analytics`

I also copied some files from `src/` into `lib` so it would be easier to reference them


Voark:
I was wondering why `lib/` shows up as added and not moved, looks like you did a copy instead of a move.

ap0nia:
> I was wondering why `lib/` shows up as added and not moved, looks like you did a copy instead of a move.

EDIT: ok I removed the original copies

Yeah I just did a copy; I'm under the impression that under people's code assumes their old locations, so I just copied the lib directory for my own convenience

ap0nia:
Closing as I'll make this a pull request to the `finality-reset` branch. Unless @EricPedley would like this to be merged directly to main instead of waiting for the overhaul to be finished?

PR #489, opened 2023-02-14
Add Share Button

stevenguyukai:
When User enter the website with paramed url, course code will automatically be added on the search form, and the search panel will appear.


stevenguyukai:
Some files in my local environment don't up to date. I Changes "SearchForm.tsx", "SectionCodeSearchBar.tsx", and "SectionTableButtons.tsx" for this share button update

alexanderl19:
Was just taking a brief look at this, and it would be super cool if the search parameters are just always appended instead of being copied with a button!

stevenguyukai:
> Was just taking a brief look at this, and it would be super cool if the search parameters are just always appended instead of being copied with a button!

OK, I will work on that!

src/components/Calendar/CourseCalendarEvent.tsx
```diff
@@ -12,7 +12,8 @@ import AppStore from '../../stores/AppStore';
 import ColorPicker from '../ColorPicker';
 import CustomEventDialog from './Toolbar/CustomEventDialog/CustomEventDialog';
 
-import locations from '../RightPane/SectionTable/static/locations.json';
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const locations = require('../RightPane/SectionTable/static/locations.json') as Record<string, string>;
```
ap0nia, line 16: require will break vite since vite uses es modules; change it to the `import` version

src/components/Calendar/CourseCalendarEvent.tsx
```diff
@@ -69,7 +70,7 @@ const styles: Styles<Theme, object> = {
 
 const genMapLink = (location: string) => {
     try {
-        const location_id = locations[location.split(' ')[0] as keyof typeof locations];
+        const location_id = locations[location.split(' ')[0]];
```
ap0nia, line 73:
```suggestion
        const location_id = locations[location.split(' ')[0] as keyof typeof locations];
```
if you want to keep typescript happy

src/components/RightPane/SectionTable/SectionTableButtons.tsx
```diff
@@ -90,6 +90,20 @@ export const ScheduleAddCell = withStyles(styles)((props: ScheduleAddCellProps)
         }
     };
 
+    const closeCopyAndAlert = () => {
+        const url = new URL(document.location.href)
```
```suggestion
        const url = new URL(window.location.href)
```
I generally prefer using window instead of document; even more so in React because there technically is no document (i.e. because it's virtual)

src/helpers.ts
```diff
@@ -353,7 +353,13 @@ export const warnMultipleTerms = (terms: Set<string>) => {
...
 export function clickToCopy(event: React.MouseEvent<HTMLElement, MouseEvent>, sectionCode: string) {
     event.stopPropagation();
-    void navigator.clipboard.writeText(sectionCode);
+
+    const tempEventTarget = document.createElement('input');
+    document.body.appendChild(tempEventTarget);
+    tempEventTarget.setAttribute('value', sectionCode);
+    tempEventTarget.select();
+    document.execCommand('copy');
+    document.body.removeChild(tempEventTarget);
```
ap0nia, line 362: use the navigator clipboard version

EricPedley (changes requested):
The build for this is failing because of a bunch of errors that I think are caused by resolving your merge conflicts with main incorrectly. Try saving your changes somewhere external to the repo, resetting the local version of your branch to main, then manually applying _just your changes_ without reverting any of the `main` code. By doing this I think you'll also address most of Brian's feedback, minus the one about using document vs. window.

PR #490, opened 2023-02-15
Import from zotcourse

### LINKED ISSUE #474: Import from ZotCourse
Allow the user to provide a ZotCourse schedule name and import it as a schedule on AntAlmanac.

This should be feasible by fetching data from their endpoint (https://zotcourse.appspot.com/schedule/load?username=[SCHEDULE_NAME]) and reformatting it. The data is pretty similar to how we structure it.

We should discuss the utility and UI aspects of this proposed feature.

Completes part of #440

JacE070:
## Summary
Added a new option in the original Import Study List Dialog to allow users to import schedules from Zotcourse.
## Test Plan
Tested schedule with sections w/ and w/o custom events.
## Issues
Closes https://github.com/icssc/AntAlmanac/issues/474

## Current Minor Issue
Peek 2023-04-09 15-08

The size of the dialog window changes as selecting different options. Guess the width is affected by the current max text width.

## Future Followup
Rename the ImportStudyList file in the source code to better indicate its functionality.


src/components/AppBar/ImportStudyList.tsx
```diff
@@ -13,9 +17,16 @@ import { ClassNameMap } from '@material-ui/core/styles/withStyles';
 import { PostAdd } from '@material-ui/icons';
 import React, { PureComponent } from 'react';
 
-import { openSnackbar } from '../../actions/AppStoreActions';
+import { addCustomEvent, openSnackbar } from '../../actions/AppStoreActions';
```
EricPedley: Can you change these imports to use the `$lib` syntax? See https://github.com/icssc/AntAlmanac/blob/main/tsconfig.json and how some of the other files do these imports.

src/helpers.ts
```diff
@@ -139,6 +144,44 @@ export function getCourseInfo(SOCObject: WebsocResponse) {
     return courseInfo;
 }
 
+export interface ZotCourseResponse {
+    codes: string[];
+    customEvents: RepeatingCustomEvent[];
+}
+export async function queryZotCourse(schedule_name: string) {
+    const url = new URL(ZOTCOURSE_ENDPOINT);
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
EricPedley: Can you explain why we had to disable a linter rule here?

src/helpers.ts
```diff
@@ -139,6 +144,44 @@ export function getCourseInfo(SOCObject: WebsocResponse) {
...
+    const response = await fetch(url, {
+        method: 'POST',
+        body: JSON.stringify({ scheduleName: schedule_name }),
+        headers: { 'Content-Type': 'application/json' },
+    }).then((r) => r.json());
+    // For custom event, there is no course attribute in each.
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
```
EricPedley: Same as previous comment, explain why we need to disable these rules.

src/helpers.ts
```diff
@@ -139,6 +144,44 @@ export function getCourseInfo(SOCObject: WebsocResponse) {
...
+    const codes = response.data
+        .filter((section: { eventType: number }) => section.eventType === 3)
+        .map((section: { course: { code: string } }) => section.course.code) as string[];
+    const days = [false, false, false, false, false, false, false];
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access
+    const customEvents: RepeatingCustomEvent[] = response.data
```
EricPedley: Same as last two comments, if we disable the linter rules it's smart to add comments explaining why.

EricPedley (changes requested):
Can you address the comments I left and resolve the merge conflicts?

src/helpers.ts
JacE070: tbh I need help with this and the following. TS raises errors since the response value has no type. 
I couldn't come up with a solution other than writing brand types that are only for these temporary variables.
What can be a good practice to solve this?

src/components/AppBar/ImportStudyList.tsx
JacE070: I now use import { addCustomEvent, openSnackbar } from '$actions/AppStoreActions'; and the import line doesn't have errors.
But for lines 67, 89, 96, the error "Unsafe call of an `any` typed value." are shown.
Why is that?

src/helpers.ts
EricPedley: The types should work themselves out if you define an interface for the entire `fetch` result. Looks like you already defined a interface called `ZotCourseResponse` for the return type of `queryZotcourse`, but I'd recommend using it to typecast your fetch request instead and let typescript infer the return type of your function. See this example from `News.tsx`: 
image

src/components/AppBar/ImportStudyList.tsx
EricPedley: The line numbers must've changed because I'm looking at those lines and there aren't function calls on any of them. Can you send the contents of the lines that are getting errors? I also opened the project in my editor and the linter didn't see any errors so IDK how to reproduce the problem you're having.

EricPedley (changes requested):
I left some comments that might help you fix the problems you're having. Go through those, fix the merge conflicts, and re-request a review. If you have any questions that you leave as comments on the PR, re-request review or ping me on Discord so I get a notification to respond.

src/components/AppBar/ImportStudyList.tsx
JacE070: ok now it's changed. The error is 
`[plugin:vite:import-analysis] Failed to resolve import "$actions/AppStoreActions" from "src/components/AppBar/ImportStudyList.tsx". Does the file exist?
` happened in the file ImportStudyList.tsx.

ap0nia: Use this updated [`vite.config.ts`](https://github.com/icssc/AntAlmanac/blob/main/apps/antalmanac/vite.config.ts). Your branch's version is missing all of the aliases for some reason.

src/helpers.ts
JacE070: IDK why Eslint stops complaining about them when I remove those disabling lines now.

Voark:
@EricPedley Gonna ping you again because I need to turn the backend changes into TypeScript in my PR so would be nice if this was merged in first.

EricPedley (approved):
Ship it!
```
       !
       !
       ^
      / \
     /___\
    |=   =|
    |     |
    |     |
    |     |
    |     |
    |     |
    |     |
    |     |
    |     |
    |     |
   /|##!##|\
  / |##!##| \
 /  |##!##|  \
|  / ^ | ^ \  |
| /  ( | )  \ |
|/   ( | )   \|
    ((   ))
   ((  :  ))
   ((  :  ))
    ((   ))
     (( ))
      ( )
       .
       .
       .
```

PR #495, opened 2023-02-16
Refactor aliases

ap0nia:
## Summary
- moved some files in the root of `src` into `src/lib` to keep a cleaner profile
- updated all relative imports to point to the right place
- re-ordered some of the imports to be in compliance with [eslint's import/order guidelines](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md)

## Notes
- set base as `refactor-root-structure` so there's less files changed
- once this has been reviewed, I'd like to merge it into `refactor-root-structure` and then merge `refactor-root-structure` into `finality-reset` without review


ap0nia:
I'm gonna go ahead and merge it since nothing substantial changed except for some files being moved to lib

PR #496, opened 2023-02-16
Refactor stores

ap0nia:
## Summary
- a store can be generally understood as "shared state" between components
- Added [zustand stores](https://github.com/pmndrs/zustand) to manage global/shared component state
- much less indirection and simpler flux pattern implementation

### Schedules
- `src/stores/schedule/index.ts` contains the (extremely simple) implementation of the schedule store, and defines any shared interfaces that can be used by external handlers
- all files next to that one export functions that will mutate the store on the caller's behalf
- when the store's stage changes, any components referring to the changed values will **magically** re-render (i.e. without having to emit events or explicitly listen/subscribing) 🪄 
- the store itself doesn't know anything about those functions; so each function provides "proxied access" to the store, simplifying the implementation of the store itself

### Search Form and Settings
- in the `settings` and `search` stores, the flux pattern is implemented directly in the stores because they're much simpler
- the stores contain values as well as the functions needed to mutate the store

## Why
- there's too many React state managers to justify not using one

## Test Plan
- after merging, gradually migrate components to invoke the functions from the new stores instead of the old ones

## TODO
- this is based off of @Voark 's schedule-refactor logic, and not everything has been fully implemented yet. But this is pretty representative of what I'd like to accomplish with this refactor.


src/actions/AppStoreActions.ts
```diff
@@ -13,15 +14,13 @@ import {
...
-import { VariantType } from 'notistack';
-
-import analyticsEnum, { logAnalytics } from '../analytics';
-import { LOAD_DATA_ENDPOINT, SAVE_DATA_ENDPOINT } from '../api/endpoints';
-import { SnackbarPosition } from '../components/AppBar/NotificationSnackbar';
-import { RepeatingCustomEvent } from '../components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog';
-import { CourseDetails, courseNumAsDecimal, getCoursesData, termsInSchedule, warnMultipleTerms } from '../helpers';
-import { Section } from '../peterportal.types';
-import AppStore, { AppStoreCourse, ShortCourseInfo, UserData } from '../stores/AppStore';
+import analyticsEnum, { logAnalytics } from '$lib/analytics';
```
EricPedley: Can you resolve the merge conflicts with `main` and re-request review? The diff viewer is showing me these "changes" in the imports that are actually the same as main, and I think it's because it isn't ready to auto-merge.

EricPedley (changes requested):

ap0nia:
I modified the theme to the previous versions used for `App.tsx` e.g. here: https://github.com/icssc/AntAlmanac/blob/ea16a75475edbef293dcd80c1a6383bf50692c3b/src/components/App.tsx

Because according to the documentation, [nesting themes overrides theme](https://github.com/icssc/AntAlmanac/blob/ea16a75475edbef293dcd80c1a6383bf50692c3b/src/index.js); so I'm only going to keep the part of the theme initialized in `App.tsx` and ignore the theme that was also initialized here: https://github.com/icssc/AntAlmanac/blob/ea16a75475edbef293dcd80c1a6383bf50692c3b/src/index.js

src/actions/AppStoreActions.ts
ap0nia: 👍

ap0nia:
Will be closing since this doesn't affect the main branch and these strategies will be employed on the alpha branch

PR #497, opened 2023-02-18
Adding Sharability

stevenguyukai:
## Updates

1. Adding a Share button at the SectionTableButton, which will append the course code into the url param
AddShareButton
2. If the user enter the website with information in url, then showLegacySearch will be set to true and the parameter will be fill automatically for user. 
Param_URL
3. User still need to press the search button.


stevenguyukai:
I have thinking about when user search a course or department or professor, the link will automatically copied to users keyboard. But this can be very unexpected to user. Some time user's clipboard can contain important imformation or complex info.

src/components/RightPane/CoursePane/SearchForm/SectionCodeSearchBar.tsx
```diff
@@ -3,6 +3,11 @@ import React, { ChangeEvent, PureComponent } from 'react';
 
 import RightPaneStore from '../../RightPaneStore';
 
+const urlParamValue = new URLSearchParams(window.location.search).get("courseCode");
+if (String(urlParamValue) != "null" && String(urlParamValue) != ""){
```
EricPedley: Can you change this so it's comparing without typecasting?

src/components/RightPane/CoursePane/SearchForm/SectionCodeSearchBar.tsx
```diff
@@ -3,6 +3,11 @@ import React, { ChangeEvent, PureComponent } from 'react';
 
 import RightPaneStore from '../../RightPaneStore';
 
+const urlParamValue = new URLSearchParams(window.location.search).get("courseCode");
```
EricPedley: @bevm0 Do you know if this line being outside any function or class creates race conditions with orders of module loading or is it guaranteed to be loaded after `RightPaneStore` since it imports it, or is it at the same time because it's in the same bundle?

EricPedley (changes requested):
One small change to make and I also want to check with @bevm0 about whether or not it's fine to have unenclosed code in a module cause I forget the specific behavior of TS/JS in this case.

src/components/RightPane/CoursePane/SearchForm/SectionCodeSearchBar.tsx
ap0nia: @EricPedley I don't know, but honestly if it's even a consideration, I'd wager that the strategy could be improved.

TLDR
If you want the store value to be a certain value when the page loads...you should do that in the store? The `window` is not exclusive to this component; inside of the store itself you should have a conditional/ternary that assigns the value during its initialization.

Additional Help
I don't fully comprehend the full strategy, but feel free to reach out if you want to talk through implementations or solutions.

Other Notes:

- Newest Proposal
the solution proposed above doesn't solve the issue of reactivity: if the form or components changes, then url will still remain static. You may prefer a solution like the one below that's slightly more involved if you want reactivity

- Initial Proposal:
I haven't tried it, but I think an idiomatic solution would be to keep track of the `urlParamValue` as a state variable, then use `useEffect` (or something similar) to listen to `window.location.search` and then update that state variable to the latest value.

stevenguyukai:
## Updates (2)

1. URL param will change automatically when user changes the value in input box
Sample
2. User can use the paramed link to auto fill the input box

src/components/RightPane/CoursePane/SearchForm/CourseNumberSearchBar.tsx
```diff
@@ -7,14 +7,44 @@ interface CourseNumberSearchBarState {
...
+        courseNumber: RightPaneStore.getUrlCourseNumValue() != "null" && RightPaneStore.getUrlCourseNumValue() != "" && RightPaneStore.getUrlCourseNumValue() != " " 
+        ?
+        this.updateCourseNumAndGetFormData()
+        : 
+        RightPaneStore.getFormData().courseNumber,
+    };
+    
+    updateCourseNumAndGetFormData(){
+        RightPaneStore.updateFormValue("courseNumber", RightPaneStore.getUrlCourseNumValue()) 
+        return RightPaneStore.getFormData().courseNumber
```
ap0nia: I don't like this, but I can see why this might be necessary if the store lives completely detached from component lifecycles.

I'd recommend consolidating this into a single method; e.g.
```ts
state = {
   courseNumber: this.getCourseNumber()
}
```

The `getCourseNumber` method will be responsible for syncing both the state of the store and the component; it will abstract away the ternaries, and nested function calls.

src/components/RightPane/CoursePane/SearchForm/DeptSearchBar/DeptSearchBar.tsx
```diff
@@ -48,8 +60,17 @@ class DeptSearchBar extends PureComponent<DeptSearchBarProps, DeptSearchBarState
...
+                ?
+                this.updatedeptValueAndGetFormData()
+                : 
+                RightPaneStore.getFormData().deptValue,
+
+                deptLabel: RightPaneStore.getUrlDeptLabel() != "null" && RightPaneStore.getUrlDeptLabel() != "" && RightPaneStore.getUrlDeptLabel() != " " 
+                ?
+                this.updatedeptLabelAndGetFormData()
+                : 
+                RightPaneStore.getFormData().deptLabel,
```
ap0nia: You can implement the recommendation for the CourseNumbeSearch here as well, e.g. create a method that will abstract away all of these steps. Another benefit of this is that you're re-calculating function calls multiple times; having a dedicated function allows you to save them into local variables

src/components/RightPane/CoursePane/SearchForm/DeptSearchBar/DeptSearchBar.tsx
```diff
@@ -85,6 +106,29 @@ class DeptSearchBar extends PureComponent<DeptSearchBarProps, DeptSearchBarState
...
+            urlParam.append('deptValue', setDeptValue.deptValue);
+            const new_url = `?${urlParam.toString()}`;
+            window.history.replaceState(stateObj, "url", "/" + new_url);
+        }else{
+            if (urlParam.toString() == "" || urlParam.toString() == null){
+                const new_url = `${urlParam.toString()}`;
+                window.history.replaceState(stateObj, "url", "/" + new_url);
+            }else{
+                const new_url = `?${urlParam.toString()}`;
+                window.history.replaceState(stateObj, "url", "/" + new_url);
```
ap0nia: It isn't necessarily wrong to save to the window state...but it's kinda hacky if we're using react-router-dom; here's relevant notes in the documentation regarding history and location for the library: https://reactrouter.com/en/main/start/concepts#locations

It may be possible to more cleanly implement this feature if we use react-router-dom's total control over history state.

The crux of this lies in the fact that this app is *client-side routed*, all  traditional `window.history` calls, navigation, etc. are intercepted and handled by the client side router, not by the browser. The browser history only has entries because the client side router actually added those entries for you.

src/components/RightPane/CoursePane/SearchForm/SearchForm.tsx
```diff
@@ -49,7 +49,22 @@ const styles: Styles<Theme, object> = {
...
+    const urlParamValue4 = new URLSearchParams(window.location.search).get("deptValue") != "" && 
+                            new URLSearchParams(window.location.search).get("deptValue") != null;
+    const urlParamValue5 = new URLSearchParams(window.location.search).get("GE") != "" && 
+                            new URLSearchParams(window.location.search).get("GE") != null;
+    const urlParamValue6 = new URLSearchParams(window.location.search).get("term") != "" && 
+                            new URLSearchParams(window.location.search).get("term") != null;
+    const [showLegacySearch, setShowLegacySearch] = useState(urlParamValue1 == true || 
+                                urlParamValue2 == true || urlParamValue3 == true || 
+                                urlParamValue4 == true || urlParamValue5 || 
+                                urlParamValue6 == true == true ? true : false);
```
bevm0, line 67: This is fine, but I think it's a little awkward to manage form logic as component state; form logic should be tied to an external store, and the component should predictably read or write from/to that source of truth.

ap0nia (approved):
TLDR: LGTM lol
The code and strategy is unidiomatic for React. But the entire app is basically unidiomatic as a whole, so it's difficult to truly address it properly. 
I'd actually prefer these changes to be pushed sooner than later, since I'm better equipped to actually address these issues on my branch. The comments are provided as "food for thought" for future implementations.

PR #498, opened 2023-02-18
Custom Buttons

ap0nia:
## Summary
- Implement a variety of different buttons used on the website.

##  Effects on App
- Doesn't effect the app or website it any way
- They'll be implemented into the app during an "integration" phase where I'll transplant sections of the app with the versions I've re-written

## Details
- Implements existing functionality using Zustand stores and a looser flux pattern with more a terser, more idiomatic approach for modern React
- Since the buttons are mostly independent from the rest of the app, I located all buttons in the same place for now
- This can be adjusted as more of the app gets implemented; e.g. by locating the buttons in the same location as their usage
- may be missing some details with logging analytics, enqueuing snackbar notifications, etc.

## Future Follow Up
- I'd like to move "logging analytics" into the store actions instead of `onClick` handlers since this will simplify component logic and readability, but it seems not all events are tied to the actual change?
- e.g. `calendarActions.changeColor` appears to be triggered when clicking on the icon button, not during the actual color change
- Use the buttons in actual components once I'm able to use the ones I've written that connect to Zustand instead of events


src/lib/download.ts
```diff
@@ -0,0 +1,281 @@
+import { termData } from '$lib/termData'
```
ap0nia, line 1: I didn't write this code; I just cleaned it up and moved the logic outside of the component

MinhxNguyen7: Where did this come from?

src/components/Buttons/Undo.tsx
```diff
@@ -0,0 +1,17 @@
+import { IconButton, Tooltip } from '@mui/material'
+import { Undo as UndoIcon } from '@mui/icons-material'
+import { undo } from '$stores/schedule/commands'
+
+export default function UndoButton() {
+  function handleClick() {
+    undo()
```
MinhxNguyen7, line 7: Why wouldn't you move this into the button, i.e., `onClick={undo}`?

src/components/Buttons/Redo.tsx
```diff
@@ -0,0 +1,17 @@
+import { IconButton, Tooltip } from '@mui/material'
+import { Redo as RedoIcon } from '@mui/icons-material'
+import { redo } from '$stores/schedule/commands'
+
+export default function RedoButton() {
+  function handleClick() {
+    redo()
```
MinhxNguyen7, line 7: Why wouldn't you move this into the button, i.e., `onClick={redo}`?

MinhxNguyen7 (commented):
I've got some minor comments, but everything seems about right

src/components/Buttons/Undo.tsx
ap0nia: Yeah, that's valid. I'm just leaving it inside a function so all of the buttons have a similar interface and there's more flexibility to add error handling, callbacks, etc. to the undo call.

src/components/Buttons/Redo.tsx
ap0nia: Yeah I can; this is just a proposal and I just decided to wrap the redo call for the reasons above

src/lib/download.ts
ap0nia: The file is here: https://github.com/icssc/AntAlmanac/blob/custom-buttons/src/lib/termData.ts

You can view the TypeScript alias here in the "paths" property: https://github.com/icssc/AntAlmanac/blob/custom-buttons/tsconfig.json

And the build time resolution is handled by Vite in the config: https://github.com/icssc/AntAlmanac/blob/custom-buttons/vite.config.ts

PR #499, opened 2023-02-18
Calendar Overhaul

ap0nia:
## Summary
- Re-wrote the calendar in a separate components folder
- Uses functional components
- Tried to mimic as much of the functionality I noticed on the current calendar while simplifying the implementation and file organization as much as possible

## App Effects
- Doesn't affect the app in any way
- I can't override the current calendar because it depends on the AppStore, which this calendar doesn't use

## Future Follow Up
- substitute the calendar into the application
- this may take a while because it's strongly dependent on Zustand stores to provide the correct, most recent state data


ap0nia:
Might re-write with consideration for immediate compatibility for the main branch

PR #500, opened 2023-02-18
added custom header

ap0nia:
## Summary
Custom header


ap0nia:
Might re-write with consideration for immediate compatibility with the main branch. But will probably just keep local to the alpha branch

PR #501, opened 2023-02-18
custom schedule renderer

ap0nia:
## Summary
- A "schedule" component renders one entire section of this:
image

## Description:
- School: the first accordion that expands with more info about a school
- Department: the second accordion that expands with more info about a department
- Course: The header (buttons) with info about the course + the body with all courses available


ap0nia:
The render strategy for displaying Websoc results has been fully realized in the alpha branch, and I'll keep it local to there

PR #502, opened 2023-02-18
Map Re-write

ap0nia:
## Notes
- not functional because it uses the newest version of react-leaflet
- I just wanted the code to be reviewed


ap0nia:
Will rewrite with consideration for immediate compatibility with the current main branch

PR #504, opened 2023-02-19
Fix staging

Voark:
## Summary
Fixes staging urls pointing to prod API. Solution shamelessly stolen from @bevm0
The main issue is Vite reads environmental variables differently from CRA, basically custom env variables are only read when they start with `VITE_`. This change removes the reference to the previous custom env variable and instead just uses the `NODE_ENV` flag, which is read by Vite. 

## Notes
- `vite-env.d.ts` defines the env interface to Vite's env variable config so we can use `import.meta.env.DEV`. 
- When setting the environmental variable `NODE_ENV` to `"development"`, Vite sets `meta.env.DEV = true` which we can use to choose the right endpoint.


src/peterportal.types.ts
```diff
@@ -1,129 +0,0 @@
-// Ripped straight from here: https://github.com/icssc/peterportal-public-api/blob/c539a62acfbfda1481152272582632a168f34302/types/websoc.types.ts
```
EricPedley: Why is this file deleted?

EricPedley (changes requested):
Can you explain the changes besides the `endpoints.ts`, `vite.config.ts`, and `stage_pr.yaml`? Those make sense but IDK what the rest of this does.

src/peterportal.types.ts
Voark: I must have accidentally selected it when trying to delete `react-app-env.d.ts`

Voark:
> Can you explain the changes besides the `endpoints.ts`, `vite.config.ts`, and `stage_pr.yaml`? Those make sense but IDK what the rest of this does.

Added in PR description.

EricPedley (approved):
🚀

PR #505, opened 2023-02-20
Fix schedule loading bugs

Voark:
- Bug 1: Querying websoc with more than 10 courses at once
- Bug 2: Querying websoc for cancelled course

## Summary
New schedule changes had issues with loading schedules that met the two conditions above.

Consolidate querying functions


EricPedley (approved):
have fun with the quickdraw github achievement

PR #506, opened 2023-02-22
Fix null

ap0nia:
## Summary
If you click around the tabs while the save data is loading, you can kill the entire app because some stuff is undefined

### DIFF
--- a/src/components/RightPane/Map/UCIMap.tsx
+++ b/src/components/RightPane/Map/UCIMap.tsx
@@ -1,7 +1,7 @@
 import '../../../../node_modules/leaflet.locatecontrol/dist/L.Control.Locate.min.js';
 
 import Leaflet, { Control, LeafletMouseEvent } from 'leaflet';
-import 'leaflet.locatecontrolj
+import 'leaflet.locatecontrol'
 import React, { PureComponent } from 'react';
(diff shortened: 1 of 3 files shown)


src/components/RightPane/Map/UCIMap.tsx
```diff
@@ -1,6 +1,7 @@
 import '../../../../node_modules/leaflet.locatecontrol/dist/L.Control.Locate.min.js';
 
 import Leaflet, { Control, LeafletMouseEvent } from 'leaflet';
+import 'leaflet.locatecontrol'
```
Voark, line 4: What is this for?

Voark (commented):
Changes look good to me, only thing is when changing to the Added Classes tab the schedule name shows up as undefined. I think we should set a default for that too when we get the current schedule name so the site looks less scuffed.

src/components/RightPane/Map/UCIMap.tsx
ap0nia: I get an error in dev because `Locate.Control isn't a constructor` importing the library extends the interface and resolves the error. I'm not sure why this isn't an issue in production, but it's better to be safe with this trivial fix

EricPedley (approved):
lgtm

PR #507, opened 2023-02-23
added split pane and fixed config

ap0nia:
## Summary
Objective: obsolete Zotcourse once and for all 😎 

Makes the desktop split view resizable (using the exact same library as Zotcourse).

Fixed some config issues, they're kinda minor so I don't think it's too bad to include them here?

## TODO
@EricPedley do you want me to do the stupid thing to add the ellipsis icon to it?

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -21,14 +21,13 @@ extends:
     - plugin:@typescript-eslint/recommended
     - plugin:@typescript-eslint/recommended-requiring-type-checking
     - plugin:react/recommended
+    - plugin:react/jsx-runtime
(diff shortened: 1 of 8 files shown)


EricPedley (changes requested):
Looks good. Can you add the ellipses thing?

EricPedley (approved):
Ship it!
```

                   ,:',:`,:'
                __||_||_||_||__
           ____["""""""""""""""]____
           \ " '''''''''''''''''''' |
    ~~~~~~^~^~^^~^~^~^~^~^~^~^~~^~^~^^~~^~^
```

PR #508, opened 2023-02-23
Run linter on entire project

EricPedley:
## Summary
When we started using eslint for stuff it never actually got run on _all_ the project files.
## Test Plan
Confirm that this didn't change anything functionally, just fixing code style everywhere.

Also confirm that there are no more linting errors on your local development environment.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -2,4 +2,3 @@
 source "$(dirname "$0")/_/husky.sh"
 
 yarn lint-staged
-yarn format
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -2,3 +2,4 @@
 source "$(dirname "$0")/_/husky.sh"
 
 yarn lint-staged
+yarn format
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
         "@vitejs/plugin-react": "^3.1.0",
         "eslint": "^8.33.0",
         "eslint-config-prettier": "^8.6.0",
+        "eslint-import-resolver-typescript": "^3.5.3",
         "eslint-plugin-import": "^2.27.5",
         "eslint-plugin-jsx-a11y": "^6.7.1",
         "eslint-plugin-react": "^7.32.2",
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
         "predeploy": "npm run build",
         "deploy": "gh-pages -d build",
         "format": "prettier --write src pages",
-        "lint": "eslint src/**/*.tsx"
+        "lint": "eslint src/**/*.ts*"
     },
     "dependencies": {
         "@date-io/date-fns": "^2.16.0",
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
         "preview": "vite preview",
         "predeploy": "npm run build",
         "deploy": "gh-pages -d build",
-        "format": "prettier --write src pages",
+        "format": "prettier --write src",
         "lint": "eslint src/**/*.ts*"
     },
     "dependencies": {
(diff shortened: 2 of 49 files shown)


package.json
```diff
@@ -17,7 +17,7 @@
         "predeploy": "npm run build",
         "deploy": "gh-pages -d build",
         "format": "prettier --write src pages",
-        "lint": "eslint src/**/*.tsx"
+        "lint": "eslint src/**/*.ts*"
```
EricPedley, line 20: This makes it lint both `.ts` and `.tsx` files

package.json
```diff
@@ -76,6 +76,7 @@
         "@vitejs/plugin-react": "^3.1.0",
         "eslint": "^8.33.0",
         "eslint-config-prettier": "^8.6.0",
+        "eslint-import-resolver-typescript": "^3.5.3",
```
EricPedley, line 79: This was to fix some errors I was getting as described by this SO post with the same problem: https://stackoverflow.com/questions/69446204/resolve-error-typescript-with-invalid-interface-loaded-as-resolver-eslint

package.json
```diff
@@ -17,7 +17,7 @@
         "predeploy": "npm run build",
         "deploy": "gh-pages -d build",
         "format": "prettier --write src pages",
```
ap0nia: Could you remove the `pages` directory? It's a left over artifact when I was trying out NextJS; we only need to format source files in `src`. You could also make it more specific about the file types like the linting command if you want

package.json
ap0nia: Oh my bad, I had the same issue but I forgot to commit that to the linting config update

package.json
EricPedley: Done.

ap0nia (commented):
No don't do that; configure this in lint-staged, `.lintstagedrc.yml` (it should already be configured)

Husky is generally for invoking lint-staged, but lint-staged is responsible for doing them

.husky/pre-commit
```diff
@@ -2,3 +2,4 @@
 source "$(dirname "$0")/_/husky.sh"
 
 yarn lint-staged
+yarn format
```
ap0nia: No don't do that; configure lint-staged to do this, `.lintstagedrc.yml` (it should already be configured)

Husky is generally for invoking lint-staged, but lint-staged is responsible for doing them

EricPedley: If it was already there then why did manually running the formatter affect so many files? Shouldn't it have already been running automatically?

ap0nia:
For better QoL with eslint + prettier, you could also install [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier#installation) which should remove any conflicting rules between eslint and prettier

ap0nia (approved):
I think these config changes and the linting/formatting process is sufficient. You can consider my note on using `eslint-config-prettier` to ensure consistencey between using the two concurrently

.husky/pre-commit
ap0nia: Pre-commit hook, e.g. the one that runs lint-staged, only affects files you're committing; formatting manually will run on the files you specified, which is the entire project.

It's a little strange if you encounter bugs too; during lint-staged, the `prettier` and `eslint` command is run individually on each file, so you might sometimes encounter errors that don't show up when you linting or formatting normally. e.g. sometimes it won't find plugins or extra config because the context is different (this mostly happened with a monorepo setup, so we should be safe from them for now)

EricPedley:
Have you ran into this problem (see screenshot) with the types of `react-leaflet` in UCIMap?
image
I'm considering whether or not to fix it in this PR or just leave it be. I'm also thinking it might just be something with my local installation of all the dependencies.

ap0nia:
Yeah don't fix that in react-leaflet; I've tried it and everything goes haywire because of conflicting versions. 

i.e. you would normally resolve it by doing `yarn add -D @types/react-leaflet` to add the type definitions

When I can push the revamped map, it'll all use modern versions of leaftlet, react-leaflet, etc. and those errors will be vanquished

PR #512, opened 2023-02-23
Fix ColorPicker not showing section color right after adding.

### LINKED ISSUE #510: Color picker colors broken in search results
When you add a class, the color picker in the section table doesn't get the right color until you reload the results. The icon is white and the color it opens to isn't the actual color of the section added.
Seems like the bug was introduced in #457.
image

EricPedley:
## Summary
Fixes a bug where the colorpicker in the sectiontable wouldn't immediately change to the color of an added course.

I figured this one out by re-reading the diff for #457 and looking specifically at the logic around addCourse. This long return chain is how we did it before and it feels kinda janky but I would rather have a patch out than spend a bunch of time architecting a more elegant solution.

## Test Plan
Verify that courses colors are displayed correctly on the color picker immediately after they're added.
## Issues

Closes #510

### DIFF
--- a/src/actions/AppStoreActions.ts
+++ b/src/actions/AppStoreActions.ts
@@ -38,7 +38,7 @@ export const addCourse = (
         section: { ...section, color: '' },
     };
 
-    AppStore.addCourse(newCourse, scheduleIndex);
+    return AppStore.addCourse(newCourse, scheduleIndex);
 };
 /**
  * @param variant usually 'info', 'error', 'warning', or 'success'
--- a/src/components/RightPane/SectionTable/SectionTableButtons.tsx
+++ b/src/components/RightPane/SectionTable/SectionTableButtons.tsx
@@ -85,7 +85,8 @@ export const ScheduleAddCell = withStyles(styles)((props: ScheduleAddCellProps) 
                     action: analyticsEnum.classSearch.actions.ADD_SPECIFIC,
                 });
             }
-            section.color = addCourse(section, courseDetails, term, scheduleIndex).section.color;
+            const newCourse = addCourse(section, courseDetails, term, scheduleIndex);
+            section.color = newCourse.section.color;
         }
     };
 
--- a/src/components/RightPane/SectionTable/SectionTableButtons.tsx
+++ b/src/components/RightPane/SectionTable/SectionTableButtons.tsx
@@ -85,7 +85,7 @@ export const ScheduleAddCell = withStyles(styles)((props: ScheduleAddCellProps) 
                     action: analyticsEnum.classSearch.actions.ADD_SPECIFIC,
                 });
             }
-            section.color = addCourse(section, courseDetails, term, scheduleIndex);
+            section.color = addCourse(section, courseDetails, term, scheduleIndex).section.color;
         }
     };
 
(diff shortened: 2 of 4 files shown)


ap0nia (changes requested):
I think `addCourse` should return the course that was added, or an object with all the necessary information about the course that can be used by other functions. I think this makes the most sense without knowing much about the function; and it also gives you more flexibility to use the returned object in other contexts. (you would then extract the desired properties from the object, i.e. the color for this patch)

Also just an optional code clarity note, you can assign values of function return and relevant intermediate steps to make it clearer about what/how you'll use it, e.g
```ts
// AppStore
const newCourse = ...
const newCourseWithColor = {...newCourse, color: 'color' }
return newCourseWithColor

// invocation/use-case
const newCourseAdded = AppStore.addCourse(...)
const newColor = newCourseAdded.color
```

Specifically returning a color makes the function coupled to this specific scenario and use-case, which might encourage hacky work arounds in the future

src/components/RightPane/SectionTable/SectionTableButtons.tsx
```diff
@@ -85,7 +85,7 @@ export const ScheduleAddCell = withStyles(styles)((props: ScheduleAddCellProps)
                     action: analyticsEnum.classSearch.actions.ADD_SPECIFIC,
                 });
             }
-            section.color = addCourse(section, courseDetails, term, scheduleIndex);
+            section.color = addCourse(section, courseDetails, term, scheduleIndex).section.color;
```
```suggestion
            const newCourse = addCourse(section, courseDetails, term, scheduleIndex).section.color;
            section.color = newCourse.section.color
```
Optional: you can a temp variable to make it clearer

src/stores/AppStore.ts
```diff
@@ -66,15 +66,15 @@ class AppStore extends EventEmitter {
...
 
     addCourse(newCourse: ScheduleCourse, scheduleIndex: number = this.schedule.getCurrentScheduleIndex()) {
-        let color: string;
+        let addedCourse: ScheduleCourse;
         if (scheduleIndex === this.schedule.getNumberOfSchedules()) {
-            color = this.schedule.addCourseToAllSchedules(newCourse);
+            addedCourse = this.schedule.addCourseToAllSchedules(newCourse);
         } else {
-            color = this.schedule.addCourse(newCourse);
+            addedCourse = this.schedule.addCourse(newCourse);
```
ap0nia:
```ts
const addedCourse = 
scheduleIndex === this.schedule.getNumberOfSchedules() ?
    this.schedule.addCourseToAllSchedules(newCourse) : 
    this.schedule.addCourse(newCourse)
```

Optional: doing this allows TypeScript to infer the return type implicitly; downside is that it's harder to read

src/actions/AppStoreActions.ts
```diff
@@ -38,7 +38,7 @@ export const addCourse = (
         section: { ...section, color: '' },
     };
 
-    AppStore.addCourse(newCourse, scheduleIndex);
+    return AppStore.addCourse(newCourse, scheduleIndex);
```
ap0nia, line 41:
```suggestion
    const addedCourse = AppStore.addCourse(newCourse, scheduleIndex);
    return addedCourse;
```
Optional: more clear what we're doing without requiring knowledge of the `addCourse` method

ap0nia (approved):
I think this is acceptable for shipping! 🚀 

IMO I don't think this "return chain" strategy is necessarily wrong; I've implemented a similar strategy in my stores re-write, but with less levels of indirection. I think clearly communicating returned interfaces is good for making existing functionality more reusable and quickly implementing features without having to go down to the root store to augment behavior for new, specific use cases

PR #514, opened 2023-02-24
Monorepo migration

Voark:
### Monorepo
Combined the 3 main AntAlmanac repos, https://github.com/icssc/antalmanac-backend, https://github.com/icssc/antalmanac-cdk and this current repo into a monorepo. For the most part the individual repos were moved verbatim into their own package in the `apps` folder. [TurboRepo](https://turbo.build/) is used for managing building of the individual packages and [pnpm](https://pnpm.io/) is used as the package manager. Pnpm was chosen as the package manager because it is fast and symlinks packages instead of installing new copies in each app. 

### Ci/CD
Reworked the Github Actions to handle the new packages, as the actions will now deploy CDK and backend changes during staging and on push to main. 

Also, the actions workflow was changed so that now in response to a trigger (ex. opening a PR), a "parent" action will call "step" actions as defined in the `step_... .yaml` files. 
Some benefits include: 
- Reusing code across staging/pushing to main
- Can deploy stacks simultaneously
- Easier to see which step fails, as each step action will show up as a new check in a pull request

## Test Plan
- This PR and the `alpha` branch (tested action that triggers on push)
- Did not test changes to the deploy to GH pages action because there is no way to test it without overwriting prod. Will fix any issues if they pop up after merging this branch into main

## Future Followup

- The prod/dev stack names are kind of horrendous so they should be changed in the future. I should've thought about it more before naming it last year, however changing the stack name is going to suck...


ap0nia:
Hey can I move my project into this monorepo lol

Also some notes:
- Idk how I feel about sharing a tsconfig in `packages`, personally I prefer all projects implementing their own tsconfig because there's less indirection
- A caveat to the point above is that _the project root should have its own tsconfig.json_ in order for VSCode to properly identify monorepo path aliases and whatnot; I usually use that as my base tsconfig to extend from since it's a lot more straightforward
- I've never opted into the shared eslint or prettier configs either, because I think it's trivial to implement your own on a per-project basis
- These are my opinions from a Svelte + Express monorepo -> all the linting and tsconfig configs are wildly different so I didn't really feel that it was worth trying to consolidate them; I also already had a good amount of shared packages between the two apps, and I didn't want to bloat that directory with even more

apps/antalmanac/.gitignore
```diff
@@ -0,0 +1,14 @@
...
+
+# misc
+.DS_Store
+.env
+.idea
+.vscode
+
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
```
ap0nia:
```suggestion
*.log*
```

apps/antalmanac/.gitignore
```diff
@@ -0,0 +1,14 @@
+node_modules/
+
+# production
+build/
```
ap0nia:
```suggestion
build/
.turbo/
```

apps/backend/.gitignore
```diff
@@ -0,0 +1,14 @@
...
+dist/
+
+# misc
+.DS_Store
+.env
+.idea
+
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
```
ap0nia:
```suggestion
*.log*
```

apps/backend/.gitignore
```diff
@@ -0,0 +1,14 @@
+node_modules/
+.serverless/
+
+# production
+dist/
```
ap0nia:
```suggestion
dist/
.turbo/
```

apps/backend/package.json
```diff
@@ -0,0 +1,32 @@
+{
+    "name": "antalmanac-backend",
+    "version": "0.0.1",
+    "description": "Backend for AntAlmanac",
+    "main": "server.js",
+    "scripts": {
+        "build": "esbuild --platform=node --target=node16 --bundle  --outfile=dist/lambda.js lambda.js",
```
ap0nia, line 7: Can we try using an esbuild script for this

apps/backend/package.json
```diff
@@ -0,0 +1,32 @@
...
+        "start": "node server.local.js",
+        "format": "prettier --write \"./**/*.{ts,tsx,js}\""
```
ap0nia:
```suggestion
        "format": "prettier --write src"
```
I usually just do something like this and call it a day lol

pnpm-workspace.yaml
```diff
@@ -0,0 +1,3 @@
+packages:
+  - "apps/*"
+  - "packages/*"
```
ap0nia:
```suggestion
  - apps/*
  - packages/*
```
I think it actually works without quotes for some reason...minimal :sunglasses:

package.json
```diff
@@ -1,102 +1,23 @@
...
+    "deploy:all": "turbo deploy"
+  },
+  "workspaces": [
+    "apps/*"
+  ],
+  "devDependencies": {
+    "husky": "^8.0.3",
+    "turbo": "latest"
+  },
+  "packageManager": "pnpm@7.27.1"
```
ap0nia:
```suggestion
  "packageManager": "pnpm@7.27.1",
    "engines": {
      "pnpm": "^7.0.0"
```
Since we're specifying a package manager, we could try specifying the engine as well?

package.json
```diff
@@ -1,102 +1,23 @@
...
+  "private": true,
+  "scripts": {
+    "prepare": "husky install",
+    "start": "turbo start --filter=antalmanac",
+    "start:all": "VITE_LOCAL_SERVER=true turbo start",
+    "build": "turbo build --filter=antalmanac",
+    "build:all": "turbo build",
+    "deploy": "turbo deploy --filter=antalmanac",
+    "deploy:cdk": "turbo deploy --filter=cdk",
+    "deploy:all": "turbo deploy"
```
ap0nia:
```suggestion
    "start:aa": "turbo start --filter=antalmanac",
    "start": "VITE_LOCAL_SERVER=true turbo start",
    "build:aa": "turbo build --filter=antalmanac",
    "build": "turbo build",
    "deploy:aa": "turbo deploy --filter=antalmanac",
    "deploy:cdk": "turbo deploy --filter=cdk",
    "deploy": "turbo deploy"
```
I think it could make sense to make the commands closer to their source; the "turbo start" command normally starts everything, and "turbo start --filter" would scope the command. It seems kinda inverted to alias "start" to do a scoped command.

apps/cdk/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "eslint-plugin-prettier": "^4.0.0",
+    "prettier": "^2.5.1",
+    "ts-node": "^9.0.0",
+    "typescript": "latest"
+  },
+  "dependencies": {
+    "aws-cdk-lib": "^2.66.1",
+    "constructs": "^10.0.0",
+    "dotenv": "^16.0.0",
+    "antalmanac": "workspace:*"
```
ap0nia:
```suggestion
    "antalmanac": "*"
```
"workspace" doesn't work in yarn or npm workspaces, so I just avoid it so I don't have to deal with any issues.

apps/backend/.env.sample
```diff
@@ -0,0 +1,4 @@
+AA_MONGODB_URI=uri
```
EricPedley: Awesome idea to include this. It'll make onboarding much easier.

apps/cdk/README.md
```diff
@@ -0,0 +1,14 @@
+# Welcome to your CDK TypeScript project!
```
EricPedley: Can we get a real README for the cdk project? Especially since our bus factor for the CDK stuff is 1 right now and we should be onboarding at least 1 more person onto devops tasks.

apps/backend/README.md
EricPedley, line 1: Can you update this to be accurate as of this PR? The backend isn't a submodule anymore and this might've changed whether the instructions for running locally are correct still.

apps/antalmanac/src/lib/api/endpoints.ts
```diff
@@ -1,4 +1,10 @@
 function endpointTransform(path: string) {
+    if (import.meta.env.VITE_ENDPOINT) {
```
EricPedley, line 2: Where does this environment variable come from? It isn't in the sample `.env` so I'm assuming it gets populated in the build step but it isn't super clear. Maybe a comment would be worthwhile?

EricPedley (changes requested):
Can you add steps to the README about installing / setting up pnpm and turborepo? 

Also can we get a tsconfig in the root directory so VSCode stops complaining? 

The other comments I put aren't that critical but I want to get at least these two things done before we merge to make the transition easier for the team.

apps/antalmanac/src/lib/api/endpoints.ts
ap0nia: Public environment variables with a certain prefix (`VITE` by default) are exposed by [vite](https://vitejs.dev/guide/env-and-mode.html)

But yeah, it would be a good idea to include any environment variables we use in the `.env` file to give a full picture of what we're working with.

ap0nia (changes requested):
Ok, I wasn't actually able to get through all of it atm lol. But I'll just post this here for now and I'll re-review it later

package.json
Voark: The reason why I did this is because it is the most used command is probably `start` as most people are just developing the frontend and everyone is probably already used to `npm start` starting the frontend. 

Yeah I agree it is a little weird but I was trying (maybe too hard) to not have too many changes for current developers. @EricPedley opinions?

apps/cdk/README.md
Voark: Forgot to update any of the READMEs, will work on all of them.

package.json
ap0nia: Two possible compromises:
1. `dev` as the start command, e.g. `npm run dev`, which will start the AA project.
2. Just request that team members `cd apps/antalmanac && npm run start` because that's generally the dev-intended strat for running scoped commands (for me at least).

apps/backend/package.json
Voark: I thought about it but it really isn't complex enough imo to justify one, maybe if we do anything extra in the future.

apps/backend/.gitignore
Voark: Is this necessary? It's in the root `.gitignore`. Speaking of, I should probably just combine them tbh.

ap0nia: Idk tbh, I just toss stuff in the gitignore to be safe. 

I don't recommend using a single root gitignore tho; standard practice is to use a gitignore for each subproject in a monorepo.

Voark: IMO it's not worth it because our gitignores are so similar I don't really see a benefit besides some extra things to manage.

EricPedley (approved):
Ship it!
```
       !
       !
       ^
      / \
     /___\
    |=   =|
    |     |
    |     |
    |     |
    |     |
    |     |
    |     |
    |     |
    |     |
    |     |
   /|##!##|\
  / |##!##| \
 /  |##!##|  \
|  / ^ | ^ \  |
| /  ( | )  \ |
|/   ( | )   \|
    ((   ))
   ((  :  ))
   ((  :  ))
    ((   ))
     (( ))
      ( )
       .
       .
       .

```

ap0nia:
LGTM! :rocket:

ap0nia (approved):

PR #517, opened 2023-03-03
Fix bug by stripping string for condition checking

### LINKED ISSUE #390: Missing waypoints for course section with fake locations

JacE070:
## Summary
Fix the remaining missing-way-point bug by stripping the name of the building in sections.
Predecessor: https://github.com/icssc/AntAlmanac/pull/471
## Test Plan
I tested the cases that it only has sections with real locations, it only has sections with fake locations, and mixed.
The map functions well.
## Issues

Closes https://github.com/icssc/AntAlmanac/issues/390

### DIFF
--- a/src/components/RightPane/Map/UCIMap.tsx
+++ b/src/components/RightPane/Map/UCIMap.tsx
@@ -1,7 +1,7 @@
 import '../../../../node_modules/leaflet.locatecontrol/dist/L.Control.Locate.min.js';
 
 import Leaflet, { Control, LeafletMouseEvent } from 'leaflet';
-import 'leaflet.locatecontrol'
+import 'leaflet.locatecontrol';
 import React, { PureComponent } from 'react';
 import { LeafletContext, Map, Marker, Polyline, TileLayer, withLeaflet } from 'react-leaflet';
 import analyticsEnum, { logAnalytics } from '$lib/analytics';
@@ -296,7 +296,7 @@ export default class UCIMap extends PureComponent {
                             !event.start.toString().includes(DAYS[day]) ||
                             courses.has(event.sectionCode) || // Remove duplicate courses that appear in the calendar
                             !courses.add(event.sectionCode) ||
-                            FAKE_LOCATIONS.includes(event.bldg)
+                            FAKE_LOCATIONS.includes(event.bldg.trim())
                         ) // Adds to the set and return false
                     )
             )
--- a/src/components/RightPane/Map/UCIMap.tsx
+++ b/src/components/RightPane/Map/UCIMap.tsx
@@ -1,11 +1,12 @@
-import '../../../../node_modules/leaflet.locatecontrol/dist/L.Control.Locate.min.js';
+import 'leaflet.locatecontrol';
 
 import Leaflet, { Control, LeafletMouseEvent } from 'leaflet';
-import 'leaflet.locatecontrol';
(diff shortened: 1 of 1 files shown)


src/components/RightPane/Map/UCIMap.tsx
```diff
@@ -1,7 +1,7 @@
 import '../../../../node_modules/leaflet.locatecontrol/dist/L.Control.Locate.min.js';
```
ap0nia:
```suggestion
```
Since you're in this file, could you try removing this line and see if the map still works? I'm pretty sure this is what the `import 'leaflet.locatecontrol'` line is supposed to do

src/components/RightPane/Map/UCIMap.tsx
```diff
@@ -296,7 +296,7 @@ export default class UCIMap extends PureComponent {
                             !event.start.toString().includes(DAYS[day]) ||
                             courses.has(event.sectionCode) || // Remove duplicate courses that appear in the calendar
                             !courses.add(event.sectionCode) ||
-                            FAKE_LOCATIONS.includes(event.bldg)
+                            FAKE_LOCATIONS.includes(event.bldg.trim())
```
ap0nia, line 300: Good patch!

Do you know how this occurs? Ideally, we'd be handling the data in such a way that doesn't require us to patch it like this

ap0nia (approved):

src/components/RightPane/Map/UCIMap.tsx
JacE070: I will inspect this! Thanks!

src/components/RightPane/Map/UCIMap.tsx
JacE070: Yeah it works! Deleting this line is safe

src/components/RightPane/Map/UCIMap.tsx
JacE070: Ok i'm sure the bldg field returned by PeterPortal api actually has a trailing whitespace when it's "TBA." So do you think this is a proper place to strip(trim) the bldg? @bevm0

ap0nia: Dang, what the heck, some of the "TBA" has embedded whitespaces in the HTML itself from the SOC, i.e. the `&nbsp;`
image

They might just be preserving that data and sending it to us. Yeah, I guess we can handle that here.

Can you just add a note somewhere, e.g. above `FAKE_LOCATIONS`, that the location data has inconsistent white spacing when received from the API, which motivates the trimming? We can merge it once that's done

JacE070: uhh I just saw your comment right after Eric merged it ; ( 

So the note hasn't been added yet...

EricPedley: Oops I didn't realize Brian wanted more changes on this 😐. 
I just read the comment and that change would be appreciated so just open another PR for it

PR #519, opened 2023-03-04
Add note for Map tba bug

JacE070:
https://github.com/icssc/AntAlmanac/pull/517#discussion_r1125139352


ap0nia (approved):
Looks good! Thanks for adding the clarification 👍

PR #523, opened 2023-03-06
Improve custom events page on mobile

### LINKED ISSUE #244: Improve custom events page on mobile
For the custom events page, it looks a bit cluttered because the names for each day take a lot of space. Is it possible to just use the abbreviations of those days (Sun, Mon, Tue, etc.) on mobile?

CustomEvent

_Originally posted by @Douglas-Hong in https://github.com/icssc-projects/AntAlmanac/pull/242#pullrequestreview-883786454_

mina03333:
## Summary
Used abbreviations (Mon, Tue, Wed...) on mobile to make it look less cluttered. 
Below is what it looks like on PC
Screenshot 2023-03-06 141028
Below is how it should look like on mobile
Screenshot 2023-03-06 141100

## Test Plan
Make sure this change applies to different kinds of mobile devices.

## Issues

Closes #244


src/components/Calendar/Toolbar/CustomEventDialog/DaySelector.tsx
```diff
@@ -11,6 +11,12 @@ interface DaySelectorProps {
 interface DaySelectorState {
     days: boolean[];
 }
+
+function detectMobile() {
+    const isMobile = window.innerWidth <= 800 && window.innerHeight <= 600;
+    return isMobile;
+}
+
```
ap0nia: Try to use the [MUI utility](https://v4.mui.com/components/use-media-query/#simple-media-query) for this if you can; using the `window` object is generally unadvisable because it's not necessarily consistent with the state of the app.

[An example of the usage in MUI5 is here](https://github.com/icssc/AntAlmanac/blob/aponia-monorepo/apps/client/src/routes/Home.tsx#L19), where I used the theme's breakpoints in conjunction with the media query. However, the examples are for functional components; you may need to adapt it for this class component.

src/components/Calendar/Toolbar/CustomEventDialog/DaySelector.tsx
```diff
@@ -30,93 +36,183 @@ class DaySelector extends PureComponent<DaySelectorProps, DaySelectorState> {
...
-                            onChange={this.handleChange(6)}
-                            value="6"
-                            color="primary"
-                        />
-                    }
-                    label="Saturday"
-                />
-            </FormGroup>
-        );
+        if (detectMobile() == false) {
```
ap0nia: Don't create every option manually (everything below this line). Define an array, and map through it.

Reference: https://reactjs.org/docs/lists-and-keys.html
Nicer guide from the beta docs: https://beta.reactjs.org/learn/rendering-lists

Here's an example from my implementation of the form:
1. [Define the array of options](https://github.com/icssc/AntAlmanac/blob/aponia-monorepo/apps/client/src/components/buttons/CustomEvent.tsx#L36)
2. [Iterate through the array and generate a JSX element for each, using the value and index](https://github.com/icssc/AntAlmanac/blob/aponia-monorepo/apps/client/src/components/buttons/CustomEvent.tsx#L192)

ap0nia (changes requested):
This component can be greatly simplified with idiomatic React

src/components/Calendar/Toolbar/CustomEventDialog/DaySelector.tsx
EricPedley: Agreed on this point. Also, `==false` is kinda clunky here you should rename the  function (see previous comment) and just do `!(function call)`

src/components/Calendar/Toolbar/CustomEventDialog/DaySelector.tsx
```diff
@@ -11,6 +11,12 @@ interface DaySelectorProps {
 interface DaySelectorState {
     days: boolean[];
 }
+
+function detectMobile() {
```
EricPedley: nit: rename this function to `isMobile` so that it reads more naturally in the conditional later.

EricPedley (dismissed):
- minor function naming nitpick
- I agree with Brian's comments
remember to re-request review once you address the stuff. As it stands right now your code works which is good but it'd be much more readable and easier to change later if we didn't have to change the form options in two places (each branch of the conditional)

ap0nia (approved):
Looks good!

Voark (approved):

PR #526, opened 2023-03-11
feat: rewrote the map

ap0nia:
I got skill diffed by eslint


ap0nia:
Ok the linting config actually completely fucks over my map so I'm gonna go overhaul it first.

This is literally the objectively correct import order in every way, Airbnb approved:
image

But why the fuck does it reorder it like this????
image

PR #527, opened 2023-03-11
Eslint refactor

ap0nia:
Simplified eslint config and cleaned up some files while I was at it


src/index.tsx
```diff
@@ -1,8 +1,3 @@
 import { render } from 'react-dom';
-
 import App from './App';
-import { unregister } from './registerServiceWorker';
```
ap0nia, line 4: I'm pretty sure this is purely an artifact of CRA, and not relevant in Vite. Removal of `registerServiceWorker` altogether should be fine

src/components/RightPane/SectionTable/SectionTableBody.tsx
```diff
@@ -95,7 +96,8 @@ const CourseCodeCell = withStyles(styles)((props: CourseCodeCellProps) => {
     return (
         <NoPaddingTableCell className={classes.cell}>
             <Tooltip title="Click to copy course code" placement="bottom" enterDelay={300}>
-                <div
+                <Button
```
ap0nia, line 98: Section code in the right course pane is now a button; doesn't seem to significantly impact the UI. But I can style it more to make it seem "invisible" like before

src/components/RightPane/CoursePane/SearchForm/AdvancedSearch.tsx
```diff
@@ -250,18 +251,19 @@ class AdvancedSearch extends PureComponent<AdvancedSearchProps, AdvancedSearchSt
     };
 
     render() {
-        const { classes } = this.props;
-
         return (
             <>
-                <div onClick={this.handleExpand} className={classes.container}>
+                <Button
```
ap0nia, line 256: Section code in the calendar popup is now a button; doesn't seem to significantly impact the UI. But I can style it more to make it seem "invisible" like before.

src/components/Calendar/CourseCalendarEvent.tsx
```diff
@@ -145,18 +145,19 @@ const CourseCalendarEvent = (props: CourseCalendarEventProps) => {
                         <tr>
                             <td className={classes.alignToTop}>Section code</td>
                             <Tooltip title="Click to copy course code" placement="right">
-                                {/* eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions */}
```
ap0nia, line 148: This can be removed since onClick was moved into the button with the section code

src/components/RightPane/CoursePane/SearchForm/AdvancedSearch.tsx
EricPedley: It's fine no need to fix the styling.

src/index.tsx
EricPedley: Yeah, confirmed with network inspector that removing this doesn't stop the page from caching itself.

.eslintrc.yml
```diff
@@ -1,54 +1,35 @@
...
-    browser: true
-    node: true
-    es6: true
-
-plugins: 
+plugins:
+    - react
+    - react-hooks
     - import
-    - simple-import-sort
```
EricPedley, line 16: I kinda liked having auto import sorting. Are there any other isort plugins that we can get that don't screw up the leaflet imports? It's fine if not.

EricPedley (changes requested):

.eslintrc.yml
ap0nia: Sure, I'll turn on this rule. https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md

ap0nia:
I added one commit where I just turned on the rule.

The next commit is the result of running `yarn lint --fix`. Any out of order imports are automatically sorted.

Caveats: they'll show up as "fixable errors", so the `--fix` is required to sort them. Screenshot below
image

Maybe it's a bit heavy-handed for them to register as errors, but I don't anticipate this being a major issue because the `.lintstagedrc.yml` config already runs `eslint --fix`, so import orders will get fixed and not throw an error during pre-commit. They'll only show up if you run `yarn lint`, and it's a straightforward process to just run `yarn lint --fix` to resolve the previous errors.

EricPedley (approved):
lgtm

PR #530, opened 2023-03-18
Fix map-linking bugs

### LINKED ISSUE #525: Map move to wrong coordinate after clicking on map link
The map will move to the right position, then snap to the wrong one. This is most apparent on mobile devices, where the popup will be off-screen. That's not an issue when activating the popup from the map search bar. It only happens when clicking from the calendar while the map is not selected/visible. 

To reproduce:
Search for a class with a real location, add it to the schedule, and click on that location link. However, this doesn't always reproduce the issue.

### LINKED ISSUE #529: Linking to map doesn't work for buildings with non-alpha buildings
E.g., BS3

This is because of this regex
image

Fix: Take everything up to the first space. From a visual scan of `locations.json`, no location contains space.

### LINKED ISSUE #531: Marker popup always pops up when opening
Even when opening the map tab without clicking on a map link, a popup will appear.

To reproduce: Add some classes, open map tab.

MinhxNguyen7:
## Summary
Map moves to the wrong coordinates sometimes. Seems to only happen when moving from the calendar (instead of the section table) and while the map is not selected. This is not a problem with events firing multiple times because that was not observed. #525 

Map-linking doesn't work on buildings with a number. #529 

Marker popup opens inappropriately. # 531

## Test Plan
Make sure that map works and looks good.

## Issues
Closes #525 
Closes #529 
Closes #531


MinhxNguyen7:
Issue with moving to the wrong location seems to be fixed when I clear the previous marker first while moving to the marker location before placing it. Not sure why that was the case.

ap0nia (approved):
LGTM

PR #533, opened 2023-03-29
Add 2023 Summer Sessions

Douglas-Hong:
## Summary

- Added the ability to search for classes in any of the 2023 summer sessions.
- The default term will still be 2023 Spring Quarter.

## Test Plan
- Make sure the class search results are correct for each quarter.
- Make sure you can add/delete/save/load courses for each quarter.
- Make sure the .ics file contains the proper calendar events for each summer class.


EricPedley (approved):
🚀

PR #534, opened 2023-04-03
Fix reading undefined noteid

Voark:
## Summary
Switching to the added courses pane will sometimes crash AntAlmanac. Can reproduce easily by loading the schedule `test` in the prod site and switch to added courses before the schedule loads in.

## Test Plan
Loaded the same schedule and others and seems fine.


Douglas-Hong (approved):
Looks good to me, thanks for catching this!

PR #535, opened 2023-04-03
Change out mapbox API token

EricPedley:
## Summary
Replaced invalid mapbox api token with a new one. When we get the chance we should figure out how to not publicly expose this but right now I don't care enough to figure it out.
## Test Plan
Make sure the map renders.

### DIFF
--- a/src/components/RightPane/Map/UCIMap.tsx
+++ b/src/components/RightPane/Map/UCIMap.tsx
@@ -52,7 +52,7 @@ class LocateControl extends PureComponent<{ leaflet: LeafletContext }> {
 const LocateControlLeaflet = withLeaflet(LocateControl);
 
 const DAYS = ['', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri'];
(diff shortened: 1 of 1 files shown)


ap0nia (approved):
I can't check if the key actually works, but LGTM :pray:

Voark (approved):
We could move it to an env variable in the future potentially but also think this is fine for now. Just checked and the map shows up in staging.

ap0nia:
> We could move it to an env variable in the future potentially but also think this is fine for now. Just checked and the map shows up in staging.

No that doesn't work because all env variables are statically replaced at build time; the final JavaScript bundle will inline all env variables we declare...front-end moment :tm:

It would just obfuscate the API key for somebody browsing the repo or contributing without the env file.

Voark:
> > We could move it to an env variable in the future potentially but also think this is fine for now. Just checked and the map shows up in staging.
> 
> No that doesn't work because all env variables are statically replaced at build time; the final JavaScript bundle will inline all env variables we declare...front-end moment ™️
> 
> It would just obfuscate the API key for somebody browsing the repo or contributing without the env file.

You're right, imagine being able to think in the morning 😔

MinhxNguyen7 (approved):
Looks good; map renders.

PR #539, opened 2023-04-10
Loading spinner

Voark:
## Summary
- Add loading spinner to the load schedule button and disable save schedule button while loading
- Helps communicate that we are actually indeed loading a schedule
- Prevents saving a schedule with messed up/no data[^1].  

## Test Plan
- Load a schedule

[^1]: Assuming the only way to trigger this bug is by saving a schedule before one finishes loading. This is the only way I've been able to reproduce it but if there is another way to cause the bug then IDK.

### DIFF
--- a/src/components/AppBar/LoadSaveFunctionality.tsx
+++ b/src/components/AppBar/LoadSaveFunctionality.tsx
@@ -18,7 +18,7 @@ import { isDarkMode } from '$lib/helpers';
 
 interface LoadSaveButtonBaseProps {
     action: typeof saveSchedule;
-    actionName: string;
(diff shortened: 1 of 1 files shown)


EricPedley (approved):
```
             ____
              ---|
  \/            /|     \/
               / |\
              /  | \        \/
             /   || \
            /    | | \
           /     | |  \
          /      | |   \
         /       ||     \
        /        /       \
       /________/         \
       ________/__________--/
 ~~~   \___________________/
         ~~~~~~~~~~       ~~~~~~~~
~~~~~~~~~~~~~     ~~~~~~~~~
                               ~~~~~~~~~

Ship it!
```

PR #540, opened 2023-04-10
Prerequisite tree modal

### LINKED ISSUE #402: Prerequisite tree modal
Right now PeterPortal has a cool graphic for each course that displays its prerequisites and classes that depend on it as a prerequisite:
image
It would be cool to have this as a modal when someone clicks on the info button for a course (see below for what I mean when I say info button). Not the prerequisite button because we're going to remove it as per #393 
image

teresa-liang:
## Summary
displays the prerequisites for a course as a tree in the info button modal
e.g.
image
aa_prereq_tree

## Test Plan
can check that the prerequisite tree matches the prerequisites for each course

## Issues

Closes #402


ap0nia:
This is some pretty neat stuff! :clap:

I'm a little hesitant to add Semantic UI to the project because we use Material-UI. Using multiple UI frameworks can cause weird styling issues sometimes, but I also don't know much about the former so idk how the two would interop.

Have you ever seen [mermaid.js](https://mermaid.js.org/)? It's dedicated to drawing out diagrams, and might be a good fit for drawing out a tree like this. I've never used it, but I can try putting together a simple hook for using it if you're interested.

src/components/RightPane/SectionTable/PrereqTree.css
```diff
@@ -0,0 +1,194 @@
...
+    justify-content: center;
+    align-items: center;
+    padding: 0.125rem 0.25rem;
+    margin: 0 0.25rem 1px 0.25rem;
+    border-radius: 0.25rem;
+    flex: 0 0 auto;
+    border: 0;
+}
+
+.dependency-needs::before {
```
EricPedley: I don't like pseudoselectors because of their complexity but if there isn't any more elegant of a way to implement whatever style you're trying to pull off here then it's fine. Just from scanning this casually is this supposed to be the lines connecting the nodes in the tree? If so add a comment because it isn't very obvious what these are supposed to be for.

src/components/RightPane/SectionTable/PrereqTree.tsx
```diff
@@ -0,0 +1,152 @@
+/* eslint-disable prefer-const */
```
EricPedley: Why do we need the eslint disable here? Which lines offend the linter and why can't they be fixed to adhere to the rule? It'd be better to disable the linter checks per-line instead of per-file.

src/lib/peterportal.types.ts
```diff
@@ -127,3 +127,9 @@ export interface SectionGQL {
     number: string;
     type: string;
 }
+
+export interface PrerequisiteJSON {
```
EricPedley: nit: I don't like this variable name. Maybe call this PrerequisiteTreeNode and call the other class PrerequisiteTreeValue? Or call this one PrereuquisiteTreeRecord or something. Usually JSON in the name of something implies data being passed as strings in JSON format somewhere but we aren't doing encoding/decoding at all here.

src/components/RightPane/SectionTable/PrereqTree.tsx
```diff
@@ -0,0 +1,152 @@
...
+import { FC } from 'react';
+import { Grid, Popup } from 'semantic-ui-react';
+
+import { PrerequisiteJSONNode, PrerequisiteJSON } from '../../../lib/peterportal.types';
```
EricPedley: Try using `$lib` here instead of the relative path. See the `import {isDarkMode...` line as an example.

EricPedley (changes requested):
I agree with Brian that it would be better to remove the dependence on Semantic UI. Can you explain why you decided to use it for this component? Here are some reasons we'd want to switch to MUI if it's a reasonable amount of work:
1. Consistency - all our stuff would be using the same UI component building blocks
2. Bundle size - we're already importing MUI so using it doesn't change the bundle size but using another UI framework adds more to the bundle, which slows down the page load. I guess we could lazy-load it since it's hidden behind a button users have to press but that complicates the app a little more.

Also I'd like to know how PeterPortal implements their prerequisite tree modal and if there are any tricks they use that we can use, or if you've already thought of that, did you reuse any code and if not, why not?

ap0nia:
> I agree with Brian that it would be better to remove the dependence on Semantic UI. Can you explain why you decided to use it for this component? Here are some reasons we'd want to switch to MUI if it's a reasonable amount of work:
> 
> 1. Consistency - all our stuff would be using the same UI component building blocks
> 2. Bundle size - we're already importing MUI so using it doesn't change the bundle size but using another UI framework adds more to the bundle, which slows down the page load. I guess we could lazy-load it since it's hidden behind a button users have to press but that complicates the app a little more.
> 
> Also I'd like to know how PeterPortal implements their prerequisite tree modal and if there are any tricks they use that we can use, or if you've already thought of that, did you reuse any code and if not, why not?

I'm gonna be honest, I don't think the second point is really valid at this point lol. The bundle size is ~2.5MB, and the generally recommended bundle size caps off at 0.5MB (500kb). I think the most important concept is just using the right tools for the right job.

I recommend MermaidJS or any well documented React alternative because it's difficult to maintain extremely customized code in a consistent, predictable way (which is my impression from the custom CSS). Using additional libraries irrespective of the bundle size is honestly fine atm until we can actually cut the bundle size under 1MB somehow (impossible challenge) and it's mostly a matter of what we choose to use and why.

If we're really concerned about bundle size, I can do another rewrite at some point and figure out which dependencies we can gut.

src/lib/peterportal.types.ts
```diff
@@ -127,3 +127,9 @@ export interface SectionGQL {
...
+    [key: string]: PrerequisiteJSONNode[];
+}
+
+export type PrerequisiteJSONNode = PrerequisiteJSON | string;
```
ap0nia:
```suggestion
export type PrerequisiteTree = Record<string, PrerequisiteNode[]>
export type PrerequisiteNode = Prerequisite | string;
```

I think something like this would look cleaner; interfaces can get annoying sometimes because they're more verbose.

src/components/RightPane/SectionTable/PrereqTree.tsx
ap0nia: It's because "let" was used inside the function components. They should all be safely converted to "const". 

A fun part of react is how const isn't const because the entire function is re-run per state update. So "let" and "const" are effectively the same unless you explicitly re-assign halfway through the function.

teresa-liang:
> > I agree with Brian that it would be better to remove the dependence on Semantic UI. Can you explain why you decided to use it for this component? Here are some reasons we'd want to switch to MUI if it's a reasonable amount of work:
> > 
> > 1. Consistency - all our stuff would be using the same UI component building blocks
> > 2. Bundle size - we're already importing MUI so using it doesn't change the bundle size but using another UI framework adds more to the bundle, which slows down the page load. I guess we could lazy-load it since it's hidden behind a button users have to press but that complicates the app a little more.
> > 
> > Also I'd like to know how PeterPortal implements their prerequisite tree modal and if there are any tricks they use that we can use, or if you've already thought of that, did you reuse any code and if not, why not?
> 
> I'm gonna be honest, I don't think the second point is really valid at this point lol. The bundle size is ~2.5MB, and the generally recommended bundle size caps off at 0.5MB (500kb). I think the most important concept is just using the right tools for the right job.
> 
> I recommend MermaidJS or any well documented React alternative because it's difficult to maintain extremely customized code in a consistent, predictable way (which is my impression from the custom CSS). Using additional libraries irrespective of the bundle size is honestly fine atm until we can actually cut the bundle size under 1MB somehow (impossible challenge) and it's mostly a matter of what we choose to use and why.
> 
> If we're really concerned about bundle size, I can do another rewrite at some point and figure out which dependencies we can gut.

I basically reused all of PeterPortal's code for its `PrereqTree` component and tweaked it a little bit so that it would be compatible with AntAlmanac. The only differences were:

- AntAlamanac uses the Rest API to obtain data about a specific course (e.g. the prerequisites) while PeterPortal uses the GraphQL API, so there were a couple differences in the data types that certain fields returned. For example, for the field `prerequisite_list`, the Rest API returns a list of strings while GraphQL returns a dictionary of dictionaries.
- For styling, I used a css file instead of a scss file (PeterPortal's implementation) because Sass was not installed and I had seen that css files were used for styling other components.

PeterPortal used Semantic UI, which is why I added it, but I didn't take into account the consistency of our components or the increased bundle size. I can definitely remove the dependency on it and I don't think it will be that difficult to do so!

@bevm0 I have never used Mermaid.js before, but I am open to using it! I think it would make the tree diagrams cleaner by using a library instead of using pure css and pseudo-classes.

EricPedley:
Seems like rewriting in mermaid.js would be a lot of work so lets just merge this ASAP and worry about refactoring later. Just fix the merge conflicts and get `package-lock.json` and `yarn.lock` changes off this PR and we're good to go.

EricPedley (approved):
Awesome job! Ship it!
```
        |
       / \
      / _ \
     |.o '.|
     |'._.'|
     |     |
   ,'|  |  |`.
  /  |  |  |  \
  |,-'--|--'-.| 
      %##%##
       #%##
       %#
        # 
        #
```

MinhxNguyen7 (approved):
Go

PR #542, opened 2023-04-11
Fix recruitment banner issues

### LINKED ISSUE #541: Recruitment banner showing up for non-ICS majors
image

No idea why this is the case. I think it didn't use to be this way.

Also, the text overflows.

Voark:
## Summary
Fixed bug where recruitment banner shows for any major, and also fixed text overflow issues.
The recruitment banner is now an alert that pops up on the bottom right.

image
image

## Test Plan
Check that recruitment banner only shows for right majors and that the text looks fine.

Closes #541 

## Future Followup
If we keep the alert/snackbar style implementation it might be worth integrating it with the rest of our snackbar and moving it to a separate component but I did not want to set up another global state variable.

### DIFF
--- a/src/components/RightPane/CoursePane/CourseRenderPane.tsx
+++ b/src/components/RightPane/CoursePane/CourseRenderPane.tsx
@@ -91,7 +91,7 @@ const flattenSOCObject = (SOCObject: WebsocResponse): (School | Department | AAC
         return accumulator;
     }, []);
 };
-const RecruitmentBanner = (classes: ClassNameMap) => {
+const RecruitmentBanner = () => {
     const [bannerVisibility, setBannerVisibility] = React.useState<boolean>(true);
 
     // Display recruitment banner if more than 11 weeks (in ms) has passed since last dismissal
@@ -99,10 +99,8 @@ const RecruitmentBanner = (classes: ClassNameMap) => {
     const dismissedRecently =
         recruitmentDismissalTime !== null &&
         Date.now() - parseInt(recruitmentDismissalTime) < 11 * 7 * 24 * 3600 * 1000;
-    const displayRecruitmentBanner =
-        bannerVisibility &&
-        !dismissedRecently &&
-        ['COMPSCI', 'IN4MATX', 'I&C SCI', 'STATS'].includes(RightPaneStore.getFormData().deptValue);
+    const isSearchCS = ['COMPSCI', 'IN4MATX', 'I&C SCI', 'STATS'].includes(RightPaneStore.getFormData().deptValue);
+    const displayRecruitmentBanner = bannerVisibility && !dismissedRecently && isSearchCS;
 
     return (
         <div style={{ position: 'fixed', bottom: 5, right: 5, zIndex: 999 }}>
@@ -279,7 +277,7 @@ class CourseRenderPane extends PureComponent<CourseRenderPaneProps, CourseRender
 
             currentView = (
                 <>
-                    <RecruitmentBanner {...classes} />
+                    <RecruitmentBanner />
                     <div className={classes.root} style={{ position: 'relative' }}>
                         <div className={classes.spacing} />
                         {this.state.courseData.length === 0 ? (
--- a/src/components/RightPane/CoursePane/CourseRenderPane.tsx
+++ b/src/components/RightPane/CoursePane/CourseRenderPane.tsx
@@ -1,4 +1,4 @@
-import { Button, Grid, IconButton, Paper, Theme } from '@material-ui/core';
+import { Button, Grid, IconButton, Paper, Theme, useTheme } from '@material-ui/core';
 import { withStyles } from '@material-ui/core/styles';
 import { ClassNameMap, Styles } from '@material-ui/core/styles/withStyles';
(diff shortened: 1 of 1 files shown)


src/components/RightPane/CoursePane/CourseRenderPane.tsx
```diff
@@ -277,27 +273,30 @@ class CourseRenderPane extends PureComponent<CourseRenderPaneProps, CourseRender
...
-                            <img src={isDarkMode() ? darkNoNothing : noNothing} alt="No Results Found" />
-                        </div>
-                    ) : (
-                        this.state.courseData.map((_: School | Department | AACourse, index: number) => {
-                            let heightEstimate = 200;
-                            if ((this.state.courseData[index] as AACourse).sections !== undefined)
-                                heightEstimate =
-                                    (this.state.courseData[index] as AACourse).sections.length * 60 + 20 + 40;
+                    <div className={classes.root} style={{ position: 'relative' }}>
+                        <div className={classes.spacing} />
```
Voark, line 282: This div is literally just there to space the back and reload buttons so they don't overlap.
image
image

If anyone else knows a better way to do this LMK.

Voark:
I just realized I forgot to turn off my dark mode extension so the colors aren't right, will fix it.

src/components/RightPane/CoursePane/CourseRenderPane.tsx
```diff
@@ -102,39 +95,42 @@ const RecruitmentBanner = (classes: ClassNameMap) => {
...
+        recruitmentDismissalTime !== null &&
+        Date.now() - parseInt(recruitmentDismissalTime) < 11 * 7 * 24 * 3600 * 1000;
     const displayRecruitmentBanner =
         bannerVisibility &&
-        (window.localStorage.getItem('recruitmentDismissalTime') === null ||
-            (Date.now() - parseInt(window.localStorage.getItem('recruitmentDismissalTime') as string) >
-                11 * 7 * 24 * 3600 * 1000 &&
-                ['COMPSCI', 'IN4MATX', 'I&C SCI', 'STATS'].includes(RightPaneStore.getFormData().deptValue)));
+        !dismissedRecently &&
+        ['COMPSCI', 'IN4MATX', 'I&C SCI', 'STATS'].includes(RightPaneStore.getFormData().deptValue);
```
MinhxNguyen7: Nitpicking, but I'd extract this to be a separate variable, e.g., `is_ICS` if for nothing other than having all boolean variables.

MinhxNguyen7 (commented):
Looks good code-wise, but I have reservations about using an alert because it is somewhat intrusive. @EricPedley, what's your opinion? 

Also, I don't love how light it is in dark mode, and (this is pretty minor) that blue color doesn't exist anywhere else in the UI, so we're introducing a whole new color. Maybe it should just be white/grey. I'm not sure though.

Voark:
> Looks good code-wise, but I have reservations about using an alert because it is somewhat intrusive. @EricPedley, what's your opinion?
> 
> Also, I don't love how light it is in dark mode, and (this is pretty minor) that blue color doesn't exist anywhere else in the UI, so we're introducing a whole new color. Maybe it should just be white/grey. I'm not sure though.

Yeah will get to the colors tomorrow, didn't feel like it today. 

Figuring out sizing in that small spacing between the buttons and the edge of the table was too hard so I didn't feel like messing with it. In terms of the intrusiveness I think because the alert won't pop up again after it's closed once it's not too bad? I can also move it to bottom left if it would be less intrusive there. Otherwise someone else will have to take over for me because I cannot style a component properly to fit in that space without overflow.

src/components/RightPane/CoursePane/CourseRenderPane.tsx
```diff
@@ -98,43 +91,49 @@ const flattenSOCObject = (SOCObject: WebsocResponse): (School | Department | AAC
...
-                            We have opportunities for experienced devs and those with zero experience!
-                        </div>
-
-                        <Button
+                <Alert
+                    icon={false}
+                    severity="info"
+                    style={{
+                        color: isDarkMode() ? '#ece6e6' : '#2e2e2e',
+                        backgroundColor: isDarkMode() ? '#2e2e2e' : '#ece6e6',
```
Voark, line 113: If there is a better way to theme this LMK.

EricPedley: I can't think of anything. Turns out the `color` prop of Alert doesn't let you enter `primary` or `secondary` like buttons and stuff, only `info`, `error`, etc.

src/components/RightPane/CoursePane/CourseRenderPane.tsx
EricPedley: 🤷‍♂️

EricPedley (approved):
Thanks for taking care of this. It's been pretty ironic that the banner advertising web dev positions was badly styled. Reminds me of this: 
image
I think by being in the bottom right it isn't that intrusive, and I just want to ship *some* fix right now so it isn't broken. If someone wants to style this to be less intrusive the change would be welcome.

PR #546, opened 2023-04-15
Add Sharability (Up-to-date version)

stevenguyukai:
# INTRO

- This has the same functionality as the issue #497 .
- If this is merged into the main, the #497 can be mark done as well. Also the branch `<Add_Sharability_YukaiGu>` can be deleted. (This branch is `<Add_Sharability_**New**_YukaiGu>`)

- I created this new up-to-date pr because the old one is having lots of conflicts with the current version. So I decided just create a new branch from main and work from the current version

# FUNCTIONALITY
## 1. Copy link from a specific Class
By clicking the `+` sign, user can use the `<Cope Link>` Button to copy the url for this course

image

A Notification will show on the left bottom 

image

A copied link is something like ⬇️

image

## 2. Copy link directly from the browser
User's inputs in search will automatically reflected into the url

Sharability_Show

After user manually copy the url, they can share with their friends and friends can use the link to find the class been shared

Cope_Link_Into_Browser


.husky/pre-commit
ap0nia, line 1: How come this was changed? I can't seem to find a `test` script in the `package.json`. I'd generally do something like this for a linting pre-commit hook

```sh
#!/bin/sh
. "$(dirname -- "$0")/_/husky.sh"

npx lint-staged
```

src/components/RightPane/RightPaneStore.ts
```diff
@@ -29,13 +29,25 @@ class RightPaneStore extends EventEmitter {
...
+    private urlCourseCodeValue: string;
+    private urlTermValue: string;
+    private urlGEValue: string;
+    private urlCourseNumValue: string;
+    private urlDeptLabel: string;
+    private urlDeptValue: string;
     constructor() {
         super();
         this.setMaxListeners(15);
         this.formData = { ...defaultFormValues }; // creates shallow copy
```
ap0nia:
```suggestion
        this.formData = structuredClone(defaultFormValues)
```
This is a built-in way to create a deep copy. I think we had issues with `structuredClone` on older browsers before tho.

src/components/RightPane/RightPaneStore.ts
```diff
@@ -29,13 +29,25 @@ class RightPaneStore extends EventEmitter {
...
         this.activeTab = 0;
         this.doDisplaySearch = true;
         this.openSpotAlertPopoverActive = false;
+        this.urlCourseCodeValue = String(new URLSearchParams(window.location.search).get('courseCode'));
+        this.urlTermValue = String(new URLSearchParams(window.location.search).get('term'));
+        this.urlGEValue = String(new URLSearchParams(window.location.search).get('GE'));
+        this.urlCourseNumValue = String(new URLSearchParams(window.location.search).get('courseNumber'));
+        this.urlDeptLabel = String(new URLSearchParams(window.location.search).get('deptLabel'));
+        this.urlDeptValue = String(new URLSearchParams(window.location.search).get('deptValue'));
```
ap0nia:
```suggestion
        this.urlCourseCodeValue = new URLSearchParams(window.location.search).get('courseCode').toString();
        this.urlTermValue = new URLSearchParams(window.location.search).get('term').toString();
        this.urlGEValue = new URLSearchParams(window.location.search).get('GE').toString();
        this.urlCourseNumValue = new URLSearchParams(window.location.search).get('courseNumber').toString();
        this.urlDeptLabel = new URLSearchParams(window.location.search).get('deptLabel').toString();
        this.urlDeptValue = new URLSearchParams(window.location.search).get('deptValue').toString();
```
I guess this is a preference; I rarely use the primitive constructors

src/components/RightPane/RightPaneStore.ts
```diff
@@ -54,6 +66,30 @@ class RightPaneStore extends EventEmitter {
...
+        return this.urlCourseNumValue;
+    };
+
+    getUrlDeptLabel = () => {
+        return this.urlDeptLabel;
+    };
+
+    getUrlDeptValue = () => {
+        return this.urlDeptValue;
+    };
```
ap0nia: I'm not sure if these would be worth it in the long run; it makes sense to support uni-directional data flows, but it seems like a lot of boilerplate. 

To compromise, here's a re-formatted version that takes up less space; I think it makes sense in this scenario because these are just boilerplate getter methods with no internal logic. 
```suggestion
    getUrlCourseCodeValue = () => this.urlCourseCodeValue;
    getUrlTermValue = () => this.urlTermValue;
    getUrlGEValue = () => this.urlGEValue;
    getUrlCourseNumValue = () =>this.urlCourseNumValue;
    getUrlDeptLabel = () => this.urlDeptLabel;
    getUrlDeptValue = () => this.urlDeptValue;
```

src/components/RightPane/CoursePane/SearchForm/TermSelector.tsx
```diff
@@ -32,6 +46,24 @@ class TermSelector extends PureComponent<TermSelectorProps> {
...
+            urlParam.append('term', event.target.value as string);
+            const new_url = `?${urlParam.toString()}`;
+            history.replaceState(stateObj, "url", "/" + new_url);
+        }else{
+            if (urlParam.toString() == "" || urlParam.toString() == null){
+                const new_url = `${urlParam.toString()}`;
+                history.replaceState(stateObj, "url", "/" + new_url);
+            }else{
+                const new_url = `?${urlParam.toString()}`;
+                history.replaceState(stateObj, "url", "/" + new_url);
```
ap0nia: If it's possible, I'd recommend manipulating react-router-dom's history instead of the browser history. Maybe this API offers the functionality you're looking for? https://reactrouter.com/en/main/hooks/use-navigate

src/components/RightPane/CoursePane/SearchForm/TermSelector.tsx
```diff
@@ -32,6 +46,24 @@ class TermSelector extends PureComponent<TermSelectorProps> {
...
+            }
+        }
```
ap0nia: Both of the bottom clauses look the same, and they all terminate similarly? So I think this might be simpler.
```suggestion
        if (typeof event.target.value === 'string' && event.target.value) {
            urlParam.append('term', event.target.value);
        } 
        const new_url = `${urlParam.toString()}`;
        history.replaceState(stateObj, "url", "/" + new_url);
```

src/components/RightPane/SectionTable/SectionTableButtons.tsx
```diff
@@ -91,6 +91,25 @@ export const ScheduleAddCell = withStyles(styles)((props: ScheduleAddCellProps)
...
+        urlParam.delete('deptValue');
+        urlParam.delete('GE');
+        urlParam.delete('term');
+        urlParam.append('courseCode', String(section.sectionCode));
+        const new_url = `${url.origin.toString()}/?${urlParam.toString()}`;
+        navigator.clipboard.writeText(new_url.toString()).then(function() {
+            openSnackbar('success', 'Course Link Copied!');
+        }, function() {
+            openSnackbar('error', 'Fail to copy the link!');
+        });
```
```suggestion
        navigator.clipboard.writeText(new_url.toString()).then(() => {
            openSnackbar('success', 'Course Link Copied!');
        }, () => {
            openSnackbar('error', 'Fail to copy the link!');
        });
```
I prefer the simpler arrow function syntax for cases like this.

src/components/RightPane/SectionTable/SectionTableButtons.tsx
```diff
@@ -109,6 +128,9 @@ export const ScheduleAddCell = withStyles(styles)((props: ScheduleAddCellProps)
                     <MenuItem onClick={() => closeAndAddCourse(scheduleNames.length, true)}>
                         Add to All Schedules
                     </MenuItem>
+                    <MenuItem onClick={() => closeCopyAndAlert()}>
```
ap0nia: I think this should suffice. In general I like to create handler functions that can be directly called by the event so I don't have to inline a function definition.
```suggestion
                    <MenuItem onClick={closeCopyAndAlert}>
```

src/components/RightPane/SectionTable/SectionTableButtons.tsx
```diff
@@ -109,6 +128,9 @@ export const ScheduleAddCell = withStyles(styles)((props: ScheduleAddCellProps)
                     <MenuItem onClick={() => closeAndAddCourse(scheduleNames.length, true)}>
```
ap0nia: Same idea as the other event handler. Is it possible to make `closeAndAddCourse` use the correct value internally, or write a handler function that will? I just try to avoid inlining function calls if I can
```suggestion
                    <MenuItem onClick={closeAndAddCourse}>
```

src/components/RightPane/SectionTable/SectionTableButtons.tsx
```diff
@@ -91,6 +91,25 @@ export const ScheduleAddCell = withStyles(styles)((props: ScheduleAddCellProps)
...
 
+    const closeCopyAndAlert = () => {
+        const url = new URL(window.location.href)
+        const urlParam = new URLSearchParams(url.search);
+        urlParam.delete('courseCode');
+        urlParam.delete('courseNumber');
+        urlParam.delete('deptLabel');
+        urlParam.delete('deptValue');
+        urlParam.delete('GE');
+        urlParam.delete('term');
```
ap0nia: Do these have any significance? i.e. are they all of the form values or just some specific ones? Here's a revised approach if it's only subset of the form fields.
```suggestion
       // somewhere at the top of the file or something, 
       // and include a description, i.e. why they need to be reset
       const fieldsToReset = ['courseCode', 'courseNumber', 'deptLabel', 'deptValue', 'GE', 'term']
       fieldsToReset.forEach(field => urlParam.delete(field))
```

src/components/RightPane/CoursePane/SearchForm/TermSelector.tsx
```diff
@@ -13,8 +13,22 @@ interface TermSelectorProps {
...
 
 class TermSelector extends PureComponent<TermSelectorProps> {
+    
+    updateTermAndGetFormData(){
+        RightPaneStore.updateFormValue("term", RightPaneStore.getUrlTermValue()) 
+        return RightPaneStore.getFormData().term
+    };
+
+    getTerm(){
+        if(RightPaneStore.getUrlTermValue() != "null" && RightPaneStore.getUrlTermValue() != "" && RightPaneStore.getUrlTermValue() != " " ){
```
ap0nia: I don't think we should invoke the function three times; just save it in a variable first and run comparisons on the calculated value. Also, if we need to ignore whitespaces, try using some variant of `String.trim` to remove the whitespaces prior to comparing.
```suggestion
       const urlTermValue = RightPaneStore.getUrlTermValue();
        if(urlTermValue != "null" && urlTermValue.trim() != "" ){
```

src/components/RightPane/CoursePane/SearchForm/SearchForm.tsx
```diff
@@ -48,7 +48,22 @@ const styles: Styles<Theme, object> = {
...
+    const urlParamValue4 = new URLSearchParams(window.location.search).get("deptValue") != "" && 
+    new URLSearchParams(window.location.search).get("deptValue") != null;
+    const urlParamValue5 = new URLSearchParams(window.location.search).get("GE") != "" && 
+    new URLSearchParams(window.location.search).get("GE") != null;
+    const urlParamValue6 = new URLSearchParams(window.location.search).get("term") != "" && 
+    new URLSearchParams(window.location.search).get("term") != null;
+    const [showLegacySearch, setShowLegacySearch] = useState(urlParamValue1 == true || 
+        urlParamValue2 == true || urlParamValue3 == true || 
+        urlParamValue4 == true || urlParamValue5 || 
+        urlParamValue6 == true == true ? true : false);
```
ap0nia: Could we try something like this? Put all of the url param values into an array, maybe transform them if you'd like (i.e. explicitly map them into boolean statements), and then use `Array.some(Boolean)` to check if any values are truthy.
```suggestion
    const allParamValues = [ urlParamValue1, urlParamValue2, urlParamValue3, ... ]
    const [showLegacySearch, setShowLegacySearch] = useState(allParamValues.some(Boolean))
```

ap0nia (changes requested):
Looking pretty good! :sunglasses:

I left comments on the `TermSelector` file, but those can be applicable to all of the other form fields. I also left some code recommendations.

src/components/RightPane/RightPaneStore.ts
stevenguyukai: I think the `String()` constructor is better here. Because the `String()` function will call the `toString()` method of the object passed at the end. And also the `toString()` method cannot handle `null` or `undefined`, which will occur a lot in this scenario.

src/components/RightPane/CoursePane/SearchForm/TermSelector.tsx
stevenguyukai: So here I am trying to replace the browser's current url to a parameter appended status. I cannot find an API for react to do this smoothly. By changing the browser history, it can seamlessly change the url show on the web browser without reloading the whole website (Say kinda like redirect the website to the parameter appended url). And I have tried using `location.replace()`, it will redirect the web page whenever user change the input value (that will be very annoying)

src/components/RightPane/CoursePane/SearchForm/TermSelector.tsx
stevenguyukai: In fact there is a little bit difference here. The the main problem is if this changing value is the first change in the url, then it has to stringed with`?` in url  to let values after it work as a parameter. And if the changing value isn't the first one, then it should not append more `?` into the url.

ap0nia: Oh oops I missed that, what about something like this? It looks kinda funny, but I think it's better than making an control branch to handle this simple case?
```ts
   const param = urlParam.toString()
   const new_url = `${param && param !== 'null' ? '?' : ''}${param}`;
   history.replaceState(stateObj, "url", "/" + new_url);
```

src/components/RightPane/SectionTable/SectionTableButtons.tsx
stevenguyukai: I didn't write it but I test it I can't just use `<MenuItem onClick={closeAndAddCourse}>`. I wrote a handler function to call the `closeAndAddCourse()`, like that ` <MenuItem onClick={addCourseHandler}>`.  It is working.

src/components/RightPane/CoursePane/SearchForm/SearchForm.tsx
```diff
@@ -48,7 +48,33 @@ const styles: Styles<Theme, object> = {
...
+        new URLSearchParams(window.location.search).get('term') != '' &&
+        new URLSearchParams(window.location.search).get('term') != null;
+    const allParamValues = [
+        urlParamValue1,
+        urlParamValue2,
+        urlParamValue3,
+        urlParamValue4,
+        urlParamValue5,
+        urlParamValue6,
+    ];
```
ap0nia: Only calculate the search params once. Also, if these params are only used once, you can just inline it into the array, or inline a conditional. The latter is more similar to what you did initially lol; just some ideas. Definitely only use `new URLSearchParams` once tho.

```suggestion
    const search = new URLSearchParams(window.location.search)
    const urlParamValue1 = search.get('courseCode') != '' && search && != null;
    const urlParamValue2 = search.get('courseNumber') != '' && search.get('courseNumber') != null;
    const urlParamValue3 = search.get('deptLabel') != '' &&  search.get('deptLabel') != null;
    const urlParamValue4 = search.get('deptValue') != '' && search.get('deptValue') != null;
    const urlParamValue5 = search.get('GE') != '' && search.get('GE') != null;
    const urlParamValue6 = search.get('term') != '' && search.get('term') != null;
    const allParamValues = [
        urlParamValue1,
        urlParamValue2,
        urlParamValue3,
        urlParamValue4,
        urlParamValue5,
        urlParamValue6,
    ];
    
    // alternatively
   const isLegacySearch = 
    (search.get('courseCode') != '' && search && != null) ||
    (search != '' && search.get('courseNumber') != null) ...
    
     const [showLegacySearch, setShowLegacySearch] = useState(isLegacySearch)
```

src/components/RightPane/SectionTable/SectionTableButtons.tsx
```diff
@@ -91,6 +91,30 @@ export const ScheduleAddCell = withStyles(styles)((props: ScheduleAddCellProps)
...
+    const addCourseHandler = () => {
+        closeAndAddCourse(scheduleNames.length, true);
+    };
+
+    const closeCopyAndAlert = () => {
+        const url = new URL(window.location.href);
+        const urlParam = new URLSearchParams(url.search);
+        // Reset these params in url becasue when copy a specific class's link, it only copy its course code
+        // if there is random value let in the url, it will mess up the url copied.
+        const fieldsToReset = ['courseCode', 'courseNumber', 'deptLabel', 'deptValue', 'GE', 'term'];
```
ap0nia: You can move this to the top level of the file so it doesn't have to be re-calculated per handler call.

src/components/RightPane/RightPaneStore.ts
ap0nia: Is this the reason why `"null"` can show up as form values? Hmm, what if we did something like this? I think all the results should either be string or nullish, and or-ing nullish results with an empty string will prevent "null" from ever being stringified. This could remove the need to check for `"null"` strings entirely?

```suggestion
        const search = new URLSearchParams(window.location.search)
        this.urlCourseCodeValue = search.get('courseCode') || '';
        this.urlTermValue = search.get('term') || '';
        this.urlGEValue = search.get('GE') || '';
        this.urlCourseNumValue = search.get('courseNumber') || '';
        this.urlDeptLabel = search.get('deptLabel') || '';
        this.urlDeptValue = search.get('deptValue') || '';
```

ap0nia (dismissed):
The strategy looks solid, I just left some final optimization notes - mainly only calculating `URLSearchParams` once and (hopefully) removing the need to compare with `"null"` by modifying the url parsing slightly

EricPedley (commented):
Gonna defer review to Brian here since he's already left some comments

stevenguyukai:
@bevm0 could you please do a last review and approval. I think this brach is good to merge into `main`

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/GESelector.tsx
```diff
@@ -30,18 +30,45 @@ const styles = {
...
+        return RightPaneStore.getFormData().ge;
+    }
+
+    getGe() {
+        const urlGEValue = RightPaneStore.getUrlGEValue();
+        if (urlGEValue != 'null' && urlGEValue.trim() != '') {
+            return this.updateGEAndGetFormData();
+        } else {
+            return RightPaneStore.getFormData().ge;
+        }
```
ap0nia:
```suggestion
        return RightPaneStore.getUrlGEValue().trim()
          ? this.updateGEAndGetFormData()
          : RightPaneStore.getFormData().ge
```

ap0nia (approved):
Thanks for your patience, everything looks good! I left one last suggestion that I'll merge in

ap0nia:
Oh lol, could you try pushing another change @stevenguyukai? It says I can't fully approve it since I last pushed it lol

stevenguyukai:
> Oh lol, could you try pushing another change @stevenguyukai? It says I can't fully approve it since I last pushed it lol

Yeah sure, although I send a request review to Eric, but yea I will push a useless change

ap0nia (approved):
LGTM, same approval as the initial one

PR #548, opened 2023-04-16
Different sections have different default colors.

### LINKED ISSUE #509: Make lectures and discussions/labs be different hues by default
Here are the helper functions @teresa-liang wrote to do color conversion.
```
export function HexToHSL(hex: string): { h: number; s: number; l: number } {
    const result = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex);

    if (!result) {
        throw new Error('Could not parse Hex Color');
    }

    const rHex = parseInt(result[1], 16);
    const gHex = parseInt(result[2], 16);
    const bHex = parseInt(result[3], 16);

    const r = rHex / 255;
    const g = gHex / 255;
    const b = bHex / 255;

    const max = Math.max(r, g, b);
    const min = Math.min(r, g, b);

    let h = (max + min) / 2;
    let s = h;
    let l = h;

    if (max === min) {
        // Achromatic
        return { h: 0, s: 0, l };
    }

    const d = max - min;
    s = l > 0.5 ? d / (2 - max - min) : d / (max + min);
    switch (max) {
        case r:
            h = (g - b) / d + (g < b ? 6 : 0);
            break;
        case g:
            h = (b - r) / d + 2;
            break;
        case b:
            h = (r - g) / d + 4;
            break;
    }
    h /= 6;

    s = s * 100;
    s = Math.round(s);
    l = l * 100;
    l = Math.round(l);
    h = Math.round(360 * h);

    return { h, s, l };
}

export function HSLToHex(hsl: { h: number; s: number; l: number }): string {
    const { h, s, l } = hsl;

    const hDecimal = l / 100;
    const a = (s * Math.min(hDecimal, 1 - hDecimal)) / 100;
    const f = (n: number) => {
        const k = (n + h / 30) % 12;
        const color = hDecimal - a * Math.max(Math.min(k - 3, 9 - k, 1), -1);

        // Convert to Hex and prefix with "0" if required
        return Math.round(255 * color)
            .toString(16)
            .padStart(2, '0');
    };
    return `#${f(0)}${f(8)}${f(4)}`;
}
```

MinhxNguyen7:
## Summary
Currently, sections of the same course have the same color. Now, they will have slightly different colors.

Apparently, color theory is not simple. Certain default colors might have to be removed if they don't look good with the algorithm.

image

## Test Plan
Check that it works for many courses and that it looks fine.

## Issues

Closes #509

### DIFF
--- a/apps/antalmanac/src/stores/Schedules.ts
+++ b/apps/antalmanac/src/stores/Schedules.ts
@@ -1,24 +1,16 @@
-import {
-    amber,
-    blue,
-    deepPurple,
-    green,
-    pink,
-    purple,
-    deepOrange,
-} from '@material-ui/core/colors';
+import {amber, blue, deepOrange, deepPurple, green, pink, purple,} from '@material-ui/core/colors';
 
-import { calendarizeCourseEvents, calendarizeCustomEvents, calendarizeFinals } from './calendarizeHelpers';
+import {calendarizeCourseEvents, calendarizeCustomEvents, calendarizeFinals} from './calendarizeHelpers';
 import {
+    HSLColor,
     Schedule,
     ScheduleCourse,
     ScheduleSaveState,
     ScheduleUndoState,
     ShortCourseSchedule,
-    HSLColor,
 } from './schedule.types';
-import { RepeatingCustomEvent } from '$components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog';
-import { combineSOCObjects, CourseInfo, getCourseInfo, queryWebsoc } from '$lib/helpers';
+import {RepeatingCustomEvent} from '$components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog';
+import {combineSOCObjects, CourseInfo, getCourseInfo, queryWebsoc} from '$lib/helpers';
 
 const defaultColors = [
     blue[500],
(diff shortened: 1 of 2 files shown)


teresa-liang (approved):
This looks really good! I like the ordering of colors - it's nice to have the first few classes you add not be similar in color. :thumbsup:

teresa-liang (commented):
Hmm looking at the picture of the example schedule you showed, I think the colors are too similar to be able to differentiate different sections. I would suggest varying the lightness between each section by like 15, so the new color should be +15 from the original color, then -15, then +30, then -30, etc. Also, I would suggest not changing the saturation because colors that are close to each other on the rainbow may start to look like each other.

For example with blue and green, you can get a schedule like this: 
Schedule (9)

MinhxNguyen7:
@EricPedley @teresa-liang Can you take a look at the PR now? It's been changed to lightness variation now, which works pretty well. I think it didn't look good in my previous testing because the color conversion code was broken.

Also, should the color-related helpers be there at the start of the file, or should I extract it to another? @bevm0, maybe you have an opinion?

ap0nia:
> @EricPedley @teresa-liang Can you take a look at the PR now? It's been changed to lightness variation now, which works pretty well. I think it didn't look good in my previous testing because the color conversion code was broken.
> 
> Also, should the color-related helpers be there at the start of the file, or should I extract it to another? @bevm0, maybe you have an opinion?

Yes, I think you should move the color helpers somewhere else. The schedule store's file should be dedicated to handling the store logic.

teresa-liang (approved):
Minh I think you did an awesome job!! It looks so good! I think the colors are perfect because now I feel like you can tell the difference between different sections but still be able to tell that they're associated with the same course.

PR #549, opened 2023-04-17
Improve schedule name dialog

### LINKED ISSUE #547: Improve schedule name dialog
The schedule name dialog is a little bit inconsistent compared to other dialogs that require text input, such as the dialogs that pop up for the Save, Load, and Import buttons. To fix this issue and make the schedule name dialog easier to use, here are some ideas:

- When the dialog opens, auto-focus on the text field.
- Allow the user to submit the schedule name with the Enter key.
- When the user wants to add a schedule and clicks on the schedule name, the default schedule name (like "Schedule 5") should not be erased.

Make sure these changes work for both the Rename and Add Schedule dialogs.

Douglas-Hong:
## Summary
To make the schedule name dialog consistent with other dialogs, such as the ones that pop up for saving/loading/importing schedules, the following changes were implemented:
- If the user presses Enter while the dialog is open, the schedule name is submitted.
- When the dialog opens, the cursor is auto-focused onto the text field. The user doesn't need to click on the field to start typing.
- Removed the following feature: if the user wants to add a schedule and clicks on the text field, the default schedule name is erased.

Some other small changes:
- The `handleAdd` and `handleRename` functions were combined into the `submitName` function.
- The `handleClose` function was renamed to `handleCancel`. `handleClose` is a bit misleading since it only runs when the cancel button is pressed.

## Test Plan
- Make sure the user can add a schedule by pressing Enter or clicking the submit button normally.
- Make sure the user can rename a schedule by pressing Enter or clicking the submit button normally.

## Issues

Closes #547


EricPedley:
Gonna start delegating PR review. I'll look at this one after John does.

johnlorenzini:
Didn't see it in the original issue description but was wondering if we should also handle ESC to cancel since we're handling ENTER to submit/rename.

Otherwise, code + implementation looks and works great, will let @EricPedley close with his final check.

Douglas-Hong:
Thanks John! I think the ESC key functionality is a good idea. Many other dialogs also don't use ESC, so we can tackle this feature in the future (#551).

Besides that, I fixed the merge conflict after the monorepo migration, so this is ready for review @EricPedley.

EricPedley (approved):
```

                                   (`-,-,
                                   ('(_,( )
                                    _   `_'
                                 __|_|__|_|_
                               _|___________|__
                              |o o o o o o o o/   lgtm
                             ~'`~'`~'`~'`~'`~'`~
```

PR #556, opened 2023-04-26
cleaned up some code

ap0nia:
Added some suggestions. Needs to be tested by Yukai because I'm not sure how to verify the behavior myself.


apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/GESelector.tsx
```diff
@@ -53,20 +55,19 @@ class GESelector extends PureComponent<GESelectorProps, GESelectorState> {
         ge: this.getGe(),
     };
 
-    handleChange = (event: ChangeEvent<{ name?: string | undefined; value: unknown }>) => {
+    handleChange: SelectProps['onChange'] = (event) => {
```
stevenguyukai: I don't think the handleChange is called when user changes the GE categories. So the GE selector is not working at all. I don't know about the SelectProps here. Think the original ChangeEvent did its job quiet well

stevenguyukai (approved):
I think everything is good to merge

PR #557, opened 2023-04-28
Fix wrong unit count

### LINKED ISSUE #536: Wrong Unit Count
Check schedule `WrongUnitCount`

image

Reproduce on different schedule by adding these classes, removing the lecture, then adding it again.

Douglas-Hong:
## Summary
- Fixed the bug described in #536. When the user removes lectures and adds them back, the total unit count is correctly updated.
- Instead of having `totalUnits` as a state, we can just derive it.
- Fixed a bug found by @Voark a couple months ago where courses that had units with decimal places (like French 1BC with 7.5 units) were displayed as integers only. We just need to do `parseFloat` instead of `parseInt`.

Bug found by @Voark:
image

Demo:

https://user-images.githubusercontent.com/78244965/235230522-224fea81-8db7-4194-8fd4-c737df381d1c.mp4

## Test Plan
- Create the schedule described in #536, remove some lectures, and add the lectures back. The total unit count should always be correct.
- Save and load the schedule, and then check if the total unit count is still correct.
- Search for French 1BC and add it to your schedule; the number of units should be 7.5 instead of 7.

## Issues

Closes #536

### DIFF
--- a/apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
+++ b/apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
@@ -48,7 +48,6 @@ interface AddedCoursePaneProps {
 interface AddedCoursePaneState {
     courses: CourseWithTerm[];
     customEvents: RepeatingCustomEvent[];
-    totalUnits: number;
     scheduleNames: string[];
     scheduleNote: string;
 }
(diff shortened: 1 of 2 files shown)


Voark:
Don't know why the action to assign reviewers ran twice but I guess it's not the end of the world.

teresa-liang:
I tested with a few other classes that have decimal unit counts by adding them to the schedule and then deleting them from the schedule and the total unit counts have been accurate. Very nice job Douglas :clap:

teresa-liang (commented):
I tested with a few other classes that have decimal unit counts by adding them to the schedule and then deleting them from the schedule and the total unit counts have been accurate. Very nice job Douglas 👏

teresa-liang:
Oh my goodness, I am so sorry. I am very new to code review and so I accidentally closed this branch and then I reopened it and then it assigned another code review to John. I am so sorry :sob:

johnlorenzini:
No worries! Since you signed off already I'll go ahead and approve, took a quick look and everything is great! Fantastic job Douglas

johnlorenzini (approved):

PR #560, opened 2023-05-06
Remove pull_request trigger and fix delete PR action

Voark:
### Reviewers Action
The assign reviewers action is triggering twice because there are two triggers. I had added the `pull_request` trigger initially because it was necessary to test the action, but now that the action is in the main branch it is no longer necessary. I didn't realize that both triggers would be triggered after, I thought only one was able to. 

### Delete PR
I forgot to add pass secrets through to the delete steps so they didn't have the credentials to execute a stack delete. Decided to add to this PR because it was so small anyways.

## Test Plan
This PR only has 1 reviewer!

### DIFF
--- a/.github/workflows/assign_random_reviewer.yaml
+++ b/.github/workflows/assign_random_reviewer.yaml
@@ -1,7 +1,5 @@
 name: "Assign Random Reviewer"
 on:
-  pull_request:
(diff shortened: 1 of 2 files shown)


MinhxNguyen7 (approved):
LGTM

PR #563, opened 2023-05-06
Rewrote the map

### LINKED ISSUE #485: Improve route hover popup
Currently, it is very difficult to hover over a route in such a way that the popup (displaying distance and duration) shows up. I think it's quite a helpful feature that should be easier to make use of. 

Making the margins for error larger for the hover would probably be good. Also, we should consider letting users click on the route to display the popup to make it usable on touchscreen devices,

ap0nia:
Rewrote the map


apps/antalmanac/package-lock.json
ap0nia, line 1: Idk why we have this

apps/antalmanac/src/App.css
ap0nia, line 1: CSS from my monorepo branch; most notable is the semicolon that'll appear on the split resizer.

apps/antalmanac/src/components/Home.tsx
ap0nia, line 1: Use CSS instead of this omega scuffed strat to put ellipsis on the split resizer.

apps/antalmanac/src/components/Map/Routes.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+  useEffect(() => {
+    routerRef.current.instance.addTo(map)
+    routerRef.current.instance.hide()
+    return () => {
+      routerRef.current.instance.remove()
+    }
+  }, [])
+
+  return null
```
ap0nia: This is supposedly the correct strategy to use react-leaflet, but it doesn't seem to do anything as far as I can tell. I thought it would help with the route ghosting that occurs, but it doesn't. I could remove it if it seems difficult to reason about the hook and context usage.

apps/antalmanac/src/stores/AppStore.ts
```diff
@@ -80,6 +80,10 @@ class AppStore extends EventEmitter {
         return this.schedule.toCalendarizedEvents();
     }
 
+    getCourseEventsInCalendar() {
+        return this.schedule.toCalendarizedCourseEvents()
+    }
+
```
ap0nia: I'm lazy so I'm just gonna monkey patch the store to do what I want

apps/antalmanac/yarn.lock
ap0nia, line 1: Idk why we have this either

apps/antalmanac/src/App.css
MinhxNguyen7: _Ahem_, I believe you are referring to a colon

ap0nia: Dang, exposed ! 😭

MinhxNguyen7: The colon's black now for some reason. I think you made it white for dark mode, and it's being automatically changed to white.

apps/antalmanac/src/components/Map/Map.tsx
```diff
@@ -0,0 +1,201 @@
...
+             * previous renders of the routes will be left behind if the keys aren't unique
+             */
+            const key = Math.random().toString(36).substring(7)
+            return <CourseRoutes key={key} latLngTuples={latLngTuples} color={color} />
+          })}
+
+        {/* draw a marker for each class */}
+        {uniqueMarkers.map((marker, index) => (
+          <Fragment key={Object.values(marker).join()}>
+            <CourseMarker {...marker} label={today ? index + 1 : undefined} stackIndex={index}>
```
MinhxNguyen7: This should be called LocationMarker or something because a course can have different sections

apps/antalmanac/src/components/Map/Map.tsx
```diff
@@ -0,0 +1,201 @@
...
+      setSelected(value)
+      const location = L.latLng(value.lat, value.lng)
+      map.current?.setView(location, 18)
+    }
+  }
+
+  /**
+   * unique buildings
+   */
+  const uniqueBuildings = Object.values(buildingCatalogue).filter(
```
MinhxNguyen7: This currently deletes the information for the other classes. There should be two columns, one for class and one for building, so that we can have multiple classes in one popup

apps/antalmanac/src/components/Map/UserLocator.tsx
```diff
@@ -0,0 +1,43 @@
+import L from 'leaflet'
+import 'leaflet.locatecontrol'
+import { useEffect } from 'react'
+import { createElementHook, createElementObject, useLeafletContext } from '@react-leaflet/core'
+import type { LeafletContextInterface } from '@react-leaflet/core'
+
+function createUserLocator(_props: any, context: LeafletContextInterface) {
```
MinhxNguyen7, line 7: I remember there being some _funny_ way that we were implementing the locator. Looks like you solved it. Very nice

apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -387,6 +387,13 @@ export class Schedules {
         ];
     }
 
+    /**
+     * Convert just courses into calendar compatible format.
+     */
+    toCalendarizedCourseEvents() {
```
MinhxNguyen7: Why isn't this `getCalendarizedCourseEvents`?

MinhxNguyen7 (changes requested):
The structure and implementation of everything look very nice. I'm glad that someone did something about the map. 

Also, good that you made the route easier to hover. I thought that was much simpler, but seems like it was not before the rewrite.

A couple of things about parity:

1. Clicking on building location links should link to the maps tab. I hope you don't think less of me after you see how that's implemented lol.
2. Since you're removing pins with the same location, we're losing information about the other classes that share the same location. I.e., someone can be directed to the map, see the popup,, and read that it's a different class than the one from which they clicked. I suggest the information in the popup be a two-column grid with the class on one side and the room on the other. 

Some UI things:

1. The route popup remains popped up after the mouse leaves the area. Also, I don't think there's a reason to have a dismiss button (if it goes away properly).
2. The kebab in the split resizer is now black in light mode. I'm guessing you didn't spot it because you use dark mode.
3. There's too much spacing between the info lines in the popup.
4. Once the location-linking is implemented, you might want to check on a mobile device to see whether the map moves to a good position to be able to see both the marker and the popup.

apps/antalmanac/src/components/Map/UserLocator.tsx
MinhxNguyen7: image
image

_nut_

apps/antalmanac/src/App.css
ap0nia: Ok fixed

apps/antalmanac/src/stores/Schedules.ts
ap0nia: Because I copy pasted the code from the methods below it that were named `to...Events` lol

apps/antalmanac/src/components/Map/Map.tsx
ap0nia: Oki

ap0nia:
Ok, I got through most of the stuff. Idk what strat you were using for the map link, so I designed my own lol.

An issue with accommodating multiple classes in the same location is that the current structure doesn't easily represent this information.
image

It renders a marker for each building. Assuming that there's only one course to worry about. A solution would be to use a record that maps building information to an array of classes associated with it. (btw the location marker accepts children, instead of using marker information as props, I think it makes the location marker more modular)

MinhxNguyen7:
> Idk what strat you were using for the map link, so I designed my own lol

[This](https://github.com/icssc/AntAlmanac/blob/ae06a74815def2ae8e78082e8c7a901f8fa97481/apps/antalmanac/src/components/RightPane/RightPaneStore.ts#L101) is how it is currently done. Actually focusing on the marker is not that difficult. Getting the popup up is done by simulating a click with a method on the marker's ref, which I think is stupid, but that's the only way that I've found after way too many hours of looking. The complicated, roundabout part is actually switching to the map, _then_ activating the focus and popup. You can build into it by just listening to the event that the map is currently listening to. If you can come up with a better way to do it, please do. It is not good.

[This](https://github.com/icssc/AntAlmanac/wiki/Feature-Checklist#map), I think, is a pretty complete list of features of the map now. I think it's pretty important that we display the classes that occur in each building, especially the room number (which is currently not done).

apps/antalmanac/src/components/Map/Routes.tsx
MinhxNguyen7: I would just include a comment that links to the source if someone's confused about why it's done this way

ap0nia: I mention the related topic on line 27; should I repeat or move this info?

MinhxNguyen7: Oh yeah. That's good then

apps/antalmanac/src/components/Map/Routes.tsx
```diff
@@ -0,0 +1,143 @@
...
+     * not turning this off will screw that up since it refuses to allow the routes to go offscreen.
+     */
+    fitSelectedRoutes: false,
+
+    plan: L.Routing.plan(waypoints, {
+      addWaypoints: false,
+      createMarker: () => false,
+    }),
+
+    routeLine(route) {
```
MinhxNguyen7: Just calling a function in here makes it confusing to read. Can you take this out, name it, and pass it in here?

apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
MinhxNguyen7, line 1: Why must you change the tab size? From a personal standpoint, 4 tabs is more readable. Additionally, it discourages you from excessive nesting. In any case, it departs from the project-wide styles, which should take precedence over "conventional" JS styles.

MinhxNguyen7 (commented):
Seems to be missing one functionality: Clicking on the location link from the section table. This should be pretty simple.

I'll remove the old mechanism of focusing on the map.

Also, the popup is a bit big on mobile because of the image. I think it should be smaller, especially so that the title is visible (and not obscured by the search bar)

Otherwise, looks good. I like it. I just fixed up some minor readability things, except for one. 

I don't understand why you have to change the tab space. It departs from the project conventions (stated or otherwise), it's less readable (IMO), and it encourages nesting.

apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
ap0nia: mb, I accidentally used my LSP formatter which doesn't respect the project's Prettier config. Let me re-format it

FWIW Prettier says 2 spaces, and I agree! For JS at least, and especially JSX. HTML with 4 space tab sounds horrendous 😔

apps/antalmanac/src/components/Map/Routes.tsx
ap0nia: Ok

MinhxNguyen7 (changes requested):
The code looks good. Everything is close to perfect. Thanks for rewriting this mess. There are just some behaviors that should/need to be changed:

1. The location linking doesn't work on my phone (Google Pixel 4 with Chrome).
2. If you open a marker popup with location link, then switch to a different day (without manually closing the popup), then click on the link again, the popup doesn't show up.
3. Hovering over the route displays the distance popup (nice), but moving your mouse away doesn't close the popup.
4. Opening the marker popup, then hovering over a route closes the popup. Logically, they should be independent of each other, and hovers should not close something that opened on-click.
5. The map should pan to the marker before the popup shows up. Maybe you'll have to use a `setTimeout()` here.
6. Clicking on a route on mobile doesn't display the route popup. This is not really a big deal, just something to consider fixing if it's not too much work.

Comment on UI (which I know you love):

The popup should be less tall. I think the building title is unnecessarily large. I agree that it is too small in the current version, but this is enormous. This is just an issue for desktop. It looks pretty good on mobile.
image

ap0nia:
> The code looks good. Everything is close to perfect. Thanks for rewriting this mess. There are just some behaviors that should/need to be changed:
> 
> 1. The location linking doesn't work on my phone (Google Pixel 4 with Chrome).
> 2. If you open a marker popup with location link, then switch to a different day (without manually closing the popup), then click on the link again, the popup doesn't show up.
> 3. Hovering over the route displays the distance popup (nice), but moving your mouse away doesn't close the popup.
> 4. Opening the marker popup, then hovering over a route closes the popup. Logically, they should be independent of each other, and hovers should not close something that opened on-click.
> 5. The map should pan to the marker before the popup shows up. Maybe you'll have to use a `setTimeout()` here.
> 6. Clicking on a route on mobile doesn't display the route popup. This is not really a big deal, just something to consider fixing if it's not too much work.
> 
> Comment on UI (which I know you love):
> 
> The popup should be less tall. I think the building title is unnecessarily large. I agree that it is too small in the current version, but this is enormous. This is just an issue for desktop. It looks pretty good on mobile. image

1. Fixed, I needed the `MobileHome` component to also listen to the URL parameters.
2. I don't want to fix that. I'd opt to make an issue for it, and then address it over summer with my overhaul.
3. I feel like I encountered inconsistent behavior when enabling that functionality, but ok I'll add it.
4. Won't fix; I think it's literally because the map can only have one popup open at a time.
5. Resolved by keying the focused `LocationMarker`; previously, the same location marker would stay rendered with the popup open from a previous building, the unique key forces a new marker to re-render, and so a new popup opens.
6. I feel like that might be because there are too many event listeners on the route lines. Oh well, this is definitely a feature! :shrug:

MinhxNguyen7 (changes requested):
1. If you go to [the map tab from a link](https://staging-563.antalmanac.com/map), either by clicking from history or by re-opening the page that was already there, the page breaks. This is the case on both mobile and desktop. This cannot be remedied by refreshing, which would be very frustrating for a user.
image
2. Location link still doesn't work on my phone. Have you tested it on yours?
3. When selecting locations (with linking) that are very far away from each other, at the very end of the pan, it zooms out and then back in. It seems minor, but that actually gives me motion sickness. The user I tested it on agrees. I think that might be a result of opening the popup, which moves the map, and flying to the location, which also moves the map. This is not 100% PR-blocking, but I would strongly prefer that we resolve it if it's feasible.
4. The popup is still hilariously tall. I know you don't like to deal with styling, so we can just delegate that to an issue if you want. It's not a big deal
image
5. I don't know why this is the case, but the marker popup and the route popup can coexist on the current version.
image

ap0nia:
I feel like some of my changes didn't make it to the staging deploy.
1.  Fixed by editing AWS CloudFront configuration 
image

2. It didn't work because of the first bullet point.
3. I turned off animations; it should only blink/flash, but not do any flying. It looks like it didn't make it to staging.I
4. This should've been fixed as well, but not on the staging. image
5. Fixed by using `addLayer` instead of `openPopup`
image

MinhxNguyen7:
Yeah changes not making it to the staging happened to me as well. Not sure why, but we should take a look at that

One more thing: clicking on the building name in the popup doesn't open the right link

ap0nia:
> Yeah changes not making it to the staging happened to me as well. Not sure why, but we should take a look at that
> 
> One more thing: clicking on the building name in the popup doesn't open the right link

Oh I see, the `building` was missing props for the location marker. I'll go fix that. I might actually need to implement the store for properly coordinating the map focus, so I'll work on that over the weekend.

apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
```diff
@@ -194,7 +196,11 @@ const CourseCalendarEvent = (props: CourseCalendarEventProps) => {
                         <tr>
                             <td className={classes.alignToTop}>Location</td>
                             <td className={`${classes.multiline} ${classes.rightCells}`}>
-                                <Link className={classes.clickableLocation} to={`/map?location=${buildingId}`}>
+                                <Link
+                                    className={classes.clickableLocation}
+                                    to={`/map?location=${buildingId}`}
+                                    onClick={focusMap}
```
MinhxNguyen7, line 202: Monkeying it a bit yeah?

MinhxNguyen7 (changes requested):
Mobile from section table still doesn't work for me.

Also, this is technically not feature parity because the popup had information about the class that clicked on it (which is especially useful for when you click on a class from search), but I know that's a lot more work, so I'm fine with losing it for now. I can't foresee an elegant way to do that though.

Otherwise everything else looks good. Very close

apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
ap0nia: Context API moment :sunglasses:
It's honestly not the worst strat, but it is most definitely a monkey patch!

MinhxNguyen7: Better than the way I did it.

MinhxNguyen7 (changes requested):
The linking from section table still doesn't work, but I think that's because the staging didn't update. I'll take a look when I have time, or you can do that.

Also, the two-space tabs sneaked back in again. It'd be good if you can undo that with a linter setting change. It's not a big deal, but it pollutes the diff and is inconsistent with the rest of the project.

Otherwise, everything else looks good.

ap0nia:
I moved the prettier config to the root of the project to enable consistent project-wide formatting. I also opted for `prettier.config.cjs` because it's more consistent with VSCode, plugins, etc. for whatever reason.
I think the preview website should be correctly updated now.

MinhxNguyen7 (approved):
__    __    __
                             |==|  |==|  |==|
                           __|__|__|__|__|__|_
                        __|___________________|___
                     __|__[]__[]__[]__[]__[]__[]__|___
                    |............................o.../
                    \.............................../
               hjw_,~')_,~')_,~')_,~')_,~')_,~')_,~')/,~')_

PR #559, opened 2023-05-06
Add section type to calendar event popup

MinhxNguyen7:
## Summary
Adds the section type to the title in the calendar event popup so that the user can check the section type when it's obscured in the calendar.

Before;
image

After:
image

## Test Plan
See if it looks good on a variety of devices

## Issues
Closes #516

Issue #414, opened 2022-07-21
Add tooltip to "Import" button


ChaseC99:
In the app bar, some of the items have tooltips explaining what the button does. For example, "Feedback" says "Give Us Feedback!"

<img width="158" alt="image" src="https://user-images.githubusercontent.com/29494270/180158275-4ff04846-b076-407b-8ace-28cf3594e504.png">

"Import" is missing one though! 
Add a tooltip to the "Import" button that says "Import a schedule from your Study List" 

Hint: wrap `<ImportStudyList />` in a [`<Tooltip>`](https://mui.com/material-ui/react-tooltip/)

KevinWu098:
I got this one :)

PR #567, opened 2023-05-12
Added a Tooltip to the Import button | Added Section Type to Calendar Popups

KevinWu098:
## Summary
Added a Tooltip to the Import Button ("Import a schedule from your Study List")

## Test Plan
- Verified that on hover, the tooltip with the appropriate message appears.

## Issues
Closes #414

PR #568, opened 2023-05-13
Update Map & Routes When Changing Schedules

### LINKED ISSUE #566: Map locations don't update when changing schedules
See title.

## Reproduction

Have a profile with multiple schedules. Open the map. Change to another schedule.

## Issues
Possibly addressed by #563

KevinWu098:
## Summary

1. Maps and routes will now update properly on both Schedule Index and Calendar Event changes. 

Demonstrates map responsiveness to schedule change

Map responsiveness on event deletion

~~2. Added tooltip to Import button~~

~~Import Tooltip~~

~~3. Didn't realize I left it in, but sectionType will now be shown in the pop-up. This appears to already be addressed by #559 , so mb 😅. Our implementation differs only in where it's shown (Heading vs with the other information), but I'll defer 🙇‍♂️~~

## Test Plan

1. Confirmed that when swapping between Schedules 1 and 2, the map's markers would change accordingly. The same does occur for planned routes.

~~2. Verified that on hover, the tooltip with the appropriate message appears.~~

## Issues
Map does not play nice with the finals schedule, but that appears to occur regardless of my PR

Closes #566   ~~& #414~~

### DIFF
--- a/apps/antalmanac/src/components/RightPane/Map/UCIMap.tsx
+++ b/apps/antalmanac/src/components/RightPane/Map/UCIMap.tsx
@@ -4,7 +4,7 @@
 import 'leaflet.locatecontrol';
 
 import Leaflet, { Control, LeafletMouseEvent } from 'leaflet';
-import React, { PureComponent } from 'react';
+import React, { PureComponent, useState } from 'react';
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore
 import { LeafletContext, Map, Marker, Polyline, TileLayer, withLeaflet } from 'react-leaflet';
@@ -97,7 +97,7 @@ export default class UCIMap extends PureComponent {
             const courses = new Set();
 
             // Filter out those in a different schedule or those not on a certain day (mon, tue, etc)
-            this.state.eventsInCalendar
+            AppStore.getEventsInCalendar()
                 .filter(
                     (event) =>
                         !(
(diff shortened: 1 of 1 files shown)


KevinWu098:
1. Gifs and Images of changes have been added :)
2. Moved #567 PR description into the summary above as well

MinhxNguyen7 (approved):
Hi @KevinWu098, thanks for the contributions. The changes look good, and I've tested it locally. 

However, I'm not sure if @EricPedley is good with merging this because it violates "atomic commits". We squash all commits in a pull request into a single commit before merging it into main, and your PR contains two distinct changes (adding tooltip to import button and updating the map when the schedule changes). 

This fix to the map will be superseded by #563, but it's small enough and is somewhat important, so I think it should be taken while that's being developed.

KevinWu098:
Sorry bout that @MinhxNguyen7 😅. I'm still getting the hang of Git and branches, so I ended up working on top of the import button PR

I'll get the hang of it for the next PR :)

EricPedley (changes requested):
Merging this isn't too time-critical so I'm gonna ask you to interactive rebase and force-push this branch to exclude the tooltip changes. See this tutorial for help: https://www.youtube.com/watch?v=ElRzTuYln0M&t=254s

KevinWu098:
@EricPedley I think I fixed it :), this PR should only have the Map bug fixes now!

ap0nia (approved):
Looks good, very cool

PR #569, opened 2023-05-14
Handle websoc death

### LINKED ISSUE #532: Users can't see schedule when WebSOC dies
@Voark observed that when WebSOC dies, people can't see their schedules, even though we have it. The UI waits for class details to display. 

We should probably show some raw version of what classes are in each schedule if WebSOC/PeterPortal doesn't respond in time, so that people can still copy the codes and register for classes.

MinhxNguyen7:
## Summary

Backup display for when PP or WebSoc goes down (but our backend is fine).

image

## Test Plan

Block peterportal with devtools. See if it works

## Issues

Closes #532


MinhxNguyen7:
@Voark I felt a great disturbance in the CI/CD pipeline.

There's no deploy.

MinhxNguyen7:
@ap0nia review pls

ap0nia:
> @ap0nia review pls

ok fine, you do my map ! :pray:

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -50,6 +50,7 @@ interface AddedCoursePaneState {
     customEvents: RepeatingCustomEvent[];
     scheduleNames: string[];
     scheduleNote: string;
+    barebonesMode: boolean;
```
ap0nia: I like the semantics of `skeleton`, i.e. `displayScheduleSkeleton`

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -266,11 +271,56 @@ class AddedCoursePane extends PureComponent<AddedCoursePaneProps, AddedCoursePan
...
         return (
-            <Grid container spacing={2}>
-                {this.getGrid()}
-            </Grid>
+            this.state.barebonesMode ?
+                this.getBarebonesSchedule()
+                :
+                <Grid container spacing={2}>
+                    {this.getGrid()}
+                </Grid>
```
ap0nia: Don't call functions in JSX, this is generally an anti-pattern because they won't be registered by React as components, and so they'll have screwed up state.
**However**, there is a caveat to this rule with class-based components for some reason. But still...don't do it :shrug:

Is this syntax possible at all? Just encapsulate both components entirely as functions, and then render one or the other as a component.

```suggestion
            this.state.barebonesMode ? <this.getBarebonesSchedule /> : <this.getGrid />
```

apps/antalmanac/src/stores/AppStore.ts
```diff
@@ -49,25 +52,29 @@ class AppStore extends EventEmitter {
     }
 
     getCurrentScheduleIndex() {
-        return this.schedule.getCurrentScheduleIndex();
+        return this.schedules.getCurrentScheduleIndex();
```
ap0nia: Just a complaint with the app store and not something you have to fix.
Why is there an abstraction over the schedules? It makes literally no sense. There's one app store in the app, and one schedule store in the app. If you want to subscribe to a particular data source, just pick the one you need.

But coercing the schedules into the schedule store just makes a really bloated class for no reason, with the guise that it's the single source of truth (it isn't and it doesn't make anything simpler).

apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -9,10 +9,20 @@ import {
...
-import { RepeatingCustomEvent } from '$components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog';
-import { combineSOCObjects, CourseInfo, getCourseInfo, queryWebsoc } from '$lib/helpers';
-
-const defaultColors = [blue[500], pink[500], purple[500], green[500], amber[500], deepPurple[500], deepOrange[500]];
+import {RepeatingCustomEvent} from '$components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog';
+import {combineSOCObjects, CourseInfo, getCourseInfo, queryWebsoc} from '$lib/helpers';
+import { openSnackbar } from '$actions/AppStoreActions';
+import AppStore from './AppStore';
+
+const defaultColors = [
```
ap0nia: Here's a funny strat I came up with to load all the possible colors we have lol; it has black and white too tho, so you'd have to deal with them too
https://github.com/icssc/AntAlmanac/blob/aponia-monorepo/apps/client/src/stores/schedule/course.ts#L17

apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -604,31 +616,39 @@ export class Schedules {
...
+                                            term: term,
+                                            sectionCodes: sectionCode.join(','),
+                                        })
+                                    )
+                            )
                         )
-                    )
-                );
-                courseInfoDict.set(term, courseInfo);
+                    );
```
ap0nia: Can we try something like this? I haven' tested it or anything, but I can follow this logic more easily.
```ts
const sectionsGroupedByTen = sectionCodes.reduce((accumulated, item, index) => {
  if (index % 10) {
   accumulated.push(item)
  } else {
    accumulated.push([])
    accumulated[index % 10].push(item)
  }
  return accumulated
}, [] as string[][])

const socObjects = await Promise.all(sectionsGroupedByTen.map(s => { ... }))

const courseInfo = getCourseInfo(combineSOCObjects(socObjects))
```

ap0nia (changes requested):
Mostly looks good, I'm fine with changes to the store stuff because all of it is unfixable anyways.
Just left a suggestion for the component rendering as well as the reducing loop

apps/antalmanac/src/stores/AppStore.ts
MinhxNguyen7: I'd imagine that's because schedule used to be a simple data structure that existed in AppStore, then it was extracted when it got too complicated.

Also, if you feel like talking OOP, it's about AppStore owning the schedule class, which is why OOP is not good

apps/antalmanac/src/stores/Schedules.ts
MinhxNguyen7: These aren't all MUI colors though. Some of them are too close, and the current ordering specifically avoids close colors for as long as possible.

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
MinhxNguyen7: I like it. I couldn't come up with anything

apps/antalmanac/src/stores/AppStore.ts
ap0nia: Based oop hater 😎

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
ap0nia: Oh, actually there's two anti patterns lol.
1. Creating a component inside of another component is the main thing that screws up state (unless it's a class component for some reason). Correct practice is to make a standalone component...even if it requires prop-drilling.

2. Using a function call instead of a React JSX tag thingy. This just prevents it from rendering on the component tree properly or something.

Conclusion : I love how React is so **easy to use** ™️

apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -604,29 +608,38 @@ export class Schedules {
...
+                    const sectionsGroupedByTen = sectionCodes.reduce((accumulated, item, index) => {
+                        if (index % 10) {
+                            accumulated.push([item]);
+                        } else {
+                            accumulated.push([]);
+                            // The ? is to placate ts-lint. If accumulated is empty, the condition above would apply
+                            accumulated.at(-1)?.push(item);
+                        }
+                        return accumulated;
+                    }, [] as string[][]);
```
ap0nia: Ok, I did some testing and I think this should work.
```suggestion
                    const sectionsGroupedByTen = sectionCodes.reduce((accumulated, item, index) => {
                        if (index % 10) {
                            accumulated.at(-1)?.push(item);
                        } else {
                            accumulated.push([item]);
                        }
                        return accumulated;
                    }, [] as string[][]);
```

apps/antalmanac/src/components/PatchNotes.tsx
ap0nia, line 1: This component was written really poorly, so I refactored it real quick.

apps/antalmanac/src/components/Calendar/Toolbar/ExportCalendar.tsx
ap0nia, line 1: This file was written poorly, so I refactored it.

apps/antalmanac/tsconfig.json
```diff
@@ -22,7 +22,7 @@
             "$lib/*": ["src/lib/*"],
             "$providers/*": ["src/providers/*"],
             "$routes/*": ["src/routes/*"],
-            "$stores/*": ["src/stores/*"],
```
ap0nia: Extra commas is invalid in JSON; TypeScript doesn't mind when parsing the file, but following the spec is good.

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
ap0nia, line 1: Ok, I just realized the UI of this component was implemented completely wrong and I fixed it.

ap0nia (approved):
Ok, I took a crack at correcting the discrepancies between this branch and main. Let me know if you think it's good and we can move forward with merging.

MinhxNguyen7:
Patch notes now keep showing up on reload. I can fix that really quickly

Otherwise, I don't see anything amiss, other than that this is a _cosmic commit_

ap0nia (approved):
I reviewed all of the stuff while I was refactoring it, and I'm fine with how it currently is lol. If you'd like anything adjusted, let me know or you can apply changes yourself. :nerd_face:

MinhxNguyen7:
Well I just noticed that the central feature of this PR actually doesn't work lol. It still does the same thing as `main`. I think the TRPC PR changed how schedules are loaded. I'll try to fix that now

apps/antalmanac/package.json
```diff
@@ -69,6 +69,7 @@
         "zustand": "^4.3.2"
     },
     "devDependencies": {
+        "@packages/antalmanac-types": "workspace:^",
```
ap0nia, line 72: I've decided that I'm ok with the pnpm default of installing workspace packages with the semantic version `workspace:^`

apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
```diff
@@ -151,11 +151,11 @@ const CourseCalendarEvent = (props: CourseCalendarEventProps) => {
...
     }, []);
 
-    const { setSelectedTab } = useContext(mobileContext);
+    const { setActiveTab } = useTabStore();
 
     const focusMap = useCallback(() => {
-        setSelectedTab(1);
-    }, [setSelectedTab]);
+        setActiveTab(2);
+    }, [setActiveTab]);
```
ap0nia, line 158: Technically, only stateful variables are needed in the dependency array, i.e. things that could change or be mutated. State setters should never change, but this is React and keeping track of what things change is stupid.

In previous commits, I did add the state setter to the dependency array to be safe. I'm gonna chalk this up to LGTM :nerd:

apps/antalmanac/src/components/RightPane/RightPaneRoot.tsx
```diff
@@ -101,12 +98,12 @@ export default function Desktop({ style }: DesktopTabsProps) {
       </Paper>
       <div
         style={{
-          padding: RightPaneStore.getActiveTab() === 2 ? '0px' : '8px 8px 0 8px',
+          padding: activeTab === 2 ? '0px' : '8px 8px 0 8px',
```
ap0nia: In general, we should try to avoid introducing specific edge cases like this. i.e. this particular div should render everything with the exact same padding, and then have the particular component position itself -- i.e. since it knows how it should be rendered.

I did address this in some other PR, so this is fine to leave as-is.

apps/antalmanac/src/stores/AppStore.ts
```diff
@@ -6,11 +6,17 @@ import { Schedules } from './Schedules';
...
+function getCurrentTheme() {
+    const theme = typeof Storage === 'undefined' ? 'auto' : window.localStorage.getItem('theme');
+    return theme === null ? 'auto' : theme;
+}
 
 class AppStore extends EventEmitter {
     schedule: Schedules;
     customEvents: RepeatingCustomEvent[];
-    colorPickers: { [key: string]: EventEmitter };
+    colorPickers: Record<string, EventEmitter>;
```
ap0nia, line 19: In general, I like to use `Record` if possible because it's more succinct. A niche edge case where this isn't very nice is when you want to map an interface's properties to alternate values. A `Record` isn't able to capture the original definition of the properties.

[Here's an example on TS playground](https://www.typescriptlang.org/play?ssl=3&ssc=12&pln=2&pc=12#code/C4TwDgpgBAsiCSA7YEBOAzAhgY2gXigG8AoKKTALigGdhUBLRAc1KgCMraHnXtO7GLAL7FioSLBAAlCNgD2qACZQCM+UoA8Aawgg56SUhQYcEADQ0BzAHyiANhGBQAHlThqFyggFdEiiOiMEMqY1OSIIKIA9FFQAKrUglAARExyUMDp-oGI9MD0cojJUIUZABbQyZjFYKhykKigUIpyENSIAOROTI5Qet7hIADuFagQxDHsspje1NDAFVAABh5KSxng0ABWs07oCj3AYSNo84sK9EyMmHZQtfVo+W1Q2JgAttDodW8Tsb7+qHK0DKcjkigseQ6YTY01m0DyUB2tBehS43mwR3IUF8BUQJQMN1uC2gOhA1AhiFoEEwyn05Ax3kJICgb0wYDASS+cjeQJKDCuiBuUFJxGcADpMKJxNA4DA2ZBFEY0FhcCoiFAANoAaSgjGFujpcCVJlwAF1+NwmFARMQHE4QG4QHL2cFjSr8Ni-AEgiEwpgItFYgkkql0plmt7cvlCsVSsSUtU7nUGk1gJgdGF+uwcFoMukjchlaYNR1MB1Tb8pq84VjWS7lGN1A2IB83jDUGF4xcBUL7inmQihqEIwwAG7BKBcn4gCVAA)

In GitHub markdown:
```ts
type MyInterface = {
  a: string
  b: string
  c: string
}

type MyRecord = Record<keyof MyInterface, string>

let x: MyRecord = undefined as any

// Using "go to definition" on the "a" property doesn't get you anywhere
// because the `Record` type just forgets where the original properties came from
// under the hood, it's because it just constructs a union of all the keys, instead of actually mapping from the original key
x.a

type MyMappedInterface = { [K in keyof MyInterface]: string }

let y: MyMappedInterface = undefined as any

// Using "go to definition" on the "a" property takes you back to MyInterface['a']
// because a mapped record remembers the original property it was derived from
y.a
```

ap0nia (approved):
I think these changes look solid, good work! 🙌 
Let me know if you're ok with finally merging everything! @MinhxNguyen7

apps/antalmanac/src/components/RightPane/RightPaneRoot.tsx
MinhxNguyen7: Yeah I thought it was janky but out of scope to change in this PR

apps/antalmanac/src/stores/AppStore.ts
MinhxNguyen7: I don't disagree, but I'd argue that it's out of scope. I'll leave it since you've already done it.

MinhxNguyen7:
I think the PR's good now, but I can't review it because I'm the author. @ap0nia, you shouldn't review it because you wrote a lot of the code here. We just need someone to look through and check that things are sound and that we haven't broken anything.

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -1,277 +1,322 @@
...
+    return (
+        <Box>
+            <Box display="flex" width={1} justifyContent="space-between" marginY={2}>
+                <Typography variant="h6">{`${scheduleName} (${scheduleUnits} Units)`}</Typography>
+
+                <Box>
+                    <PopupState variant="popover">
+                        {(popupState) => (
+                            <>
+                                <Button variant="outlined" {...bindTrigger(popupState)}>
```
KevinWu098, line 225: Color should probably be `inherit` like on main

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -1,277 +1,322 @@
...
+                                    </MenuItem>
+                                </Menu>
+                            </>
+                        )}
+                    </PopupState>
+
+                    <Button
+                        sx={{ marginLeft: '4px', marginRight: '4px' }}
+                        variant="outlined"
+                        color="secondary"
```
KevinWu098:
```suggestion
                        color="inherit"
```
Currently invisible on light mode

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
KevinWu098:
```suggestion
                                <Button variant="outlined" color="inherit" {...bindTrigger(popupState)}>
```
Currently it's blue colored, but it should match what's on live

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -1,277 +1,322 @@
...
+        return result;
+    }, [courses]);
+
+    const scheduleName = useMemo(() => {
+        return scheduleNames[scheduleIndex];
+    }, [scheduleNames, scheduleIndex]);
+
+    return (
+        <Box>
+            <Box display="flex" width={1} justifyContent="space-between" marginY={2}>
```
KevinWu098: Margins are also a little weird around this title section:

Screenshot 2023-09-06 at 3 01 11 AM

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -1,277 +1,322 @@
...
-                                scheduleNames={this.state.scheduleNames}
+                                scheduleNames={scheduleNames}
                             />
                         </Grid>
                     );
                 })}
-                {this.state.customEvents.length > 0 && <Typography variant="h6">Custom Events</Typography>}
-                {this.state.customEvents.map((customEvent) => {
+
+                {customEvents.length > 0 && <Typography variant="h6">Custom Events</Typography>}
```
KevinWu098, line 272: This is not updating when custom events are created:

chrome-capture-2023-8-6

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -1,277 +1,322 @@
...
+
+                {customEvents.map((customEvent) => {
```
KevinWu098: This is not updating when custom events are created:

chrome-capture-2023-8-6

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
KevinWu098: Styling is off here:

Screenshot 2023-09-06 at 3 04 36 AM

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -1,277 +1,328 @@
...
+        }, {} as Record<string, string[]>);
+
+        return Object.entries(result);
+    }, [skeletonSchedule.courses]);
+
+    return (
+        <>
+            {
+                // Sections organized under terms, in case the schedule contains multiple terms
+                sectionsByTerm.map(([term, sections], index) => (
```
KevinWu098: Sorry about all the styling nitpicks, but can we add a little padding here:

Screenshot 2023-09-06 at 1 39 25 PM

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -1,277 +1,328 @@
...
+            accumulated[course.term] ??= [];
+            accumulated[course.term].push(course.sectionCode);
+            return accumulated;
+        }, {} as Record<string, string[]>);
+
+        return Object.entries(result);
+    }, [skeletonSchedule.courses]);
+
+    return (
+        <>
```
KevinWu098:
```suggestion
        <>
                        <Typography variant="h6">{`${skeletonSchedule.scheduleName}`}</Typography>

```

Looking at `ShortCourseScheduleSchema`, we should be able to improve the user experience by also supporting displaying the name of the schedule. 

_You may need to restyle the term typography a little bit, depending on how you decide to implement the typography of the schedule name._

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
ap0nia: Oh, I meant to wrap a padded box around the entire thing, but I forgot.

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -1,277 +1,328 @@
...
+            <Grid container spacing={2} padding={0}>
+                {courses.map((course) => {
                     return (
                         <Grid item md={12} xs={12} key={course.deptCode + course.courseNumber}>
                             <SectionTableLazyWrapper
-                                classes={this.props.classes}
                                 courseDetails={course}
                                 term={course.term}
-                                allowHighlight={false}
+                                highlightAdded={false}
```
KevinWu098:
```suggestion
                                allowHighlight={false}
```

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
KevinWu098: In a similar vein, I think it... should be possible to display custom events and schedule notes as well?

The base functionality is great and I think adding on this additional support would really make it awesome

ap0nia: Can we add that later lol; I don't wanna debug more features

KevinWu098: Sure, let's hold off on custom events and notes. But the one thing I will really really push for right now is having the schedule name shown.

Otherwise, there's no real way for users to figure out what's what based off just the codes.

ap0nia: Did I not do that? https://github.com/icssc/AntAlmanac/blob/de7f86b5ca63b22e5bdc1f952144448ccfe2a826/apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx#L110

KevinWu098: 💀 sorry, I'm on mobile and didn't see it 😅

I'll check it out once I'm home; Thanks for adding in the schedule name functionality :))

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneRoot.tsx
```diff
@@ -71,7 +71,7 @@ class RightPane extends PureComponent {
 
     render() {
         return (
-            <>
+            <div style={{ padding: 8 }}>
```
KevinWu098: Not sure where the source of the issue is, but the GIF has been shifted up in the Search pane:

Screenshot 2023-09-06 at 8 07 06 PM

KevinWu098:
```typescript
export const ShortCourseScheduleSchema = type([
    {
        scheduleName: 'string',
        courses: arrayOf(ShortCourseSchema),
        customEvents: arrayOf(RepeatingCustomEventSchema),
        'scheduleNote?': 'string',
    },
    '|>',
    (s) => ({ scheduleNote: '', ...s }),
]);
export type ShortCourseSchedule = typeof ShortCourseScheduleSchema.infer;
```

~~I'm noticing that Typescript can't read the type of `ShortCourseSchedule`. My guess is because of the syntax used for scheduleNote? Is there a way for us to nudge Typescript into inferring properly?~~

Screenshot 2023-09-06 at 8 16 52 PM

~~This currently affects `Schedules.ts` and `AddedCoursePane.tsx`~~

Nvm I'm trolling, fixing this is out of scope I think, but we should probably fix the types in a soon <sup>tm</sup> PR since this has been antagonizing `Schedules.ts` for some time 😅

KevinWu098:
Honestly, I really love this support for when Websoc dies, and I think future follow-up could make it an even better user experience.

For now, the comments above are my only remaining issues (I hope), but just to list out some future targets:
1. Support custom events and schedule notes
2. Disable the calendar toolbar (since you can't do anything with it anyways)
3. Support toggling between multiple schedules in skeleton mode

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneRoot.tsx
ap0nia: Is it supposed to be in the middle?

ap0nia:
The type issues should a problem on your end; not a universal one. You could try asking for help, opening an issue to start a discussion, etc. but strictly speaking this isn't "solvable", since everything works fine on my machine :tm:

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneRoot.tsx
KevinWu098: Yeah, center of the pane. You can see the intended behavior on the Map pane

KevinWu098 (approved):
I would nitpick the styling here, but I think it's not worth holding this PR up over; It'll get obliterated anyways after this gets merged in and I add the intended changes from #665 

Screenshot 2023-09-07 at 4 06 45 AM

LGTM, awesome feature 💖

I'll leave the merging to y'alls, just in case anybody wants to double-check my review 🫶

KevinWu098:
> The type issues should a problem on your end; not a universal one. You could try asking for help, opening an issue to start a discussion, etc. but strictly speaking this isn't "solvable", since everything works fine on my machine ™️

💀 I'll figure it out one of these days

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -1,277 +1,329 @@
-import { Button, Grid, Menu, MenuItem, Paper, TextField, Typography } from '@material-ui/core';
-import { withStyles } from '@material-ui/core/styles';
-import { ClassNameMap } from '@material-ui/core/styles/withStyles';
+import { useCallback, useEffect, useMemo, useState, Fragment } from 'react';
```
KevinWu098:
```suggestion
import { useCallback, useEffect, useMemo, useState } from 'react';
```

Unused import

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -1,277 +1,329 @@
...
 import PopupState, { bindMenu, bindTrigger } from 'material-ui-popup-state';
-import { PureComponent } from 'react';
 
+import { Box, Button, Chip, Grid, Menu, MenuItem, Paper, TextField, Tooltip, Typography } from '@mui/material';
 import { AACourse } from '@packages/antalmanac-types';
-import { RepeatingCustomEvent } from '../../Calendar/Toolbar/CustomEventDialog/CustomEventDialog';
+
+import { float } from 'html2canvas/dist/types/css/property-descriptors/float';
```
KevinWu098:
```suggestion
```
Unused import

KevinWu098 (approved):
LGTM 🚀

PR #572, opened 2023-05-16
trpc

Voark:
## Summary

Rewrote backend in TS so we could use TRPC. Because now we are sharing types I moved a lot of the shared types to their own package.

Also migrated the PP API endpoint to v2 as I consume their new types anyways. The main change is the `bldgs` field going from `String` to `String[]` so I did some scuffed 0 indexing, don't know what the best option for this is and I think it can be looked into more in the future.

## Issues

Closes #

## Future Followup

With caching, PP will be able to return the co-classes of every GE instead of us having to make a separate request.


apps/antalmanac/src/stores/calendarizeHelpers.ts
```diff
@@ -42,7 +42,7 @@ export const calendarizeCourseEvents = (currentCourses: ScheduleCourse[] = []) =
                             term: course.term,
                             title: course.deptCode + ' ' + course.courseNumber,
                             courseTitle: course.courseTitle,
-                            bldg: meeting.bldg,
+                            bldg: meeting.bldg[0],
```
EricPedley, line 45: Opened an issue for the fact that meeting.bldg is an array now: #592. Don't want to scope-creep this PR.

apps/backend/package.json
```diff
@@ -2,26 +2,47 @@
...
+        "concurrently": "^8.0.1",
+        "esbuild": "^0.17.19",
+        "eslint": "^8.34.0",
+        "eslint-config-prettier": "^8.6.0",
+        "eslint-plugin-import": "^2.27.5",
+        "husky": "^8.0.3",
+        "lint-staged": "^13.1.2",
+        "nodemon": "^2.0.22",
+        "prettier": "^2.8.4",
+        "tsx": "^3.12.3",
```
EricPedley: why does the backend depend on tsx? Not important enough to block this PR but I'm curious.

EricPedley (approved):
lgtm

apps/antalmanac/src/components/RightPane/SectionTable/CourseInfoBar.tsx
```diff
@@ -78,20 +78,22 @@ const CourseInfoBar = (props: CourseInfoBarProps) => {
...
-                            title: jsonResp.title,
-                            prerequisite_tree: jsonResp.prerequisite_tree,
-                            prerequisite_list: jsonResp.prerequisite_list,
-                            prerequisite_text: jsonResp.prerequisite_text,
-                            prerequisite_for: jsonResp.prerequisite_for,
-                            description: jsonResp.description,
-                            ge_list: jsonResp.ge_list.join(', '),
+                            id: data.id,
+                            title: data.title,
+                            prerequisite_tree: data.prerequisiteTree,
```
EricPedley, line 91: Sike, doesn't look good to me. The migration breaks this part because `data.prerequisiteTree` used to be a string-serialized JSON object but now it's a real object. @teresa-liang or @Voark can you fix this?

EricPedley (changes requested):

apps/backend/package.json
Voark: I probably copy pasted a little too hard

apps/antalmanac/src/components/RightPane/SectionTable/CourseInfoBar.tsx
teresa-liang: This issue has been resolved in commit f7c41696766521aac31fb36076d7165c75513800!

EricPedley (approved):
lgtm

For real this time

PR #573, opened 2023-05-16
Fix course adding bug

Voark:
## Summary
When a course is already added to a schedule, adding it to another didn't do anything. The issue was literally I forgot to add the course in the code.

## Test Plan
- Adding courses in 2 schedules
- Confirm changing color in one schedule changes in both
- Adding a course for the 3rd time keeps the new colors.

## Issues

Closes #


ecxyzzy (approved):
`bGd0bQ==`

(lgtm in base64)

PR #574, opened 2023-05-16
Adjusted color algorithm for the same sectionType

KevinWu098:
## Summary
- Duplicate classes of the same sectionCode & sectionType (ie ICS 32A LAB at 4PM or 6PM) will generate the same color.
- Additionally, an existing class of the same sectionType will generate the same color, preventing overuse of colors. 
- Currently, I'm only aware of courses with 3 total sectionTypes (ICS 51 w/ LEC LAB DIS), but the color algorithm should comfortably fit four close colors, so it should work for this case as well

Tentative Fix

## Test Plan
- Confirmed that when selecting new courses, the same sectionType (ie ICS 32A LAB) would maintain the same color. 
- When there are 3 types (DIS, LEC, LAB), functionality is as intended
Three sectionTypes

## Issues
Follows up from #548 & #509

- Noticed previously that the "nuclear" use case of selecting all available courses of a particular course title (ie every single ICS 32A discussion course) resulted in overuse of the hue, resulting in: Pink -> Light Pink -> ... -> White -> Black

Issue Screenshot

## Future Followup
Regarding UI design, I'm unsure how the behavior should be implemented for the case demonstrated by Stats 67 in which courses are the same but have different sections.

Stats 67 Case

1. Should identical courses but different sections (NOT sectionType, eg Section A1, A2 versus B1, B2) have the same color?
2. If the answer to 1 is yes, how should colors be implemented?
3. Is there any clean, visually clear way to distinguish classes in different sections? Event shape, borders, etc?

### DIFF
--- a/apps/antalmanac/src/stores/Schedules.ts
+++ b/apps/antalmanac/src/stores/Schedules.ts
@@ -179,17 +179,14 @@ function getColorForNewSection(newSection: ScheduleCourse, sectionsInSchedule: S
     const existingSectionsType = existingSections.filter((course) => course.section.sectionType === newSection.section.sectionType)
 
     const usedColors = new Set(sectionsInSchedule.map((course) => course.section.color));
-
-    if (existingSections.length > 0) {
-        let originalColor = existingSections[0].section.color;
-
-        //if another course of the same sectionType exists, return that color
-        if (existingSectionsType.length > 0){
-            originalColor = existingSectionsType[0].section.color;
-            return originalColor;
-        }
-        return (generateCloseColor(originalColor, usedColors));
-    }
+    
+    //If the same sectionType exists, return that color
+    if(existingSectionsType.length > 0) 
+        return (existingSectionsType[0].section.color);
+        
+    //If the same courseTitle exists, but not the same sectionType, return a hue
+    if(existingSections.length > 0)
+        return (generateCloseColor(existingSections[0].section.color, usedColors));
 
     // If there are no existing sections with the same course title, generate a new color
     return defaultColors.find((materialColor) => !usedColors.has(materialColor)) || '#5ec8e0';
--- a/apps/antalmanac/src/stores/Schedules.ts
+++ b/apps/antalmanac/src/stores/Schedules.ts
@@ -184,7 +184,7 @@ function getColorForNewSection(newSection: ScheduleCourse, sectionsInSchedule: S
     if(existingSectionsType.length > 0) 
         return (existingSectionsType[0].section.color);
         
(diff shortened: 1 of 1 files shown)


apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -175,14 +175,18 @@ function getColorForNewSection(newSection: ScheduleCourse, sectionsInSchedule: S
...
-        // Generate a slightly different color that does not conflict with any other section in the schedule
-        return generateCloseColor(originalColor, usedColors);
-    }
+    const usedColors = new Set(sectionsInSchedule.map((course) => course.section.color));
+    
+    //If the same sectionType exists, return that color
+    if(existingSectionsType.length > 0) 
+        return (existingSectionsType[0].section.color);
+        
+    //If the same courseTitle exists, but not the same sectionType, return a hue
```
MinhxNguyen7:
```suggestion
    //If the same courseTitle exists, but not the same sectionType, return a close color
```

It's the same hue and saturation with a different lightness (currently). I prefer not to say that the function returns a hue of the original color, which would imply that it's a different hue.

apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -175,14 +175,18 @@ function getColorForNewSection(newSection: ScheduleCourse, sectionsInSchedule: S
                 Math.abs(parseInt(b.section.sectionCode) - parseInt(newSection.section.sectionCode))
         );
 
-    const usedColors = new Set(sectionsInSchedule.map((course) => course.section.color));
-
-    if (existingSections.length > 0) {
-        const originalColor = existingSections[0].section.color;
+    //New array of courses that share the same sectionType & courseTitle
+    const existingSectionsType = existingSections.filter((course) => course.section.sectionType === newSection.section.sectionType)
```
MinhxNguyen7, line 179: I would combine this with the other filtering that `existingSections` does. There's not really a reason to do two passes

MinhxNguyen7 (dismissed):
Makes sense. Just a few changes

apps/antalmanac/src/stores/Schedules.ts
KevinWu098: I created a second filtering because otherwise I'm not sure how I would be able to distinguish between courseTitle versus sectionType.

Cases:
1. courseTitle && sectionType => Same Color
2. courseTitle => Close Color
3. nothing => New Color

If I don't have two arrays, one which checks for courseTitle and one for sectionType, I'm not sure I would be able to fulfill case 1 AND 2.

MinhxNguyen7 (approved):
Tested and LGTM!

PR #576, opened 2023-05-17
Fix for Unremoved Markers - Unique Keys for Children

### LINKED ISSUE #575: Duplicate Markers due to Duplicate Courses(?)
When there are multiple duplicate courses, markers remain on the map even when the schedule is changed. Will need to investigate to figure out why this is happening/why markers are not unrendering.

Edit: Likely introduced (?) or revealed by #568 as past markers of duplicate children seem not to derender, which was previously unseen due to the bug resolved in #568 

Duplicate Marker Nonsense

KevinWu098:
## Summary
Following #568 , the map would now not remove markers that had duplicate keys. For instance, if two STATS 67 LEC were in the calendar, the second of the two event markers would not be removed due to their duplicate keys. 

To resolve this, sectionCode & term was added to the key generator:

Before:

Duplicate Marker Nonsense

After:
Fixed Marker Nonsense

## Test Plan
- As far as I can tell, `key` is only used here as an identifier, so no issues should arise elsewhere.
- Checked functionality for multiple courses & combinations

## Issues
~~Frankly, I've got no idea how courses end up working term to term, in the context of AntAlmanac and this key, so... _theoretically could there be some edge case where **the same class with the same section code** BUT **in different terms** appear on the same calendar? Would that create this bug again?_~~
Solved in commit 996ffbff9246eb7f60daeddbf6cf14bf104f1f09 by incorporating terms into the key!

<sub> 99 little bugs in the code, 99 little bugs in the code. Take one down, patch it around 117 little bugs in the code. <sub>

Closes #575

### DIFF
--- a/apps/antalmanac/src/components/RightPane/Map/UCIMap.tsx
+++ b/apps/antalmanac/src/components/RightPane/Map/UCIMap.tsx
@@ -430,8 +430,8 @@ export default class UCIMap extends PureComponent {
             const courses = pins[buildingCode];
             for (let index = courses.length - 1; index >= 0; index--) {
                 const [event, eventIndex] = courses[index];
-                const courseString = `${event.title} ${event.sectionType} @ ${event.bldg} : ${event.sectionCode} @ ${event.term}`;
                 if (locationData === undefined) return;
+                const key = `${event.title} ${event.sectionType} @ ${event.bldg} : ${event.sectionCode} @ ${event.term}`;
 
                 // Acronym, if it exists, is in between parentheses
                 const acronym = locationData.name.substring(
@@ -441,7 +441,7 @@ export default class UCIMap extends PureComponent {
 
                 markers.push(
                     <MapMarker
-                        key={courseString}
+                        key={key}
                         image={locationData.imageURLs[0]}
                         markerColor={event.color}
                         location={locationData.name}
--- a/apps/antalmanac/src/components/RightPane/Map/UCIMap.tsx
+++ b/apps/antalmanac/src/components/RightPane/Map/UCIMap.tsx
@@ -430,7 +430,7 @@ export default class UCIMap extends PureComponent {
             const courses = pins[buildingCode];
             for (let index = courses.length - 1; index >= 0; index--) {
                 const [event, eventIndex] = courses[index];
-                const courseString = `${event.title} ${event.sectionType} @ ${event.bldg} : ${event.sectionCode}`;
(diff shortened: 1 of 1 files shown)


MinhxNguyen7:
Hey @KevinWu098, I'll take a look at this, but I recommend not working on the map for now because #563 will replace its entirety

KevinWu098:
🫡 Aye Aye Capn' @MinhxNguyen7 

It was a quick one-line fix for the bug, so I thought it was worth PRing, since the bug is pretty visible :)

MinhxNguyen7:
Oh ok. I haven't looked at it, so I didn't know

apps/antalmanac/src/components/RightPane/Map/UCIMap.tsx
```diff
@@ -430,7 +430,7 @@ export default class UCIMap extends PureComponent {
             const courses = pins[buildingCode];
             for (let index = courses.length - 1; index >= 0; index--) {
                 const [event, eventIndex] = courses[index];
-                const courseString = `${event.title} ${event.sectionType} @ ${event.bldg}`;
+                const courseString = `${event.title} ${event.sectionType} @ ${event.bldg} : ${event.sectionCode} @ ${event.term}`;
                 if (locationData === undefined) return;
```
ap0nia: - Could we define `courseString` later? We shouldn't pre-emptively evaluate it and then not use it lol
- Also I think I'd prefer it named as `key`, since that's the only thing it's used for (also re-name its reference in the prop)
```suggestion
                if (locationData === undefined) return;
                const key = `${event.title} ${event.sectionType} @ ${event.bldg} : ${event.sectionCode} @ ${event.term}`;
```

ap0nia (dismissed):
Just some suggestions. Neat fix! :sunglasses:

ap0nia:
Oh also as a reference, my version of `key` for each marker is supremely brute force lol. I just join all the values together. This isn't that efficient...but React isn't either :shrug:
image

KevinWu098:
> Just some suggestions. Neat fix! :sunglasses:

Looks good to me @bevm0  :). I'll make changes when I get back to my laptop

apps/antalmanac/src/components/RightPane/Map/UCIMap.tsx
KevinWu098: Done!

MinhxNguyen7 (approved):
LGTM! I tested locally, and nothing broke (as expected)

MinhxNguyen7:
🚢

PR #579, opened 2023-05-17
Escape Dialog Fix #2

### LINKED ISSUE #551: Allow all dialogs to close when the user presses Esc
Currently, many dialogs (such as adding/renaming a schedule, saving a schedule, loading a schedule, and importing a study list) do not close when the Esc key is pressed.

To accomplish this, we possibly have to add `onClose` props to each dialog.

KevinWu098:
## Summary
- Added in the capability to use the escape key and/or click out of the Dialogs in the AppBar & the Add Custom Event dialog. 
- Edit (5/19): Virtually all popups/dialogs that I can find have escape functionality now 
    - Includes add, rename, delete schedule popups
 
Escape Dialog Responsiveness

## Test Plan
- Tested `AppBar`'s Save, Load, Import, and About dialogs for responsiveness on Esc and clicks outside.
- Also tested the `CustomEventDialog` button for the same
- Tested Add & Rename Schedule Dialogs for the same

## Issues
Closes #551 

## Additional Notes
- For the Add/Rename Schedule Dialogs, it currently retains the last inputted strings before `onClose`. Not sure what the "best practice" behavior would be for it, so leaving it up to y'alls input! _It does clear itself when the initial popup (eg the initial rename/delete schedule popup) fully closes though!_

Current Input Behavior

<sub> I have no doubt there's some implementation of mine that isn't ~~best practice~~ idiomatic 😵‍💫. Would love pointers on how to improve 💖 <sub>

## Future Followup
- Schedule Index/Schedule Selection (in the top left) popups <-- Don't want this to hold up the PR (and quite frankly `FormGroup` is driving me nuts) so I'm gonna let this simmer

### DIFF
--- a/apps/antalmanac/src/components/AppBar/AboutPage.tsx
+++ b/apps/antalmanac/src/components/AppBar/AboutPage.tsx
@@ -8,6 +8,18 @@ class AboutPage extends PureComponent {
     state: { isOpen: boolean } = {
         isOpen: false,
     };
+
+    handleClose = (wasCancelled: boolean) => {
+        if(wasCancelled)
+            this.setState({ isOpen: false}, () => {
+                document.removeEventListener('keydown', this.enterEvent, false); //this.enterEvent seems like nonsense
+                this.setState({ userID: '' })
+            });
+    }
+    
+    enterEvent = (event: KeyboardEvent) => {
+    };
+
     render() {
         return (
             <>
@@ -24,7 +36,7 @@ class AboutPage extends PureComponent {
                 >
                     About
                 </Button>
-                <Dialog open={this.state.isOpen}>
+                <Dialog open={this.state.isOpen} onClose={this.handleClose}>
                     <DialogTitle>About</DialogTitle>
                     <DialogContent>
                         <DialogContentText>
--- a/apps/antalmanac/src/components/AppBar/AboutPage.tsx
+++ b/apps/antalmanac/src/components/AppBar/AboutPage.tsx
@@ -9,16 +9,9 @@ class AboutPage extends PureComponent {
         isOpen: false,
     };
 
-    handleClose = (wasCancelled: boolean) => {
-        if(wasCancelled)
-            this.setState({ isOpen: false}, () => {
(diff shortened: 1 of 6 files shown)


KevinWu098:
Currently, still working on the escape functionality for the schedule selector in the top left, but the rest should be good for review :) 💖

MinhxNguyen7 (changes requested):
Mostly LGTM!

I would say that the rename should recover the schedule's name. I think that should be pretty straightforward, and that would, at the very least, match the behavior of the other inputs. I also think that that makes the most sense, regardless.

Also, I would like you to reformat the PR description so that it's easier to skim, though. The idea is that someone can come back in the future to see why you made the changes in this PR.

The summary should be a summary of the changes that you made. There are some in the issues section that you crossed out because you resolved them. 

If you'd like to keep the writing, I'd add an additional notes section at the bottom. In any case, I think the `Issues` section is to reference issues in the repository that this PR addresses. It's not to document the issues you had writing it.

The same idea applies to the `Future Folloup` section. If it's been done, it wouldn't be a _future_ follow-up.

KevinWu098:
> Mostly LGTM!
> 
> I would say that the rename should recover the schedule's name. I think that should be pretty straightforward, and that would, at the very least, match the behavior of the other inputs. I also think that that makes the most sense, regardless.
> 
> Also, I would like you to reformat the PR description so that it's easier to skim, though. The idea is that someone can come back in the future to see why you made the changes in this PR.
> 
> The summary should be a summary of the changes that you made. There are some in the issues section that you crossed out because you resolved them.
> 
> If you'd like to keep the writing, I'd add an additional notes section at the bottom. In any case, I think the `Issues` section is to reference issues in the repository that this PR addresses. It's not to document the issues you had writing it.
> 
> The same idea applies to the `Future Folloup` section. If it's been done, it wouldn't be a _future_ follow-up.

Aye Aye cap'n 🫡! PR has been trimmed down and I left the rename recovery as is!

MinhxNguyen7 (approved):
👍

PR #580, opened 2023-05-18
Close calendar event popups with [esc]

KevinWu098:
## Summary
<sub> _Because I'm very shaky on React Hooks, I made a new branch with just the changes to `CourseCalendarEvent.tsx` for review and feedback specifically on this, since it isn't a Dialog & `onClick`!_ <sub>

- Previously, you had to click out of Calendar Event popups (which may cover other events). 
- Implemented React Hooks that listens for the "escape" keyCode and sets the visibility of the `Paper` popup to `'none'`.

Escape Functionality

## Test Plan
Checked that on "escape", the Paper/popup disappears.

## Issues
Extends upon #579 which resulted from #551

### DIFF
--- a/apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
+++ b/apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
@@ -3,6 +3,7 @@ import { Theme, withStyles } from '@material-ui/core/styles';
 import { ClassNameMap, Styles } from '@material-ui/core/styles/withStyles';
 import { Delete } from '@material-ui/icons';
 import { Event } from 'react-big-calendar';
+import { useEffect, useRef } from 'react'; 
 
 import RightPaneStore, { BuildingFocusInfo } from '../RightPane/RightPaneStore';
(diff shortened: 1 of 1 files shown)


MinhxNguyen7 (commented):
LGTM and works. I think that this is the idiomatic way of implementing an event listener.

I would just nitpick and say that `event.keyCode === 27` should be documented, either with a comment or with `const escapeKeyCode = 27`...

KevinWu098:
> LGTM and works. I think that this is the idiomatic way of implementing an event listener.
> 
> I would just nitpick and say that `event.keyCode === 27` should be documented, either with a comment or with `const escapeKeyCode = 27`...

Added in the documentation w/ a comment!

MinhxNguyen7 (changes requested):
I should've said this last time, but I would rename `isShown` to something like `paperRef`. It's not a bool. After that, I think it's ready to merge. Sorry about that

KevinWu098:
> I should've said this last time, but I would rename `isShown` to something like `paperRef`. It's not a bool. After that, I think it's ready to merge. Sorry about that

Done! 💖

MinhxNguyen7 (approved):
Tested and LGTM!

PR #586, opened 2023-05-23
custom event - building

### LINKED ISSUE #249: Add Location (Lat/Long) or building field to custom events
User feedback request:
> It would be nice to be able to put a specific location/ building on our custom events so we can see the paths.

Let's add a field to the custom events box to add location. Maybe either a building dropdown option or input the lat/long manually. These should show up on the map too.
The info in `client/src/components/Map/static/buildingCatalogue.js` might be helpful.

stevenguyukai:
### 1. Add Building Search block at the custom event
image

### 2. Show the Location when clicking the event on the calendar 
image

### 3. Add an EVENT marker on the custom events
image

### 4. Fix Inconsistency
screenshot
image

### 5. Add location for the custom event in ADDED CLASSES
image

### 6. Add the custom event location on the map and join the route plan
image

### 7. Quick show of the location by clicking on the location link in custom event 
May-23-2023 21-03-06

### 8. Support complex and non-teaching buildings or places
image

## Issues
Closes #249


apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
```diff
@@ -315,4 +333,4 @@ class ScheduleCalendar extends PureComponent<ScheduleCalendarProps, ScheduleCale
     }
 }
 
-export default withStyles(styles)(ScheduleCalendar);
+export default withStyles(styles)(ScheduleCalendar);
```
MinhxNguyen7:
```suggestion
export default withStyles(styles)(ScheduleCalendar);

```

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/LocationSelector.tsx
```diff
@@ -0,0 +1,89 @@
...
+                        : null}
+                    onChange={this.props.handleSearch}
+                    renderInput={(params) => <TextField {...params} label="Search for a place" variant="filled" />}
+                />
+            </Paper>
+        );
+    }
+}
+
+export default withStyles(styles)(LocationSelector);
```
MinhxNguyen7:
```suggestion
export default withStyles(styles)(LocationSelector);

```

apps/antalmanac/src/components/RightPane/AddedCourses/CustomEventDetailView.tsx
```diff
@@ -68,17 +74,17 @@ const CustomEventDetailView = (props: CustomEventDetailViewProps) => {
...
                     <Delete fontSize="small" />
                 </IconButton>
-                <CustomEventDialog customEvent={customEvent} scheduleNames={props.scheduleNames} />
             </CardActions>
         </Card>
     );
 };
 
-export default withStyles(styles)(CustomEventDetailView);
+export default withStyles(styles)(CustomEventDetailView);
```
MinhxNguyen7:
```suggestion
export default withStyles(styles)(CustomEventDetailView);

```

apps/antalmanac/src/components/RightPane/Map/UCIMap.tsx
```diff
@@ -515,3 +643,5 @@ export default class UCIMap extends PureComponent {
         );
     }
 }
+
+export default withStyles(styles)(UCIMap);
```
MinhxNguyen7:
```suggestion
export default withStyles(styles)(UCIMap);

```

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
MinhxNguyen7: It's best-practice to leave a new-line at the end of files. I.e., every line in the file should have new line at the end of it, including the last one.

MinhxNguyen7 (changes requested):
I haven't looked at the code much, only at the functionality. Overall, it looks great, and I look forward to the feature. I would suggest some UI changes, though:

1. NAME in the popup doesn't need to be capitalized. Also, for the sake of consistency, it would make more sense for it to be called "Event", parallelling "Class"
2. I don't see a compelling reason to have the big EVENT box in the map marker popup. Just saying "Event: [Event name]" would suffice, IMO. 
3. Similarly, the calendar event should just have "Event" where "Dis", "Lec", etc. is. 

Also, the commits here are not atomic, which I'd say is considered bad practice. That's an issue with descriptiveness, either right now, when a reviewer (like me) looks through the changes you made, or when someone comes back to see why/how you made certain changes. It took be a little bit to get used to, but it made me more focused and productive as well. There are other reasons as well if @EricPedley wants to chime in. I don't take enormous issue here, but you should definitely keep it in mind in the future.

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -16,11 +16,15 @@ import React, { PureComponent } from 'react';
 
 import DaySelector from './DaySelector';
 import ScheduleSelector from './ScheduleSelector';
+import LocationSelector from './LocationSelector';
 import { addCustomEvent, editCustomEvent } from '$actions/AppStoreActions';
 import analyticsEnum, { logAnalytics } from '$lib/analytics';
 import { isDarkMode } from '$lib/helpers';
 import AppStore from '$stores/AppStore';
 
+// import MapMarker from '../../../RightPane/Map/MapMarker';
```
EricPedley: If you accidentally left this comment in here from trying different stuff in development, can you remove it?

On the other hand, if there is a good reason for leaving it, include that reason as another line comment.

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/LocationSelector.tsx
```diff
@@ -0,0 +1,89 @@
...
+    padding: 0,
+});
+
+
+interface LocationSelectorProps {
+    handleSearch: (event: React.ChangeEvent<unknown>, value: Building | null) => void;
+    previousOption: string;
+}
+
+class LocationSelector extends PureComponent<LocationSelectorProps>  {
```
EricPedley, line 20: At first glance, this component seems really similar in functionality to the `RightPane/Map/MapMenu` component. Can we combine their functionality into a single component that can be reused in both the Map and Custom Event creator menus? One potential solution would be replacing `LocationSelector` with a `MapMenu` that has a custom `handleSearch` function passed in. Or if these are different enough that they need to be separate, let me know why and include it as a comment in the source code.

apps/antalmanac/src/components/RightPane/Map/UCIMap.tsx
```diff
@@ -107,27 +114,62 @@ export default class UCIMap extends PureComponent {
...
-                    // Get building code, get id of building code, which will get us the building data from buildingCatalogue
-                    const buildingCode = event.bldg.split(' ').slice(0, -1).join(' ') as keyof typeof locations;
-                    const id = locations[buildingCode] as keyof typeof buildingCatalogue;
-                    const locationData = buildingCatalogue[id];
+                    if (event.isCustomEvent) {
+                        // Get building code, get id of building code, which will get us the building data from buildingCatalogue
+                        var buildingCode;
+                        var id;
+                        var locationData;
+                        if (event.bldg.includes("(")){
```
EricPedley, line 131: I don't see why a lot of this stuff is specific to custom events. Is there any way we can reuse this stuff between both cases (custom and non-custom events) so that the stuff inside `if(event.isCustomEvent)` is very clearly only stuff that custom events need? And on that note, how are the custom event markers even different than the normal markers? Can we just use the same logic for both and not even have that conditional?

apps/antalmanac/src/components/RightPane/Map/UCIMap.tsx
```diff
@@ -461,6 +545,50 @@ export default class UCIMap extends PureComponent {
                 );
             }
         }
+        for (const buildingCode in customEventPins) {
```
EricPedley, line 534: In the same vein as my comment earlier in this file, I think a lot of this stuff is the same for custom and normal events so we should reuse more code. Is there any way you can reduce the diff here so we aren't adding 40 more lines to deal with custom event logic?

EricPedley (changes requested):
Some code reuse comments to address, plus I agree with Minh's suggestions for the UI, so make sure to address those as well. Of course, if you have some compelling reasons to leave the UI as-is or take it in a different direction, feel free to push back and start a discussion here.

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
stevenguyukai: OK! I will add an empty line on the file that has not.

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
stevenguyukai: Yes, I accidentally left this comment. I will delete it.

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/LocationSelector.tsx
stevenguyukai: I changed this class to a location selector class, so wherever we want to add a location selector, we can use this class. This is now used by the MapMenu & CustomEventDialog. I delete the unnecessary part of this class.

apps/antalmanac/src/components/RightPane/Map/UCIMap.tsx
stevenguyukai: Yes, for this file, I can put most of the same code together.

apps/antalmanac/src/components/RightPane/Map/UCIMap.tsx
stevenguyukai: For this file, it is a bit hard to do so. Regular class pin contains regular `locationData` and detailed information. But for custom event, it has special detailed information to show (for example, no `room` info and the it is called `event` rather then `class`). Also the calculation of locationData for customevent is bit different, because some location for custom event has no `()`. so the formula for calculate regular class's locationData is not applicable here. So I will still remain two part like that.

stevenguyukai:
Updated the code following the all above comments

EricPedley:
Closing because this will be superseded by #726

PR #587, opened 2023-05-23
Multiple STAFF display on separate lines

### LINKED ISSUE #585: Two staff instructors incorrect placement
When there's a section with two instructors named "STAFF", they get appended. E.g.:

image

We should either make them different lines or just remove multiple mentions of STAFF

KevinWu098:
## Summary
Added a `<div>` tag to the profName, making STAFFSTAFF display on separate lines.
STAFF STAFF

## Test Plan
Looks fine at all widths, row resizes to fit for 2+ STAFF as well.

## Issues
Closes #585

### DIFF
--- a/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
+++ b/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
@@ -1,4 +1,14 @@
-import { Button, Popover, TableCell, TableRow, Theme, Tooltip, Typography, useMediaQuery } from '@material-ui/core';
+import {
+    Box,
+    Button,
+    Popover,
+    TableCell,
+    TableRow,
+    Theme,
+    Tooltip,
+    Typography,
+    useMediaQuery,
+} from '@material-ui/core';
 import { withStyles } from '@material-ui/core/styles';
 import { ClassNameMap, Styles } from '@material-ui/core/styles/withStyles';
 import classNames from 'classnames';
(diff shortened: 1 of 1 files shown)


apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
```diff
@@ -166,7 +166,7 @@ const InstructorsCell = withStyles(styles)((props: InstructorsCellProps) => {
                     </div>
                 );
             } else {
-                return profName;
+                return <div key={profName}> {profName} </div>;
```
ap0nia: Incorrect! Never use `div`, `p`, or literally any native HTML tag in MUI...they have their own version for everything. I didn't choose this UI library, but that's just how the game is played :pensive: .

```suggestion
                return <Box key={profName}>{profName}</Box>;
```

On a technical note, the reason you would legitimately prefer `Box` over `div` is because it'll be styled properly according to whatever theme settings we applied for MUI, which results in greater visual consistency (supposedly, even though they're literally both `div` at the end of the day).

Since you're here, you could also replace all the other `div` tags with `Box` instead. It shouldn't really affect the UI at all, but it's just more aligned with the spirit of MUI :shrug: I already happened to do this, so you can use this as a reference (MUIv5) https://github.com/icssc/AntAlmanac/blob/aponia-monorepo/apps/client/src/components/School/Department/Course/Section/Section.tsx

ap0nia (changes requested):
Never use native HTML elements in MUI! Should've used TailwindCSS if you wanted to use regular HTML tags! :pensive:

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
KevinWu098: Box Box it is!

ap0nia (approved):
I think this is good, just tiny changes in the UI components

PR #590, opened 2023-05-24
trpc + auth + some react-query

ap0nia:


apps/backend/src/env.ts
```diff
@@ -1,16 +1,22 @@
 import { type } from 'arktype';
-import 'dotenv/config';
+
+import { config } from 'dotenv';
+
+config({ path: '../../.env' })
```
ap0nia: I put the env in the root of the project :^)

apps/backend/src/routers/index.ts
```diff
@@ -1,12 +1,12 @@
 import { router } from '../trpc';
 import newsRouter from './news';
 import usersRouter from './users';
-import zotcourseRouter from "./zotcours";
+import zotcourseRouter from './zotcours';
```
Voark, line 4: Did the pre-commit not work on my end? Surprised this didn't fix itself...

PR #594, opened 2023-05-25
Fix RMP url scheme

### LINKED ISSUE #593: RMP Linking Broken
Clicking on instructor link leads to a 404 page on RMP. This is probably because of RMP changing their URL routing that is no longer compatible. I haven't found how we can fix this yet

MinhxNguyen7:
## Summary
Change the RMP link's scheme so that RMP will take it. It doesn't automatically open the professor page anymore and just opens the search, but I don't think there's anything straightforward that we can do about that.

## Test Plan
Click on professor names and see if it works.

## Issues

Closes #593 

## Future Followup

1. This is quite some to do, but we could save RMP's professor IDs in our DB and serve that instead of the query link.
image

2. We link to PeterPortal because they have a professor rating feature. Unfortunately, they don't have that many ratings, which somewhat kneecaps the utility of the feature.
2. We could ask PeterPortal/API to give us an endpoint. I have suggested that they should scrape reviews from RMP such that RMP reviews are a subset of PP reviews. Otherwise, they can never get past the chiken-and-egg problem where you don't have enough reviews to be useful. Travel booking websites do this for the same reason.

### DIFF
--- a/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
+++ b/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
@@ -167,7 +167,7 @@ const InstructorsCell = withStyles(styles)((props: InstructorsCellProps) => {
                 return (
                     <Box key={profName}>
                         <a
-                            href={`https://www.ratemyprofessors.com/search/professors?sid=U2Nob29sLTEwNzQ=&q=${lastName}`}
+                            href={`https://www.ratemyprofessors.com/search/professors/1074?q=${lastName}`}
                             target="_blank"
                             rel="noopener noreferrer"
                         >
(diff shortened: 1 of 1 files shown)


EricPedley:
I think you'd have to change the entire query to be something like https://www.ratemyprofessors.com/search/professors/1074?q=klefstad instead. Just tried your change as-is and it only searches up professors at UCI instead of the specific query, so I did a search manually to see what the generated URL looks like

EricPedley (changes requested):
See previous comment, re-request review when it's addressed.

MinhxNguyen7:
@EricPedley There's funny shit going on. The link in the PR is `"/professors?sid=U2Nob29sLTEwNzQ=&q=[LAST NAME]"`, but the staging instance is using the old schema `'/teachers?sid=U2Nob29sLTEwNzQ=&query=[LAST NAME]"`. I've tested it locally, and it does work.

I'll change it to the format that you gave because it's cleaner. That'll trigger a re-run of the Actions, and hopefully it'll work

EricPedley:
LGTM

EricPedley (approved):

PR #598, opened 2023-05-29
Added toggle for upper div classes

### LINKED ISSUE #597: Suggestion: Add toggle for Upper Divs
> When filtering for classes, you should consider adding an option that separates all of the upper division courses from the lower division so that it makes it easier for students to browse through the class list without having to scroll really far down

This would/could be implemented both on the AA side and the PeterPortal API side. On the AA side, just filter by course numbers is probably the most direct way

Example Implementation with toggle:
Example Toggle

KevinWu098:
## Summary
Added toggle to display only upper-level courses by having it input 100-199 into the Course Code manual search input. 

_If we want to include graduate level, we can just change it to 100-999_

Toggle Functionality

## Test Plan
- Displays only courses between 100-199, including ones like STATS 120A. 
- When toggled on/off the url is also updated properly.

## Issues
Closes #597 

## Future Followup
Currently, I have the Upper Div toggle with `labelPlacement='top'`, but the Online toggle is defaulted to `'labelPlacement='end'`. I changed the positioning for UI reasons (it looks too cramped otherwise), but this might be worth changing to maintain a consistent style.

Different Label Placement

### DIFF
--- a/apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/CourseNumberSearchBar.tsx
+++ b/apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/CourseNumberSearchBar.tsx
@@ -24,42 +24,33 @@ class CourseNumberSearchBar extends PureComponent<Record<string, never>, CourseN
     };
 
     handleChange = (name: string) => (event: ChangeEvent<HTMLInputElement>) => {
+        const url = new URL(window.location.href);
+        const urlParam = new URLSearchParams(url.search);
+        urlParam.delete('courseNumber');
+
         if (name === 'upper') {
             if (event.target.checked) {
                 this.setState({ courseNumber: '100-199' });
                 RightPaneStore.updateFormValue('courseNumber', '100-199');
-                const url = new URL(window.location.href);
-                const urlParam = new URLSearchParams(url.search);
-                urlParam.delete('courseNumber');
-
                 urlParam.append('courseNumber', '100-199');
-                const param = urlParam.toString();
-                const new_url = `${param.trim() ? '?' : ''}${param}`;
-                history.replaceState({ url: 'url' }, 'url', '/' + new_url);
             } else {
                 this.setState({ courseNumber: '' });
                 RightPaneStore.updateFormValue('courseNumber', '');
-                const url = new URL(window.location.href);
-                const urlParam = new URLSearchParams(url.search);
                 urlParam.delete('courseNumber');
-
-                const param = urlParam.toString();
-                const new_url = `${param.trim() ? '?' : ''}${param}`;
-                history.replaceState({ url: 'url' }, 'url', '/' + new_url);
             }
-        } else {
+        }
+
+        if (name === 'number') {
             this.setState({ courseNumber: event.target.value });
             RightPaneStore.updateFormValue('courseNumber', event.target.value);
-            const url = new URL(window.location.href);
-            const urlParam = new URLSearchParams(url.search);
-            urlParam.delete('courseNumber');
             if (event.target.value) {
                 urlParam.append('courseNumber', event.target.value);
             }
-            const param = urlParam.toString();
-            const new_url = `${param.trim() ? '?' : ''}${param}`;
-            history.replaceState({ url: 'url' }, 'url', '/' + new_url);
         }
+
+        const param = urlParam.toString();
+        const new_url = `${param.trim() ? '?' : ''}${param}`;
+        history.replaceState({ url: 'url' }, 'url', '/' + new_url);
     };
 
     componentDidMount() {
--- a/apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/CourseNumberSearchBar.tsx
+++ b/apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/CourseNumberSearchBar.tsx
@@ -1,7 +1,6 @@
 import { FormControlLabel, Switch, TextField } from '@material-ui/core';
 import { ChangeEvent, PureComponent } from 'react';
 
-import { Box } from '@mui/material';
 import RightPaneStore from '../../RightPaneStore';
 
 interface CourseNumberSearchBarState {
--- a/apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/CourseNumberSearchBar.tsx
+++ b/apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/CourseNumberSearchBar.tsx
@@ -23,29 +23,27 @@ class CourseNumberSearchBar extends PureComponent<Record<string, never>, CourseN
         courseNumber: this.getCourseNumber(),
     };
 
-    handleChange = (name: string) => (event: ChangeEvent<HTMLInputElement>) => {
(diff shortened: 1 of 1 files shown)


apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/CourseNumberSearchBar.tsx
```diff
@@ -14,27 +14,52 @@ class CourseNumberSearchBar extends PureComponent<Record<string, never>, CourseN
...
 
-    handleChange = (event: ChangeEvent<HTMLInputElement>) => {
-        this.setState({ courseNumber: event.target.value });
-        RightPaneStore.updateFormValue('courseNumber', event.target.value);
-        const url = new URL(window.location.href);
-        const urlParam = new URLSearchParams(url.search);
-        urlParam.delete('courseNumber');
-        if (event.target.value) {
-            urlParam.append('courseNumber', event.target.value);
+    handleChange = (name: string) => (event: ChangeEvent<HTMLInputElement>) => {
```
EricPedley: Can you refactor this into two functions for upper and number? The higher order function here seems unnecessarily complex since we're only using it twice.

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/CourseNumberSearchBar.tsx
```diff
@@ -14,27 +14,52 @@ class CourseNumberSearchBar extends PureComponent<Record<string, never>, CourseN
...
+        if (name === 'upper') {
+            if (event.target.checked) {
+                this.setState({ courseNumber: '100-199' });
+                RightPaneStore.updateFormValue('courseNumber', '100-199');
+                const url = new URL(window.location.href);
+                const urlParam = new URLSearchParams(url.search);
+                urlParam.delete('courseNumber');
+
+                urlParam.append('courseNumber', '100-199');
+                const param = urlParam.toString();
```
EricPedley: Seems like this is duplicate code between the two parts of this conditional statement. Can you move the stuff that's repeated outside the if/else?

EricPedley (changes requested):
Functionality looks good, just make a few code quality changes and this PR will be ready to merge.

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/CourseNumberSearchBar.tsx
```diff
@@ -14,24 +14,40 @@ class CourseNumberSearchBar extends PureComponent<Record<string, never>, CourseN
...
+    handleChange = (name: string) => (event: ChangeEvent<HTMLInputElement>) => {
         const url = new URL(window.location.href);
         const urlParam = new URLSearchParams(url.search);
         urlParam.delete('courseNumber');
-        if (event.target.value) {
-            urlParam.append('courseNumber', event.target.value);
+
+        if (name === 'upper') {
+            if (event.target.checked) {
+                this.setState({ courseNumber: '100-199' });
```
EricPedley: Good changes, this function is more readable now. Now that I can see more clearly what it's doing, I think a better design would be to leave `handleChange` as it was before this PR, and make another function `handleUpperDivChange` that has some simple logic like 
```
if(event.target.checked) {
    handleChange({target: {value: '100-199'}})
else {
    handleChange({target:{}})
}
```
This makes it more modular, and you could also rename `handleChange` to be something like `handleNumberChange` to be more specific since it only has one job.
This would also let you try to make the type on `event` more strict on both of the functions since they're different types of input elements (TextField vs Switch instead of a generic HTMLInputElement). I'm not sure if more fine-grained types exist but if they do we should use them.

Let me know what you think of these suggestions. I get that this is kinda overkill for a small PR but the idea is to practice code quality so you can write code that's more readable for bigger changes.

EricPedley (changes requested):

KevinWu098:
> Let me know what you think of these suggestions. I get that this is kinda overkill for a small PR but the idea is to practice code quality so you can write code that's more readable for bigger changes.

I always appreciate feedback 🫶 💖. 

In regards to the suggestions:
- I had a real hard time getting the string from `handleUpperDivChange` to play nice with `handleChange` which wanted `ChangeEvent<HTMLInputElement`. 
- I also wasn't 100% on how `{target: {value: '100-199'}}` worked and I couldn't finagle it into playing nice
- [Type Predicates](https://stackoverflow.com/questions/51528780/typescript-check-typeof-against-custom-type?newreg=a2ed99479b654ca5b849829cee1870da) and checking were also hassly (and eventually didn't work either)

However, to make it more modular, I ended up splitting it into three functions: `handleUpperDivChange`, `handleNumbersChange`, `handleCourseNumbers`. The first two, associated with the text input and the switch, respectively, pass strings (`event.target.value` or `'100-199'`) to the `handleCourseNumbers` which executes all the relevant URL & state business

And for making the type stricter... there is `ChangeEvent<HTMLTextAreaElement>`, which does work fine, but I left it out of the most recent commits for your input on whether it's worth changing.

LMK whatcha think!

EricPedley:
> * I had a real hard time getting the string from `handleUpperDivChange` to play nice with `handleChange` which wanted `ChangeEvent<HTMLInputElement`.
>
> * I also wasn't 100% on how `{target: {value: '100-199'}}` worked and I couldn't finagle it into playing nice

It'd look like this (see screenshot), but I think your solution is cleaner because passing in our own "event" is kinda janky compared to just writing a function that actually takes a string.
image

> * [Type Predicates](https://stackoverflow.com/questions/51528780/typescript-check-typeof-against-custom-type?newreg=a2ed99479b654ca5b849829cee1870da) and checking were also hassly (and eventually didn't work either)

Seems cursed, glad we aren't using it.


> However, to make it more modular, I ended up splitting it into three functions: `handleUpperDivChange`, `handleNumbersChange`, `handleCourseNumbers`. The first two, associated with the text input and the switch, respectively, pass strings (`event.target.value` or `'100-199'`) to the `handleCourseNumbers` which executes all the relevant URL & state business

Good changes, better than what I had in mind 👍

> And for making the type stricter... there is `ChangeEvent<HTMLTextAreaElement>`, which does work fine, but I left it out of the most recent commits for your input on whether it's worth changing.

Yeah, do this.

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/CourseNumberSearchBar.tsx
```diff
@@ -14,24 +14,38 @@ class CourseNumberSearchBar extends PureComponent<Record<string, never>, CourseN
...
-        RightPaneStore.updateFormValue('courseNumber', event.target.value);
+    handleUpperDivChange = (event: ChangeEvent<HTMLInputElement>) => {
+        if (event.target.checked) {
+            this.handleCourseNumbers('100-199');
+        } else {
+            this.handleCourseNumbers('');
+        }
+    };
+
+    handleNumbersChange = (event: ChangeEvent<HTMLInputElement>) => {
```
EricPedley: Change the type here to be stricter like you mentioned in your comment

EricPedley (changes requested):

EricPedley (approved):
lgtm 🚀🚀🚀🚀🚀

PR #603, opened 2023-05-31
Removed onClose from Menu, adjusted function accordingly

### LINKED ISSUE #602: Inapproriate snackbar popup with TBA classes
For a TBA class, when you click the dropdown to add to a schedule but then click out of it without adding to a schedule, it still says "Added Online/TBA Class" (or something along these lines)

KevinWu098:
## Summary
Bug fix to prevent snackbar popup when clicking out of the Add Course menu

Before:
Before

After:
After

## Test Plan
All functionality appears to be the same, and I'm not entirely sure what the original intent was.

## Issues
Closes #602


MinhxNguyen7 (approved):
🚀

PR #605, opened 2023-06-01
Fixed schedule adding bug

### LINKED ISSUE #604: Adding to Schedule from Menu doesn't add to correct schedule
When adding a course to a schedule from the dropdown menu, it adds to the current schedule, not the actually selected schedule.

Inappropriate Add

KevinWu098:
## Summary
Adding to a specific schedule using the menu while on another schedule would always add to the current schedule. By removing `this.schedule.getCurrentScheduleIndex()`, this bug is resolved.

Functionality is maintained because the correct ScheduleIndex was already being properly passed by both the Add button and the Menu Buttons, but it was being overwritten.

Proper functionality

## Test Plan
All functionality remains the same.

## Issues
Closes #604

### DIFF
--- a/apps/antalmanac/src/stores/Schedules.ts
+++ b/apps/antalmanac/src/stores/Schedules.ts
@@ -1,6 +1,6 @@
-import {amber, blue, deepOrange, deepPurple, green, pink, purple,} from '@material-ui/core/colors';
+import { amber, blue, deepOrange, deepPurple, green, pink, purple } from '@material-ui/core/colors';
 
-import {calendarizeCourseEvents, calendarizeCustomEvents, calendarizeFinals} from './calendarizeHelpers';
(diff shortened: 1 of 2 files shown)


apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -1,6 +1,6 @@
-import {amber, blue, deepOrange, deepPurple, green, pink, purple,} from '@material-ui/core/colors';
+import { amber, blue, deepOrange, deepPurple, green, pink, purple } from '@material-ui/core/colors';
```
EricPedley, line 1: Are these styling changes made by one of our autoformatting or linting configs? Not sure why they're popping up here when this file has been edited a couple times since we added Prettier and ESLint.

EricPedley (approved):
Looks good, but do address the comment I left about the styling. It's not bad enough to block my approval so I marked it as "resolved" but I'm curious since these formatting changes make reviewing larger PRs harder.
lgtm

PR #608, opened 2023-06-04
Resolved Custom Event Day Selector Bug

### LINKED ISSUE #607: Custom events day selector doesn't respond until name change
When you press a day selector, nothing appears to happen, and the change is only reflected after you change the event name

KevinWu098:
## Summary
The custom day selector now works and updates properly when it's being used.

Previously the `handleChange` / `setSelectedDays` function was doing something which I don't 100% understand, but now it just directly calls `setSelectedDays` and updates DaySelector properly.

_I believe the result of the old implementation was that the `checked` prop of `<Checkbox>` would not update when `handleChange` was called, resulting in a state where the `days?: boolean[]` was accurate, but not being displayed._

Clicky Clicky Worky

## Test Plan
All functionality is the same.

## Issues
Closes #607


johnlorenzini (approved):
Tested and works great! Great job. Not sure if I'm allowed to go ahead and merge so I'll leave the merge pending.

KevinWu098:
> Tested and works great! Great job. Not sure if I'm allowed to go ahead and merge so I'll leave the merge pending.

:) Pinging @EricPedley then!

Issue #440, opened 2022-09-17
ZotCourse Feature Parity


EricPedley:
I've seen a lot of people using [ZotCourse](https://zotcourse.appspot.com/), which is a similar schedule planning website to ours. It has some features that would fit well in AntAlmanac that should be on our roadmap to implement:

- [x] resizable menus: #431
- [x] gpa and grade distribution in search results #671 
- [x] ~~sort by columns~~
- [x] hide certain columns #624 
- [ ]~~it has "search on eaterevals". we removed this feature, but it might be worth investigating how ZotCourse implements it.~~
- [x] ~~"recent schedules": shows the most recent schedules loaded in the load menu. Depending on how we change the project with Google sign-ins, this might be incompatible with our sign-in system.~~
- [x] donation button #892 
- [x] import from zotcourse: #474

 

We can resolve this issue once we've implemented all those features. The last 3 need further investigation for utility and feasibility, but the first ones we should definitely do.

MinhxNguyen7:
Search on EaterEvals doesn't work on ZotCourse. The URL format is currently https://evaluations.eee.uci.edu/eaterevals/browse/instructor/[INSTRUCTOR_ID]. Instructor ID might be searchable per info in #174 

I like the donation button. Maybe eventually we'll get enough to buy a pizza.

I'm looking into importing from ZotCourse. It seems pretty feasible. We should discuss the utility more.

EricPedley:
We're gonna shelve sort by columns because there's only really two columns we'd want to be able to sort by: GPA and RMP. GPA would be useful but a pain in the ass to implement because of how it has to manipulate the table intrustively, and automatically getting the RMP ratings isn't working because RMP changed how the IDs work.

If anyone really wants to implement GPA sorting they can do it but we're de-prioritizing it.

EricPedley:
Recent schedules will be covered under the google auth changes.

PR #624, opened 2023-06-13
Hide Columns

KevinWu098:
## Summary
Columns can now be shown/hidden through a dropdown menu in the Course Render Pane

Demo Hide Column feature

### Columns
- Columns can now be hidden by deselecting them in a new Dropdown Menu housed in `CoursePaneButtonRow`
  - Active columns are stored and emitted in `RightPaneStore.ts`

### Removing ConditionalWrapper
- `ConditionalWrapper.tsx` has been removed in favor of plain conditional rendering logic.
- >Just use conditional rendering logic instead of a goofy component. JSX is already a pain to read, there's no need to obscure component logic even more.

### Removing colorAndDelete
- The property `colorAndDelete`, used in SectionTableBody (and related files) has been removed in favor of more reasonably calculating the proper value for `addedCourse` in `sectionTableBody.tsx`

Previously, `sectionTableBody.tsx` would always pass `colorAndDelete` as false, but then immediately recalculate it with the addedCourse state when it rendered:

```typescript
const SectionTableBody = withStyles(styles)((props: SectionTableBodyProps) => {
    const { classes, section, courseDetails, term, 
      colorAndDelete /*ALWAYS false in courseRenderPane*/, 
      allowHighlight, scheduleNames } = props;

    const [addedCourse, setAddedCourse] = useState(colorAndDelete);

{...}

    // Immediately resets addedCourse
    useEffect(() => {
        const updateHighlight = () => {
            const doAdd = AppStore.getAddedSectionCodes().has(`${section.sectionCode} ${term}`);
            setAddedCourse(doAdd);
        };
```

The new callback doesn't immediately setAddedCourse, so instead we should just set the default useState to:
```typescript
const [addedCourse, setAddedCourse] = useState(
        AppStore.getAddedSectionCodes().has(`${section.sectionCode} ${term}`)
    );
```

Doing so eliminates the need for the `colorAndDelete` prop!

### Index.ts changes
- The way the app is rendered has been changed in `index.tsx` to an async function

### Refactor scheduleConflict
- Refactored and added new helper functions into `calendarizeHelpers.ts` to assist in calculating scheduleConflicts
- Utilize useMemo and useCallback to more correctly manage state

## Test Plan
- [ ] Columns should correctly be derendered when they're deselected in the Menu
- [ ] Make sure it functions well on all devices

## Issues
Zotcourse parity a la #440


apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
```diff
@@ -24,14 +27,52 @@ const styles: Styles<Theme, object> = {
...
+    { value: 'location', label: 'Places' },
+    { value: 'sectionEnrollment', label: 'Enrollment' },
+    { value: 'restrictions', label: 'Restrictions' },
+    { value: 'status', label: 'Status' },
+];
+
 interface CoursePaneButtonRowProps {
     classes: ClassNameMap;
     showSearch: boolean;
+    columnSelector: boolean;
```
MinhxNguyen7: I would rename this to `displayColumnsSelector`

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
```diff
@@ -48,6 +89,36 @@ class CoursePaneButtonRow extends PureComponent<CoursePaneButtonRowProps> {
...
+                                <MenuItem key={column.value} value={column.value} style={{ maxWidth: '200px' }}>
+                                    <Checkbox
+                                        checked={this.state.activeColumns.indexOf(column.value) >= 0}
+                                        color="default"
+                                    />
+                                    <ListItemText primary={column.label} />
+                                </MenuItem>
+                            ))}
+                        </Select>
+                    </FormControl>
```
MinhxNguyen7: I would extract this to be a separate component so that it has its own name so that its function would be immediately obvious

apps/antalmanac/src/components/RightPane/RightPaneStore.ts
```diff
@@ -19,6 +19,18 @@ const defaultFormValues: Record<string, string> = {
     room: '',
 };
 
+// If there's a best practice that doesn't use 'let', LMK!
+const activeColumns = [
```
MinhxNguyen7: This doesn't appear to be used

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
```diff
@@ -48,6 +89,36 @@ class CoursePaneButtonRow extends PureComponent<CoursePaneButtonRowProps> {
...
+                        <MoreVert />
+                    </IconButton>
+                </Tooltip>
+
+                {this.state.columnSelector && (
+                    <FormControl className={classes.formControl}>
+                        <Select
+                            style={{ width: '0', zIndex: -1, color: isDarkMode() ? '#303030' : '#FAFAFA' }}
+                            inputProps={{ IconComponent: () => null }} // some styling nonsense to Thanos snap away the little arrow icon: https://aguidehub.com/blog/2023-02-25-how-to-hide-arrow-from-mui-select-in-react-js/
+                            open={this.state.columnSelector}
```
MinhxNguyen7: Wouldn't this always be true because, otherwise, it wouldn't be rendered?

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
```diff
@@ -24,14 +27,52 @@ const styles: Styles<Theme, object> = {
...
     showSearch: boolean;
+    columnSelector: boolean;
     onDismissSearchResults: () => void;
     onRefreshSearch: () => void;
 }
 
 class CoursePaneButtonRow extends PureComponent<CoursePaneButtonRowProps> {
+    state = {
+        columnSelector: false,
+        activeColumns: [
```
MinhxNguyen7: I would make column names an enum and make this an enum array so that all the naming is consistent. You can then translate it in `handleChange`. Alternatively, you can have an interface of `Record<ColumnName, Boolean>`.

Also, I don't get why the column names can't be the name that is displayed

apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
```diff
@@ -64,12 +66,52 @@ const styles = {
...
+    { value: 'status', label: 'Status' },
+];
+
 const SectionTable = (props: SectionTableProps) => {
     const { classes, courseDetails, term, colorAndDelete, highlightAdded, scheduleNames, analyticsCategory } = props;
     const courseId = courseDetails.deptCode.replaceAll(' ', '') + courseDetails.courseNumber;
     const encodedDept = encodeURIComponent(courseDetails.deptCode);
     const isMobileScreen = useMediaQuery(`(max-width: ${MOBILE_BREAKPOINT}`);
 
+    const [columns, setColumns] = useState([
```
MinhxNguyen7: I think that having a consolidated mapping of the selected columns would be better.

apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
```diff
@@ -64,12 +66,52 @@ const styles = {
     },
 };
 
+const TableHeadColumns: { value: string; label: string }[] = [
```
MinhxNguyen7: This should be camelcase

apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
```diff
@@ -64,12 +66,52 @@ const styles = {
...
+            setColumns(columns);
+            setActiveTableHeadColumns(TableHeadColumns.filter((column) => columns.includes(column.value)));
+        });
+
+        return () => {
+            RightPaneStore.removeListener('columnChange', (columns) => {
+                setColumns(columns);
+                setActiveTableHeadColumns(TableHeadColumns.filter((column) => columns.includes(column.value)));
+            });
+        };
```
MinhxNguyen7: Extract callback to `changeColumns()` or something along these lines so that you're not duplicating code

MinhxNguyen7 (dismissed):
Sorry for taking so long to review. I missed the notification. 

Great feature and works well. There are only some code quality issues.

My main issue is that there's so much separate state management and duplication going on. It's difficult to parse logic that should be relatively straightforward. Instead of having slightly different and separate states, try to put it all in one place, have named types, and translate it at the very end when you use it if really necessary.

Also, remember to merge. (I know it's my fault that there's a commit on `main`)

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
KevinWu098: you damn right

apps/antalmanac/src/components/RightPane/RightPaneStore.ts
KevinWu098: you also is damn right here

apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
KevinWu098: Hopefully resolved by me removing that nonsense activeTableColumn code and instead just filtering during render

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
KevinWu098: I mentioned this on Disc, but lemme try to reword it a bit better. I'm not entirely sure how to implement this because my current logic is as follows:

1. `CoursePaneButtonRow` is updated, emitting the new set of activeColumns to `SectionTable` & `SectionTableBody`. 
2. Both SectionTables then translate the activeColumns based on what they're rendering (Header strings vs Components)

Because there's slight variations, (CPBR has Restrictions but ST has Rstr, STB is components not strings), I'm not sure how to transfer the states through RightPaneStore without an in-between through value

Screenshot 2023-06-29 at 10 13 44 PM

apps/antalmanac/src/components/AppBar/CustomAppBar.tsx
```diff
@@ -38,6 +37,14 @@ interface CustomAppBarProps {
     classes: ClassNameMap;
 }
 
+const components = [
+    <SettingsMenu key="settings" />,
+    <ImportStudyList key="studylist" />,
+    <Feedback key="feedback" />,
+    <News key="news" />,
+    <AboutPage key="about" />,
+];
```
ap0nia, line 46: Initializing an array of components is most consistent when done outside a component -- i.e. just like this, especially when concerning function components. Because of React's rendering paradigm, initializing components within components in non-standard ways can mess up state badly (another React footgun :tm:)

apps/antalmanac/src/components/ConditionalWrapper.tsx
ap0nia, line 1: Good riddance to this abomination of a component. :sob: 💢

apps/antalmanac/src/components/PatchNotes.tsx
```diff
@@ -31,29 +31,27 @@ const PatchNotes = () => {
             >
                 <DialogTitle>{"What's New - May 2023"}</DialogTitle>
                 <DialogContent>
-                    <DialogContentText>
```
ap0nia: This component is a `p` tag, don't nest all this random stuff into a `p` tag.

apps/antalmanac/src/components/AppBar/CustomAppBar.tsx
```diff
@@ -60,36 +67,18 @@ const CustomAppBar = ({ classes }: CustomAppBarProps) => {
...
-                    wrapper={(children) => (
-                        <div>
-                            <MenuIcon onClick={handleClick} />
-                            <Menu anchorEl={anchorEl} keepMounted open={Boolean(anchorEl)} onClose={handleClose}>
-                                {children}
-                            </Menu>
-                        </div>
-                    )}
-                >
+                {isMobileScreen ? (
```
ap0nia, line 70: Just use conditional rendering logic instead of a goofy component. JSX is already a pain to read, there's no need to obscure component logic even more.

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
```diff
@@ -1,56 +1,161 @@
...
+        [RightPaneStore.setActiveColumns]
+    );
+
+    const handleOpen = useCallback(() => {
+        setOpen(true);
+    }, [setOpen]);
+
+    const handleClose = useCallback(() => {
+        setOpen(false);
+    }, [setOpen]);
```
ap0nia, line 86: Technically all callback functions should be memo-ized if you want to write "idiomatic React", but the performance improvements are negligible if you consider that React is one of the worst performant frameworks in existence. I'll just leave these callbacks here and let inconsistencies exist in the codebase where callbacks haven't been memo-ized :shrug:

apps/antalmanac/src/index.tsx
```diff
@@ -1,3 +1,23 @@
-import { render } from 'react-dom';
+import { createRoot } from 'react-dom/client';
 import App from './App';
-render(<App />, document.getElementById('root'));
+
+/**
+ * This function runs in the browser to load the React single-page application.
+ */
+async function main() {
```
ap0nia, line 7: I prefer explicitly creating and invoking asynchronous functions for executable scripts because Promises can be scoped without any issues and it's easy to tell what the file wants to accomplish, i.e. run a function when it's imported.

ap0nia (approved):
I think the changes look fine. I'll wait for the preview deployment to finish before merging.

EricPedley:
Bumping this PR, seems like all that's left is to fix some merge conflicts

EricPedley (approved):

ap0nia:
wait, stop

ap0nia:
What are you doing, don't do this
```ts
  useEffect(() => {
        const updateHighlight = useCallback(() => {
          setAddedCourse(AppStore.getAddedSectionCodes().has(`${section.sectionCode} ${term}`));
        }, [setAddedCourse, AppStore.getAddedSectionCodes]);

        const checkAndDisplayScheduleConflict = () => {
            if (AppStore.getCourseEventsInCalendar().length < 1) {
                setScheduleConflict(false);
                return;
            }
```

ap0nia:
Don't invoke hooks inside of other hooks

KevinWu098:
💀 I did a whoopsie

ap0nia:
Ok, fine, give me a bit to fix it

KevinWu098:
Sorry 'bout all that nonsense, hopefully this latest commit should resolve the issues, but I'll test on the staging instance // read over the code again right now!

ap0nia:
Ok no like. The entire strategy for figuring out conflicts is just wrong. There shouldn't even be a useEffect + useState. The entire logic should be a single useMemo

apps/antalmanac/src/stores/calendarizeHelpers.ts
```diff
@@ -144,8 +144,8 @@ export const calendarizeCustomEvents = (currentCustomEvents: RepeatingCustomEven
  * @returns the start and end time of a course in a 24 hour time with a leading zero (##:##)
  * @returns undefined if there is no WebSOC time (e.g. 'TBA', undefined)
  */
-export const translateWebSOCTimeTo24HourTime = (section: AASection) => {
-    const timeString = section.meetings[0].time.replace(/\s/g, '');
+export const translateWebSOCTimeTo24HourTime = (time: string) => {
```
ap0nia: Can anybody confirm if changing this interface breaks any part of the app? 

The rationale for the API change is ***don't pass in the entire object  to the function if all it needs is a string*** to do its work

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
```diff
@@ -349,97 +362,119 @@ interface SectionTableBodyProps {
...
+    const sectionDetails = useMemo(() => {
+        return {
+            // If there already exists a more well-written way to translate secton.meetings days (string) into a number, LMK
+            // Converts SuTuTh -> [Su, Tu, Th] -> [0, 2, 4]
+            day: section.meetings[0].days.match(/[A-Z][a-z]*/g)?.map((day: string) => DAYS_TO_NUMS[day]),
+            translatedTime: translateWebSOCTimeTo24HourTime(section.meetings[0].time),
+            startTime: '',
+            endTime: '',
+        };
+    }, [section.meetings]);
```
ap0nia: Instead of recalculating this arbitrarily, we know that it should be updated whenever `section.meetings` changes, so memoize it with that

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
```diff
@@ -349,97 +362,119 @@ interface SectionTableBodyProps {
...
-            checkAndDisplayScheduleConflict();
+    useEffect(() => {
+        AppStore.on('addedCoursesChange', updateHighlight);
+        AppStore.on('currentScheduleIndexChange', updateHighlight);
+
+        return () => {
+            AppStore.removeListener('addedCoursesChange', updateHighlight);
+            AppStore.removeListener('currentScheduleIndexChange', updateHighlight);
         };
+    }, [updateHighlight]);
```
ap0nia: For clarity, all event handlers are attached and removed in their own useEffect blocks

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
```diff
@@ -349,97 +362,119 @@ interface SectionTableBodyProps {
...
         return () => {
-            AppStore.removeListener('addedCoursesChange', updateCourseState);
-            AppStore.removeListener('currentScheduleIndexChange', updateCourseState);
+            AppStore.removeListener('addedCoursesChange', updateCalendarEvents);
+            AppStore.removeListener('currentScheduleIndexChange', updateCalendarEvents);
         };
-    }, [section.sectionCode, term]); //should only run once on first render since these shouldn't change.
+    }, [updateCalendarEvents]);
+
+    const scheduleConflict = useMemo(() => {
```
ap0nia, line 438: This value is derived from other stateful values, so it shouldn't be recorded with `useState` and `useEffect`. If it's declared as an un-memoized value, then it's re-calculated on every single render anyways. It's memoized because that's literally the correct way to write this dumb shit.

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
```diff
@@ -19,6 +22,16 @@ import { mobileContext } from '$components/MobileHome';
 import locationIds from '$lib/location_ids';
 import { translateWebSOCTimeTo24HourTime } from '$stores/calendarizeHelpers';
 
+const DAYS_TO_NUMS: { [key: string]: number } = {
```
ap0nia: This is moved outside the function component because it's a const value and we don't really need it re-calculated on every render.

ap0nia (approved):
Ok, code is fixed. I have to go so I can't immediately verify the behavior live. Somebody else please review :pray:

TLDR:
Please don't write bad code. And if you don't know React, try consulting somebody that does before approving React code.

KevinWu098:
> Deployed staging instance to https://staging-624.antalmanac.com

I'll eventually move this into the original PR body, but for now, this is why I removed colorAndDelete:

Previously, `sectionTableBody.tsx` would always pass `colorAndDelete` as false, but then immediately recalculate it with the addedCourse state when it rendered:

```typescript
const SectionTableBody = withStyles(styles)((props: SectionTableBodyProps) => {
    const { classes, section, courseDetails, term, 
      colorAndDelete /*ALWAYS false in courseRenderPane*/, 
      allowHighlight, scheduleNames } = props;

    const [addedCourse, setAddedCourse] = useState(colorAndDelete);

{...}

    // Immediately resets addedCourse
    useEffect(() => {
        const updateHighlight = () => {
            const doAdd = AppStore.getAddedSectionCodes().has(`${section.sectionCode} ${term}`);
            setAddedCourse(doAdd);
        };
```

The new callback doesn't immediately setAddedCourse, so instead we should just set the default useState to:
```typescript
const [addedCourse, setAddedCourse] = useState(
        AppStore.getAddedSectionCodes().has(`${section.sectionCode} ${term}`)
    );
```

Doing so eliminates the need for the `colorAndDelete` prop!

ap0nia (approved):
Ok good enough, if it breaks prod I'll fix it later. :rocket:

PR #628, opened 2023-06-15
Select Component for Course Level Selection

### LINKED ISSUE #627: Course Level Selection (Parity a la Zotcourse)
I'm a fan of how they've got it done, and it's more flexible compared to how I first implemented it in #598 with just an on/off switch.

I'm planning to swap the component from a on/off to a dropdown or something

Screenshot 2023-06-14 at 12 24 47 AM

KevinWu098:
## Summary
Converted the Upper Div Switch from #598 to a full Select menu for Any, Lower, Upper, and Grad. This menu will reset to Any when courseNumber is edited.

chrome-capture-2023-5-15

## Test Plan
URL updates correctly, searches still work fine, styling is alright on mobile and desktop (if a little cramped)

## Issues
Closes #627

## Additional Notes
1. Following commit (ef98a89a830d339b00de0d9a52b8d21f21c57c25) there's a case to be made that the courseLevel selector should be in the advanced section, seeing as it does not affect the URL and it's default state is undefined, kinda like Ends Before/After etc. Thoughts?

2. Currently, I have lower levels set to 1-99, but if there exists courses with 0, LMK and I can adjust accordingly.

### DIFF
--- a/apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/CourseNumberSearchBar.tsx
+++ b/apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/CourseNumberSearchBar.tsx
@@ -92,9 +92,10 @@ class CourseNumberSearchBar extends PureComponent<Record<string, never>, CourseN
                     value={this.state.courseNumber}
                     onChange={this.handleNumbersChange}
                     helperText="ex. 6B, 17, 30-40"
+                    style={{ marginRight: 15 }}
                 />
 
-                <FormControl style={{ marginLeft: 15, marginRight: 15, width: 125 }}>
+                <FormControl style={{ minWidth: 75 }}>
                     <InputLabel>Course Level</InputLabel>
                     <Select
                         value={this.state.courseLevel}
--- a/apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/CourseNumberSearchBar.tsx
+++ b/apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/CourseNumberSearchBar.tsx
@@ -1,10 +1,18 @@
-import { FormControlLabel, Switch, TextField } from '@material-ui/core';
+import { FormControl, InputLabel, ListItemText, MenuItem, Select, TextField } from '@material-ui/core';
 import { ChangeEvent, PureComponent } from 'react';
 
 import RightPaneStore from '../../RightPaneStore';
 
+const courseLevelList: { level: string; label: string }[] = [
+    { level: '0', label: 'Any Course Division' },
+    { level: '1-99', label: 'Lower Division Only' },
+    { level: '100-199', label: 'Upper Division Only' },
+    { level: '200-999', label: 'Graduate/Professional Only' },
+];
+
 interface CourseNumberSearchBarState {
     courseNumber: string;
+    courseLevel: string;
 }
 
 class CourseNumberSearchBar extends PureComponent<Record<string, never>, CourseNumberSearchBarState> {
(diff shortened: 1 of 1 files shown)


apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/CourseNumberSearchBar.tsx
```diff
@@ -72,20 +96,28 @@ class CourseNumberSearchBar extends PureComponent<Record<string, never>, CourseN
...
-                    }
-                    labelPlacement="top"
-                    label="Upper Div Only"
-                />
+                <FormControl style={{ minWidth: 100 }}>
+                    <InputLabel>Course Level</InputLabel>
+                    <Select
+                        value={this.state.courseLevel}
+                        renderValue={(selected) =>
+                            courseLevelList.find((item) => item.level === selected)?.label as string
```
EricPedley, line 107: super-duper-ultra-nit: I don't like the parenthesis around the sole parameter of the arrow function when it's nested immediately inside another function because it makes it _slightly_ harder to see where the function ends. Feel free to ignore if you disagree.

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/CourseNumberSearchBar.tsx
```diff
@@ -21,18 +29,34 @@ class CourseNumberSearchBar extends PureComponent<Record<string, never>, CourseN
 
     state = {
         courseNumber: this.getCourseNumber(),
+        courseLevel: '',
```
EricPedley, line 32: Can we make the default value `undefined` instead of an empty string?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/CourseNumberSearchBar.tsx
```diff
@@ -1,10 +1,18 @@
-import { FormControlLabel, Switch, TextField } from '@material-ui/core';
+import { FormControl, InputLabel, ListItemText, MenuItem, Select, TextField } from '@material-ui/core';
 import { ChangeEvent, PureComponent } from 'react';
 
 import RightPaneStore from '../../RightPaneStore';
 
+const courseLevelList: { level: string; label: string }[] = [
+    { level: '0', label: 'Any Course Division' },
```
EricPedley, line 7: I think it makes more sense for the level here to be the same as the default level, and the drop-down should have this option selected by default.

EricPedley (changes requested):
Three code comments to address, one being a nit. 

In response to the additional notes:
1. I think based on the points you made, it does make sense for the course level selector to be in the advanced options, and you should lump that change in with this PR. Given that users already have to expand the manual search thing to find this dropdown, I think it's fine to expect them to dig another level deeper to find settings they need. 
2. I don't think there's any courses that are 0 but having magic numbers like this kinda sucks for readability IMO. My suggestion would be using `undefined` as the level instead of a magic value like `0`

KevinWu098:
> Three code comments to address, one being a nit. 
> 
> 
> 
> In response to the additional notes:
> 
> 1. I think based on the points you made, it does make sense for the course level selector to be in the advanced options, and you should lump that change in with this PR. Given that users already have to expand the manual search thing to find this dropdown, I think it's fine to expect them to dig another level deeper to find settings they need. 
> 
> 2. I don't think there's any courses that are 0 but having magic numbers like this kinda sucks for readability IMO. My suggestion would be using `undefined` as the level instead of a magic value like `0`

Aye aye cap'n, I'll hop to as soon as I can :)

KevinWu098:
Gonna redo all of this nonsense because I just realized PP-API literally has a thing for this

KevinWu098:
This is super scuffed and old, so I'm gonna open a new PR :D

PR #632, opened 2023-06-23
Add Tabs for Sat/Sun on Map

### LINKED ISSUE #588: Map Filters Don't Display Sat/Sun
No Sat/Sun

KevinWu098:
## Summary
Map will now display Menu Tabs for Sat/Sun & associated markers when there are weekend courses in your calendar.

Weekend Tabs

## Test Plan
Checked that functionality works swapping between schedules and when removing the weekend course

## Issues
Closes #588

### DIFF
--- a/apps/antalmanac/src/components/Map/Map.tsx
+++ b/apps/antalmanac/src/components/Map/Map.tsx
@@ -1,6 +1,6 @@
 import './Map.css';
 
-import { Fragment, useEffect, useRef, useCallback, useState, createRef } from 'react';
+import { Fragment, useEffect, useRef, useCallback, useState, createRef, useMemo } from 'react';
 import { useSearchParams, useNavigate } from 'react-router-dom';
 import L, { type Map, type LatLngTuple } from 'leaflet';
 import { MapContainer, TileLayer } from 'react-leaflet';
@@ -22,8 +22,9 @@ const ATTRIBUTION_MARKUP =
 
 const url = `https://api.mapbox.com/styles/v1/mapbox/streets-v11/tiles/{z}/{x}/{y}?access_token=${ACCESS_TOKEN}`;
 
-const work_week = ['All', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri'];
-const full_week = ['All', 'Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat'];
+const WORK_WEEK = ['All', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri'];
+const FULL_WEEK = ['All', 'Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat'];
+const weekendIndices = [0, 6];
 
 interface MarkerContent {
     key: string;
@@ -94,25 +95,43 @@ export default function CourseMap() {
     const [searchParams] = useSearchParams();
     const [selectedDayIndex, setSelectedDay] = useState(0);
     const [markers, setMarkers] = useState(getCoursesPerBuilding());
+    const [calendarEvents, setCalendarEvents] = useState(AppStore.getCourseEventsInCalendar());
 
     const updateMarkers = useCallback(() => {
         setMarkers(getCoursesPerBuilding());
     }, [setMarkers, getCoursesPerBuilding]);
 
+    const updateCalendarEvents = useCallback(() => {
+        setCalendarEvents(AppStore.getCourseEventsInCalendar());
+    }, [setCalendarEvents]);
+
     useEffect(() => {
         AppStore.on('addedCoursesChange', updateMarkers);
         AppStore.on('currentScheduleIndexChange', updateMarkers);
+
         return () => {
             AppStore.removeListener('addedCoursesChange', updateMarkers);
             AppStore.removeListener('currentScheduleIndexChange', updateMarkers);
         };
-    }, [AppStore, updateMarkers]);
+    }, [updateMarkers]);
+
+    useEffect(() => {
+        AppStore.on('addedCoursesChange', updateCalendarEvents);
+        AppStore.on('currentScheduleIndexChange', updateCalendarEvents);
+
+        return () => {
+            AppStore.removeListener('addedCoursesChange', updateCalendarEvents);
+            AppStore.removeListener('currentScheduleIndexChange', updateCalendarEvents);
+        };
+    }, [updateCalendarEvents]);
 
     useEffect(() => {
         const locationID = Number(searchParams.get('location') ?? 0);
         const building = locationID in buildingCatalogue ? buildingCatalogue[locationID] : undefined;
 
-        if (building == null) return;
+        if (building == null) {
+            return;
+        }
 
         setTimeout(() => {
             map.current?.flyTo([building.lat + 0.001, building.lng], 18, { duration: 250, animate: false });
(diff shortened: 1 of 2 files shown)


apps/antalmanac/src/components/Map/Map.tsx
```diff
@@ -176,13 +180,15 @@ export default function CourseMap() {
...
                     <Tabs value={selectedDayIndex} onChange={handleChange} variant="fullWidth" sx={{ minHeight: 0 }}>
-                        {days.map((day) => (
-                            <Tab
-                                key={day}
-                                label={day || 'All'}
-                                sx={{ padding: 1, minHeight: 'auto', minWidth: '10%', p: 1 }}
-                            />
-                        ))}
+                        {days
+                            .filter((day) => hasWeekendCourse || (day !== 'Sat' && day !== 'Sun'))
```
EricPedley: Instead of doing the filter on `days` here can you move the variable to be initialized after `hasWeekendCourse`? Looking at this rn, it seems like `days` is scoped globally in the file for no reason, and IMO this nested filter+map is slightly harder to read than just a `map`. 

Also while you're at it, can you change `day || 'All'` to just `day` and replace the empty string in the initial list with `'All'`?

EricPedley (changes requested):
One comment to address about refactoring stuff for readability before merging, other than that looks good.

KevinWu098:
> Instead of doing the filter on days here can you move the variable to be initialized after hasWeekendCourse?

Wasn't 100% sure how you wanted it done, but it should be fine now :D

apps/antalmanac/src/components/Map/Map.tsx
ap0nia: Days is correctly scoped globally. If it's defined inside the component, then it's reconstructed on every single render, i.e. whenever the store or state is mutated.

apps/antalmanac/src/components/Map/Map.tsx
```diff
@@ -146,6 +141,14 @@ export default function CourseMap() {
           }
         : undefined;
 
+    const hasWeekendCourse = AppStore.getCourseEventsInCalendar().some(
+        (event) => event.start.getDay() === 0 || event.start.getDay() === 6
+    );
+
+    const days = hasWeekendCourse
+        ? ['All', 'Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat']
+        : ['All', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri'];
```
ap0nia: Move both of these arrays into global scope so they don't get recomputed on every render. 

_Also change "All" back to ""._

This is used in subsequent lines of code for comparison; all event days will contain an empty string, which is why it's used to represent "All". Any non-empty string will filter out certain days; filtering by "All" would literally filter out all the days since no day starts with "All". You can leave a note if this wasn't clear.

```suggestion
        ? ['', 'Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat']
        : ['', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri'];
```

```ts
const workWeek = ['', 'Mon', 'Tue', ...]
const fullWeek = ['', 'Sun', 'Mon', "Tue', ...]
```

apps/antalmanac/src/components/Map/Map.tsx
```diff
@@ -177,11 +180,7 @@ export default function CourseMap() {
                 <Paper sx={{ zIndex: 400, position: 'relative', my: 2, mx: 6.942, marginX: '15%', marginY: 8 }}>
                     <Tabs value={selectedDayIndex} onChange={handleChange} variant="fullWidth" sx={{ minHeight: 0 }}>
                         {days.map((day) => (
-                            <Tab
-                                key={day}
-                                label={day || 'All'}
-                                sx={{ padding: 1, minHeight: 'auto', minWidth: '10%', p: 1 }}
-                            />
+                            <Tab key={day} label={day} sx={{ padding: 1, minHeight: 'auto', minWidth: '10%', p: 1 }} />
```
ap0nia: Revert this. Also you can remove either "p" or "padding", they're the same thing.

```suggestion
                            <Tab key={day} label={day || 'All'} sx={{ padding: 1, minHeight: 'auto', minWidth: '10%'  }} />
```

ap0nia (changes requested):

apps/antalmanac/src/components/Map/Map.tsx
```diff
@@ -146,6 +141,14 @@ export default function CourseMap() {
           }
         : undefined;
 
+    const hasWeekendCourse = AppStore.getCourseEventsInCalendar().some(
+        (event) => event.start.getDay() === 0 || event.start.getDay() === 6
+    );
```
ap0nia: Also, just a nitpick but this (and a lot of the other consts in this component) should be memoized. e.g.
```ts
    const hasWeekendCourse = useMemo(() => AppStore.getCourseEventsInCalendar().some(
        (event) => event.start.getDay() === 0 || event.start.getDay() === 6
    ), [AppStore.courseEventsInCalendar])
```
I'm not actually sure if this works properly because the store isn't a conventional observable. But that's the idea at least.

apps/antalmanac/src/components/Map/Map.tsx
ap0nia: Oh, and "" instead of "All" is a deliberate choice because the filtering of days does a match, e.g.

```ts
event.day.includes(day)
```
"" is chosen to represent "All days" because all strings will contain an empty string.

apps/antalmanac/src/components/Map/Map.tsx
KevinWu098: Fixed

apps/antalmanac/src/components/Map/Map.tsx
KevinWu098:
```typescript
Global: 
const workWeek = ['', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri'];
const fullWeek = ['', 'Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat'];
...
Export:
const days = hasWeekendCourse ? fullWeek : workWeek;
```

Ended up doing that ^. I'm not sure if doing that ternary recomputes, thereby making this whole thing pointless, but if it does, then I'd advocate for just filtering out Sat and Sun like I had originally

ap0nia: No, this is the best approach. It only does the ternary, but not the array construction. Filtering is worse than this simple recomputation. Being poorly optimized is the hallmark of React.

KevinWu098: Aye Aye cap'n 🫡

apps/antalmanac/src/components/Map/Map.tsx
```diff
@@ -24,8 +24,12 @@ const url = `https://api.mapbox.com/styles/v1/mapbox/streets-v11/tiles/{z}/{x}/{
 
 /**
  * empty day is alias for "All Days"
+ * This is used in subsequent lines of code for comparison; all event days will contain an empty string,
+ * which is why it's used to represent "All". Any non-empty string will filter out certain days;
+ * filtering by "All" would literally filter out all the days since no day starts with "All".
```
ap0nia:
```suggestion
* When filtering for events that occur on a specific day, every event's start datetime (string)
* is checked if it includes the day. All strings contain an empty string, so "" is an alias for "all days".
```

ap0nia (approved):
Looks good! Just accept the suggestion with a more succinct explanation

apps/antalmanac/src/components/Map/Map.tsx
EricPedley: Good point since I think it is used like that in other places in the codebase, but for this instance I'm pretty sure it can be "All" because the `Tabs` selector only passes on the index of the choice, and having the string there accurately represent the label is better IMO as it's slightly more readable. I also think the `event.day.includes(day)` matching empty string is kind of a hack and should just be like `day=='all' || event.day.includes(day)` for clarity but changing that everywhere in the codebase is out of scope for this PR

ap0nia: @KevinWu098 I think this is reasonable

EricPedley: Also I kinda disagree about days being scoped globally, because I don't think the performance difference really matters here since it's just initializing a list with strings, but the current solution in this PR is a bit better than it was before.

Regardless of where they're scoped, I think we should make is making those variables all caps snakecase as is the convention for hard-coded constants.

EricPedley (dismissed):
see the two comments I added on one of the threads then re-request

KevinWu098:
> see the two comments I added on one of the threads then re-request

donezo :)

ap0nia (approved):
Looks good to me :rocket:

PR #634, opened 2023-07-09
Multiple Locations on Courses

### LINKED ISSUE #592: Deal with multiple buildings
Right now we only display the first one.
(snippet from `apps/antalmanac/src/stores/calenderizeHelpers.ts`):
image

alanchangxyz:
## Summary
Adds support on calendar events (data, formatting) for multiple buildings if a course lists more than one meeting location
image
image
_(see Test Plan re: DBH 1100 appearing twice)_

Map code also tweaked to be compatible with multiple locations, but currently only shows one per class (see Future Followup)

## Test Plan
- Mostly verified on frontend with a temporary location duplication line on course store change (hence why DBH 1100 is shown twice in one of the screenshots), removed before commit to maintain proper functionality. Could not find an actual course with 2+ locations so this was used to stub one for testing
- Not completely tested with save/load due to a CORS error with the dev backend on initial load; I only set up for frontend development for this PR

## Issues

Closes #592 

## Future Followup

Might be confusing for users if a class with multiple locations does not distinguish which days it is held at each of those locations
- the map could show markers for all locations, but this PR leaves it to show only a marker for the first location (which could be the wrong meeting location on a given day). would require some sort of validation in the future to make sure it is showing the correct location(s) for the correct day


Voark:
Wasn't actually a CORS error on the backend (even though it displayed as one), it was actually bc of a bug causing the lambda to spin up a localhost server. I think I have fixed it manually, LMK if it works and I'll add the code to main.

alanchangxyz:
update: tried testing again with stubs for ELH 110 & MSTB 120 and nothing seems to be breaking (although the additional fake locations don't show in the added courses section ~~and map~~ - is that because course data for those components is pulled from api instead of store @EricPedley)

**edit#1: forgot i only made one marker per class on the map in the original description so never mind for the map - but question still applies for added courses section**

also ran into the same cors/500 error sequence after another few queries but addressing & resolving w/ fred separately

apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
```diff
@@ -86,7 +86,7 @@ interface CommonCalendarEvent extends Event {
 }
 
 export interface CourseEvent extends CommonCalendarEvent {
-    bldg: string; // E.g., ICS 174, which is actually building + room
+    bldg: string[]; // E.g., ICS 174, which is actually building + room
```
EricPedley: Idea: replace `bldg: string[]` with `locations: Location[]` and add an
```
interface Location {
    building: string
    room: string
}
```
If you agree this is a good idea, please include the change in this PR. Otherwise lmk what problems you think this would introduce.

EricPedley (changes requested):
One comment to address. I also have a question about what you wrote in the Future Followup section: how would we know which days correspond to which location?

EricPedley:
> the additional fake locations don't show in the added courses section and map - is that because course data for those components is pulled from api instead of store

It should be loading them from the store. This is the relevant snippet from `AddedCoursePane.tsx`: 
image
This is worth figuring out before merging because in theory it should work, so this might be indicative of a bug somewhere, or it could be a quirk with the way you're adding the extra location.

apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
alanchangxyz: re: showing correct locations for given days - I have no clue how we could verify this programmatically since it could be set arbitrarily by profs/schedules.

+1 re: introducing the location type (will commit that soon) although I think we can also add a `days: string[]` attribute (and potentially an enum too) for if the days of week end up being supplied programmatically later

EricPedley: IDK if we can even scrape that information about which location is for which day. @ecxyzzy do you know?

ecxyzzy: If there's a one location for one set of days (see screenshot), then the first set of days is suppposed to correspond to the first location, the second set to the second location, etc. If not, then it's arbitrary.
image
The issue is that in this case, the API supplies two meeting objects in the `meetings` array, whereas in some cases one meeting object has multiple buildings.

alanchangxyz: just curious, what class is this for testing purposes 🤔 i probably should have asked but i assumed (incorrectly) classes were searchable by only building/room

ecxyzzy: The screenshot above corresponds to section code `44100` for 2022 Spring. Another test case that might be useful is `17085` for the same term, since it has two buildings in one meeting object.

alanchangxyz:
this newest commit should be pretty encompassing with functionality that might come into contact with the multiple meetings; I made sure to test with saving/loading & exporting to ical and they all seem to be working fine even after adding the `Location` type.

@EricPedley Testing with eddy's two cases from spring '22 (`44100` for multiple meetings and `17085` for multiple buildings per meeting) seemed to work as intended so I think the persistence issue I was seeing was just the system soft-failing my attempts to insert fake locations into each course. With real locations it seems to be alright 👍

alanchangxyz:
#696 (this is gonna be a pain to rebase anyways)

Issue #611, opened 2023-06-04
Section table elements all add their own listener to AppStore, which uses lots of memory


EricPedley:
As mentioned in #133, every `SectionTableBody` adds two listeners to AppStore, which causes a warning so we had to manually increase the max listeners. I did a little experiment with the Firefox Profiler in devtools to see if it actually uses significant amounts of memory, and my preliminary results suggest the answer is yes.

To get these results, I started the profiler with "JS Allocations" turned on in settings, then searched bio sci and scrolled down through the results to get every section to load.

With listeners - 8 mb:
![image](https://github.com/icssc/AntAlmanac/assets/48658337/a4e95bfd-6ad4-4ba5-bcfa-1af88bd64b35)

Without listeners (commented out all instances of `AppStore.on` in `CourseRenderPane` and `SectionTableBody`) - <1 mb:
![image](https://github.com/icssc/AntAlmanac/assets/48658337/e3501716-0f95-43cd-856d-f71739f4a5ba)

Before anyone tackles this, make a summary of the entire website's memory usage before you do this, as this issue might actually be insignificant compared to the total usage. If it's insignificant, post the results and close this issue.

PR #636, opened 2023-07-15
UI Warning for Schedule Conflicts

KevinWu098:
## Summary
Right Pane Courses (`coursePaneEvent` in the code) will display with a grayed-out overlay (i.e. a warning) on them IF the course would conflict with another course already in the user's calendar (`calendarEventTimes`).

For example, ICS 42 at 11:00AM - 11:50 AM will be grayed-out/pseudo-hidden IF you already added ICS 68 which occurs at 11:20AM - 12:40 PM

Feature Demo

Dark Mode:
Screenshot of Feature

Light Mode:
image

## Test Plan
1. Courses should properly update when there are schedule conflicts.
2. The gray background and opacity should convey the pseudo-hidden state but still be readable in both dark and light mode

## Issues
Closes #

### DIFF
--- a/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
+++ b/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
@@ -385,13 +385,13 @@ const SectionTableBody = withStyles(styles)((props: SectionTableBodyProps) => {
         };
 
         const checkAndDisplayScheduleConflict = () => {
-            if (AppStore.getEventsInCalendar().length < 1) {
+            if (AppStore.getCourseEventsInCalendar().length < 1) {
                 setScheduleConflict(false);
                 return;
             }
 
             // An array of lists of time information on every added event
-            const calendarEventTimes = AppStore.getEventsInCalendar().map((event) => {
+            const calendarEventTimes = AppStore.getCourseEventsInCalendar().map((event) => {
                 const courseDay = event.start.getDay();
 
                 // courseStart/EndTime is normalized to ##:## (i.e. leading zero, no seconds)
--- a/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
+++ b/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
@@ -52,7 +52,7 @@ const styles: Styles<Theme, object> = (theme) => ({
             background: isDarkMode() ? '#b0b04f' : '#fcfc97',
         },
         '&.scheduleConflict': {
-            background: isDarkMode() ? '#121212' : '#bdbdbd',
+            background: isDarkMode() ? '#121212' : '#bdbdbd', // The light mode color might not be readable enough
             opacity: 0.6,
         },
     },
--- a/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
+++ b/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
@@ -52,7 +52,7 @@ const styles: Styles<Theme, object> = (theme) => ({
             background: isDarkMode() ? '#b0b04f' : '#fcfc97',
         },
         '&.scheduleConflict': {
-            background: isDarkMode() ? '#121212' : '#bdbdbd', // The light mode color might not be readable enough
+            background: isDarkMode() ? '#121212' : '#7c7c7c',
             opacity: 0.6,
         },
     },
--- a/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
+++ b/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
@@ -54,8 +54,8 @@ const styles: Styles<Theme, object> = (theme) => ({
                     45deg,
                     #f44336,
                     #f44336 2px,
-                    #515151 2px,
-                    #515151 80px
+                    ${theme.palette.action.hover} 2px,
+                    ${theme.palette.action.hover} 80px
                 )`,
             },
             '&:nth-of-type(even)': {
@@ -63,8 +63,8 @@ const styles: Styles<Theme, object> = (theme) => ({
                     45deg,
                     #f44336,
                     #f44336 2px,
-                    #424242 2px,
-                    #424242 80px
+                    ${theme.palette.background.paper} 2px,
+                    ${theme.palette.background.paper} 80px
                 )`,
             },
         },
--- a/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
+++ b/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
@@ -387,7 +387,15 @@ const SectionTableBody = withStyles(styles)((props: SectionTableBodyProps) => {
     // Google Chrome console is telling me I have trememdnous amounts of event listeners pepege
     const [timingWarning, setTimingWarning] = useState(false);
 
-    const translateDaysToNums = { Su: '0', M: '1', Tu: '2', W: '3', Th: '4', F: '5', Sa: '6' };
+    const translateDaysToNums: { [key: string]: string } = {
+        Su: '0',
+        M: '1',
+        Tu: '2',
+        W: '3',
+        Th: '4',
+        F: '5',
+        Sa: '6',
+    };
 
     useEffect(() => {
         const toggleHighlight = () => {
@@ -412,6 +420,7 @@ const SectionTableBody = withStyles(styles)((props: SectionTableBodyProps) => {
             const coursePaneEvent = {
                 // If there already exists a more well-written way
                 // to translate secton.meetings into a day number LMK
+                // Converts SuTuTh -> [Su, Tu, Th]
                 day: section.meetings[0].days.match(/[A-Z][a-z]*/g)?.map((day: string) => translateDaysToNums[day]),
                 startTime: '',
                 endTime: '',
--- a/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
+++ b/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
@@ -393,7 +393,8 @@ const SectionTableBody = withStyles(styles)((props: SectionTableBodyProps) => {
             // An array of lists of time information on every added event
             const calendarEventTimes = AppStore.getEventsInCalendar().map((event) => {
                 const courseDay = event.start.getDay();
-                //courseStart/EndTime is normalized to ##:## (i.e. leading zero, no seconds)
+
+                // courseStart/EndTime is normalized to ##:## (i.e. leading zero, no seconds)
                 const courseStartTime = event.start.toString().split(' ')[4].slice(0, -3);
                 const courseEndTime = event.end.toString().split(' ')[4].slice(0, -3);
                 return { day: courseDay, startTime: courseStartTime, endTime: courseEndTime };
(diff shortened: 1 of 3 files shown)


apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
```diff
@@ -362,26 +385,129 @@ const SectionTableBody = withStyles(styles)((props: SectionTableBodyProps) => {
...
+        Su: 0,
+        M: 1,
+        Tu: 2,
+        W: 3,
+        Th: 4,
+        F: 5,
+        Sa: 6,
+    };
+
+    // the following timeString code is stolen from ${calendarizeCourseEvents}
```
EricPedley: Can you refactor this into its own function in `calendarizeHelpers` and just import it here?

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
```diff
@@ -362,26 +385,129 @@ const SectionTableBody = withStyles(styles)((props: SectionTableBodyProps) => {
     const { classes, section, courseDetails, term, colorAndDelete, highlightAdded, scheduleNames } = props;
     const [addedCourse, setAddedCourse] = useState(colorAndDelete);
 
+    const [scheduleConflict, setScheduleConflict] = useState(false);
+
+    const translateDaysToNums: { [key: string]: number } = {
```
EricPedley: This could have a better name to suggest it's a hard-coded constant and is an object instead of a function. My suggestion is `DAYS_TO_NUMS` or something. Important properties of the name are that it's upper snake case and doesn't contain a verb since it isn't a function.

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
```diff
@@ -362,26 +385,129 @@ const SectionTableBody = withStyles(styles)((props: SectionTableBodyProps) => {
...
+            };
+
+            // Wasn't sure if running the function constantly in useEffect was bad, so moved it out just in case
+            const translatedTimeString = translateTimeString();
+            if (translatedTimeString) {
+                coursePaneEvent.startTime = translatedTimeString.startTime;
+                coursePaneEvent.endTime = translatedTimeString.endTime;
+            }
+
+            for (let i = 0; i < calendarEventTimes.length; i++) {
```
EricPedley: I think this could be a `for of` loop.

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
```diff
@@ -362,26 +385,129 @@ const SectionTableBody = withStyles(styles)((props: SectionTableBodyProps) => {
...
+                // Check if there is day overlap
+                if (!coursePaneEvent?.day?.includes(calendarEventTimes[i]?.day)) {
+                    continue;
+                }
+
+                // Then, IF the course ( starts AND ends BEFORE) OR ( starts AND ends AFTER), it's good
+                // Currently, it's coded to be greater/less than or equal to, but... if class A ends at 1:00 PM and class B starts at 1:00 PM, you're kinda boned, no?
```
EricPedley: I don't think this is possible because of how classes always end 10 minutes before the even multiples of 30 minutes, but always start at those even multiples. Even if it was possible, I think this feature should convey that WebReg won't let you sign up for a class, and back-to-back classes would suck but are probably technically a valid schedule.

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
```diff
@@ -49,7 +48,31 @@ const styles: Styles<Theme, object> = (theme) => ({
...
+                background: `repeating-linear-gradient(
+                    45deg,
+                    rgba(244, 67, 54, 0.5),
+                    rgba(244, 67, 54, 0.5) 2px,
+                    ${theme.palette.action.hover} 2px,
+                    ${theme.palette.action.hover} 80px
+                )`,
+            },
+            '&:nth-of-type(even)': {
+                background: `repeating-linear-gradient(
```
EricPedley: mfw the Cascading Style Sheets can't be cascaded to reuse some of these options 😭

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
```diff
@@ -362,26 +385,129 @@ const SectionTableBody = withStyles(styles)((props: SectionTableBodyProps) => {
...
+                if (!(happensBefore || happensAfter)) {
+                    setScheduleConflict(true);
+                    return;
+                }
+            }
+            setScheduleConflict(false);
+            return;
+        };
+
+        const handleEventListeners = () => {
```
EricPedley: More naming nitpicks for this and some other functions:
1. `handleEventListeners` is hella generic. Maybe `updateFormatting`?
2. `toggleHighlight` doesn't actually always toggle the highlight. Should be `updateHighlight`
3. `checkScheduleConflict` not only checks for the schedule conflict, but updates the overlay as well. Can't think of a good name off the top of my head, but it should convey all the actions that the function takes.

EricPedley (changes requested):

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
KevinWu098: checkAndDisplayScheduleConflict ?

EricPedley:
I really like this feature. I'm a bit iffy about the exact look of the red stripes though. Maybe a solid but low-opacity red highlight across the whole row and the red bars should be thicker. I think we should definitely put a lot of thinking into the UI/UX of this feature, because it might confuse people if they can't figure out what it means when there is red over a section. 

In addition, since this is a pretty big feature addition, it's worth updating the patch notes again, which should be included as a change in this PR. You can reuse most of the PR description and the gif/pics for this, plus take a look at the merged PR list and throw in anything else that's new since the last patch notes.

KevinWu098:
UI/UX wise, a possible way to communicate the reason for the red would be a tooltip over the add button if scheduleConflict is true.

Something along the lines of:
"This course overlaps with another event in your calendar!"

EricPedley:
The tooltip sounds like a good idea, I think you should add it

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
KevinWu098: I'll leave as is then!

KevinWu098:
Everything should be cleaned up now! Your input on some of the refactored/renamed names would be appreciated! I chose some that are... a little verbose, but I think work

@EricPedley

KevinWu098:
I ran into a couple of additional notes / potential implementations that I wanted to ask about, so here they are! 

-----

Currently, courses will be considered to be in conflict if either another event, either a course or a custom event, shares the same timeframe.

However, the issue is that custom events will not trigger the listeners and the `updateCourseState` function because there **is no listener for custom events** in `sectionTableBody.tsx`. We can resolve this by adding another listener, but then we would have 2 -> 3 listeners per row, which in #611 is described as already a lot of memory at 2 listeners.

The current options would be:
1. Add the listener, memory be damned
2. Don't have custom events be factored into this feature
3. Keep things as is, where the conflict with custom events will only update when a course is added / removed. (This last one seems the most unintuitive for the user!) 

------------
Regarding how we want to handle the implementation of features (i.e. grayed-out state, then later the hovering feature), would we want to have the tooltip as an in-between measure to more directly explain to users what the new grayed out indicator is? 

------------
Another note about refactoring how WebSOC times are translated. There's already a To-do note in `calendarizeHelpers.ts` regarding refactoring two functions: `calendarizeCourseEvents` and `calendarizeFinals`. My new function `translateWebSOCTimeTo24HourTime` would/should also be refactored as part of that, but I think this would deviate too far from the original scope of this PR. 

Block in question:
``` typescript
            let startHour = parseInt(startStr, 10);
            const startMin = parseInt(startMinStr, 10);
            let endHour = parseInt(endStr, 10);
            const endMin = parseInt(endMinStr, 10);
            const weekdayInclusion: boolean[] = [
                date.includes('Sat'),
                date.includes('Sun'),
                date.includes('Mon'),
                date.includes('Tue'),
                date.includes('Wed'),
                date.includes('Thu'),
                date.includes('Fri'),
            ];
            if (ampm === 'pm' && endHour !== 12) {
                startHour += 12;
                endHour += 12;
                if (startHour > endHour) startHour -= 12;
            }
``` 

Once this PR goes through, I'd be happy to open up another PR refactoring `calendarizeHelpers.ts`

EricPedley:
Definitely add the tooltip in this PR. Idk how I feel about the immediate feedback upon hover, I could use some more convincing before I agree we should implement it.

EricPedley:
For the custom events thing, the ideal would be add the listeners and use a memory profiler to see if it actually makes a difference. If it's a significant impact then I think we just shouldn't support custom events. Although I haven't given this too much thought so don't discount the possibility of another way to do this that doesn't add hundreds of listeners.

EricPedley:
No comment on the websoc time thing until I get a chance to look at the code in context on my laptop later

KevinWu098:
> For the custom events thing, the ideal would be add the listeners and use a memory profiler to see if it actually makes a difference. If it's a significant impact then I think we just shouldn't support custom events. Although I haven't given this too much thought so don't discount the possibility of another way to do this that doesn't add hundreds of listeners.

There's perhaps a... sneaky way of reducing our listener usage by making it so AppStore would emit anytime something "schedule" related occurs.

I'm definitely spitballing here, and the memory profiler that you bring up should come first, but my proposal would be something like:

When adding a course, currently the courseAdded emitter triggers. In addition to that, we could add another emitter that's along the lines of "scheduleUpdated" / "scheduleChanged" etc that conveys some aspect of the schedule has changed. 

This could allow us to cut listeners down to one per row, seeing as scheduleIndex change, custom event change, and added course change would fall under the aforementioned category so we'd only need to listen to this new, general event.

In short: Create a new emitter that would be called in addition to each of the three currently used emitters. It wouldn't affect anything that specifically needs addCourse or scheduleIndexChange or customEvent, but then we wouldn't need to listen to each of the three individually.

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
EricPedley: sounds good

EricPedley (approved):
I think this is ready to merge in. Don't sweat it on supporting custom events with this feature because they only comprise less than 3 percent of total events on the site (big analytics W here), so it isn't a huge deal that we don't support them. I'm leaving the final merge button up to you because I've forgotten a lot of the context around this PR so lmk if there's anything you wanted me to pay more attention to that I've failed to address, and if not just merge this.

KevinWu098:
> I think this is ready to merge in. Don't sweat it on supporting custom events with this feature because they only comprise less than 3 percent of total events on the site (big analytics W here), so it isn't a huge deal that we don't support them. I'm leaving the final merge button up to you because I've forgotten a lot of the context around this PR so lmk if there's anything you wanted me to pay more attention to that I've failed to address, and if not just merge this.

I'll do a final sweep and get back to you! Off the top of my head I'll see if I can get around the custom events thing, and there may be some other code quality things lying around :)

KevinWu098:
Okie Dokie! Everything should be cleaned up! I resolved the custom event support issue by just getting course events instead of all events!

chrome-capture-2023-7-16

---------
I've also updated the original PR to be more accurate!

---------
The very last thing I can think of would be the light-mode sided implementation, and I just want to make sure that the "psuedo-disabled" state is readable among the lighter grays! @EricPedley 

image

EricPedley (approved):
lgtm ship it 🚢🚢🚢🚢🚢🚢

PR #638, opened 2023-07-17
Schedule Select element closes with [esc]

### LINKED ISSUE #551: Allow all dialogs to close when the user presses Esc

KevinWu098:
## Summary
The schedule selector (top left) will now close when [escape] is pressed.

Feature Demo

## Test Plan
All functionality related to that top left corner (renaming, deleting, adding) still work. Closing by clicking outside also still works.

## Issues
Closes #551, finishes up my PR from #579


apps/antalmanac/src/components/Calendar/CalendarToolbar.tsx
```diff
@@ -88,6 +88,11 @@ const CalendarPaneToolbar = ({
         setOpenSchedules((prev) => !prev);
     };
 
+    const handleClose = (event: React.ChangeEvent<unknown>) => {
```
EricPedley: How is the intended effect of this different than `handleMenuClose` and why are their implementations so different, desipite almost identical names? Might be worth adding comments to this component to make it clearer, since it's hard to tell why this new function is necessary.

EricPedley (dismissed):

apps/antalmanac/src/components/Calendar/CalendarToolbar.tsx
KevinWu098: Definitely was too vague in my first commit! 

I've gone and renamed handleClose to handleScheduleClose to be more specific. There are now two pairs of handleClick & handleClose, with one set for the Menu (which is the mobile menu that has the download, screenshot, and custom event adder) and another set for the Schedule selector element which is what I'm focusing on in this PR.

If we want to be more verbose and clear about what's what, handleMenuClose/Click could be renamed to handleMobileMenuClose/Click. handleScheduleClose/Click could become handleScheduleSelectClose/Click as well!

LMK what you think and I'll get right on it!

ap0nia (approved):
LGTM :sparkles:

PR #641, opened 2023-07-18
fix: backend development

### LINKED ISSUE #640: Improve Local Backend Development Experience
Action items:
1. Add a link to the backend's README in the root README
2. Figure out why `pnpm start` fails on Windows, and add instructions to the README on how to get it to work
3. Replace values in `.env.sample` with valid placeholders that will run, put all accepted values of the STAGE variable in a comment, and update instructions in the backend README on exactly how the `.env` should be set up to run locally. 
4. Get the local backend server to run on the same port as the frontend expects (hard-coded in `apps/antalmanac/src/lib/api/endpoints.ts`)

Context:
Running any of the dev commands that invokes sub-commands fails on windows. For example, `pnpm start` tries calling `VITE_LOCAL_SERVER=true turbo start`, which fails because it gets run in CMD for some reason, and you can't prepend env variable declarations like that. Even if you run all 3 commands that `pnpm start` indirectly invokes yourself, starting the backend fails because of missing environment variables if you don't have a `.env`. If you simply rename the `.env.sample` to `.env` it also fails because the placeholders aren't valid values. Even when you format your `.env` to get the server to start, it defaults to port 3000, when the frontend expects it at port 8080.

ap0nia:
# Summary
- [x] Added links to both sub-projects in the root README
- [x] used `cross-env` to reliably set environment variables for commands, i.e. `pnpm start`
- [x] used valid placeholder values in `.env.sample` (except for mongodb uri, developer should contact project lead first)
- [x] both frontend and backend expect the local dev server on port 8080

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,15 +14,21 @@ AntAlmanac is a schedule planner website for classes at UC Irvine. Here are some
 
 ### Tech Stack
 
-Our site is a React single page application bootstrapped with create-react-app and hosted on Github Pages. We use the Material UI component library for most of our styling. The site gets its course data from the [PeterPortal API](https://api.peterportal.org/), which is also maintained by ICS Student Council (ICSSC).
+Our site is a React single page application bootstrapped with create-react-app and hosted on Github Pages.
+We use the Material UI component library for most of our styling.
+The site gets its course data from the [PeterPortal API](https://api.peterportal.org/),
+which is also maintained by ICS Student Council (ICSSC).
 
 ### History
 
 AntAlmanac was created in 2018 by a small group of students under the leadership of @the-rango.  
-They formed an AntAlmanac club to recruit other students and work on new features, so that the website would live on even after its makers graduated.  
-In 2019, @devsdevsdevs took over as AntAlmanac Project Lead and oversaw a massive rewrite of the codebase, laying the foundation for the AntAlmanac that we know and love today.  
+They formed an AntAlmanac club to recruit other students and work on new features,
+so that the website would live on even after its makers graduated.  
+In 2019, @devsdevsdevs took over as AntAlmanac Project Lead and oversaw a massive rewrite of the codebase,
+laying the foundation for the AntAlmanac that we know and love today.  
 In 2020, AntAlmanac was adopted by the ICSSC Projects Committee.  
-ICSSC continues to provide funding, marketing, and engineering to support the growing number of users and open-source developers that make up our AntAlmanac Community.
+ICSSC continues to provide funding, marketing, and engineering 
+to support the growing number of users and open-source developers that make up our AntAlmanac Community.
 Since then, the project has continued to evolve and grow with successive generations of projects committee members!
 
 | Year        | Project Lead         |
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -87,3 +95,31 @@ git checkout main
 git pull
 pnpm run deploy
 ```
+
+## Further Reading
+Sub-projects have their own READMEs with more in-depth information.
+
+[AntAlamanc](apps/antalmanac)
```
EricPedley: `apps/antalmanac` doesn't actually have it's own README

README.md
```diff
@@ -87,3 +95,31 @@ git checkout main
...
+[Backend Server](apps/backend/README.md)
```
EricPedley: IMO this should be mentioned under the "Running the Backend" section

README.md
```diff
@@ -87,3 +95,31 @@ git checkout main
...
+
+## Troubleshooting
```
EricPedley, line 101: kudos: adding a Troubleshooting section was a good idea.

apps/backend/package.json
```diff
@@ -3,16 +3,16 @@
     "version": "0.0.1",
     "description": "Backend for AntAlmanac",
     "scripts": {
-        "dev": "esbuild --watch --bundle --sourcemap --outdir=dist --platform=node src/index.ts",
-        "build": "esbuild --bundle --minify --outdir=dist --platform=node src/lambda.ts",
-        "start": "concurrently 'npm run dev' 'nodemon dist'",
+        "dev": "tsx src/index.ts",
```
EricPedley: Does `tsx` also live-reload the program when the source changes? I think that was the goal of the convoluted dev command before.

apps/backend/scripts/build.mjs
```diff
@@ -0,0 +1,15 @@
...
+    minify: true,
+    platform: 'node',
+    outdir: 'dist',
+    entryPoints: {
+      index: 'src/index.ts',
+    }
+  })
+}
+
+main()
```
EricPedley, line 15: Why does this need to be a module? From my limited understanding of modules vs scripts, a small program like this counts as a script and would be a `.js` file.

EricPedley (changes requested):

apps/backend/package.json
ap0nia: Oh, yeah it can

`tsx watch src/index.ts`

apps/backend/scripts/build.mjs
ap0nia: No, that's not what the distinction between `mjs` and `cjs` files are. `mjs` files use the new es6 syntax for importing, `cjs` is the old CommonJS system.

```ts
// esm
import path from 'path'

// CommonJS
const path = require('path')
```

A plain `js` file is assumed to be CommonJS unless `"type": "module"` is set in the closest `package.json`

The reason I'm using `mjs` is because there's no `"type": "module"` in the `package.json` and I don't like CommonJS because it's cringe!

README.md
ap0nia: Ok, I'll just move the backend README section and nuke the frontend README link.

apps/backend/scripts/build.mjs
EricPedley: Makes sense, thx for the explanation

EricPedley (changes requested):
Staging got CORS'd
image

image

I don't think changes introduced in this PR broke it though, seems like the difference between this and the working environments is that the working ones hit `dev.api.antalmanac.com` and this one is hitting `staging-641.api.antalmanac.com`. 

Potential fix could be rearranging the API staging endpoints URLs to be `api.staging-N.antalmanac.com`?

ap0nia:
Oh shoot, I built the wrong file for the lambda lol

EricPedley:
How is staging this goddamn fast?

EricPedley (approved):
lgtm

ap0nia:
We've achieved vercel at home :tm:

Oh it's because I'm not changing the frontend; the frontend messes with CloudFormation, which takes like years lol

EricPedley:
> Oh it's because I'm not changing the frontend; the frontend messes with CloudFormation, which takes like years lol

The backend also goes through CloudFormation though. This got me curious and I looked through the build logs for both, seems like the bottleneck for the frontend is updating our S3 bucket deployment resource, which takes ~2 minutes, as opposed to the lambda updating in <10 seconds. Also on the first deploy to staging, it has to create _all_ of the 15 staging resources (lambda, s3 bucket deployment, cloudfront, etc.), while subsequent staging deploys only re-create the parts that've been changed.

To illustrate, here are examples from two deployments that only touched either the frontend or backend:

Lambda update:
https://github.com/icssc/AntAlmanac/actions/runs/5583563589/jobs/10204045807
image

S3 Bucket Deployment Update:
https://github.com/icssc/AntAlmanac/actions/runs/5566378004/jobs/10167451315
image

Entire stack creation:
https://github.com/icssc/AntAlmanac/actions/runs/5563145002/jobs/10161921287

PR #643, opened 2023-07-19
feat: add support for new prereq tree

ecxyzzy:
## Summary

Update the prereq tree rendering code to handle the API's new prereq tree format. Also add additional metadata to the leaf nodes if present.

Note that this is also `no deploy` because this will not work with API prod, since I manually uploaded the old prereq trees to the prod DB to prevent AA prod from breaking. Whoever decides to review this will need to checkout this branch locally and point the `PETERPORTAL_REST_ENDPOINT` at https://staging-58.api-next.peterportal.org/.

## Test Plan

Start dev server locally and make sure things don't break.

Some test cases I ran:

image

image

image

## Issues

N/A


apps/antalmanac/src/components/RightPane/SectionTable/PrereqTree.tsx
```diff
@@ -40,30 +46,37 @@ interface TreeProps {
...
+    } else {
+        const prereqTree = prerequisite as Record<string, PrerequisiteNode[]>;
         return (
             <div className={'prerequisite-node'}>
                 <div style={{ display: 'inline-flex', flexDirection: 'row', padding: '0.5rem 0' }}>
                     <span style={{ margin: 'auto' }}>
                         <div className={'prereq-branch'}>
-                            {Object.prototype.hasOwnProperty.call(prerequisite, 'OR') ? 'one of' : 'all of'}
+                            {
+                                Object.entries(phraseMapping).filter(([x, _]) =>
```
EricPedley: super-nit: one-letter variable name `x`. Maybe do something like `propertyName`. I'm even having trouble coming up with something because I barely understand what `x` is supposed to be here.

apps/antalmanac/src/components/RightPane/SectionTable/PrereqTree.tsx
```diff
@@ -40,30 +46,37 @@ interface TreeProps {
 
 const Tree: FC<TreeProps> = (props) => {
```
EricPedley: Not part of your diff, but I feel like `Tree` sucks as a name for this class. Maybe throw in a name change to `PrereqTreeDisplayNode` or something else if you can think of a better one.

EricPedley (changes requested):

EricPedley:
Super minor changes requested but I didn't approve because I think they're so easy we should just knock them out of the way first. It'd be more overhead to create separate follow-up issues.

EricPedley (approved):
Approving rn but I'm assuming we need to closely coordinate when to merge this into prod so we can release the API change at the same time. Will coordinate on Discord.

EricPedley (approved):
lgtm

PR #644, opened 2023-07-22
Change hard-coded color to random choice from palette

EricPedley:
## Summary
When we choose a color for a course it's from a fixed-size palette. When we run out of unique colors from the palette, previously we just chose a hard-coded color. This changes it to choose a random one from the palette.
## Test Plan
Add courses until you run out of unique colors (at least 8)


KevinWu098 (approved):
LGTM! 🚀

PR #649, opened 2023-07-31
Add back Dependency and Script from #620

### LINKED ISSUE #648: Dependency and Script from #620 removed by #563
Merging #563 removed the following which were added as part of #620:

1. `"@packages/antalmanac-types": "*"`
2. `"start:local": "VITE_LOCAL_SERVER=true vite"`

Regarding # 1, the removal of this package resulted in DX problems during development, but didn't actually break any builds or prevent work from being done. 

image

Regarding # 2:

>VITE_LOCAL_SERVER env variable gets set so when developing the frontend uses localhost for the api, otherwise it will just use the dev.api.antalmanac.com endpoint.
-- Fred

The dependency and script were removed due to #563, so closer inspection of packages when submitting a PR would help to resolve/prevent these issues from reoccuring!

KevinWu098:
## Summary
1. Updated package.json with the dependency and script.
2. Then ran `pnpm i` which also updated `pnpm-lock.yaml`
3. In additional to the dependency and script, I believe yaml is now updated with the latest version of turbo (1.10.8 -> 1.10.12) in addition to both `/memory-pager@1.5.0:` and `/sparse-bitfield@3.0.3:` gaining `requiresBuild: true`.

## Test Plan
Everything still seems to run fine.

## Issues
Closes #648

### DIFF
--- a/apps/antalmanac/package.json
+++ b/apps/antalmanac/package.json
@@ -10,6 +10,7 @@
     "repository": "github:icssc/antalmanac",
     "scripts": {
         "start": "vite",
+        "start:local": "VITE_LOCAL_SERVER=true vite",
         "dev": "vite",
         "build": "vite build",
         "preview": "vite preview",
@@ -19,6 +20,7 @@
         "lint": "eslint src"
     },
     "dependencies": {
+        "@packages/antalmanac-types": "*",
         "@date-io/date-fns": "^2.16.0",
         "@emotion/react": "^11.10.5",
         "@emotion/styled": "^11.10.5",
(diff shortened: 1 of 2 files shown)


apps/antalmanac/package.json
```diff
@@ -19,6 +20,7 @@
         "lint": "eslint src"
     },
     "dependencies": {
+        "@packages/antalmanac-types": "*",
```
EricPedley, line 23: how did this being excluded not break our builds?

EricPedley (approved):

apps/antalmanac/package.json
KevinWu098: I wish I knew, but that might be something @Voark would know!

PR #651, opened 2023-08-02
Prereq Tree is now within a Popover Component

### LINKED ISSUE #650: Prerequisite Tree can get very large
chrome-capture-2023-7-2

For classes with lots of dependencies, such as Math 3A, the prereq tree gets so large that other information is hard to view.

KevinWu098:
## Summary
1. Added a Button and Popover component which contains the Prereq Tree. Previously, for courses such as Math 3A, the Prereq Tree was very large and made for a less friendly UX. 
2. Although this adds an additional step for the user, I think that because the prereq tree is obscuring other info such as GE categories (or even the prereqs/dependencies as just text), it's alright for it to be put behind another action.

Before:

After:
chrome-capture-2023-7-3

## Test Plan
1. Make sure that the styling and functionality work as intended across all devices.

## Issues
Closes #650

### DIFF
--- a/apps/antalmanac/src/components/RightPane/SectionTable/PrereqTree.css
+++ b/apps/antalmanac/src/components/RightPane/SectionTable/PrereqTree.css
@@ -206,6 +206,7 @@
     margin-top: 2em;
 }
 
+/* Styling that targets the Popover component (not sure why it can't be done inline)*/
 .MuiPopover-paper {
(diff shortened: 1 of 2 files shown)


EricPedley (approved):
🚢🚢🚢🚢🚢🚢🚢🚢🚢 ship it

PR #654, opened 2023-08-16
[WIP] Trpc + auth

Voark:
Yes this PR is extremely bloated. Also it is not completely finished yet (task items below), but probably should start the review process earlier because of how big it is...

### TRPC
- Added [TRPC](https://trpc.io/) to the repo to handle requests to and from the backend with schema validator [ArkType](https://arktype.io/) to enforce better type safety
- Thanks to @ap0nia for help with ArkType and setting up the schemas
- **We now have our own copy of PeterPortal API schemas/types because the schemas are not currently in the PeterPortal API repo.** It was set up this way because I needed it for this specific PR but this probably should be changed in the future so there is less opportunities for code smell 

### Auth
- Added Google OAuth login
- Users can login/logout with the new button in the top right
- Term Changes
  - In order to improve user experience, schedules are now only allowed to have a single term
    - This obviously only affects new schedules, old schedules still load in fine but they will show a "Multiple Terms" term and until it is fixed no new courses can be added
    - Adding new courses is not completely disallowed when terms do not match
    - The schedule selector now lists and orders the schedules by term
    - The TermViewer (term selector in the calendar pane) will autoselect the first schedule from a term when the term is changed
   - [ ] The schedule dropdown should probably be changed to allow reordering of schedules and quick deleting of schedules through some edit mode
   - A lot of RightPaneStore and AppStore logic was added in order to handle new dependencies on terms and term changes, this could potentially lead to issues but I don't want to think deep enough to find any
- Thanks again to @ap0nia for writing a whole entire auth library and then doing the implementation in this repo for me 😄 

### Misc
- Update actions, CDK and env variables to add new Google OAuth secrets
- [ ] The OAuth secrets are currently tied to my UCI email and they definitely need to be moved to another account
- Changed the calendar toolbar to reactively hide and unhide the buttons in response to scaling of the calendar pane
  - Also moved finals button to right side
- Fixed some issues with the new dev scripts not working properly

- There are probably more changes I just forgot and I don't properly name my commits 
- Big thanks to @falsidge and @ChaseC99 for all their work on the first auth attempt, many ideas were stolen from that branch

## Test Plan
- Manual testing, need more help with this

## Issues

Closes #
#229


EricPedley:
Would it be easy to split this up into multiple PR's? Based on your description it seems like it could be a PR for the TRPC changes separately from everything else that we could merge in first. It might be possible to interactive rebase to reorder your commits,  create a new branch at the point where TRPC changes end and auth starts, then open another PR.

Testing the TRPC changes alone also makes sense because we can just verify everything works the same as it did before. But if it's too hard to split this thing up we can just bite the bullet and read the entire diff (but i really don't want to 🥺)

Voark:
> Would it be easy to split this up into multiple PR's? Based on your description it seems like it could be a PR for the TRPC changes separately from everything else that we could merge in first. It might be possible to interactive rebase to reorder your commits, create a new branch at the point where TRPC changes end and auth starts, then open another PR.
> 
> Testing the TRPC changes alone also makes sense because we can just verify everything works the same as it did before. But if it's too hard to split this thing up we can just bite the bullet and read the entire diff (but i really don't want to 🥺)

I did initially start out just working on the TRPC changes and there currently already is a branch for that, but at some point I guess I made this branch and started working on auth. It's been many months since that happened so I have no idea why...

Anyways I would love to split this branch up but I don't want to spend the time dealing with all the hassles that will probably follow... At some point I did a big merge conflict fix but that commit is after the transition from the TRPC to Auth changes and I would prefer to not have to redo that. If it is really too unwieldy I can try but after like 15 min it doesn't seem trivial unfortunately.

MinhxNguyen7 (changes requested):
This is not a full review. These are just some areas that I spotted that need to be worked on. I'll update it as I find more things.

- [ ] Upon opening the app, you are loaded into "Any Term" by default. When I tried to search for a course and hit enter, I got `TypeError: Cannot read properties of undefined (reading 'replace') at https://staging-654.antalmanac.com/assets/SectionTable-8c79322c.js:85:36652`. I think this is the result of the API change to time formats. This also means that I can't do much testing.
- [ ] It's not possible to select anything other than "Any Term" in the calendar dropdown if you're not logged in, and you can only select Fall 2023 when you're logged in.

apps/antalmanac/src/components/Calendar/CalendarToolbar.tsx
```diff
@@ -88,49 +109,62 @@ const CalendarPaneToolbar = ({
...
+
+        const resizeObserver = new ResizeObserver(handleResize);
+        if (paperRef.current) {
+            resizeObserver.observe(paperRef.current);
+        }
+
+        return () => {
+            window.removeEventListener('resize', handleResize);
+        };
+    }, [paperRef]);
```
ap0nia: No, please just use a media query. You can import one from MUI, i.e. `import { useMediaQuery } from '@mui/material'`

apps/antalmanac/src/components/Calendar/CalendarToolbar.tsx
```diff
@@ -88,49 +109,62 @@ const CalendarPaneToolbar = ({
...
+                    open={openSchedules}
+                    onClick={handleScheduleClick}
                 >
-                    Finals
-                </Button>
-            </Tooltip>
-
-            <div className={classes.spacer} />
+                    {Array.from(scheduleMap.entries()).flatMap(([term, schedules]) => {
+                        return [
```
ap0nia: Why are you returning an array lol, just return a JSX element.

apps/antalmanac/src/components/Calendar/CalendarToolbar.tsx
```diff
@@ -38,17 +40,35 @@ const styles: Styles<Theme, object> = {
...
+        paddingLeft: '1em',
+        width: '10rem',
+    },
+    termSelector: {
+        flexGrow: 1,
+        flexShrink: 1,
+        display: 'flex',
+        overflow: 'hidden',
+        paddingLeft: '1rem',
+        paddingRight: '1rem',
```
ap0nia: There should be a padding inline CSS variable which does padding left and right https://developer.mozilla.org/en-US/docs/Web/CSS/padding-inline

In MUIv5, you can also use `px` or `paddingX`

apps/antalmanac/src/components/Calendar/TermViewer.tsx
```diff
@@ -0,0 +1,75 @@
...
+};
+
+interface TermViewerProps {
+    classes: ClassNameMap;
+}
+
+const TermViewer = ({ classes }: TermViewerProps) => {
+    const getTerm = () => {
+        return AppStore.schedule.getCurrentScheduleTerm();
+    };
```
ap0nia, line 24: This doesn't even have to be inside the component; this function can be moved outside and prevented from being re-calculated on every render.

apps/backend/src/routers/index.ts
```diff
@@ -1,12 +1,14 @@
 import { router } from '../trpc';
 import newsRouter from './news';
 import usersRouter from './users';
-import zotcourseRouter from "./zotcours";
+import zotcourseRouter from './zotcours';
```
ap0nia, line 4: typo! skill issue!

ap0nia (commented):
I reviewed

apps/antalmanac/src/components/Calendar/CalendarToolbar.tsx
Voark: Just replaced this all with main now that the resizing is responsive I'm not going to touch it.

Voark:
> This is not a full review. These are just some areas that I spotted that need to be worked on. I'll update it as I find more things.
> 
> * [ ]  Upon opening the app, you are loaded into "Any Term" by default. When I tried to search for a course and hit enter, I got `TypeError: Cannot read properties of undefined (reading 'replace') at https://staging-654.antalmanac.com/assets/SectionTable-8c79322c.js:85:36652`. I think this is the result of the API change to time formats. This also means that I can't do much testing.
> * [ ]  It's not possible to select anything other than "Any Term" in the calendar dropdown if you're not logged in, and you can only select Fall 2023 when you're logged in.

These issues should ™️  be resolved in the most recent commit, let me know if you have any issues. I did some quick testing but definitely did not hit all user flows.

MinhxNguyen7:
image

PR #657, opened 2023-08-20
Patch Notes now display if they haven't been viewed. 

### LINKED ISSUE #656: Patch Notes modal only displays on first-ever page visit
Currently, the Patch Notes will only be shown on the user's first-ever visit to AA, meaning that future patch notes would never be displayed to the user until their localstorage is cleared.

This could be changed to some non-binary value correlating to the last patch notes that were viewed by the user!

    // show modal only on first visit
    useEffect(() => {
        if (localStorage.getItem('visitedCount') == 'y') {
            setIsOpen(false);
        } else {
            localStorage.setItem('visitedCount', 'y');
        }
    }, []);

KevinWu098:
## Summary
1. Changed the "hard-coded" `if (localStorage.getItem('visitedCount') == 'y')` to a new arbitrary string.
2. This new arbitrary string is in the format YYYYMMDD and would also need to be updated anytime the patch notes are updated. (Technically means you could just make it any random string, but making it based on the date should prevent non-unique values)

Patch Note demo

Description Above: _When localStorage's `latestValue` is not 20230819, the patch notes show up again!_

## Test Plan
1. Do patch notes still display as intended (i.e. if the user hasn't seen it yet)?

## Issues
Closes #656

### DIFF
--- a/apps/antalmanac/src/components/PatchNotes.tsx
+++ b/apps/antalmanac/src/components/PatchNotes.tsx
@@ -5,12 +5,15 @@ import { useEffect, useState } from 'react';
 const PatchNotes = () => {
     const [isOpen, setIsOpen] = useState(true);
 
-    // show modal only on first visit
+    // show modal only if the current patch notes haven't been shown
+    // This is denoted by a date string YYYYMMDD (e.g. 20230819)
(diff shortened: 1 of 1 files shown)


MinhxNguyen7 (approved):
Seems right

PR #662, opened 2023-08-21
Added Course Pane will not Display ScheduleConflict

### LINKED ISSUE #660: Courses in "Added Courses" greyed out
Per #636, sections in the class search should be greyed out if it conflicts with a section in the schedule. However, because the section table is also used in the added courses pane, every class that has a "real" time gets greyed out. @KevinWu098 I assume you'd know how to fix this, so I'm tagging you.

KevinWu098:
## TLDR Summary
1. Fixed ternary which always resolved to `false` in `AddedCoursePane.tsx`, resulting in `scheduleConflict` always being checked, and thus always being displayed.
2. Refactored `highlightAdded` to `allowHighlight` for clarity.

Demoing fixed Added Course Pane

-----------

## In Depth Summary
Corrected bad logic from this line:

```typescript
                addedCourse && highlightAdded
                    ? { addedCourse: addedCourse && highlightAdded }
                    : { scheduleConflict: scheduleConflict }
```

I misunderstood what the purpose of highlightAdded was, and assumed it was a state variable for displaying the yellow-highlight for added courses. highlightAdded is **actually** a prop passed by `AddedCoursePane.tsx` and `SectionTableBody.tsx` (also `GEDataFetchProvider.tsx`, but that's not relevant here) to allow highlighting or not (i.e., highlighting shouldn't be allowed on Added Course Pane). 

This means that `highlightAdded` was always `true` on Course Render Pane and always `false` on Added Course Pane. This meant that the ternary above with `highlightAdded` would always resolve to false, therefore allowing `scheduleConflict` to be added to the CSS (since `scheduleConflict` would always be true). 

For context, this was the code prior to #636:
```typescript
className={classNames(classes.tr, { addedCourse: addedCourse && highlightAdded })}
```

For future clarity, I've also decided to refactor `highlightAdded` to `allowHighlight`, which feels more appropriate. If you've got input regarding the naming, I'm all ears :) !

## Test Plan
Check that scheduleConflict functionality remains the same, except for it now not displaying on the Added Course Pane.

## Issues
Closes #660

## Misc
If the refactoring makes more sense as a separate PR, just LMK and I'll undo them!

### DIFF
--- a/apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
+++ b/apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
@@ -230,7 +230,7 @@ class AddedCoursePane extends PureComponent<AddedCoursePaneProps, AddedCoursePan
                                 courseDetails={course}
                                 term={course.term}
                                 colorAndDelete={true}
-                                highlightAdded={false}
+                                allowHighlight={false}
                                 analyticsCategory={analyticsEnum.addedClasses.title}
                                 scheduleNames={this.state.scheduleNames}
                             />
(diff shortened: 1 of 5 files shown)


MinhxNguyen7 (approved):
LGTM!

PR #663, opened 2023-08-21
Light Mode Accessibility - Schedule Conflict

### LINKED ISSUE #661: Accessibility for greyed out courses in section table
This is very hard to read. @KevinWu098 I assume that, as many of us do, you use dark mode exclusively. However, we need to test color changes in both light and dark mode.
image

KevinWu098:
## Summary
1. Opacity will only apply to dark mode
2. Light mode's "pseudo-disabled" gray coloring was lightened
3. Dis is now colored orange #ff6e00 instead of #8d63f0 (violet) for readability

Before:

After:

## Test Plan
1. Changes are limited to styling only, so... does it look cute? 💖

## Issues
Closes #661


MinhxNguyen7 (approved):
LGTM!

PR #668, opened 2023-08-22
Change "Copy Section Code" to Chip

### LINKED ISSUE #667: Course code Button in calendar should be a Chip
The changes in #626 changed the Button component in the Course Render Pane to a Chip component in order to clean up messy styling. It overlooked the identical component also present in the Calendar Course Event popover. This should be fixed for visual consistency.

Screenshot 2023-08-22 at 3 01 16 AM

KevinWu098:
## Summary
1. Change button component to Chip component in Calendar Event popup.
2. Makes the "Copy Button" (or Chip, I suppose) consistent across both the Calendar popup and the Right Pane.
3. Refactor `clickToCopy` to an `async` function.

Screenshot 2023-08-22 at 3 07 00 AM

## Test Plan
- [ ] Styling should still work across devices
- [ ] Functionality should still work (Click it -> is it on your clipboard ?)

## Issues
Closes #667

### DIFF
--- a/apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
+++ b/apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
@@ -1,5 +1,5 @@
 import { Link } from 'react-router-dom';
-import { Button, IconButton, Paper, Tooltip } from '@material-ui/core';
+import { Chip, IconButton, Paper, Tooltip } from '@material-ui/core';
 import { Theme, withStyles } from '@material-ui/core/styles';
 import { ClassNameMap, Styles } from '@material-ui/core/styles/withStyles';
 import { Delete } from '@material-ui/icons';
@@ -38,6 +38,7 @@ const styles: Styles<Theme, object> = {
         display: 'flex',
         justifyContent: 'space-between',
         alignItems: 'center',
+        marginBottom: '0.25rem',
     },
     table: {
         border: 'none',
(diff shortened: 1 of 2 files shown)


apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
```diff
@@ -170,18 +171,18 @@ const CourseCalendarEvent = (props: CourseCalendarEventProps) => {
                             <td className={classes.alignToTop}>Section code</td>
                             <Tooltip title="Click to copy course code" placement="right">
                                 <td className={classes.rightCells}>
-                                    <Button
-                                        size="small"
-                                        onClick={(e) => {
+                                    <Chip
+                                        onClick={(event) => {
+                                            clickToCopy(event, sectionCode);
```
ap0nia, line 176: Since you're here, could you refactor `clickToCopy` to be a proper async function? I think we used `void` to get ESLint to ignore it so we can consume it as a synchronous function, but it should be async for clarity.

```ts
export async function clickToCopy(event: React.MouseEvent<HTMLElement, MouseEvent>, sectionCode: string) {
    event.stopPropagation();
    await navigator.clipboard.writeText(sectionCode)
    openSnackbar('success', 'WebsocSection code copied to clipboard');
}
```

The function is here: https://github.com/icssc/AntAlmanac/blob/main/apps/antalmanac/src/lib/helpers.ts#L333

The two consumers shouldn't be affected:
- https://github.com/icssc/AntAlmanac/blob/main/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx#L110
- https://github.com/icssc/AntAlmanac/blob/main/apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx#L180

ap0nia (changes requested):

apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
KevinWu098: Done, also tested locally and it works!

ap0nia (approved):
Good

PR #669, opened 2023-08-22
Add support for new API time formats

KevinWu098:
## Summary
Updated and refactored code + types related to times and dates used across AA to support the breaking changes of the API's new time formats.

On the user-side, nothing has changed aside from meeting time text which went from:
`MWF 12:00 - 12:50p` to `MWF 12:00 - 12:50 PM`

Functionality Demo

## Detailed Summary
1. Peterportal types updated to [Staging Instance 68](https://staging-68.api-next.peterportal.org/)
2. Types across AA have been updated in accordance to # 1 ^
3. As mentioned above, formatting of times in `SectionTableBody.tsx` looks like this now: `MWF 12:00 - 12:50 PM`. AM will now also be included in the string.
4. Calendarization has been updated, removing all the nasty regex

## To-Do
- [x] Convert 24 hour to 12 hour in a sane way
- [x] Get correct types
- [x] Update months to 0-indexing (pending PP-API changes)

## Test Plan
- [ ] Please don't crash 🫠

## Issues
Closes #

### DIFF
--- a/apps/antalmanac/src/stores/calendarizeHelpers.ts
+++ b/apps/antalmanac/src/stores/calendarizeHelpers.ts
@@ -12,7 +12,7 @@ export const calendarizeCourseEvents = (currentCourses: ScheduleCourse[] = []) =
             const endHour = parseInt(meeting.endTime.hour, 10);
             const endMin = parseInt(meeting.endTime.minute, 10);
 
-            const dates = [
+            const dates: boolean[] = [
                 meeting.days.includes('Su'),
                 meeting.days.includes('M'),
                 meeting.days.includes('Tu'),
@@ -53,15 +53,16 @@ export const calendarizeFinals = (currentCourses: ScheduleCourse[] = []) => {
 
     for (const course of currentCourses) {
         const finalExam = course.section.finalExam;
-        if (finalExam.length > 5) {
-            const [, date, , , startStr, startMinStr, endStr, endMinStr, ampm] = finalExam.match(
-                /([A-za-z]+) ([A-Za-z]+) *(\d{1,2}) *(\d{1,2}):(\d{2})-(\d{1,2}):(\d{2})(am|pm)/
-            ) as RegExpMatchArray;
+        console.log(finalExam);
+        if (finalExam.examStatus == 'SCHEDULED_FINAL') {
+            const date = finalExam.dayOfWeek;
+
             // TODO: this block is almost the same as in calenarizeCourseEvents. we should refactor to remove the duplicate code.
-            let startHour = parseInt(startStr, 10);
-            const startMin = parseInt(startMinStr, 10);
-            let endHour = parseInt(endStr, 10);
-            const endMin = parseInt(endMinStr, 10);
+            const startHour = parseInt(finalExam.startTime.hour, 10);
+            const startMin = parseInt(finalExam.startTime.minute, 10);
+            const endHour = parseInt(finalExam.endTime.hour, 10);
+            const endMin = parseInt(finalExam.endTime.minute, 10);
+
             const weekdayInclusion: boolean[] = [
                 date.includes('Sat'),
                 date.includes('Sun'),
@@ -71,11 +72,6 @@ export const calendarizeFinals = (currentCourses: ScheduleCourse[] = []) => {
                 date.includes('Thu'),
                 date.includes('Fri'),
             ];
-            if (ampm === 'pm' && endHour !== 12) {
-                startHour += 12;
-                endHour += 12;
-                if (startHour > endHour) startHour -= 12;
-            }
 
             weekdayInclusion.forEach((shouldBeInCal, index) => {
                 if (shouldBeInCal)
--- a/apps/antalmanac/src/stores/calendarizeHelpers.ts
+++ b/apps/antalmanac/src/stores/calendarizeHelpers.ts
@@ -11,7 +11,7 @@ const FINALS_WEEK_DAYS = ['Sat', 'Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri'];
 export function calendarizeCourseEvents(currentCourses: ScheduleCourse[] = []): CourseEvent[] {
     return currentCourses.flatMap((course) => {
         return course.section.meetings
-            .filter((meeting) => meeting.timeIsTBA && meeting.startTime && meeting.endTime && meeting.days)
+            .filter((meeting) => !meeting.timeIsTBA && meeting.startTime && meeting.endTime && meeting.days)
             .flatMap((meeting) => {
                 const startHour = meeting.startTime?.hour;
                 const startMin = meeting.startTime?.minute;
--- a/apps/antalmanac/src/stores/calendarizeHelpers.ts
+++ b/apps/antalmanac/src/stores/calendarizeHelpers.ts
@@ -186,21 +186,23 @@ export function normalizeTime(meeting: WebsocSectionMeeting): NormalizedWebSOCTi
     }
 }
 
-export function translate24To12HourTime(section: WebsocSectionMeeting | WebsocSectionFinalExam): string | undefined {
-    if (section.startTime && section.endTime) {
-        const timeSuffix = section.endTime.hour >= 12 ? 'PM' : 'AM';
-
-        const formattedStartHour12 = section.startTime.hour > 12 ? section.startTime.hour - 12 : section.startTime.hour;
-        const formattedEndHour12 = section.endTime.hour > 12 ? section.endTime.hour - 12 : section.endTime.hour;
-
-        // TO-DO: See if a leading zero can be added to minute
-        // prettier-ignore
-        const meetingStartTime = `${formattedStartHour12}:${section.startTime?.minute === 0 ? '00' : section.startTime?.minute}`;
-        // prettier-ignore
-        const meetingEndTime = `${formattedEndHour12}:${section.endTime?.minute === 0 ? '00' : section.endTime?.minute}`;
-
-        const timeString = `${meetingStartTime} - ${meetingEndTime} ${timeSuffix}`;
-
-        return timeString;
+export function translate24To12HourTime(startTime?: HourMinute, endTime?: HourMinute): string | undefined {
+    if (!startTime || !endTime) {
+        return;
     }
+    const postMeridian = endTime.hour >= 12;
+
+    const timeSuffix = postMeridian ? 'PM' : 'AM';
+
+    const formattedStartHour = `${startTime.hour - +(postMeridian && 12)}`
+
+    const formattedEndHour = `${endTime.hour - +(postMeridian && 12)}`
+
+    const meetingStartTime = `${formattedStartHour}:${formattedStartHour.padStart(2, '0')}`;
+
+    const meetingEndTime = `${formattedEndHour}:${formattedEndHour.padStart(2, '0')}`;
+
+    const timeString = `${meetingStartTime} - ${meetingEndTime} ${timeSuffix}`;
+
+    return timeString;
 }
--- a/apps/antalmanac/src/stores/calendarizeHelpers.ts
+++ b/apps/antalmanac/src/stores/calendarizeHelpers.ts
@@ -7,55 +7,41 @@ export const calendarizeCourseEvents = (currentCourses: ScheduleCourse[] = []) =
 
     for (const course of currentCourses) {
         for (const meeting of course.section.meetings) {
-            const timeString = meeting.time.replace(/\s/g, '');
+            const startHour = parseInt(meeting.startTime.hour, 10);
+            const startMin = parseInt(meeting.startTime.minute, 10);
+            const endHour = parseInt(meeting.endTime.hour, 10);
+            const endMin = parseInt(meeting.endTime.minute, 10);
 
-            if (timeString !== 'TBA') {
-                const [, startHrStr, startMinStr, endHrStr, endMinStr, ampm] = timeString.match(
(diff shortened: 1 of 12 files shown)


EricPedley:
Idk if you knew about this before, but with git you can modify your previous commit with the --amend flag so you don't have to create a separate commit to clean up mistakes

apps/antalmanac/src/stores/calendarizeHelpers.ts
```diff
@@ -200,3 +158,49 @@ export function parseDaysString(daysString: string): number[] {
...
+        // prettier-ignore
+        const meetingStartTime = `${formattedStartHour12}:${section.startTime?.minute === 0 ? '00' : section.startTime?.minute}`;
+        // prettier-ignore
+        const meetingEndTime = `${formattedEndHour12}:${section.endTime?.minute === 0 ? '00' : section.endTime?.minute}`;
+
+        const timeString = `${meetingStartTime} - ${meetingEndTime} ${timeSuffix}`;
+
+        return timeString;
+    }
+}
```
ap0nia: Alternate implementation

```suggestion
export function translate24To12HourTime(startTime?: HourMinute, endTime?: HourMinute): string | undefined {
    if (!startTime || !endTime) {
        return;
    }
    const postMeridian = endTime.hour >= 12;

    const timeSuffix = postMeridian ? 'PM' : 'AM';

    const formattedStartHour = `${startTime.hour - +(postMeridian && 12)}`

    const formattedEndHour = `${endTime.hour - +(postMeridian && 12)}`

    const meetingStartTime = `${formattedStartHour}:${formattedStartHour.padStart(2, '0')}`;

    const meetingEndTime = `${formattedEndHour}:${formattedEndHour.padStart(2, '0')}`;

    const timeString = `${meetingStartTime} - ${meetingEndTime} ${timeSuffix}`;

    return timeString;
}
```

Notes
- Pass the smallest amount of information into the function as possible; don't pass the entire section object if it just needs two numbers
- Return early, i.e. guard clause, if it's applicable to keep blocks organized and concise
- The subtraction syntax does look stupid; it just means that you'll take the boolean value and convert it to a number (i.e. 1 or 0) and then subtract it from the original hour
- Use padStart to do the padding to 2 characters
- Everything is represented as a string so I don't have to nest string literals

apps/antalmanac/src/stores/calendarizeHelpers.ts
```diff
@@ -200,3 +158,49 @@ export function parseDaysString(daysString: string): number[] {
...
+        const startHour = `${meeting.startTime.hour < 10 ? `0${meeting.startTime.hour}` : meeting.startTime.hour}`;
+        const endHour = `${meeting.endTime.hour < 10 ? `0${meeting.endTime.hour}` : meeting.endTime.hour}`;
+
+        const startTime = `${startHour}:${meeting.startTime.minute}`;
+        const endTime = `${endHour}:${meeting.endTime.minute}`;
+
+        return { startTime: startTime, endTime: endTime };
+    }
+}
+
```
ap0nia:
```suggestion
interface NormalizeTimeOptions {
  timeIsTBA?: boolean;
  startTime?: HourMinute | null;
  endTime?: HourMinute | null;
}

// This also works.
type NormalizeTimeOptions = Pick<WebsocSectionMeeting, 'timeIsTBA' | 'startTime' | 'endTime'>

/**
 * @param section
 * @returns The start and end time of a course in a 24 hour time with a leading zero (##:##).
 * @returns undefined if there is no WebSOC time (e.g. 'TBA', undefined)
 */
export function normalizeTime(options: NormalizeTimeOptions): NormalizedWebSOCTime | undefined {
    if (options.timeIsTBA) {
        return;
    }

    if (!options.startTime || !options.endTime) {
        return;
    }

    // Times are normalized to ##:## (10:00, 09:00 etc)
    const startHour = `${options.startTime.hour}`.padStart(2, '0')
    const endHour = `${options.endTime.hour}`.padStart(2, '0')

    const startTime = `${startHour}:${options.startTime.minute}`;
    const endTime = `${endHour}:${options.endTime.minute}`;

    return { startTime, endTime };
}
```

Notes:
- Same idea as before with reducing the amount of required information needed to pass to the function. Sometimes it makes sense to pass the entire thing, but making function interfaces simple is always a good idea
- Use guard clauses to exit the function early
- Also, for non-library projects, explicitly typing return types is less important because using inferred type definitions can be more accurate.

ap0nia:
```suggestion
```suggestion
interface NormalizeTimeOptions {
  timeIsTBA?: boolean;
  startTime?: HourMinute | null;
  endTime?: HourMinute | null;
}

// This also works.
type NormalizeTimeOptions = Pick<WebsocSectionMeeting, 'timeIsTBA' | 'startTime' | 'endTime'>

/**
 * @param section
 * @returns The start and end time of a course in a 24 hour time with a leading zero (##:##).
 * @returns undefined if there is no WebSOC time (e.g. 'TBA', undefined)
 */
export function normalizeTime(options: NormalizeTimeOptions): NormalizedWebSOCTime | undefined {
    if (options.timeIsTBA) {
        return;
    }

    if (!options.startTime || !options.endTime) {
        return;
    }

    // Times are normalized to ##:## (10:00, 09:00 etc)
    const startHour = `${options.startTime.hour}`.padStart(2, '0')
    const endHour = `${options.endTime.hour}`.padStart(2, '0')

    const startTime = `${startHour}:${options.startTime.minute}`;
    const endTime = `${endHour}:${options.endTime.minute}`;

    return { startTime, endTime };
}
```

Notes:
- Same idea as before with reducing the amount of required information needed to pass to the function. Sometimes it makes sense to pass the entire thing, but making function interfaces simple is always a good idea
- Use guard clauses to exit the function early
- Also, for non-library projects, explicitly typing return types is less important because using inferred type definitions can be more accurate.
```

File: apps/antalmanac/src/stores/calendarizeHelpers.ts [hunk 3]
```diff
@@ -139,43 +130,6 @@ export const calendarizeCustomEvents = (currentCustomEvents: RepeatingCustomEven
     return customEventsInCalendar;
```
ap0nia:
```suggestion
export function calendarizeCustomEvents(currentCustomEvents: RepeatingCustomEvent[] = []): CustomEvent[] {
    return currentCustomEvents.flatMap((customEvent) => {
        return customEvent.days.filter(Boolean).map((_day, index) => {
            const startHour = parseInt(customEvent.start.slice(0, 2), 10);
            const startMin = parseInt(customEvent.start.slice(3, 5), 10);
            const endHour = parseInt(customEvent.end.slice(0, 2), 10);
            const endMin = parseInt(customEvent.end.slice(3, 5), 10);

            return {
                customEventID: customEvent.customEventID,
                color: customEvent.color ?? '#000000',
                start: new Date(2018, 0, index, startHour, startMin),
                isCustomEvent: true,
                end: new Date(2018, 0, index, endHour, endMin),
                title: customEvent.title,
            };
        });
    });
}
```

This isn't on the diff, but this function should be implemented like this

apps/antalmanac/src/stores/calendarizeHelpers.ts
```diff
@@ -67,46 +58,46 @@ export const calendarizeFinals = (currentCourses: ScheduleCourse[] = []) => {
 
     for (const course of currentCourses) {
```
ap0nia:
```suggestion
const WEEK_DAYS = ['Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat'];

/**
 * Given a string or string array of weekdays, return an array of booleans indicating whether
 * each weekday occurs in the input.
 */
function getWeekdaysOcurring(weekDays?: string | string[] | null): boolean[] {
    return weekDays ? WEEK_DAYS.map((day) => weekDays.includes(day)) : WEEK_DAYS.map(() => false);
}

export function calendarizeFinals(currentCourses: ScheduleCourse[] = []): CourseEvent[] {
    return currentCourses
        .filter((course) => course.section.finalExam.examStatus === 'SCHEDULED_FINAL')
        .filter(
            (course) =>
                course.section.finalExam.startTime &&
                course.section.finalExam.endTime &&
                course.section.finalExam.dayOfWeek
        )
        .flatMap((course) => {
            const finalExam = course.section.finalExam;
            const startHour = finalExam.startTime?.hour;
            const startMin = finalExam.startTime?.minute;
            const endHour = finalExam.endTime?.hour;
            const endMin = finalExam.endTime?.minute;

            const weekdaysOccurring = getWeekdaysOcurring(course.section.finalExam.dayOfWeek);
            return weekdaysOccurring.filter(Boolean).map((_day, index) => {
                return {
                    color: course.section.color,
                    term: course.term,
                    title: course.deptCode + ' ' + course.courseNumber,
                    courseTitle: course.courseTitle,
                    bldg: course.section.meetings[0].bldg[0],
                    instructors: course.section.instructors,
                    sectionCode: course.section.sectionCode,
                    sectionType: 'Fin',
                    start: new Date(2018, 0, index - 1, startHour, startMin),
                    end: new Date(2018, 0, index - 1, endHour, endMin),
                    finalExam: course.section.finalExam,
                    isCustomEvent: false,
                };
            });
        });
}
```
This function should be implemented like this

ecxyzzy:
@KevinWu098 Months are already zero-indexed FYI, my bad for not letting you know sooner

KevinWu098:
> @KevinWu098 Months are already zero-indexed FYI, my bad for not letting you know sooner

Updated! :D

apps/antalmanac/src/stores/calendarizeHelpers.ts
KevinWu098: This makes sense to me, but there were some minor kinks in the implementation here, which is why I ended up reverting it: 
`return weekdaysOccurring.filter(Boolean).map((_day, index) => {`

I think this change, as well as the similar one for `calendarizeFinals` would be best done in a separate PR where `calendarizeCourseEvents` and `CalendarizeFinals` are made a little less redundant.

apps/antalmanac/src/stores/calendarizeHelpers.ts
ap0nia: Ok, I was trolling a bit. Could you change the `+` to `Number`. It'll be less stupid to read lol

apps/antalmanac/src/stores/calendarizeHelpers.ts
ap0nia: Wait, what was the issue?

apps/antalmanac/src/stores/calendarizeHelpers.ts
```diff
@@ -198,18 +195,18 @@ export function translate24To12HourTime(startTime?: HourMinute, endTime?: HourMi
...
+    const postMeridian = endTime.hour >= 12;
 
-    const formattedStartHour = `${startTime.hour > 12 ? startTime.hour - 12 : startTime.hour}`;
-    const formattedEndHour = `${endTime.hour > 12 ? endTime.hour - 12 : endTime.hour}`;
+    const timeSuffix = postMeridian ? 'PM' : 'AM';
 
-    const formattedStartMinute = `${startTime.minute}`.padStart(2, '0');
-    const formattedEndMinute = `${endTime.minute}`.padStart(2, '0');
+    const formattedStartHour = `${startTime.hour - Number(postMeridian && 12)}`;
+    const formattedEndHour = `${endTime.hour - Number(postMeridian && 12)}`;
```
KevinWu098: The problem with subtracting 12 based on postMeridian is that when it's 12 PM we don't subtract 12 to get 0 PM, it stays 12 PM.

Screenshot 2023-08-25 at 11 48 54 AM

I'm not sure what would look cleanest, but  it seems like we would have to have two separate variables, postMeridian and past12 (?).

apps/antalmanac/src/stores/calendarizeHelpers.ts
```diff
@@ -200,3 +153,60 @@ export function parseDaysString(daysString: string): number[] {
...
+    if (!startTime || !endTime) {
+        return;
+    }
+
+    const postMeridian = endTime.hour >= 12;
+
+    const timeSuffix = postMeridian ? 'PM' : 'AM';
+
+    const formattedStartHour = `${startTime.hour - Number(postMeridian && 12)}`;
+    const formattedEndHour = `${endTime.hour - Number(postMeridian && 12)}`;
```
KevinWu098:
```suggestion
    const pastNoon = endTime.hour > 12;

    const timeSuffix = endTime.hour >= 12 ? 'PM' : 'AM';

    const formattedStartHour = `${startTime.hour - Number(pastNoon && 12)}`;
    const formattedEndHour = `${endTime.hour - Number(pastNoon && 12)}`;

```

The problem with subtracting based on postMeridian (`>= 12`) is that 12:00 PM is correct, and shouldn't be "formatted." 

Screenshot 2023-08-25 at 11 59 16 AM

Note: `pastNoon` is also, in my opinion, not a particularly good name since it doesn't really convey that it's non-inclusive of 12... (?), but `postMeridian` doesn't seem like it would work if it doesn't apply to 12.

ap0nia: :sunglasses:

```ts
`${(startTime.hour - Number(pastNoon && 12)}) || 12`;
```

Ok yeah, let's just keep the ternary lol

Unless we want to just create a separate conditional.
```ts
`${startTime.hour - Number(endTime.hour > 12 && 12)}`;
```

KevinWu098: I feel like another conditional is one layer too deep for me 😵‍💫. I think I'll just do the ternary 😅

ap0nia (approved):
Looks good enough :rocket:

PR #670, opened 2023-08-22
refactor: handle websoc death

ap0nia:
## Issues

Closes #

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+charset = utf-8
+insert_final_newline = true
+end_of_line = lf
+indent_style = space
+indent_size = 4
+max_line_length = 100
--- a/apps/antalmanac/package.json
+++ b/apps/antalmanac/package.json
@@ -66,6 +66,7 @@
         "zustand": "^4.3.2"
     },
     "devDependencies": {
+        "@packages/antalmanac-types": "workspace:^",
         "@babel/core": "^7.20.12",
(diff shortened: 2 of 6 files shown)


.editorconfig
ap0nia, line 1: I need this badly, because I use 2 space indent by default. I was gonna open a dev-ops PR but I just need this now :shrug:

apps/antalmanac/package.json
ap0nia, line 1: Idk why this wasn't already installed

apps/antalmanac/src/components/Calendar/Toolbar/ExportCalendar.tsx
ap0nia, line 1: Normalized JSDoc commenting/documentation.

apps/antalmanac/tsconfig.json
ap0nia, line 1: It had extra commas, which is invalid JSON syntax.

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
ap0nia, line 1: I converted everything into function components.

MinhxNguyen7 (approved):
Seems about right. It's not merging into `main` so I didn't look too closely. We can test it along with the main PR.

I do think that this refactor makes the original PR un-atomic. The refactor should really be its own PR to `main`, but I won't protest too much unless @EricPedley takes issue

PR #671, opened 2023-08-23
GPA and Grade Distribution in Search Results

### LINKED ISSUE #599: GPA and Grade Distribution in Search Results
Like ZotCourse

image

Part of #440

MinhxNguyen7:
## Summary

Adds GPA as a section table column. Clicking on the GPA number shows the grade distribution of that class taught by the particular instructor. If there are multiple instructors, the first one on which we have data will be used.

image

Note:

- There's some creep with changing how the table is styled, but it was necessary because the GPA column would make the table too cramped on mobile otherwise. 
- There's a refactor of the columns store to Zustand in a separate module instead of RightPaneStore. I considered it justified because I needed to change the logic somewhat. I have some empathy for Brian's refactoring zeal now.

## TODO

- [x] GPA column and cell
- [x] Styling
    - [x] GPA number
    - [x] Give table min width depending on number of columns shown and force side scroll
- [x] Close GPA graph on clickaway
- [x] ~~Make sure added courses pane doesn't send duplicate requests due to race condition in caching.~~ Disable GPA column in added courses pane.
- [x] Update PP API endpoint to prod

## Test Plan
See that the GPA data shows up correctly for all kinds of searches and the graph reliably shows and hides intuitively/as expected. 

See if styling needs improvement on both mobile and desktop.

## Issues

Closes #599 
Part of #440

## Future Follow-up

Have more information in graph like ZotCourse
image


apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
```diff
@@ -85,13 +77,10 @@ export function ColumnToggleButton() {
         setOpen(false);
     }, [setOpen]);
```
ap0nia: You can safely remove this from the dependency array. You can also do this where applicable in other memo-ized callbacks.

```suggestion
    }, []);
```

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
```diff
@@ -58,24 +60,14 @@ function renderEmptySelectValue() {
  * e.g. show/hide the section code, instructors, etc.
  */
 export function ColumnToggleButton() {
-    const [activeColumns, setActiveColumns] = useState(RightPaneStore.getActiveColumns());
+    const { selectedColumns, setSelectedColumns } = useColumnStore();
```
ap0nia: You can also slice the store to reduce the amount of data the component subscribes to by default. The `useStore` function accepts a callback which can return specific properties from the store. You can return the setter/getter for columns in an array, sort of like a regular React `useState`.

```suggestion
    const [ selectedColumns, setSelectedColumns ] = useColumnStore(store => [store.selectedColumns, store.setSelectedColumns]);
```

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
```diff
@@ -103,19 +92,16 @@ export function ColumnToggleButton() {
...
-                    value={activeColumns}
+                    value={selectedColumnNames}
                     open={open}
                     onChange={handleChange}
                     onClose={handleClose}
                     renderValue={renderEmptySelectValue}
                     sx={{ visibility: 'hidden' }}
                 >
-                    {Object.entries(columnLabels).map(([column, label]) => (
+                    {Object.entries(columnLabels).map(([column, label], idx) => (
```
ap0nia: Can we compute the object entries for column labels at the top level, this saves on some computation for every render.

```suggestion
                    {columnLabelEntries.map(([column, label], index) => (
```

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
```diff
@@ -85,13 +77,10 @@ export function ColumnToggleButton() {
...
 
-    useEffect(() => {
-        RightPaneStore.on('columnChange', handleColumnChange);
-
-        return () => {
-            RightPaneStore.removeListener('columnChange', handleColumnChange);
-        };
-    }, [handleColumnChange]);
+    const selectedColumnNames = useMemo(
+        () => SECTION_TABLE_COLUMNS.filter((_, idx) => selectedColumns[idx]),
```
ap0nia: Nit: I think I prefer either `i` or `index` for the variable name.
```suggestion
        () => SECTION_TABLE_COLUMNS.filter((_, index) => selectedColumns[index]),
```

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
```diff
@@ -12,7 +12,8 @@ import {
     type SxProps,
 } from '@mui/material';
 import { ArrowBack, Visibility, Refresh } from '@mui/icons-material';
-import RightPaneStore, { type SectionTableColumn } from '../RightPaneStore';
+import RightPaneStore from '../RightPaneStore';
+import useColumnStore, { SECTION_TABLE_COLUMNS, type SectionTableColumn } from '$stores/ColumnStore';
```
ap0nia: I prefer to use named exports because there's slightly better LSP support for finding them.

```suggestion
import { useColumnStore, SECTION_TABLE_COLUMNS, type SectionTableColumn } from '$stores/ColumnStore';
```

apps/antalmanac/src/components/RightPane/SectionTable/GradesPopup.tsx
```diff
@@ -1,116 +1,137 @@
...
+        console.error(e);
+        return undefined;
+    });
+
+    if (!courseGrades) {
+        return undefined;
+    }
+
+    /**
+     * Format data for displayiing in chart.
```
ap0nia: Typo -- probably by me lol

```suggestion
     * Format data for displaying in chart.
```

apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
```diff
@@ -10,109 +12,122 @@ import {
...
+interface EnrollmentColumnHeaderProps {
+    label: string;
+}
 
-const SectionTable = (props: SectionTableProps) => {
-    const { classes, courseDetails, term, allowHighlight, scheduleNames, analyticsCategory } = props;
-    const courseId = courseDetails.deptCode.replaceAll(' ', '') + courseDetails.courseNumber;
-    const encodedDept = encodeURIComponent(courseDetails.deptCode);
+function EnrollmentColumnHeader(props: EnrollmentColumnHeaderProps) {
     const isMobileScreen = useMediaQuery(`(max-width: ${MOBILE_BREAKPOINT}`);
```
ap0nia: I think it's missing a parentheses. It also might be preferable to query MUI's built-in breakpoints, since those will correspond with the global theme settings.

```suggestion
    const isMobileScreen = useMediaQuery(`(max-width: ${MOBILE_BREAKPOINT})`);
    
    // alternatively
    const isMobileScreen = useMediaQuery(theme.breakpoints.up('sm'));
```

apps/antalmanac/src/stores/ColumnStore.ts
```diff
@@ -0,0 +1,79 @@
...
+] as const;
+
+export type SectionTableColumn = (typeof SECTION_TABLE_COLUMNS)[number];
+
+interface ColumnStore {
+    // Columns that are enabled. Sometimes columns can be hidden depending on the tab, or perhaps screen size in the future.
+    enabledColumns: boolean[];
+
+    // Columns that are selected in the dropdown
+    selectedColumns: boolean[];
```
ap0nia, line 31: Very neat JSDoc👌

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
```diff
@@ -183,6 +200,100 @@ const InstructorsCell = withStyles(styles)((props: InstructorsCellProps) => {
...
+                anchorEl={anchorEl}
+                anchorOrigin={{ vertical: 'bottom', horizontal: 'left' }}
+                transformOrigin={{ vertical: 'top', horizontal: 'left' }}
+                disableRestoreFocus
+            >
+                <GradesPopup
+                    deptCode={deptCode}
+                    courseNumber={courseNumber}
+                    instructor={instructor}
+                    isMobileScreen={useMediaQuery(`(max-width: ${MOBILE_BREAKPOINT}`)}
```
ap0nia, line 288: Same as before; maybe consider using MUI's built-in breakpoints.

```suggestion
                    isMobileScreen={useMediaQuery(`(max-width: ${MOBILE_BREAKPOINT})`)}
```

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
```diff
@@ -352,7 +463,9 @@ const StatusCell = withStyles(styles)((props: StatusCellProps) => {
     //         </NoPaddingTableCell>
     //     )
     return (
-        <NoPaddingTableCell className={`${classes[status.toLowerCase()]} ${classes.cell}`}>{status}</NoPaddingTableCell>
+        <NoPaddingTableCell className={`${classes[status.toLowerCase()]} ${classes.cell} ${classes.statusCell}`}>
```
ap0nia, line 464: Idk when/if this should be adjusted, but I don't get why we have to lowercase the status. Like, why don't we just make the keys uppercase to match the provided status? i.e. if we know that the status will always be capitalized in a certain way.

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
```diff
@@ -403,13 +517,6 @@ const SectionTableBody = withStyles(styles)((props: SectionTableBodyProps) => {
...
-    const updateColumns = useCallback(
-        (newActiveColumns: SectionTableColumn[]) => {
-            setColumns(newActiveColumns);
-        },
-        [setColumns]
-    );
-
     const updateHighlight = useCallback(() => {
         setAddedCourse(AppStore.getAddedSectionCodes().has(`${section.sectionCode} ${term}`));
     }, [setAddedCourse]);
```
ap0nia: We can remove this from the dependency array.

```suggestion
    }, []);
```

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
```diff
@@ -103,19 +92,16 @@ export function ColumnToggleButton() {
             <FormControl>
                 <Select
                     multiple
-                    value={activeColumns}
+                    value={selectedColumnNames}
                     open={open}
                     onChange={handleChange}
                     onClose={handleClose}
                     renderValue={renderEmptySelectValue}
                     sx={{ visibility: 'hidden' }}
```
ap0nia: This was recently committed to the main branch, but it's missing here.

```suggestion
                    sx={{ visibility: 'hidden', position: 'absolute' }}
```

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
```diff
@@ -230,17 +231,24 @@ class CourseRenderPane extends PureComponent<CourseRenderPaneProps, CourseRender
...
+
             try {
-                let jsonResp;
-                if (params.units.includes(',')) {
-                    jsonResp = await queryWebsocMultiple(params, 'units');
-                } else {
-                    jsonResp = await queryWebsoc(params);
-                }
+                // Query websoc for course information and populate gradescache
+                const [websocJsonResp, _] = await Promise.all([
```
ap0nia: I think it's fine to omit the dummy/placeholder variable.

```suggestion
                const [websocJsonResp] = await Promise.all([
```

apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
```diff
@@ -10,109 +12,122 @@ import {
...
+        </Box>
     );
+}
+
+function SectionTable(props: SectionTableProps) {
+    const { courseDetails, term, allowHighlight, scheduleNames, analyticsCategory } = props;
+
+    const { selectedColumns, getActiveColumns } = useColumnStore();
+
+    const isMobileScreen = useMediaQuery(`(max-width: ${MOBILE_BREAKPOINT}`);
```
ap0nia:
```suggestion
    const isMobileScreen = useMediaQuery(`(max-width: ${MOBILE_BREAKPOINT})`);
```

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
```diff
@@ -382,7 +496,7 @@ const tableBodyCells: Record<SectionTableColumn, React.ComponentType<any>> = {
 const SectionTableBody = withStyles(styles)((props: SectionTableBodyProps) => {
     const { classes, section, courseDetails, term, allowHighlight, scheduleNames } = props;
 
-    const [activeColumns, setColumns] = useState(RightPaneStore.getActiveColumns());
+    const { getActiveColumns } = useColumnStore();
```
ap0nia:
```suggestion
    const getActiveColumns = useColumnStore(store => store.getActiveColumns);
```

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableButtons.tsx
```diff
@@ -35,7 +38,7 @@ export const ColorAndDelete = withStyles(styles)((props: ColorAndDeleteProps) =>
     const isMobileScreen = useMediaQuery(`(max-width: ${MOBILE_BREAKPOINT}`);
```
ap0nia, line 38:
```suggestion
    const isMobileScreen = useMediaQuery(`(max-width: ${MOBILE_BREAKPOINT})`);
```

apps/antalmanac/src/lib/api/endpoints.ts
```diff
@@ -15,7 +15,7 @@ export const LOOKUP_NOTIFICATIONS_ENDPOINT = endpointTransform('/api/notificatio
 export const REGISTER_NOTIFICATIONS_ENDPOINT = endpointTransform('/api/notifications/registerNotifications');
 
 // PeterPortal API
-export const PETERPORTAL_GRAPHQL_ENDPOINT = 'https://api-next.peterportal.org/v1/graphql';
+export const PETERPORTAL_GRAPHQL_ENDPOINT = 'https://staging-88.api-next.peterportal.org/v1/graphql';
```
ap0nia: Are we planning on ever updating this? Maybe we could open an issue to remind us or something.

ap0nia (changes requested):
I think these changes are looking pretty solid :rocket:

I left some stylistic suggestions and some minor corrections.

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
```diff
@@ -183,6 +200,100 @@ const InstructorsCell = withStyles(styles)((props: InstructorsCellProps) => {
...
+                    color: isDarkMode() ? 'dodgerblue' : 'blue',
+                    padding: 0,
+                    minWidth: 0,
+                    fontWeight: 400,
+                    fontSize: '1rem',
+                }}
+                onClick={handleClick}
+                variant="text"
+            >
+                {gpa}
```
KevinWu098, line 274: Should this be underlined like how instructors and locations are? I'm not sure of the styling conventions here, but I feel that since it's clickable, underlining might be a good touch.

KevinWu098:
This is a pretty dang awesome PR, and I don't want to hold it up with additional feature requests but I do want to bring up points for followup.

Currently Zotcourse handles their popups like so:

Screenshot 2023-09-11 at 2 34 20 AM

I think having the `Last Taught:` text would also be great!

--------------
I'm far less convinced about this second followup, but I might as well bring it up:

Additionally, Zotcourse  distinguishes course GPA vs instructor GPA which isn't super relevant here, but I do think a parenthesis after `Grade Distribution` like `Grade Distribution (instructor)` or `Grade Distribution (course)` could be nice. I personally wasn't initially sure why the individual GPAs were different from the GPA in the zotcourse button popup. 

Again, this probably isn't super necessary, but I wanted to throw it out there.

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
MinhxNguyen7: Interesting. What's the benefit of that?

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
MinhxNguyen7: Huh. I didn't know you could do that. Python would not be happy if you did that. I'm inclined to leave the underscore in though to indicate that we're ignoring the second return value.

apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
MinhxNguyen7: Isn't that the same code?

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
MinhxNguyen7: I don't have the braincells right now to think about this, but I deem it out of scope, so you can change it some other time.

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
MinhxNguyen7: I was just thinking about that. Since it doesn't lead away from the site, the location link should not be underlined so it doesn't look as much like a url. We should change that. I think that was the accidental result of Brian changing the location link to be a <`Link`> component.

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
ap0nia: If you don't slice, then the component will react to all changes within the store (which isn't necessarily a big deal if you manage small stores properly, but generally an anti-pattern). [Reference](https://docs.pmnd.rs/zustand/recipes/recipes#fetching-everything)

apps/antalmanac/src/lib/api/endpoints.ts
MinhxNguyen7: Well I'd have to wait for PP to push the change to `main`. My plan is to wait for that, then change it, then push this PR. I'll leave this comment up unresolved as reminder.

apps/antalmanac/src/stores/ColumnStore.ts
MinhxNguyen7: The power of Copilot

apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
ap0nia: There should be a missing right parenthesis

apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
MinhxNguyen7: I can whip up a PR that changes all of these mobile media queries since it includes files outside of this PR.

MinhxNguyen7:
> Currently Zotcourse handles their popups like so:

We could do well to have that nice graph, but I don't want to bother with changing up the graph implementation right now

> I think having the `Last Taught:` text would also be great!

Yeah I think that could be a pretty good feature, but I also don't want to implement that right now. I'll justify it by saying it's a little out of scope.

> Additionally, Zotcourse distinguishes course GPA vs instructor GPA which isn't super relevant here, but I do think a parenthesis after Grade Distribution like Grade Distribution (instructor) or Grade Distribution (course) could be nice. I personally wasn't initially sure why the individual GPAs were different from the GPA in the zotcourse button popup.

I think that's a pretty good idea

MinhxNguyen7:
@ap0nia review pls so I can finally be done with this

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
```diff
@@ -59,45 +59,33 @@ function renderEmptySelectValue() {
...
-                RightPaneStore.setActiveColumns(e.target.value);
-            }
-        },
-        [RightPaneStore.setActiveColumns]
-    );
+    const handleChange = (e: SelectChangeEvent<SectionTableColumn[]>) => {
+        if (typeof e.target.value !== 'string') {
+            setSelectedColumns(e.target.value);
+        }
+    };
```
ap0nia: For consistency, you can memoize this like the other handler functions.

```suggestion
    const handleChange = useCallback((e: SelectChangeEvent<SectionTableColumn[]>) => {
        if (typeof e.target.value !== 'string') {
            setSelectedColumns(e.target.value);
        }
    }, [])
```

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
```diff
@@ -59,45 +59,33 @@ function renderEmptySelectValue() {
...
+        () => SECTION_TABLE_COLUMNS.filter((_, index) => selectedColumns[index]),
+        [selectedColumns]
+    );
 
-    useEffect(() => {
-        RightPaneStore.on('columnChange', handleColumnChange);
+    const columnLabelEntries = useMemo(
+        () => Object.entries(columnLabels),
+        []
+    );
```
ap0nia: A better location for this might be at the top level of the module, where it's "automatically memoized" since it's separate from rendering logic.

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
```diff
@@ -109,19 +97,16 @@ export function ColumnToggleButton() {
...
-                    value={activeColumns}
+                    value={selectedColumnNames}
                     open={open}
                     onChange={handleChange}
                     onClose={handleClose}
                     renderValue={renderEmptySelectValue}
                     sx={{ visibility: 'hidden', position: 'absolute' }}
                 >
-                    {Object.entries(columnLabels).map(([column, label]) => (
+                    {columnLabelEntries.map(([column, label], idx) => (
```
ap0nia: I like either `index` or `i`

```suggestion
                    {columnLabelEntries.map(([column, label], index) => (
```

apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
```diff
@@ -10,109 +12,122 @@ import {
...
+                </Tooltip>
+            )}
+        </Box>
     );
+}
+
+function SectionTable(props: SectionTableProps) {
+    const { courseDetails, term, allowHighlight, scheduleNames, analyticsCategory } = props;
+
+    const { selectedColumns, getActiveColumns } = useColumnStore();
```
ap0nia: You can slice the store. A minor visual improvement is that the left side looks more like a typical useState call.

```suggestion
    const [ selectedColumns, getActiveColumns ] = useColumnStore(store => [store.selectedColumns, store.getActiveColumns]);
```

apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
```diff
@@ -10,109 +12,122 @@ import {
...
+    const encodedDept = useMemo(() => {
+        return encodeURIComponent(courseDetails.deptCode);
+    }, [courseDetails.deptCode]);
+
+    /**
+     * Limit table width to force side scrolling.
+     */
+    const tableMinWidth = useMemo(() => {
+        const width = isMobileScreen ? 600 : 780;
+        const numActiveColumns = getActiveColumns().length;
```
ap0nia: Because `selectedColumns` is being memoized it would be best to use it directly as part of the calculation (i.e. since it's guaranteed to be affecting when this is calculated). And if necessary, the column store should be refactored to accommodate this, i.e. by setting the `selectedColumns` to the desired structure directly, and/or creating separate data structures to hold slightly differently formatted info.

```suggestion
        const numActiveColumns = selectedColumns.length;
```

apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
```diff
@@ -152,49 +168,29 @@ const SectionTable = (props: SectionTableProps) => {
...
-                                                    )}
-                                                </div>
-                                            )}
-                                        </TableCell>
-                                    );
-                                })}
+                            <TableCell padding="none" />
+
+                            {tableHeaderColumnEntries
+                                .filter(([column]) => getActiveColumns().includes(column as SectionTableColumn))
```
ap0nia: Same idea here. I don't really like calling "getter" functions explicitly because they should ideally be calculated and captured in state. The state variable should properly react to changes (which is a given when using Zustand) and the component will update accordingly. A raw function call can't do that, and it's just run on every render. Which means that it can miss updates or over-render depending on the component's other state variables.

```suggestion
                                .filter(([column]) => selectedColumns.includes(column as SectionTableColumn))
```

apps/antalmanac/src/stores/ColumnStore.ts
```diff
@@ -0,0 +1,84 @@
...
+}
+
+/**
+ * Store of columns that are currently being displayed in the search results.
+ */
+export const useColumnStore = create<ColumnStore>((set, get) => {
+    return {
+        activeColumns: [...SECTION_TABLE_COLUMNS],
+        enabledColumns: SECTION_TABLE_COLUMNS.map(() => true),
+        selectedColumns: SECTION_TABLE_COLUMNS.map(() => true),
```
ap0nia: It looks like all these properties start off as copies of the array, so you can either use `Array.from` or the spread operator to shallow copy them.

```suggestion
        activeColumns: Array.from(SECTION_TABLE_COLUMNS),
        enabledColumns: Array.from(SECTION_TABLE_COLUMNS),
        selectedColumns: Array.from(SECTION_TABLE_COLUMNS),
```

apps/antalmanac/src/stores/ColumnStore.ts
```diff
@@ -0,0 +1,84 @@
...
+        setSelectedColumns: (columns: SectionTableColumn[]) => {
+            set(() => {
+                const selectedColumns = SECTION_TABLE_COLUMNS.map((column) => columns.includes(column));
+                return { selectedColumns: selectedColumns };
+            });
+            logAnalytics({
+                category: analyticsEnum.classSearch.title,
+                action: analyticsEnum.classSearch.actions.TOGGLE_COLUMNS,
+            });
+        },
```
ap0nia: How come `getActiveColumns` is needed? Could `setSelectedColumns` not set the `selectedColumns` property to this calculated value? Another property could also be added to capture this info if needed.

Otherwise, the best implementation of this particular property would be as a [derived store](https://github.com/pmndrs/zustand/issues/108). It should always be in sync with some other value in the store, so it would be best to either commit to deriving it, or just setting it at the same time as the other values when they change.

ap0nia (changes requested):
Everything looks very clean for the most part ✨ 

The column store and its usage can be tweaked to improve the reactivity.

apps/antalmanac/src/stores/ColumnStore.ts
MinhxNguyen7: The latter two aren't copies.

apps/antalmanac/src/stores/ColumnStore.ts
MinhxNguyen7: Isn't this the closest you can get to a derived store? Zustand doesn't seem to natively support it.

ap0nia: You could make a property like `currentlyActiveColumns` and then also set that property whenever `setSelectedColumns` or `setColumnEnabled` is called. 

Derived state is automatically updated whenever other state is changed; invoking a function to get the value isn't the same because when multiple components need the derived value, they'll need to individually re-calculate it instead of the value being readily available. It also makes building dependency arrays less clear because functions can't be memo-ized, and it may not be immediately obvious to memo-ize with the two arrays in the store.

MinhxNguyen7: Yeah I think the double update is the way to go, yeah. I didn't want to do that because it violates single-source of truth, but that's better than forcing the consumer to to update when `setSelectedColumns` or `setColumnEnabled` is called.

apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
```diff
@@ -124,7 +121,7 @@ function SectionTable(props: SectionTableProps) {
      */
     const tableMinWidth = useMemo(() => {
         const width = isMobileScreen ? 600 : 780;
-        const numActiveColumns = getActiveColumns().length;
+        const numActiveColumns = activeColumns.length;
         return (width * numActiveColumns) / TOTAL_NUM_COLUMNS;
     }, [isMobileScreen, selectedColumns]);
```
ap0nia: Don't forget to include `activeColumns` in the dependency array.

ap0nia (approved):
Perfect! 💯

PR #674, opened 2023-08-26
UI: Move Calendar Exports to AppBar

KevinWu098:
## The Issue
Screenshot 2023-08-26 at 1 11 24 AM

Having `Download`, `Screenshot`, and `Add Custom` as text within buttons makes resizing the Calendar pretty unwieldy. 

Additionally, Export feels like it makes more sense being housed in the App Bar alongside `Import` and `Load` (but I do understand the reason for why it was housed within the Calendar).

<sub> I probably should've opened a issue first, but I got carried away with just seeing what it might look like and ended up just building out the whole thing </sub>

## Summary
1.  Move download and screenshot into `Export` in the App Bar
2. Change `Add Custom` button to just an Icon Button

Desktop Demo

Mobile Demo

## Test Plan
- Screenshot and download still work
- Changes are all UI/UX based, so it comes down to what makes sense for users

## Issues
Closes #

## Thoughts and Misc
1. Tweaked the padding on mobile of the Menu Icon to let it breathe a little more
2. Pretty unrelated, but the App Bar does get pretty cramped at smaller tablet sizes, but the Logo resizes to allow it to fit (*up until the mobile breakpoint). 
Cramped App Bar

3. The one reservation I have about this change is mainly that the "Custom" part of "Add Custom Events" isn't particularly well conveyed by just a simple Add icon. Some solutions might be to make the Add icon a variant that's a little more pronounced (such as a Plus in a Circle, etc), hinting to the user that it's not simply a basic Add Button.


ap0nia:
Cool, I can help with this PR. The entire AppBar was supposed to be refactored from scratch, but I never got around to pushing my changes into the main branch. Just request a review whenever you're ready and I'll take a look.

The `AppBar` should be renamed to `Header`. Here's a reference of [how it should actually be built](https://github.com/icssc/AntAlmanac/blob/aponia/src/components/Header/Header.tsx). 

My refactor also addresses a UI issue on the live website where hovering over a button in the mobile menu causes a double outline (this is because it currently nests a button inside of a component that was already styled to be a button). 
image

apps/antalmanac/src/components/AppBar/NotificationSnackbar.tsx
KevinWu098, line 1: What happened to this component? I don't think it's in the Header anymore

apps/antalmanac/src/routes/Home.tsx
```diff
@@ -12,13 +12,17 @@ import NotificationSnackbar from '$components/AppBar/NotificationSnackbar';
...
             <CssBaseline />
+
             <PatchNotes />
-            <Bar />
+
+            <Header />
+
             {isMobileScreen ? (
                 <MobileHome />
             ) : (
```
KevinWu098: Not sure why, but the app isn't seamlessly swapping between Mobile and Desktop views anymore when the schedule has unsaved courses.

Resizing to Mobile creates warning

KevinWu098 (commented):
Looks pretty darn solid, but there are two main issues:
1. Snackbar got obliterated and wasn't added back into `Header`
2. Resizing the desktop app now creates a warning when swapping between desktop and mobile

apps/antalmanac/src/components/buttons/Download.tsx
```diff
@@ -0,0 +1,59 @@
...
+            ref.current.download = 'schedule.ics';
+            ref.current.click();
+
+            enqueueSnackbar('Schedule downloaded!', { variant: 'success' });
+        });
+    }, [enqueueSnackbar]);
+
+    return (
+        <>
+            <Tooltip title="Download Schedule (.ics)">
```
KevinWu098: Text and Icon color is blue here, as well as for `Screenshot.tsx` and isn't readable on Dark Mode.

Screenshot 2023-09-01 at 2 33 35 PM

apps/antalmanac/src/components/buttons/Screenshot.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+        ref.current.href = canvas.toDataURL('image/png');
+        ref.current.download = 'Schedule.png';
+        ref.current.click();
+    };
+
+    return (
+        <>
+            <Tooltip title="Get a screenshot of your schedule">
+                {iconOnly ? (
```
KevinWu098: Same styling issue as here: https://github.com/icssc/AntAlmanac/pull/674/files#r1313549071

apps/antalmanac/src/components/buttons/Download.tsx
ap0nia: Oh, yeah the component design is correct; it's supposed to be a minimal abstraction over the original `Button` component. But I just forgot to apply `color="inherit"` when I use them.

apps/antalmanac/src/routes/Home.tsx
ap0nia: It was switching between two components and both were loading the schedule automatically when mounting. I created a store to keep track of the first time the app ever attempts to load the schedule automatically, and check that before automatically loading.

apps/antalmanac/src/components/buttons/Screenshot.tsx
ap0nia: This should be resolved by applying `color="inherit"` where the screenshot button is used.

apps/antalmanac/src/components/AppBar/NotificationSnackbar.tsx
ap0nia: I accidentally obliterated it because some genius put it in the `AppBar` folder despite not being related to the previous header at all

EricPedley:
UI/UX comment: I agree that the add custom event just being a plus sign is kinda confusing, and I think the strat might be to move it to the class search tab in the right pane instead of where it is right now, and bring back the full "Add Custom" text to make it obvious. But I don't want to scope creep this PR any further than it already is with the refactor and everything, so lmk what you think and I can create an issue for it.

EricPedley:
Also, respectfully, can we aim to keep PR's more limited in scope in the future? Like, moving the exports to the app bar should be a completely separate concern from _a total refactor of the entire component_, so we can merge in the feature change super quick then spend more time worrying about the refactor without it blocking the UI/UX change.

ap0nia:
> Also, respectfully, can we aim to keep PR's more limited in scope in the future?

Sorry, the header's been bothering me for a while and I finally got the chance to scratch that itch 😅 

Idk if the devops PR can be easily split up either, since all the tooling is just kind of interconnected; there are definitely files that can be pruned from the diff, but I just want to include them so I don't have to go back :^) . 

I'll work on splitting up the tRPC and auth PR though

apps/antalmanac/src/components/Header/About.tsx
```diff
@@ -0,0 +1,63 @@
+import { useCallback, useState } from 'react';
```
EricPedley: Which of these files are just renamed and which had their contents change?

apps/antalmanac/src/components/PatchNotes.tsx
```diff
@@ -1,73 +1,72 @@
-import { Button, Dialog, DialogActions, DialogContent, DialogContentText, DialogTitle } from '@material-ui/core';
-import { useEffect, useState } from 'react';
+import { useCallback, useState } from 'react';
+import { Button, Dialog, DialogActions, DialogContent, DialogTitle } from '@mui/material';
+import { MuiMarkdown } from '$components/MuiMarkdown';
```
EricPedley: I like this change but would've been easier if it was it's own PR. Like if you opened a PR that _only_ included this markdown rendering for the patch notes, it would be reviewed and merged in like 10 minutes instead of needing to wait for the rest of this stuff to get reviewed.

EricPedley:
> > Also, respectfully, can we aim to keep PR's more limited in scope in the future?
> 
> Sorry, the header's been bothering me for a while and I finally got the chance to scratch that itch 😅
> 
> Idk if the devops PR can be easily split up either, since all the tooling is just kind of interconnected; there are definitely files that can be pruned from the diff, but I just want to include them so I don't have to go back :^) .
> 
> I'll work on splitting up the tRPC and auth PR though

Yeah I'm not saying don't refactor stuff, I'm just saying that you should try to keep your PR's as small as possible to make them easier to review, and try to refrain from tacking on changes that increase the scope of existing PR's.

apps/antalmanac/src/components/Header/About.tsx
EricPedley: Btw I'm putting the rest of the review on hold until this gets answered because knowing this will make it significantly easier.

ap0nia: All of them had their contents changed lol. They've all been migrated to MUIv5 and are written in "idiomatic" React functional components (i.e. with a stupid amount of memo-ization)

apps/antalmanac/src/components/PatchNotes.tsx
ap0nia: Not really a defense: I didn't know this was going to be in the PR, I just Googled it and thought it would be fun to add lol

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
```diff
@@ -11,6 +11,7 @@ import { Calendar, DateLocalizer, momentLocalizer, Views } from 'react-big-calen
 import CalendarToolbar from './CalendarToolbar';
 import CourseCalendarEvent, { CalendarEvent } from './CourseCalendarEvent';
 import AppStore from '$stores/AppStore';
+import { screenshotElementId } from '$lib/constants';
```
ap0nia: We refer to the calendar element by ID to take the screenshot, so it seems like a good idea to set it as a constant.

A future TODO might be to prop drill the calendar's ref to the screenshot button somehow; working with refs is more "idiomatic" in React, but it could increase component complexity for trivial benefits (classic React :tm:)

apps/antalmanac/src/components/Header/About.tsx
ap0nia, line 1: In order to accommodate whether the buttons is going to be a `MenuItem` or just a regular `Button`, I've just made two separate components with similar logic for all the header buttons. The only thing they're supposed to do is encapsulate the logic for opening/closing the dialogs.

There's not really an easy way to initialize `const [open, setOpen] = useState(false)` without creating a component that can manage that itself (there are ways, but I feel that they touch the lines of advanced React shenanigans, and I don't want to deal with it)

apps/antalmanac/src/components/PatchNotes.tsx
EricPedley: Eh, it's fine this time, but in the future if you just checkout back to main before adding random stuff like this, it'll get merged in way faster.

apps/antalmanac/src/components/Header/Header.tsx
```diff
@@ -0,0 +1,86 @@
...
+            position="static"
+            color="primary"
+            sx={{ backgroundColor: theme.palette.primary.main, backgroundImage: 'none' }}
+        >
+            <Toolbar variant="dense">
+                {isMobileScreen ? <MobileLogo height={32} /> : <Logo height={32} />}
+
+                <Box flexGrow={1} />
+
+                {isMobileScreen ? (
```
ap0nia: When it's a mobile screen, create a menu and nest all of the respective menu items inside of it. I've chosen to **not** use an array for these cases because it's more legible as plain HTML/JSX. And I don't have to deal with adding a key to each component in the array.

apps/antalmanac/src/components/Header/News.tsx
```diff
@@ -0,0 +1,153 @@
...
+interface NewsPopoverProps {
+    anchorEl: Element | undefined;
+    setAnchorEl: React.Dispatch<React.SetStateAction<Element | undefined>>;
+    setShowDot: React.Dispatch<React.SetStateAction<boolean>>;
+}
+
+function NewsPopover(props: NewsPopoverProps) {
+    const { anchorEl, setAnchorEl, setShowDot } = props;
+
+    const query = trpc.news.findAll.useQuery();
```
ap0nia: A future TODO is to change this to an infinite query, and have it automatically load more news items as the user scrolls

```suggestion
    const query = trpc.news.findAll.useInfiniteQuery();
```

apps/antalmanac/src/components/Header/Settings.tsx
```diff
@@ -0,0 +1,142 @@
...
+import analyticsEnum, { logAnalytics } from '$lib/analytics';
+
+interface SettingsPopoverProps {
+    colorScheme: string;
+    anchorEl: Element | undefined;
+    setAnchorEl: React.Dispatch<React.SetStateAction<Element | undefined>>;
+}
+
+function SettingsPopover(props: SettingsPopoverProps) {
+    const { colorScheme, anchorEl, setAnchorEl } = props;
```
ap0nia: Technically, this component could manage its own subscription to the `AppStore` as well, and get its theme directly. But it's a lot more boilerplate with the callbacks and state stuff, and I'm ok with just having the parent component pass down this information, and the child reacting to it.

apps/antalmanac/src/components/NotificationSnackbar.tsx
```diff
@@ -52,9 +52,6 @@ class NotificationSnackbar extends PureComponent<NotificationSnackbarProps> {
     openSnackbar = () => {
         this.props.enqueueSnackbar(AppStore.getSnackbarMessage(), {
             variant: AppStore.getSnackbarVariant(),
-            // shitty hack because notistack says it doesn't support `duration`, but this still runs without errors 🤷‍♂️
-            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-            // @ts-ignore
```
ap0nia: This was resolved with declaration merging in the new `src/global.d.ts`. Basically, if a library declares interfaces, you can do [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) to override or extend their exported interfaces. It's dumb, but is a way to monkey patch incorrect type definitions.

apps/antalmanac/src/components/buttons/LoadSchedule.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+export function LoadScheduleButton(props: ButtonProps = {}) {
+    const [loadedSchedule, setLoadedSchedule] = useLoadingStore((state) => [
+        state.loadedSchedule,
+        state.setLoadedSchedule,
+    ]);
+
+    const [open, setOpen] = useState(false);
+
+    const [rememberMe, setRememberMe] = useState(true);
```
ap0nia: TODO is to use a form library that can manage the states of the inputs for us, instead of declaring a new `useState` for every single input.

apps/antalmanac/src/components/buttons/LoadSchedule.tsx
```diff
@@ -0,0 +1,140 @@
...
+        const userId = getSavedUserId();
+
+        if (userId !== null) {
+            handleLoadSchedule(userId, rememberMe);
+        }
+
+        setLoadedSchedule(true);
+
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [loadedSchedule]);
```
ap0nia: eslint wants us to include all stateful variables, i.e. `userId, rememberMe`, but we don't really want `useEffect` to run in those cases, since only the first app load matters. Technically we could include the dependencies and run it, and it would be expected to exit immediately (i.e. after the first type guard, basically acting like a noop)

apps/antalmanac/src/components/buttons/SaveSchedule.tsx
ap0nia, line 1: Previously, `LoadSchedule` and `SaveSchedule` buttons used to be the same component, but I'm going to split them up even though they have similar logic. This is mainly because I don't want the component to perform different functions based on a boolean flag. i.e. I consider `LoadSchedule` and `SaveSchedule` to have similar logic by coincidence, not by conscious choice.

apps/antalmanac/src/components/buttons/Screenshot.tsx
```diff
@@ -0,0 +1,71 @@
...
+        });
+
+        const canvas = await html2canvas(elementToScreenshot, {
+            scale: 2.5,
+            backgroundColor: theme.palette.background.paper,
+        });
+
+        ref.current.href = canvas.toDataURL('image/png');
+        ref.current.download = 'Schedule.png';
+        ref.current.click();
```
ap0nia: We can remove our dependency on the `saveAs` library now too :nerd:

apps/antalmanac/src/components/dialogs/About.tsx
```diff
@@ -0,0 +1,58 @@
...
+Projects Committee and built by students from the UCI community.
+
+<a target="_blank" href="https://github.com/icssc/AntAlmanac/contributors">
+    <img
+        src="https://contrib.rocks/image?repo=icssc/antalmanac"
+        width='100%'
+        alt="AntAlmanac Contributors"
+    />
+</a>
+`;
```
ap0nia: A neat DX improvement because of this is that we can store our patch notes as markdown files somewhere, and have a process of importing them as raw text into the app. Then we can just render each patch notes with the markdown component

apps/antalmanac/src/components/dialogs/Import.tsx
```diff
@@ -0,0 +1,264 @@
...
+        const sectionsAdded = addCoursesMultiple(courseInfo, term, currentScheduleIndex);
+
+        logAnalytics({
+            category: analyticsEnum.nav.title,
+            action: analyticsEnum.nav.actions.IMPORT_STUDY_LIST,
+            value: sectionsAdded / (sectionCodes.length || 1),
+        });
+
+        if (sectionsAdded === sectionCodes.length) {
+            enqueueSnackbar(`Successfully imported ${sectionsAdded} of ${sectionsAdded} classes!`, {
```
ap0nia: We should try to migrate away from the `AppStore`'s notification handler, and just use `enqueueSnackbar` directly for simplicity.

apps/antalmanac/src/lib/trpc.ts
ap0nia, line 1: This is a tRPC client that's integrated with React-Query, which enables server-side state management.

apps/antalmanac/src/routes/Home.tsx
```diff
@@ -1,28 +1,36 @@
...
 import DesktopTabs from '$components/RightPane/RightPaneRoot';
-import NotificationSnackbar from '$components/AppBar/NotificationSnackbar';
+import NotificationSnackbar from '$components/NotificationSnackbar';
 
 export default function Home() {
     const isMobileScreen = useMediaQuery('(max-width: 750px)');
+
     const theme = useTheme();
 
+    const backgroundColor = useMemo(() => theme.palette.primary.main, [theme.palette.primary.main]);
```
ap0nia: Made the color of the split pane responsive.

apps/antalmanac/src/vite-env.d.ts
ap0nia, line 1: Just formatted the JSDoc of this file.

apps/backend/src/routers/news.ts
```diff
@@ -1,5 +1,5 @@
 import { router, procedure } from '../trpc';
-import NewsModel, { News } from '$models/News';
+import NewsModel, { News } from '../models/News';
```
ap0nia: Long story short, aliases will cause TS to break when importing the server's type definition into the frontend, since the frontend's tsconfig doesn't have these aliases.

apps/antalmanac/src/global.d.ts
ap0nia, line 1: Override the interfaces provided by notistack because they're wrong :^)

apps/antalmanac/src/components/Header/index.tsx
ap0nia, line 1: Barrel export for convenience, so you can do
`import Header from '$components/Header'` instead of `import Header from '$components/Header/Header'`

KevinWu098:
> UI/UX comment: I agree that the add custom event just being a plus sign is kinda confusing, and I think the strat might be to move it to the class search tab in the right pane instead of where it is right now, and bring back the full "Add Custom" text to make it obvious. But I don't want to scope creep this PR any further than it already is with the refactor and everything, so lmk what you think and I can create an issue for it.

I think moving the Custom Event button to the Search tab would cause too large of a UI/UX disconnect, so I've opted instead to revert it's design back to its previous `Button` form.

Screenshot 2023-09-05 at 8 56 37 PM

Previous upgrades to the toolbar now allows the components to wrap, making it look pretty nice on mobile screens as well:

chrome-capture-2023-8-5

ap0nia:
Oh, yeah I added flexbox to the toolbar during the refactor. 🤓

KevinWu098:
I know we're a tad far along into this PR, but I'm pretty strongly in favor of some sort of rebase (split?, sorry not sure of terminology) back to commit #3f1c3ca (pre-refactor) and moving the 2 or 3 other scopes (Appbar/Header, Markdown, etc) included within this PR into their own PRs.

My reasoning mainly being that:
1. We all agree this PR is **too big** as described above
2. Styling goes somewhat out of whack when upgrading from MUI 4 -> 5, particularly around the light mode styling of AppBar/Header and, **particularly befuddling**, the links in `sectionTableBody.tsx` ⁉️

@EricPedley @ap0nia Thoughts?

ap0nia:
> I know we're a tad far along into this PR, but I'm pretty strongly in favor of some sort of rebase (split?, sorry not sure of terminology) back to commit #3f1c3ca (pre-refactor) and moving the 2 or 3 other scopes (Appbar/Header, Markdown, etc) included within this PR into their own PRs.

Sure let's do that so we can get it merged faster

KevinWu098:
Not sure why Github-Actions isn't deploying a new staging instance @ap0nia @EricPedley

apps/antalmanac/src/components/AppBar/Export.tsx
```diff
@@ -0,0 +1,91 @@
...
+    betaBadge: { transform: 'scale(1) translate(95%, -50%)' },
+};
+
+interface SettingsState {
+    anchorEl?: HTMLElement;
+    theme: string;
+    screenshotting: boolean;
+}
+
+class SettingsMenu extends PureComponent<{ classes: ClassNameMap }, SettingsState> {
```
EricPedley: Why is this class named `SettingsMenu`? 👁️👄👁️

apps/antalmanac/src/components/AppBar/Export.tsx
```diff
@@ -0,0 +1,91 @@
...
+        flexDirection: 'column',
+        gap: '0.5rem',
+    },
+    betaBadge: { transform: 'scale(1) translate(95%, -50%)' },
+};
+
+interface SettingsState {
+    anchorEl?: HTMLElement;
+    theme: string;
+    screenshotting: boolean;
```
EricPedley: Why do we need to store this `screenshotting` state? Where are we using it?

EricPedley (changes requested):

EricPedley:
image

Me when short PR

apps/antalmanac/src/components/AppBar/Export.tsx
KevinWu098: 💀 Relic of when I copied over `SettingsMenu.tsx` It will be dealt with 🫡

apps/antalmanac/src/components/AppBar/Export.tsx
KevinWu098: Yeah, I genuinely have no recollection of writing this. It is very very strange code. I'm obliterating it soon.

EricPedley (approved):
lgtm

PR #675, opened 2023-08-26
Rename tabs for descriptiveness and concision

MinhxNguyen7:
## Summary

Rename tab names for descriptiveness and concision.

"Class Search" -> "Search".
"Added Courses" -> "Added" (You can already see the naming inconsistencies that plague us).
"Search" -> "Classes" in mobile home. There's no reason to call a tab search when a sub-tab is the actual search tab. I feel less confident in my judgment on this one, but I am certain that the name needs to change

## Test Plan
See that it looks alright. Not much to test


KevinWu098:
My two cents are that I'm a fan!

ap0nia (approved):
LGTM :rocket:

PR #676, opened 2023-08-26
refactor: section table components

ap0nia:
# Summary
Refactored the components of the section table that are affected by the PR.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+charset = utf-8
+insert_final_newline = true
+end_of_line = lf
+indent_style = space
+indent_size = 4
+max_line_length = 100
--- a/apps/antalmanac/package.json
+++ b/apps/antalmanac/package.json
@@ -31,6 +31,7 @@
         "@mui/icons-material": "^5.11.0",
         "@mui/lab": "^5.0.0-alpha.118",
         "@mui/material": "^5.11.7",
+        "@mui/styles": "^5.14.6",
         "@packages/antalmanac-types": "*",
         "@react-leaflet/core": "^2.1.0",
         "@tanstack/react-query": "^4.24.4",
--- a/apps/antalmanac/package.json
+++ b/apps/antalmanac/package.json
@@ -94,7 +94,7 @@
         "gh-pages": "^5.0.0",
         "husky": "^8.0.3",
         "lint-staged": "^13.1.1",
-        "peterportal-api-next-types": "1.0.0-beta.2",
+        "peterportal-api-next-types": "1.0.0-rc.2.68.0",
         "prettier": "^2.8.4",
         "typescript": "^4.9.5",
         "vite": "^4.2.1",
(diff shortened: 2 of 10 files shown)


MinhxNguyen7 (changes requested):
The section table does feel somewhat faster after some rough and unscientific testing, which is definitely nice considering it does chug when you have a lot of classes in there,

I haven't looked at the code itself, but I noticed some issues with styling:
- [x] The table doesn't have a proper min-width to force side-scrolling, so when the right pane is narrow, everything is squished.
- [x] The colors and general style of the status column are gone.
- [x] The GPA numbers are tiny.
- [x] The scroll bar is back in the grades graph.
- [x] The times column's font size is extra large.
- [x] The table headers are inconsistently styled.

Just pull this up versus the base branch and open the same search and you'll see the discrepancies quite clearly. There might be other ones as well that I didn't notice. 

I'm sure you'll have so much fun with CSS.

ap0nia:
Ok, thanks for letting me know about the styling issues. I took a crack at them, let me know what you think :pray:

MinhxNguyen7:
The styling is still not quite right.

base
here

The GPA text is small and bolded (?), and the spacings are just not right. The text for "Enrollment" is not styled consistently.

If you take a look at how I did the styling, I actually didn't force the width for any cell except for the leftmost one, and I think it works pretty decently. I also made the minWidth of the table depend on the number of columns displayed [here](https://github.com/icssc/AntAlmanac/pull/671/files#:~:text=force%20side%20scrolling-,const%20tableMinWidth,-%3D). That is an important QOL feature on mobile because imagine you're looking at like three columns and you're still forced to scroll sideways.

IMO, there's no need to stay true to exactly how I styled the base PR or even how the app is currently styled, but it still needs to look decent, and it doesn't right now.

ap0nia:
Ok, sorry I made a typo when I was calculating the minWidth because I wasn't referring to the component's local state when I was memo-izing. If you don't set any widths, then the table columns can fall out of alignment between different sections, which is different from what is on prod. Is this what you want?

Staging 671: 
image

Prod:
image

It could also be helpful if you could give me some concrete feedback on some of the code I refactored, since I'm trying to reproduce what you wrote.

MinhxNguyen7:
We've voted to keep the columns aligned between tables. I'm putting it down here for the sake of posterity. As we agreed, in the future, we might add an option to disalign them in the settings menu to come. I'm putting it down here for the sake of posterity.

Considering the fact that this is bigger than the base PR, I strongly think that this should be a normal PR to `main` after `gpa` is merged. I'll review it as soon as this PR is ready to merge. 

I understand that you want me to give better feedback, but I honestly don't want to read through ~1800 lines of changes now and then again later. That's why I only comment on its shortcomings in terms of _user experience_. After all, that is what matters the most. In any case, I'm likely to say it looks good because you have a much better idea of what idiomatic code looks like. I can only spot bugs.

ap0nia:
I'll just go ahead and close this so it's easier to keep track of what we want merged directly to main. I'll keep the branch open so I can preserve the overall refactoring goals.

PR #677, opened 2023-08-27
Add Course Level Filtering, Restyle Advanced Search

### LINKED ISSUE #627: Course Level Selection (Parity a la Zotcourse)

KevinWu098:
## Summary
- Add Course Level filtering functionality. It's housed in the Advanced Search Options because it doesn't change the URL.
- I've also taken the liberty of moving the Online Classes label to `top` instead of the default (`right`), which fits more cleanly.

Course Level Demo
_It's pretty blurry in this GIF, but upon selecting Upper Division Only, returned results are only those between 100 and 199 :D_

-----------------
##  Advanced Search Styling Changes:

Before:
Before Advanced Search Layout

After:
After Advanced Search Layout

After Demo Gif

I'm a really big fan of the new look with it's left-sided alignment and consistent gap. My main gripe is that MUI doesn't support row-only gap, so the gaps vertically can be a tad much. You can see it happening in the after image to the Room element, which looks a little orphaned. _Worthy tradeoff in my opinion_

## Test Plan
- Make sure it works

## Issues
Closes #627, the successful sibling to #628

### DIFF
--- a/apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
+++ b/apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
@@ -227,6 +227,7 @@ class CourseRenderPane extends PureComponent<CourseRenderPaneProps, CourseRender
                 fullCourses: formData.coursesFull,
                 building: formData.building,
                 room: formData.room,
+                division: formData.division,
             };
 
             try {
--- a/apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch.tsx
+++ b/apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch.tsx
@@ -1,4 +1,5 @@
 import {
+    Box,
     Button,
     Collapse,
     FormControl,
@@ -19,16 +20,24 @@ import { ChangeEvent, PureComponent } from 'react';
 import RightPaneStore from '../../RightPaneStore';
 
 const styles: Styles<Theme, object> = {
+    fieldContainer: {
+        display: 'flex',
+        gap: '1.5rem',
+        flexWrap: 'wrap',
+        paddingLeft: '8px',
+        paddingRight: '8px',
+        marginBottom: '1rem',
+    },
     units: {
         width: '80px',
     },
     timePicker: {
         width: '130px',
     },
-    smallTextFields: {
-        display: 'flex',
-        justifyContent: 'space-around',
-        flexWrap: 'wrap',
+    onlineSwitch: {
+        margin: 0,
+        justifyContent: 'flex-end',
+        left: 0,
     },
 };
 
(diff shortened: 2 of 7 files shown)


MinhxNguyen7 (changes requested):
I haven't looked at the code yet, but I have some suggestions for the UI.

Some renaming for concision and taking up less space, especially on a mobile screen:
- "Course Level" -> "Level"
- "Any Course Division" -> "Any"
- "Lower Division Only" -> "Lower Division"
- "Upper Division Only" -> "Upper Division"
These shouldn't affect the comprehensibility of the feature because the removed words can be easily implied from context

The advanced search menu now looks pretty disorganized. It's especially hard to parse on mobile. I'd say that they should be more vertically organized or at least grouped more intentionally. You'll notice that it is the latter in the current version on mobile, especially.
image

The filter doesn't apply to the fuzzy search. I realized that this is consistent with the other fields, so this should be changed (or declared a feature) in a separate PR.

KevinWu098:
> I haven't looked at the code yet, but I have some suggestions for the UI.
> 
> Some renaming for concision and taking up less space, especially on a mobile screen:
> 
> * "Course Level" -> "Level"
> * "Any Course Division" -> "Any"
> * "Lower Division Only" -> "Lower Division"
> * "Upper Division Only" -> "Upper Division"
>   These shouldn't affect the comprehensibility of the feature because the removed words can be easily implied from context
> 
> The advanced search menu now looks pretty disorganized. It's especially hard to parse on mobile. I'd say that they should be more vertically organized or at least grouped more intentionally. You'll notice that it is the latter in the current version on mobile, especially. image
> 
> The filter doesn't apply to the fuzzy search. I realized that this is consistent with the other fields, so this should be changed (or declared a feature) in a separate PR.

I agree with you about the concision, but I do want to push back on `Course Level -> Level` and `Any Course Division -> Any`. 

- I think `Course Level -> Level` sacrifices too much readability for being concise. Compared to its next-door neighbors like `Online Classes Only` and `Class Full Option`, I think it's probably the least offensive of the three. 

  - Side note, but `Online Classes Only` could probably be renamed to `Online Only`... It's also somewhat out of place as a switch element amongst a few of Selects, but I don't have any ideas currently about making it feel more cohesive.

- This is more of a nitpick, but `Any` is just so much shorter than its counterparts and just makes it feel... unbalanced? This is probably a little over-subjective, but I think it's worth bringing up. 

In the meanwhile, I'll fiddle around with some divs and see if we can clean this up :)

KevinWu098:
I'm pretty happy with how it's been cleaned up by getting rid of space-around in favor of a left-side oriented wrapping using gap. LMK how it looks :D @MinhxNguyen7

MinhxNguyen7 (approved):
LGTM!

PR #678, opened 2023-08-27
Revert "Add support for new API time formats"

KevinWu098:
Reverts icssc/AntAlmanac#669

The staging instance is using the Dev DB and so it broke prod


ecxyzzy (approved):
lgtm

PR #679, opened 2023-08-27
Add support for new API time formats (#669)

KevinWu098:
## Important
Teehee the staging instance was using the Dev DB so it broke prod 😅. #678 reverted #669, and this is just bringing it back :)

Before merging, this block of code needs to be cleaned up:

```typescript
// PeterPortal API
export const PETERPORTAL_GRAPHQL_ENDPOINT = 'https://api-next.peterportal.org/v1/graphql';
export const PETERPORTAL_REST_ENDPOINT = 'https://api-next.peterportal.org/v1/rest';

// Testing API
export const PETERPORTAL_REST_ENDPOINT_68 = 'https://staging-68.api-next.peterportal.org/v1/rest';
export const PETERPORTAL_WEBSOC_ENDPOINT = `${PETERPORTAL_REST_ENDPOINT_68}/websoc`;
```
### Finals information in Calendar Popups now always includes the location(s)
Previously, if a course had a finals location(s) that differed from its regular meeting time, the Calendar Course Popup would note that. 

In my opinion, regardless if the location differs or not, I think it's better for the user to know with certainty where the finals location is. Or, to put it another way: A user could assume that the finals are in the same place, but they would probably prefer to be assured that it is.

Mobile Demo

Desktop Demo

### Finals Information logic is processed differently

 The API now passes it as an object, not a string, meaning that prior logic (simply using the returned string from API) needed to be changed to process the various props that are now returned.

## Test Plan
- Should be fine, but check staging :)

## Issues
Closes #


apps/antalmanac/src/lib/utils.ts
```diff
@@ -0,0 +1,30 @@
+export function notNull<T>(value: T): value is NonNullable<T> {
```
EricPedley, line 1: how is this not built in to the language 😭. Also, is the `!=` instead of `!==` on purpose here?

ap0nia: Yes, `!=` will match with `null` and `undefined`, which is synonymous for this case. `!==` will match only with `null`, but not `undefined` because JavaScript feature :tm:

And yeah, this should ideally be integrated with `Boolean` by JS. e.g. `arrayWithNulls.filter(Boolean)`
There's [this project](https://github.com/total-typescript/ts-reset) that aims to monkey patch JS's built-in types, but I'd caution against using it seriously in a big project because the type overrides are global.

KevinWu098:
Edit: Also included in the PR body now

------------
Previously, if a course had a finals location(s) that differed from its regular meeting time, the Calendar Course Popup would note that. 

In my opinion, regardless if the location differs or not, I think it's better for the user to know with certainty where the finals location is. Or, to put it another way: A user could assume that the finals are in the same place, but they would probably prefer to be assured that it is.

So in addition to the approved changes from #669, I've appended the location(s) to `finalExamString`.

Mobile Demo

Desktop Demo

.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 4
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
```
EricPedley: You don't need to change it on this PR, but it's better to put these extra changes in a separate PR instead of lumping them in. Count-intuitively it makes it faster to separate stuff because when the PR only has a single purpose it's easier to just figure out if it's accomplishing that one thing. Like if this editorconfig were it's own PR, it would take less than a minute to review because it's so simple.

apps/antalmanac/tests/calendarize-helpers.test.ts
```diff
@@ -0,0 +1,227 @@
+import { describe, test, expect } from 'vitest';
```
EricPedley, line 1: Massive, MASSIVE W for including unit tests in your PR. This is a historic moment for AntAlmanac. For a second I was having a hang-up on why we're using vitest because I was confusing Vue and Vite 🤦‍♂️🤦‍♂️🤦‍♂️. Anyways, if you read that before I edited this just ignore it.

Edit: I stand corrected, just realized you had already merged in a test before this PR. How did nobody make a big deal out of that or ask about it in the review for that PR??

apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
```diff
@@ -146,6 +165,21 @@ const CourseCalendarEvent = (props: CourseCalendarEventProps) => {
...
+        if (finalExam.examStatus == 'NO_FINAL') {
+            finalExamString = 'No Final';
+        } else if (finalExam.examStatus == 'TBA_FINAL') {
+            finalExamString = 'Final TBA';
+        } else {
+            if (finalExam.startTime && finalExam.endTime && finalExam.month && finalExam.bldg) {
+                const timeString = translate24To12HourTime(finalExam.startTime, finalExam.endTime);
+                const locationString = `at ${finalExam.bldg.join(', ')}`;
+
+                // prettier-ignore
```
EricPedley: why does this need a prettier-ignore?

apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
```diff
@@ -146,6 +165,21 @@ const CourseCalendarEvent = (props: CourseCalendarEventProps) => {
 
         const buildingId = locationIds[buildingName] ?? 69420;
 
+        let finalExamString = '';
```
EricPedley, line 168: Same sentiment as the previous comment, it's confusing why we're changing the final exam logic on a PR that's supposed to be about supporting a new API time format. I'm assuming this version of the API returns the final as an object instead of a string now, which is the reason for this change, but make sure to mention that in the PR description.

apps/antalmanac/tests/calendarize-helpers.test.ts
```diff
@@ -0,0 +1,227 @@
...
+        }
+    }
+
+    return finalsEventsInCalendar;
+}
+
+/**
+ * TODO: Remove this and replace with an array of expected values.
+ */
+export function oldClendarizeCustomEvents(currentCustomEvents: RepeatingCustomEvent[] = []): CustomEvent[] {
```
EricPedley: This function name seems like a typo.

apps/antalmanac/tests/calendarize-helpers.test.ts
```diff
@@ -0,0 +1,227 @@
+import { describe, test, expect } from 'vitest';
+import type { ScheduleCourse } from '@packages/antalmanac-types';
+import type { CourseEvent, CustomEvent } from '$components/Calendar/CourseCalendarEvent';
+import type { RepeatingCustomEvent } from '$components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog';
+import { calendarizeCourseEvents, calendarizeCustomEvents, calendarizeFinals } from '$stores/calendarizeHelpers';
+
+describe('calendarize-helpers', () => {
+    const courses: ScheduleCourse[] = [
+        {
+            courseComment: 'string',
```
EricPedley: Can you make these more obviously placeholders, like "foobar" or "placeholder1" or something? I was confused for a couple of minutes looking at this because I thought it was a janky type definition 💀

apps/antalmanac/tests/calendarize-helpers.test.ts
```diff
@@ -0,0 +1,227 @@
...
+        const newResult = calendarizeCustomEvents(customEvents);
+        const oldResult = oldClendarizeCustomEvents(customEvents);
+        expect(newResult).toStrictEqual(oldResult);
+    });
+});
+
+/**
+ * TODO: Remove this and replace with an array of expected values.
+ */
+export function oldCalendarizeCourseEvents(currentCourses: ScheduleCourse[] = []): CourseEvent[] {
```
EricPedley: Can you replace these with the array of expected values in this PR? Less code = better, and if you can already get these tests to pass then it should be pretty easy to just hard-code the results.

EricPedley (changes requested):
Seems functionally alright as far as I can tell, just have some comments to address. Oh yeah, also can you get rid of that bullet point list of commit messages in the PR description? I could directly edit it and delete them myself but I was wondering if you had a good reason to keep those there? IMO its just a waste of space.

apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
KevinWu098: Gotcha, I'll add it into the PR!

apps/antalmanac/tests/calendarize-helpers.test.ts
ap0nia: What? This concern literally makes no sense lol. Jest, Mocha, Jasmin, Vitest, etc. are all testing frameworks. They're not responsible for testing UI by themselves. You setup integrations and stuff with the framework of your choice. i.e. here's an example of react-testing-library https://www.robinwieruch.de/vitest-react-testing-library/

TLDR: this is like saying "but Vite is targeted for first-class support for Vue, so it might not be the best choice for React"

.editorconfig
ap0nia: Oh, they're not supposed to be different. I'm supposed to be copy/pasting the same stuff, but there's a diff for some reason :shrug:

Let me try copy/pasting better.

apps/antalmanac/tests/calendarize-helpers.test.ts
ap0nia: I'll remove the old functions altogether and replace them with the desired hardcoded values. I kept them around because I was literally checking for compatibility between my re-written version and the one in prod.

apps/antalmanac/tests/calendarize-helpers.test.ts
EricPedley: Yeah, I agree it made no sense. Embarrassing moment there 😳

apps/antalmanac/tests/calendarize-helpers.test.ts
ap0nia: Lol, I took copied the type definition and used it as the example; so yeah it basically is a janky type def. I'll replace the values with more random placeholders.

apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
KevinWu098: Prettier formats the string like so, which is not awesome

```typescript
  finalExamString = `${finalExam.dayOfWeek} ${MONTHS[finalExam.month]} ${
      finalExam.day
  } ${timeString} ${locationString}`;
```

apps/antalmanac/tests/calendarize-helpers.test.ts
ap0nia: Ok wait, actually. There's a lot more code lol. And some of the formatting is hard to capture. Like `finalExam` should be a duplicate of the original `finalExam` property, and `start` and `end` should be the original strings passed into `Date` objects. Idk, any thoughts? @EricPedley 

My concern is that if we wrote a transformer, i.e. "transform the old result to the new, expected result", we literally just did the same thing twice lol. But not doing that means there's a lot of awkward manual labor to verbalize what we want.

image

apps/antalmanac/tests/calendarize-helpers.test.ts
ap0nia: Imagine confusing names between JS frameworks. I can't imagine not knowing the difference between React and Preact, or Svelte and SvelteKit, or Vuepress and Vitepress. :pensive:

apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
ap0nia: I'm pretty sure it breaks on the braced property accessor, i.e. something like this
```ts
const finalExamString = `${finalExam.dayOfWeek} ${MONTHS[
  finalExam.month
]} ${ finalExam.day } ${timeString} ${locationString}`;
```

a simple workaround is just to declare a variable to hold it; prettier won't forcefully break up the variable unless it's able to do so without affecting the string, i.e. when you use braces

```ts
const finalExamMonth = MONTHS[finalExam.month]
const finalExamString = `${finalExam.dayOfWeek} ${finalExamMonth} ${finalExam.day} ${timeString} ${locationString}`;
```

EricPedley: ok, agreed that looks gross 🤮

KevinWu098: > ```ts
> const finalExamMonth = MONTHS[finalExam.month]
> ```

Yup. That was the issue. I've pushed your suggestion :)

apps/antalmanac/tests/calendarize-helpers.test.ts
ap0nia: > Edit: I stand corrected, just realized you had already merged in a test before this PR. How did nobody make a big deal out of that or ask about it in the review for that PR??

Oh, and it's because I added the tests and reviewed the same PR lol

apps/antalmanac/tests/calendarize-helpers.test.ts
EricPedley: ok but that isn't more "code" per se, it's an array of hard-coded values, which is orders of magnitude easier to read than an entire other function. IMO way better than leaving an old function in there.

EricPedley (approved):
lgtm 🚢 it

EricPedley:
@ap0nia needs your review because I added some commits to clean up stuff

README.md
```diff
@@ -73,6 +73,10 @@ If you ever need help, feel free to ask around on our [Discord server](https://d
    and the backend server at http://localhost:8080 (if started).
    As you make changes to the React application in `src`, those changes will be automatically reflected on the website locally.
 
+#### Running Tests
+1. go into `apps/antalmanac`
+2. `pnpm run test`
```
ap0nia, line 78: `pnpm test` also works, since pnpm will assumes "run" if there's no valid matching command.

Also, it would be good if we setup vitest from the root; vitest can be configured for workspaces with a `vitest.workspace.ts` config file. But that can be done in a different PR.

If we set it up from the root, then you can just do `pnpm test` from the root and it'll run all the tests in the monorepo.

.editorconfig
ap0nia, line 1: I think this should match with the one on main, idk why it's showing up as a diff

ap0nia:
Looks good! :rocket:

README.md
EricPedley: We can work this out in https://github.com/icssc/AntAlmanac/pull/682

KevinWu098:
> Seems functionally alright as far as I can tell, just have some comments to address. Oh yeah, also can you get rid of that bullet point list of commit messages in the PR description? I could directly edit it and delete them myself but I was wondering if you had a good reason to keep those there? IMO its just a waste of space.

Whoops sorry I just saw this comment; they just carried over from the reverting process and I just left it as is. 

I'll delete it rn

ap0nia (approved):
LGTM :rocket:

PR #681, opened 2023-08-31
refactor: schedule selector popover

ap0nia:
# Summary
Implemented the schedule selector properly with a popover component. Using a popover instead of an autocomplete or select component gives us more flexibility in customizing how it works in the future, i.e. with re-ordering drag-and-drop, etc.

## TODO (after merging)
Verify and remove the following files -- they should be extraneous.
This can be done as the "noop" commit after this is merged.

- The directory `apps/antalamanc/src/components/Calendar/Toolbar/EditSchedule`

### DIFF
--- a/apps/antalmanac/src/components/dialogs/AddSchedule.tsx
+++ b/apps/antalmanac/src/components/dialogs/AddSchedule.tsx
@@ -18,6 +18,11 @@ function AddScheduleDialog(props: ScheduleNameDialogProps) {
      */
     const { onKeyDown, ...dialogProps } = props;
 
+    /**
+     * This is destructured separately for memoization.
+     */
+    const { onClose } = props;
+
     const [scheduleNames, setScheduleNames] = useState(AppStore.getScheduleNames());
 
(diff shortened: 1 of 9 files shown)


apps/antalmanac/src/components/dialogs/AddSchedule.tsx
```diff
@@ -0,0 +1,104 @@
...
+    const { onKeyDown, ...dialogProps } = props;
+
+    /**
+     * This is destructured separately for memoization.
+     */
+    const { onClose } = props;
+
+    const [scheduleNames, setScheduleNames] = useState(AppStore.getScheduleNames());
+
+    const defaultScheduleName = useMemo(() => `Schedule ${scheduleNames.length + 1}`, [scheduleNames]);
```
KevinWu098: I think the useState and useMemo aren't playing nice in `AddSchedule` since we're not rerendering the popover

chrome-capture-2023-8-1 (2)

KevinWu098 (changes requested):
Check out the one comment about the Add Schedule `defaultScheduleName`, otherwise LGTM

PR #682, opened 2023-09-04
Add Vitest to GitHub Actions

EricPedley:
## Summary
Add a GitHub Actions check that runs all unit tests in the project.

To get this to work I had to delete the test npm action from our cdk package because it was making turbo fail and not actually doing anything

I copied some of the config necessary from this page: https://turbo.build/repo/docs/handbook/testing

## Test Plan
Verify that it runs successfully on this PR

## Issues
no related issue


turbo.json
```diff
@@ -25,6 +25,7 @@
       "dependsOn": [
         "lint"
       ]
-    }
+    },
+    "test": {}
```
ap0nia: No, remove this change. Turbo doesn't run the tests; only Vitest does.

apps/cdk/package.json
ap0nia, line 1: You  don't have to change this file either.

apps/antalmanac/package.json
```diff
@@ -18,7 +18,8 @@
         "deploy": "gh-pages -d build",
         "format": "prettier --write src",
         "lint": "eslint src",
-        "test": "vitest"
+        "test:watch": "vitest",
+        "test": "vitest run"
```
ap0nia: I think this is sufficient.

```suggestion
        "test": "vitest"
```

ap0nia (changes requested):
Hang on, let me revamp the devops setup for the repo. The setup should be similar to PeterPortal API's  https://github.com/icssc/peterportal-api-next/tree/main/.github/workflows

Also, we  don't use turborepo to run the tests, we should use vitest with their workspace configuration https://vitest.dev/guide/workspace.html

The reasoning is because all the sub-projects should be using vitest and the same root config; if they weren't then they would have separate test setups and that's why we would use turborepo to run their testing suites in parallel

turbo.json
EricPedley: This runs all test steps in the subpackages though, so if we ever wanted to add tests to the backend it would run both

EricPedley: We also get caching via doing it through Turbo

apps/cdk/package.json
EricPedley: Conditonal on keeping the tests run through Turbo, we do need to change this file. If we decide not to do that then yeah I guess we don't need to change this, but it wasn't doing anything in the first place so I don't think we need to keep it there.

apps/antalmanac/package.json
EricPedley: `vitest` runs interactively by default so if we have it in our CI then it'll block everything and never finish. This is what the Turbo handbook recommended.

turbo.json
ap0nia: Hmm, honestly I think I'd prefer opting out of using turborepo for this; I'd rather that all the tests ran all the time so that way we could always know that everything is working properly. Until we have enough tests for the caching to matter, I think it would be easier to do more work, but be more consistent.

apps/cdk/package.json
ap0nia: The jest stuff is part of the "cdk --init" command, so it's just put there by AWS people who wrote the template. Yeah, it doesn't really matter if it's there or not I guess :shrug:

package.json
```diff
@@ -10,14 +10,17 @@
     "build": "turbo build",
     "deploy:aa": "turbo deploy --filter=antalmanac",
     "deploy:cdk": "turbo deploy --filter=cdk",
-    "deploy": "turbo deploy"
+    "deploy": "turbo deploy",
+    "test": "vitest run",
+    "test:watch": "vitest"
```
ap0nia: For simplicity, I think I'll do this for the lightest abstraction; if you want to specifically invoke `vitest run`, then you can do `pnpm test run`

```suggestion
    "deploy": "turbo deploy",
    "test": "vitest"
```

apps/cdk/package.json
```diff
@@ -4,7 +4,6 @@
   "scripts": {
     "build": "tsc && cdk synth",
     "watch": "tsc -w",
-    "test": "jest",
```
ap0nia: I'm going to keep this line the same so that way the diff is smaller; I'm overhauling devops and project specs in a different PR to handle these sorts of small things.

apps/antalmanac/package.json
ap0nia: I'm going to use this command instead because it's the simplest; to invoke `vitest run`, use `pnpm test run`

```suggestion
        "test": "vitest"
```

.github/workflows/step_setup_and_build.yaml
```diff
@@ -54,6 +54,9 @@ jobs:
             echo "apiSubDomain=dev" >> $GITHUB_ENV
           fi
 
+      - name: test project
+        run: pnpm test
```
ap0nia: Generally, testing is a separate pipeline. i.e. It's possible to successfully build the project, but not pass all the tests. 

A general approach for coordinating this is to build and deploy the website in one pipeline so you can get immediate feedback and preview. Then run tests, type checking, quality control in a separate pipeline because these are issues that should be addressed by the developer as they get around to it.

 i.e. A separation of concerns exists between fixing the website and fixing the tests, and making them dependent on each other can slow down both processes.

ap0nia (changes requested):
I'm going to be doing a comprehensive refactor / cleanup of the project in #686, but I'll port a minimal set of necessary changes to this PR to get it merged first. I'll push any necessary changes directly to this branch.

ap0nia (approved):
I don't know why it's running ghost tests that don't exist on this branch.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,66 @@
...
+                API_SUB_DOMAIN: ${{ env.apiSubDomain }}
+
+                # Turborepo credentials.
+                TURBO_API: ${{ vars.TURBO_API }}
+                TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
+                TURBO_TEAM: "team_antalmanac"
+              run: pnpm build
+
+            - name: Test
+              run: echo "all tests passed!!" # pnpm test run
```
EricPedley: Why does this not do the real thing? Can I change it to actually running the tests??

ap0nia: Idk, you tell me lol
https://github.com/icssc/AntAlmanac/actions/runs/6086266424/job/16512262907
image

ap0nia: If you run `pnpm test` locally on this branch, there is no such test as `download-ics.test.ts`, this only exists on the main branch. And if you did put in the test on this branch, it runs perfectly fine locally, but not during CI/CD ??

Edit (clarification): When testing locally, there is no test `download-ics.test.ts` and everything passes. When it's run in CI/CD, it seems to pick up this test from the main branch (??) and then runs it, and then fails. If you added this test locally, the test command would still succeed. So why does CI/CD pick this test up ***and*** fail, when replicating these conditions locally works fine??

EricPedley: idk man, works on my machine :tm: without putting that test in. I wanna see for myself what happens when we put it through the CI/CD again.

ap0nia: That's literally exactly what I'm saying happens lol. It works locally, but not in CI/CD.

EricPedley: Mb I thought you meant that it failed locally when you said
> If you run pnpm test locally on this branch, there is no such test as download-ics.test.ts

as in it failed with an error message like "no such test as download..."
Anyways I think we should fix the root cause, which seems to be that the build gets cached from a different branch, before merging this in, otherwise we're gonna see this problem over and over on every new PR

ap0nia: Ok hang on. Bruh, we're using Node 16 https://github.com/icssc/AntAlmanac/blob/ci-tests/.github/workflows/test.yaml#L36; the current Node 18 (LTS) is the minimum that polyfills fetch globally. Let me try using that instead

ap0nia: Ok, we're in the clear. I love devops 🥲

EricPedley: I figured out the problem.
image
d1423 is the current `main` HEAD, and `609292` is one of the commits on this branch, when you activated the CI last. When it checks out the branch, it automatically is merging into main.

EricPedley: Yeah I guess that would solve that test specifically failing in Actions, but now we also know why it's running tests that aren't on the branch the PR is based on.

ap0nia: Ok cool, you can feel free to clean up the Git history if you want (or push some other miscellaneous change). I can merge it once you're done with everything :+1:

EricPedley: Sounds good, I just pushed an update to the README and after the checks pass feel free to approve and merge.

ap0nia (approved):
Cool, I'll make sure this workflow gets implemented in the devops overhaul PR.

I'm gonna ask Fred to help review that one, and then the tRPC + auth PR will be built on top of the devops refactor, which you can review if you'd like.

PR #684, opened 2023-09-04
Fix Light Mode Coloring

KevinWu098:
## Summary
Schedule Dropdown, Finals Button, and Clear Schedule did not have the correct `isDarkMode()` ternary applied to them to give them proper light mode colors.

## Test Plan
Only styling changes, make sure everything is correctly viewable on light mode

## Issues
Closes #

### DIFF
--- a/apps/antalmanac/src/components/Calendar/CalendarToolbar.tsx
+++ b/apps/antalmanac/src/components/Calendar/CalendarToolbar.tsx
@@ -30,6 +30,7 @@ import RenameScheduleDialog from '$components/dialogs/RenameSchedule';
 import DeleteScheduleDialog from '$components/dialogs/DeleteSchedule';
 import analyticsEnum, { logAnalytics } from '$lib/analytics';
 import AppStore from '$stores/AppStore';
+import { isDarkMode } from '$lib/helpers';
 
 function handleScheduleChange(index: number) {
     logAnalytics({
@@ -55,7 +56,7 @@ function handleUndo() {
     });
     undoDelete(null);
 }
-    
+
 function handleClearSchedule() {
     if (window.confirm('Are you sure you want to clear this schedule?')) {
         clearSchedules();
@@ -188,7 +189,7 @@ function SelectSchedulePopover(props: { scheduleNames: string[] }) {
         <Box>
             <Button
                 size="small"
-                color="secondary"
+                color={isDarkMode() ? 'secondary' : 'inherit'}
                 variant="outlined"
                 onClick={handleClick}
                 sx={{ minWidth, maxWidth, justifyContent: 'space-between' }}
@@ -313,7 +314,7 @@ function CalendarPaneToolbar(props: CalendarPaneToolbarProps) {
                 <SelectSchedulePopover scheduleNames={scheduleNames} />
                 <Tooltip title="Toggle showing finals schedule">
                     <Button
-                        color={showFinalsSchedule ? 'primary' : 'secondary'}
+                        color={showFinalsSchedule ? 'primary' : isDarkMode() ? 'secondary' : 'inherit'}
                         variant={showFinalsSchedule ? 'contained' : 'outlined'}
                         onClick={handleToggleFinals}
                         size="small"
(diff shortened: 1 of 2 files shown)


apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -170,7 +171,11 @@ class AddedCoursePane extends PureComponent<AddedCoursePaneProps, AddedCoursePan
                         <PopupState variant="popover">
                             {(popupState) => (
                                 <>
-                                    <Button variant="outlined" {...bindTrigger(popupState)}>
+                                    <Button
+                                        variant="outlined"
+                                        color={isDarkMode() ? 'secondary' : 'inherit'}
```
ap0nia:
```suggestion
                                        color="inherit"
```

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -203,7 +208,7 @@ class AddedCoursePane extends PureComponent<AddedCoursePaneProps, AddedCoursePan
                         <Button
                             className={this.props.classes.clearSchedule}
                             variant="outlined"
-                            color="secondary"
+                            color={isDarkMode() ? 'secondary' : 'inherit'}
```
ap0nia:
```suggestion
                            color="inherit"
```

apps/antalmanac/src/components/Calendar/CalendarToolbar.tsx
```diff
@@ -313,7 +314,7 @@ function CalendarPaneToolbar(props: CalendarPaneToolbarProps) {
                 <SelectSchedulePopover scheduleNames={scheduleNames} />
                 <Tooltip title="Toggle showing finals schedule">
                     <Button
-                        color={showFinalsSchedule ? 'primary' : 'secondary'}
+                        color={showFinalsSchedule ? 'primary' : isDarkMode() ? 'secondary' : 'inherit'}
```
ap0nia:
```suggestion
                        color={showFinalsSchedule ? 'primary' : 'inherit'}
```

apps/antalmanac/src/components/Calendar/CalendarToolbar.tsx
```diff
@@ -188,7 +189,7 @@ function SelectSchedulePopover(props: { scheduleNames: string[] }) {
         <Box>
             <Button
                 size="small"
-                color="secondary"
+                color={isDarkMode() ? 'secondary' : 'inherit'}
```
ap0nia:
```suggestion
                color="inherit"
```

ap0nia (dismissed):
I'm pretty sure inherit covers the light/dark mode compatibility. Apparently "secondary" is very inconsistent with compatibility.

MinhxNguyen7 (changes requested):
The code looks fine, but the outline should be grey like the buttons on the right side. Let me know if you want to go to bed and do this later. I can merge this in the meantime

KevinWu098:
> The code looks fine, but the outline should be grey like the buttons on the right side. Let me know if you want to go to bed and do this later. I can merge this in the meantime

Ah, so I do agree on that point @MinhxNguyen7 , but the main discussion I had with Brian was that using `inherit` made for a much simpler coding experience when working with both modes. 

> I prefer not having to do complex teraries everytime we want to do compatible colors for different modes
> Just adjust the “inherit” color in the theme provider
 -- Brian

Any thoughts? I'm happy to do the ternaries for now, before we clean it up with a theme provider overhaul, but I'm also happy to just keep it as `inherit`.

ap0nia:
I think it's correct to use "inherit" as the default color for buttons when attempting to target light/dark mode compatibility. But by default, it's only black/white to contrast the current mode. I think the most correct solution would be to make the "inherit" the desired color (i.e. gray) in the theme provider and use it consistently.

Because this issue and other ones with theming just sort of exist intrisically in the app, i.e. dark blue colors in dark mode aren't very visible, I was thinking of opening an RFC and a dedicated PR addressing the theming. So I'm personally ok with inconsistent theming atm (since there's already some inconsistencies lol).

MinhxNguyen7 (approved):
Sounds good. Ship it

MinhxNguyen7:
@KevinWu098 I can't merge because someone needs to commit something before I can approve it for merging.

PR #685, opened 2023-09-04
Refactor .ics exports for new time formats

KevinWu098:
## Summary
Fixes the Export Calendar as .ics file feature. I forgor 💀 about it on #679

## Test Plan
Check that downloading the .ics file works
- [ ] Courses occur on the correct days of the week
- [ ] Courses occur at the right time
- [ ] Courses **begin** and **end** at the right times in the year (date-wise)
- [ ] Finals occur at the right time and date

## Issues
Closes #


ap0nia (changes requested):
Please write a test case for this.

apps/antalmanac/src/lib/download.ts
ap0nia, line 1: This file was re-formatted in #674 , but it looks like this PR will be merged first since it's smaller.

apps/antalmanac/src/stores/AppStore.ts
```diff
@@ -41,11 +41,13 @@ class AppStore extends EventEmitter {
             return theme === null ? 'auto' : theme;
         })();
 
-        window.addEventListener('beforeunload', (event) => {
-            if (this.unsavedChanges) {
-                event.returnValue = `Are you sure you want to leave? You have unsaved changes!`;
-            }
-        });
+        if (typeof window !== 'undefined') {
```
ap0nia, line 44: Vitest runs in Node, not the browser, so do this check so it doesn't break. Maybe TODO is to find a way to get Vitest to pretend it's running in the browser.

apps/antalmanac/src/components/Calendar/Toolbar/ExportCalendar.tsx
ap0nia, line 1: I moved this code to `$lib/download.ts` because most of it doesn't have any component logic.

apps/antalmanac/src/lib/download.ts
```diff
@@ -0,0 +1,318 @@
...
+    return [
+        [...date, ...classStartTime],
+        [...date, ...classEndTime],
+    ];
+}
+
+/**
+ * Get the start and end datetime of an exam
+ *
+ * @example ("Mon Jun 7 10:30-12:30pm", "2019") -> [[2019, 6, 7, 10, 30], [2019, 6, 7, 12, 30]]
```
KevinWu098: I believe this example should be WebsocFinalExam (the type) instead of the string

ap0nia: Ok, I made the example more comprehensive.

KevinWu098:
@ap0nia , I've made the unit test use Fall 2023 as its placeholder since `quarterStartDates` gets its data from a hardcoded array, so `placeholderTerm` would return `undefined`. 

Would it be more correct to add a placeholderTerm to `termData` so that this term runs correctly, or to keep the test case as is with `Fall 2023`?

ap0nia (approved):
Thanks for the patch, LGTM :pray:

PR #686, opened 2023-09-05
Refactor devops

ap0nia:
## Summary
Overhaul devops.


.github/workflows/deploy_production.yml
```diff
@@ -0,0 +1,123 @@
...
+            - name: Install dependencies
+              run: pnpm install --frozen-lockfile
+
+            - name: Build frontend
+              run: pnpm --filter "antalmanac" build
+
+            - name: Upload frontend artifacts
+              uses: actions/upload-pages-artifact@v2
+              with:
+                  path: apps/antalmanac/build
```
Voark: Feel like you forgot to do something here

.github/workflows/deploy_production.yml
```diff
@@ -0,0 +1,123 @@
...
+              run: pnpm install --frozen-lockfile
+
+            - name: Build backend
+              run: pnpm --filter "antalmanac-backend" build
+
+            - name: Deploy production backend CloudFormation stack
+              run: pnpm --filter "antalmanac-cdk" production-backend deploy
+
+    # A shadow/mirror of the production backend used for development.
+    deploy_production_shadow_backend:
```
Voark: Very nit: I think for the dev backend referring to it as dev backend makes more sense imo

.github/workflows/deploy_staging.yml
```diff
@@ -0,0 +1,135 @@
...
+    deploy_staging_frontend:
+        name: Deploy staging frontend CloudFormation stack and environment
+
+        needs: get_staging_configuration
+
+        runs-on: ubuntu-latest
+
+        environment:
+            name: staging-frontend-${{ github.event.pull_request.number }}
+            url: https://staging-${{ github.event.pull_request.number }}.antalmanac.com
```
Voark: What's the reason to use an environment and the comment staging url?

.gitignore
```diff
@@ -1,28 +1,50 @@
...
+.idea/workspace.xml
+.idea/usage.statistics.xml
+.idea/shelf
+
+## VSCode settings. Based on the following recommendation: https://stackoverflow.com/a/57749909
+.vscode/**/*
+!.vscode/settings.json
+!.vscode/tasks.json
+!.vscode/launch.json
+!.vscode/extensions.json
```
Voark, line 50: I feel like it's worth just ignoring all of these do people actually share settings like this through git

Voark: Also what was this generated by just curious

apps/antalmanac/package.json
```diff
@@ -3,21 +3,25 @@
     "version": "1.0.0",
     "private": true,
     "homepage": "https://antalmanac.com",
-    "description": "A course exploration and scheduling tool for UCI Anteaters",
+    "description": "the premiere course exploration and planning tool for UCI anteaters",
```
Voark, line 6: Why no caps 🤔

apps/cdk/src/legacy/app.ts
Voark, line 1: I feel like we can just remove all these legacy things if someone needs to reference they can always look in the git history

apps/cdk/src/constants.ts
```diff
@@ -0,0 +1 @@
+export const zoneName = 'antalmanac.com';
```
Voark, line 1: Nit but I think this can just be hardcoded in the two places its used because it will literally never ever change so dunno if it is worth extracting to a variable

.github/workflows/deploy_production.yml
```diff
@@ -0,0 +1,123 @@
...
+
+    # Turborepo credentials.
+    TURBO_API: ${{ vars.TURBO_API }}
+    TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
+    TURBO_TEAM: team_antalmanac
+
+    # AWS credentials.
+    AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+    AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+    AWS_DEFAULT_REGION: us-east-1
```
Voark: Prod is deployed to us-west-1
```suggestion
    AWS_DEFAULT_REGION: us-west-1
```

apps/cdk/src/app/production-backend.ts
```diff
@@ -0,0 +1,19 @@
+import { App } from 'aws-cdk-lib';
+
+import { BackendStack } from '../stacks/backend';
+import { waitForStackIdle } from '../wait-for-stack-idle';
+
+/**
+ * Deploy the production backend.
+ */
+async function main() {
+    const stackName = 'antalmanac-backend';
```
Voark: 1. nit: I feel like the name should have prod in it somewhere
2. not a nit: The name of our current prod backend (which is very much my fault but anyways) is `prod-us-west-1-Backend` which means that there is no way this stack will deploy as it isn't going to overwrite the current stack. The main problem is that the prod user data db is in that stack so we can't just delete it and redeploy. The database will need to be extracted from the stack no matter what before it is deleted, after that we can either try and import it with Cloudformation (very not fun) or figure out how to copy paste the data from one DynamoDB to another (uncertain how not fun I have never done this).

There is a similar issue with the dev backend but that can just be deleted as the data on there isn't really important.

Voark (changes requested):

.github/workflows/deploy_production.yml
ap0nia: No, this is correct for uploading to GitHub Pages if you publish with the [new GitHub Actions](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-with-a-custom-github-actions-workflow)

This is exactly what I've done at least [10 other times](https://github.com/Elysium-Everlasting/bronya.js/blob/main/.github/workflows/github-pages.yml)

.github/workflows/deploy_staging.yml
ap0nia: Using a deployment seems like the most correct way to indicate a deployment. Like I'm pretty sure commenting the staging URL is literally just a scuffed version of a real deployment. PPA's implemented deployments properly; you just click on "view deployment"
image

I could remove the commenting altogether unless anybody really liked that feature.

.gitignore
ap0nia: Stop reading the gitignore! :nerd_face: I wrote it all from scratch myself :sunglasses:

apps/antalmanac/package.json
ap0nia: Idk, I just looked through a lot of package.jsons and felt that they used really informal punctuation. Like no capitalization or period. I'd do both so it's formal, but idk :shrug:

.github/workflows/deploy_production.yml
ap0nia: Ok fine

apps/cdk/src/constants.ts
ap0nia: Oh so my motivation for this is literally, so the stacks look more similar lol. The `frontend` and `backend` stacks are structured very similarly, but the `zoneName` placement would be different in them, since it needs to be declared at different places. So I just put it in a separate file lol :nerd_face:

apps/cdk/src/app/production-backend.ts
ap0nia: Hmm: https://www.npmjs.com/package/copy-dynamodb-table :eyes:

.github/workflows/deploy_production.yml
Voark: I can't read I thought the last action was just uploading the artifacts for use in another step not upload to pages 🤡

.github/workflows/deploy_staging.yml
Voark: I did the commenting because I couldn't get the environments to delete before but also like I think we literally aren't using any feature of environments besides indicating deploys so I think it just ends up clogging stuff so I don't know if it's worth it 🤷. That being said, I can see it either way so I will just @EricPedley for his opinion.

apps/cdk/src/constants.ts
Voark: if anything I feel like you could've just extracted the entire zone component (not that that is really worth it either) but obviously not a big deal either way.

apps/cdk/src/app/production-backend.ts
Voark: Gonna leave this comment open as a reminder when merging this PR

ap0nia:
For my next trick, I will close this PR and see if the staging deploy gets obliterated.

ap0nia:
It didn't work! :sob:

ap0nia:
Will be closing because it's stale, but I'll keep the branch open as a reference.

PR #688, opened 2023-09-06
Fix helper function: removeDuplicateMeetings

KevinWu098:
## Summary
Overlooked `removeDuplicateMeetings` in `helpers.ts`, now it has the correct properties and types. I'm guessing that it wasn't caught in testing mainly because courses at the same time but at different buildings are pretty rare.

## Test Plan
Does everything work?

## Issues
Closes #

## Future Followup
On the topic of `removeDuplicateMeetings`, I'm wondering if this note still applies, or if the API now handles this issue.
>    //The data from the API will duplicate a section if it has multiple locations.
>    //I.e., if there's a Tuesday section in two different (probably adjoined) rooms,
>    //courses[i].sections[j].meetings will have two entries, despite it being the same section.
>    //For now, I'm correcting it with removeDuplicateMeetings, but the API should handle this


ap0nia (approved):
LGTM :rocket:

PR #690, opened 2023-09-08
fix: conflicting concurrency group with testing

ap0nia:
# Summary
The testing workflow has the exact same concurrency group as the staging PR workflow. Just make them distinct and both can deploy at the same time.


EricPedley (approved):
lgtm

PR #691, opened 2023-09-08
UI: Update Button Row in Added Pane

### LINKED ISSUE #664: Support Show/Hide Columns in Added Courses
## Problem
Show/Hide column functionality affects `SectionTable.tsx` and `SectionTableBody.tsx`, both of which are used by the Course Render Pane and the Added Course Pane.

A possible, common user-flow might be to:
1. Hide some columns, let's say **time** and **places**, while they're scrolling through
2. They add the classes they want
3. After, they look at Added Course Pane, but can't see time and place because it was hidden previously.
4. The user now has to go back to Course Render Pane, reenable **time** and **place**, then return to Added Course Pane.

Steps 3 and 4 are no bueno for the user experience.

Bad UX Demo

## Solution
**Added Course Pane should also support Showing/Hiding columns**

Additionally, Zotcourse supports hiding columns on their equivalent of Added Course Pane as well!

Screenshot 2023-08-21 at 1 58 01 AM

KevinWu098:
## Summary
The reasoning for this change is to both add Show/Hide functionality to Added Courses, but the Copy/Clear changes are so that visually Show/Hide is in the same position. (See below for a gif of what it would look like without the Copy/Clear changes)

1. Show/Hide Column button is now also accessible on the Added Course Pane, fixing the user-flow issue referenced in the original pull request
2. Copy Schedule and Clear Schedule have been turned into IconButton components.
3. Clean up styling
4. Copy Schedule will now also copy custom events

Before: 
Before

After:
After

Copy and Clear Demo:
Copy and Clear demo

Show/Hide Columns Demo:
Show/Hide Columns Demo

### Show/Hide
1. As described in issue #664, the lack of a Show/Hide column button in the Added Course Pane could/would create a poor user experience because they wouldn't be able to reenable (or disable) columns from the Added Course Pane that they had adjusted in Course Render Pane.
2. Adding it into Added Course Pane resolves the above issue, but creates its own visual issues as well (see below)...

### Moving Copy + Clear Schedules into IconButtons
1. The Show/Hide button initially worked fine on its own, but the lack of visual consistency between the Search and Added Course panes was poor because the Show/Hide button would've been 3rd on Course pane and 1st on Added pane.

Demo:
Demo of Visual Issue

2. By moving Copy & Clear, visual consistency can be maintained (and, in my opinion, improved) while also cleaning up the styling
3. Visual consistency is better with these three new "action" buttons at all time. But more specifically, the consistency is due to the Show/Hide button being able to maintain the same position on the screen.

## Test Plan
- [ ] Functional
- [ ] Functional on all devices

## Issues
Closes #664

### DIFF
--- a/apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
+++ b/apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
@@ -1,7 +1,21 @@
 import { useCallback, useEffect, useMemo, useState } from 'react';
 import PopupState, { bindMenu, bindTrigger } from 'material-ui-popup-state';
 
-import { Box, Button, Chip, Grid, Menu, MenuItem, Paper, TextField, Tooltip, Typography } from '@mui/material';
+import {
+    Box,
+    Button,
+    Chip,
+    Grid,
+    IconButton,
+    Menu,
+    MenuItem,
+    Paper,
+    SxProps,
+    TextField,
+    Tooltip,
+    Typography,
+} from '@mui/material';
+import { ContentCopy, DeleteOutline } from '@mui/icons-material';
 import { AACourse } from '@packages/antalmanac-types';
 
 import SectionTableLazyWrapper from '../SectionTable/SectionTableLazyWrapper';
@@ -10,6 +24,21 @@ import AppStore from '$stores/AppStore';
 import analyticsEnum, { logAnalytics } from '$lib/analytics';
 import { clearSchedules, copySchedule, updateScheduleNote } from '$actions/AppStoreActions';
 import { clickToCopy } from '$lib/helpers';
+
+/**
+ * All the interactive buttons have the same styles.
+ */
+const buttonSx: SxProps = {
+    backgroundColor: 'rgba(236, 236, 236, 1)',
+    marginRight: 1,
+    padding: 1.5,
+    boxShadow: '2',
+    color: 'black',
+    '&:hover': {
+        backgroundColor: 'grey',
+    },
+    pointerEvents: 'auto',
+};
 
 interface CourseWithTerm extends AACourse {
     term: string;
@@ -60,17 +89,23 @@ function getCourses() {
 }
 
 function handleClear() {
-    if (
-        window.confirm(
-            'Are you sure you want to clear this schedule? You cannot undo this action, but you can load your schedule again.'
-        )
-    ) {
+    if (window.confirm('Are you sure you want to clear this schedule?')) {
         clearSchedules();
         logAnalytics({
             category: analyticsEnum.addedClasses.title,
             action: analyticsEnum.addedClasses.actions.CLEAR_SCHEDULE,
         });
     }
+}
+
+function ClearScheduleButton() {
+    return (
+        <Tooltip title="Clear Schedule">
+            <IconButton sx={buttonSx} onClick={handleClear}>
+                <DeleteOutline />
+            </IconButton>
+        </Tooltip>
+    );
 }
 
 function createCopyHandler(index: number) {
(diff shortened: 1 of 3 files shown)


apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -73,12 +98,64 @@ function handleClear() {
...
+            </IconButton>
+        </Tooltip>
+    );
+}
+
 function createCopyHandler(index: number) {
     return () => {
         copySchedule(index);
     };
 }
```
ap0nia: Can you switch the order of the function declarations? I like to have helper functions in one area, and function components in another. It just helps me distinguish which parts of the file are responsible for rendering UI.

```suggestion
function createCopyHandler(index: number) {
    return () => {
        copySchedule(index);
    };
}

function ClearScheduleButton() {
    return (
        <Tooltip title="Clear Schedule">
            <IconButton sx={buttonSx} onClick={handleClear}>
                <DeleteOutline />
            </IconButton>
        </Tooltip>
    );
}
```

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -73,12 +98,64 @@ function handleClear() {
...
 
+function CopyScheduleButton() {
+    const [scheduleNames, setScheduleNames] = useState(AppStore.getScheduleNames());
+
+    useEffect(() => {
+        const handleScheduleNamesChange = () => {
+            setScheduleNames([...AppStore.getScheduleNames()]);
```
ap0nia, line 125: **maybe** we could leave a note here explaining why the array has to be shallow copied? I didn't include one when I wrote it, but the reason is because React's state setter won't re-render the component if the value is the same, and the store internally pushes to the same array, so we need to clone it here.

```suggestion
            /**
             * A shallow copy needs to be made so the array reference is different and the component re-renders.
             */
            setScheduleNames([...AppStore.getScheduleNames()]);
```

apps/antalmanac/src/stores/Schedules.ts
ap0nia, line 1: Could you make this a separate PR? I'm not entirely sure what the functionality of copying is intended to be, and it would be clearer with a PR like "copying schedules also copies custom events"

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
ap0nia, line 1: Since you're in this file, could you fix my mistake with the `buttonSx`? All those props should be applied to `IconButton`, not the `ToolTip`.

ap0nia (changes requested):
Looks good!

TLDR:
- swap two functions
- fix my styling mistakes
- adjust "copy" functionality in a different PR.

apps/antalmanac/src/stores/Schedules.ts
KevinWu098: Sure, I can move the copying changes to another PR. The intended functionality, according to Fred, is that it should also copy over custom events.

> I was never really a custom event user so I think what most likely happened is i just forgot about custom events when working on that method
> I think we don't need multiple methods, just the fixed one shoudl be fine ??!!

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
ap0nia, line 1: A future TODO for this component might be to break up the buttons like the other file you changed, and to stop prop-drilling the event handlers. But for this PR, these changes are good :fire:

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -215,61 +295,31 @@ function AddedSectionsGrid() {
...
-                        onClick={handleClear}
-                    >
-                        Clear Schedule
-                    </Button>
-                </Box>
+        <Box display="flex" flexDirection="column" gap={1}>
+            <Box display="flex" width={1}>
+                <CopyScheduleButton />
+                <ClearScheduleButton />
+                <ColumnToggleButton />
```
ap0nia, line 302: Very clean 🙌

ap0nia (approved):
Good work :rocket:

PR #692, opened 2023-09-09
Fix: Patch Notes now set Local Storage value

KevinWu098:
## Summary
The refactored patch notes in #569 didn't set the local storage value when the patch notes dialog was closed (slipped by me, I should've tested harder :')).

Fix Demo
Caption: Setting `latestPatchSeen` to `20230817` makes the dialog reappear and local storage to be set to `20230819` on refresh. 

1. onClick replaced onClose
2. Tests were added

## Test Plan
- [] Is local storage being set properly
- [] Dialog still closes fine
- [] Dialog doesn't reappear

## Issues
Closes #


ap0nia (changes requested):
Write the data to local storage when the "close" button is clicked. When they hit "close" it's like a confirmation that they've "read the patch notes". Also, would you like to try writing a test for this, or would you like me to write one?

### Mini-rant
In general, it's good to be mindful of what certain actions mean, and what's happening in the code in relation to what the user is doing. 

For example (a real one that some smooth brain committed to the codebase), if a user clicks on a button to load a schedule, and the dialog closes after this process is done ... You don't "load the schedule" in relation to when the dialog closes. You load the schedule when the button is clicked, because that's when the user expects something to happen; the dialog closing is a side effect and not indicative of what the user originally did. For your sanity as much as the user's, make things happen in the program when you'd expect them to happen as a user.

ap0nia:
Honestly, I'm pretty sure you can use `handleClick` for both. i.e. remove the original  `handleClose`, rename `handleClick` to `handleClose`, and use that as the `onClose` prop for the `Dialog`, and as the `onClick` prop for the `Button`.

KevinWu098:
> Write the data to local storage when the "close" button is clicked. When they hit "close" it's like a confirmation that they've "read the patch notes". Also, would you like to try writing a test for this, or would you like me to write one?
> 
> ### Mini-rant
> In general, it's good to be mindful of what certain actions mean, and what's happening in the code in relation to what the user is doing. 
> 
> For example (a real one that some smooth brain committed to the codebase), if a user clicks on a button to load a schedule, and the dialog closes after this process is done ... You don't "load the schedule" in relation to when the dialog closes. You load the schedule when the button is clicked, because that's when the user expects something to happen; the dialog closing is a side effect and not indicative of what the user originally did. For your sanity as much as the user's, make things happen in the program when you'd expect them to happen as a user.

It'll take me a little bit as I'm away from my laptop, but I'd like to attempt writing a tests for this! 

In regards to writing when it's closed, I'm unsure why handleClose, more specifically useCallback, isn't writing to local storage.

ap0nia:
That's because we literally forgot to put it there lol 
image

KevinWu098:
> That's because we literally forgot to put it there lol 
> 
> image

It was there originally:

image

apps/antalmanac/src/components/PatchNotes.tsx
```diff
@@ -22,20 +22,8 @@ function isOutdated() {
...
-            }
-        }) satisfies DialogProps['onClose'],
-        [setOpen]
-    );
-
-    const handleClick = useCallback(() => {
+    const handleClose = useCallback(() => {
+        localStorage.setItem('latestPatchSeen', latestPatchNotesUpdate);
         setOpen(false);
     }, [setOpen]);
```
ap0nia:
```suggestion
    }, []);
```

apps/antalmanac/tests/setup.ts
ap0nia, line 1: You can remove this file if it's not being used.

apps/antalmanac/vitest.config.ts
```diff
@@ -13,4 +13,8 @@ export default defineConfig({
             $stores: resolve(__dirname, './src/stores'),
         },
     },
+    test: {
+        environment: 'jsdom',
+        // setupFiles: './tests/setup.ts',
```
ap0nia: You can remove this if we don't need it.

```suggestion
```

apps/antalmanac/tests/patch-notes.test.tsx
```diff
@@ -0,0 +1,79 @@
+import { describe, expect, it } from 'vitest';
+import { render, screen, fireEvent } from '@testing-library/react';
+import PatchNotes from '$components/PatchNotes';
+
+const setLocalStorage = (key: string, value: string) => {
+    window.localStorage.setItem(key, value);
+};
+
+describe('PatchNotes Component', () => {
```
ap0nia: I think this description is fine.
```suggestion
describe('patch notes', () => {
```

apps/antalmanac/tests/patch-notes.test.tsx
```diff
@@ -0,0 +1,79 @@
+import { describe, expect, it } from 'vitest';
```
ap0nia: I prefer `test` instead of `it`.

```suggestion
import { describe, test, expect } from 'vitest';
```

apps/antalmanac/tests/patch-notes.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+describe('PatchNotes Component', () => {
+    describe('when latestPatchSeen is not equal to latestPatchNotesUpdate', () => {
+        it('should show the dialog', () => {
+            setLocalStorage('latestPatchSeen', '20230513');
+
+            const mockLatestPatchNotesUpdate = '20230819';
+            if (mockLatestPatchNotesUpdate != localStorage.getItem('latestPatchSeen')) {
+                render(<PatchNotes />);
+            }
```
ap0nia: No, this shouldn't be handling any sort of rendering logic; that defeats the purpose of testing the component. The value of local storage should be manipulated, and the component should be verified to do the correct behavior depending on the value in local storage.

i.e. if we change the value in local storage to be different from the latest patch notes, then we expect no dialog to appear if the patch notes is rendered.

ap0nia (changes requested):
Good job getting the tests setup :fire:

Left some suggestions on implementation.

apps/antalmanac/vitest.config.ts
KevinWu098: Whoops, I forgot that I had it commented out. The setup.ts file is needed as the tests will break if the JSDom isn't cleaned up after each test 😅

KevinWu098:
Not sure why tests are failing on the GH-Actions; My guess is that it's not running them in `apps/AntAlmanac`?

Works on my machine<sup>tm</sup>. LMK if there's anything I can do/fix to make the tests run!

README.md
```diff
@@ -74,8 +74,7 @@ If you ever need help, feel free to ask around on our [Discord server](https://d
    As you make changes to the React application in `src`, those changes will be automatically reflected on the website locally.
 
 #### Running Tests
-1. go into `apps/antalmanac`
-2. `pnpm test`
+1. Run `pnpm test` in the root directory
```
ap0nia, line 77: You can also run `pnpm -w test` from any subdirectory. But I don't really care about the readme lol. Figure it out or it's a skill issue 🤓

ap0nia (approved):
Epic :rocket:

PR #695, opened 2023-09-11
fix: no error when loading undefined schedule name

### LINKED ISSUE #554: Website crashes when editing schedule too quickly
If you open the website and immediately try to click the Edit Schedule button, the website crashes because `scheduleName` is undefined.

We resolved a similar issue here: #534 

https://user-images.githubusercontent.com/78244965/232614127-6d00cf20-35bc-484e-927f-5fbb7467fef1.mp4

ap0nia:
## Summary
The schedule store's `getScheduleName` can access a schedule at an undefined index in the schedules array.

Closes #554

### DIFF
--- a/apps/antalmanac/src/lib/course-helpers.ts
+++ b/apps/antalmanac/src/lib/course-helpers.ts
@@ -1,0 +1,158 @@
+import type { WebsocAPIResponse, WebsocSectionMeeting } from 'peterportal-api-next-types';
+import { PETERPORTAL_WEBSOC_ENDPOINT } from './api/endpoints';
+import type { CourseInfo } from './helpers';
+
+interface CacheEntry extends WebsocAPIResponse {
+    timestamp: number;
+}
+
+const websocCache: { [key: string]: CacheEntry } = {};
+
+export function clearCache() {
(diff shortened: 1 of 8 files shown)


apps/antalmanac/src/lib/course-helpers.ts
ap0nia, line 1: Some dipshit setup a circular dependency between the AppStore and ScheduleStores, so they're imported in the wrong order somehow; I can't do any testing for schedule store unless I fix it.
image

KevinWu098:
I believe this should close #554

apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -50,7 +51,7 @@ export class Schedules {
      * @return a specific schedule name
      */
     getScheduleName(scheduleIndex: number) {
-        return this.schedules[scheduleIndex].scheduleName;
+        return this.schedules[scheduleIndex]?.scheduleName;
```
EricPedley, line 54: This is is the actual fix right? Everything else was just a refactor to make writing the test work?

ap0nia: Yes, I just wanted to add a question mark and a test for that lol

EricPedley (approved):

PR #696, opened 2023-09-12
feat: multiple locations on courses

### LINKED ISSUE #592: Deal with multiple buildings

ap0nia:
## Summary
A continuation of #634

### DIFF
--- a/apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
+++ b/apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
@@ -88,8 +88,25 @@ interface CommonCalendarEvent extends Event {
     title: string;
 }
 
+export interface Location {
+    /**
+     * @example 'ICS'
+     */
+    building: string;
+
+    /**
+     * @example '174'
+     */
+    room: string;
+
+    /**
+     * If the location only applies on specific days, this is non-null.
+     */
+    days?: string[];
+}
+
 export interface CourseEvent extends CommonCalendarEvent {
-    bldg: string; // E.g., ICS 174, which is actually building + room
+    bldg: Location[];
     finalExam: {
         examStatus: 'NO_FINAL' | 'TBA_FINAL' | 'SCHEDULED_FINAL';
         dayOfWeek: 'Sun' | 'Mon' | 'Tue' | 'Wed' | 'Thu' | 'Fri' | 'Sat' | null;
@@ -158,14 +175,12 @@ const CourseCalendarEvent = (props: CourseCalendarEventProps) => {
     }, [setActiveTab]);
 
     const { classes, courseInMoreInfo } = props;
+
     if (!courseInMoreInfo.isCustomEvent) {
         const { term, instructors, sectionCode, title, finalExam, bldg, sectionType } = courseInMoreInfo;
 
-        const [buildingName = ''] = bldg.split(' ');
-
-        const buildingId = locationIds[buildingName] ?? 69420;
-
         let finalExamString = '';
+
         if (finalExam.examStatus == 'NO_FINAL') {
             finalExamString = 'No Final';
         } else if (finalExam.examStatus == 'TBA_FINAL') {
@@ -229,15 +244,19 @@ const CourseCalendarEvent = (props: CourseCalendarEventProps) => {
                             <td className={`${classes.multiline} ${classes.rightCells}`}>{instructors.join('\n')}</td>
                         </tr>
                         <tr>
-                            <td className={classes.alignToTop}>Location</td>
+                            <td className={classes.alignToTop}>Location{bldg.length > 1 && 's'}</td>
                             <td className={`${classes.multiline} ${classes.rightCells}`}>
-                                <Link
-                                    className={classes.clickableLocation}
-                                    to={`/map?location=${buildingId}`}
-                                    onClick={focusMap}
-                                >
-                                    {bldg}
-                                </Link>
+                                {bldg.map((location) => (
+                                    <div key={`${sectionCode} @ ${location.building} ${location.room}`}>
+                                        <Link
+                                            className={classes.clickableLocation}
+                                            to={`/map?location=${locationIds[location.building] ?? 0}`}
+                                            onClick={focusMap}
+                                        >
+                                            {location.building} {location.room}
+                                        </Link>
+                                    </div>
+                                ))}
                             </td>
                         </tr>
                         <tr>
--- a/apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
+++ b/apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
@@ -106,7 +106,8 @@ export interface Location {
 }
 
 export interface CourseEvent extends CommonCalendarEvent {
-    bldg: Location[];
+    locations: Location[];
+    showLocationInfo: boolean;
     finalExam: {
         examStatus: 'NO_FINAL' | 'TBA_FINAL' | 'SCHEDULED_FINAL';
         dayOfWeek: 'Sun' | 'Mon' | 'Tue' | 'Wed' | 'Thu' | 'Fri' | 'Sat' | null;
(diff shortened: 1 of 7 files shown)


apps/antalmanac/src/components/Map/Map.tsx
```diff
@@ -97,33 +99,33 @@ export default function CourseMap() {
...
-    }, [setMarkers, getCoursesPerBuilding]);
-
-    const updateCalendarEvents = useCallback(() => {
-        setCalendarEvents(AppStore.getCourseEventsInCalendar());
-    }, [setCalendarEvents]);
-
     useEffect(() => {
+        const updateMarkers = () => {
+            setMarkers(getCoursesPerBuilding());
+        };
```
ap0nia, line 104: Minor housekeeping to clean this up.

apps/antalmanac/src/components/Map/Map.tsx
```diff
@@ -97,33 +99,33 @@ export default function CourseMap() {
...
             AppStore.removeListener('addedCoursesChange', updateMarkers);
             AppStore.removeListener('currentScheduleIndexChange', updateMarkers);
         };
-    }, [updateMarkers]);
+    }, []);
 
     useEffect(() => {
+        const updateCalendarEvents = () => {
+            setCalendarEvents(AppStore.getCourseEventsInCalendar());
+        };
```
ap0nia, line 118: Housekeeping.

apps/antalmanac/src/components/Map/Map.tsx
```diff
@@ -41,7 +41,9 @@ interface MarkerContent {
 export function getCoursesPerBuilding() {
     const courseEvents = AppStore.getCourseEventsInCalendar();
 
-    const allBuildingCodes = courseEvents.map((event) => event.bldg.split(' ').slice(0, -1).join(' '));
+    const allBuildingCodes = courseEvents
+        .map((event) => event.bldg.map((location) => location.building))
+        .reduce((prev, cur) => prev.concat(cur), []);
```
EricPedley: Can we just replace this reduce with a `flat()` call? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat

apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
```diff
@@ -88,8 +88,25 @@ interface CommonCalendarEvent extends Event {
...
+
+    /**
+     * If the location only applies on specific days, this is non-null.
+     */
+    days?: string[];
+}
+
 export interface CourseEvent extends CommonCalendarEvent {
-    bldg: string; // E.g., ICS 174, which is actually building + room
+    bldg: Location[];
```
EricPedley: Can we rename this to `locations`?

apps/antalmanac/src/stores/calendarizeHelpers.ts
```diff
@@ -1,13 +1,18 @@
...
-import { CourseEvent, CustomEvent } from '$components/Calendar/CourseCalendarEvent';
+import { CourseEvent, CustomEvent, Location } from '$components/Calendar/CourseCalendarEvent';
 import { RepeatingCustomEvent } from '$components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog';
 import { notNull, getReferencesOccurring } from '$lib/utils';
 
 const COURSE_WEEK_DAYS = ['Su', 'M', 'Tu', 'W', 'Th', 'F', 'Sa'];
 
 const FINALS_WEEK_DAYS = ['Sat', 'Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri'];
 
+export function getBuilding(bldg: string): Location {
```
EricPedley: Same idea was one of my previous comments, can we keep the naming consistent so everything that previously referred to Building but now is actually a Location gets renamed properly?

apps/antalmanac/src/components/Map/Map.tsx
```diff
@@ -249,9 +251,14 @@ export default function CourseMap() {
                 {/* Draw a marker for each class that occurs today. */}
                 {markersToDisplay.map((marker, index) => {
                     // Find all courses that occur in the same building prior to this one to stack them properly.
+                    // TODO Handle multiple buildings between class comparisons on markers.
```
EricPedley, line 253: Is this referring to the fact that we're only using `marker.bldg[0]` here and ignoring all others if they don't share the same `building` attribute?

apps/antalmanac/src/components/Map/Map.tsx
EricPedley: Thx for these comments to explain why these changes are here. Although again, would be better in their own PR that is just for refactoring. A little git hack that might make it easier:

1. selectively stage only the refactor changes to your index
2. commit them as their own atomic commit
3. stash
4. revert the refactor commit
5. checkout to main, make a new branch, then reflog to find the id and cherry-pick that commit back onto your new branch
6. open new PR
7. checkout back to your feature branch and pop the stash back on

Now that I lay it out like that it sounds slightly like a pain in the ass, but if are experienced enough to not need to look up any of the commands it's pretty fast, and probably saves more time during review than it takes during writing.

EricPedley (changes requested):
Minor stuff, plus the big thing is renaming all stuff that used to just be a building string to now have "location" in the name.

EricPedley:
Also, seems like this just does everything #634 does. Are #634 and #592 closable after this? If so, can you try rebasing in at least one of Alan's commits so he gets co-authorship on this? Definitely want to assign credit since a lot of the stuff in this is just copy/paste from his. Also I'm not entirely sure if the rebasing in other author's commits is easy or not so lmk if you're having trouble with that.

apps/antalmanac/src/components/Map/Map.tsx
ap0nia: Oh, I didn't realize what it was doing; I was just copying the code lol. But I can do you one better with `flatMap`

```suggestion
        courseEvents.flatMap(event => event.bldg.map(location => location.building))
```

ap0nia:
Yeah, it's just #634 but up to date with main. I was gonna let him push some commits to address the comments and stuff so he could get some participation credit. I just opened this PR because the previous one was getting pretty stale and seemed kinda difficult to get up to speed. I still don't know how to rebase lol, so you can take care of that if you want.

apps/antalmanac/src/components/Map/Map.tsx
ap0nia: I don't remember making this comment, it might've been Alan. But yeah, we can only take one building because that's how the routing will calculate the route. When we show the routes, we show how to get from class 1 to class 2, class 2 to class 3, etc. and we don't have a way to represent class 1 to multiple locations for class 2.

apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
alanchangxyz: looking into this and some of the queries from api also return `bldg` instead of `locations`; going to see if i can also make those changes in this pr or if there's going to be a need for a related backend pr/intermediate formatting

alanchangxyz: update: decided to use intermediate formatting for replacing parts where api gives `bldg` instead of `locations`, yall can do a separate migration for backend later if this needs to become standard across all parts of the stack

alanchangxyz:
a PR on a PR, since i cannot commit directly to this icssc/ branch - #697

EricPedley:
For testing purposes, using ENGRMAE 295 Fall 2023 (section code 19177) as it's the first course I found that has two locations. Leaving this comment here so nobody needs to dig up an example on their own.

EricPedley (changes requested):
Ok basically looks fine, tested it out and everything works, just fix the failing tests and we're good here.

EricPedley (approved):
🚢🚢🚢🚢🚢

PR #697, opened 2023-09-13
Syncing fork changes for multiple buildings (#696-a)

alanchangxyz:
Notable:
- `CourseEvent` type now has `showLocationInfo` bool, mostly used because events in finals mode show "x locations" and don't pop up details for what those locations actually are. When true, a calendar event shows the actual locations comma-separated; when false, it shows "x locations" or the only location.
- `bldg` has been changed to `locations` everywhere for semantic consistency; a few cases remain for `bldg` where it comes directly from the api and there would be more cross-cutting concerns to deal with beyond the scope of multiple locations
- calendarize tests have also been adjusted to use new `locations` terminology


EricPedley (approved):
lgtm good work

PR #699, opened 2023-09-16
Remove underline from location link

### LINKED ISSUE #673: Calendar map link in dark mode is unreadable
image

### LINKED ISSUE #698: Location links should not be underlined
Since they don't lead away from the page, they should look less like URLs.

MinhxNguyen7:
## Summary
Location links should not be underlined. Since they don't lead away from the page, they should look less like URLs. I also fixed the class name for the link, so it changes color in dark mode appropriately now.

## Test Plan
See that it looks as expected.

## Issues

Closes #698 #673


KevinWu098 (approved):
LGTM; I'm gonna let the staging deploy then I'll merge

PR #702, opened 2023-09-19
Add Custom Events and Schedule Notes to Skeleton Mode

### LINKED ISSUE #700: Support Custom Events and Schedule Notes in Skeleton Mode
This would be a follow-up for #569 

The ability to support custom events and schedule notes should be possible when in skeleton mode.

### LINKED ISSUE #720: Too much gap between courses in Added Pane
Screenshot 2023-10-03 at 1 08 41 PM

Padding / gap needs to be adjusted

KevinWu098:
## Summary
1. Custom Events are now viewable in Skeleton Mode. Creation and Editing functionality is disabled.
2. Schedule Notes are now viewable in Skeleton Mode. Creation and Editing functionality is disabled.
3. Styling has been cleaned up for both skeleton and live mode.
4. Schedule Select Popover in the Calendar Toolbar is disabled.

### Empty Schedule
Before
Before - Empty Schedule
After
After - Empty Schedule

### Custom Events Box
Before
Before - Custom Events
After
After - Custom Events

### Skeleton Mode
Before
Before - Skeleton Mode
After
After - Skeleton Mode

## Test Plan
1. Load a schedule with courses, custom events, and a note in the *first* schedule (the one at the top of the dropdown)
2. Change `PETERPORTAL_WEBSOC_ENDPOINT` to trigger skeleton mode
3. Check that custom events are there and *are not editable*
4. Check that notes are there and *are not editable*
5. For the Calendar Toolbar, check that the Schedule Dropdown, Finals Button, Undo, Delete, and Add Custom Event buttons are disabled
6. For the App Bar, check that Save/Load, and Export are disabled. Notably, Import remains enabled because refactoring it to subscribe to the AppStore properly is somewhat out of scope. Import will still fail though!
7. For the edge case where there are no course events in any schedule, Skeleton Mode **will not trigger.** This is the current intended behavior until a better "Ping Websoc" option is created. 
8. `CustomEventDialog` was refactored. Test that custom events can still be created as long as day and schedule are checked. Make sure that editing the event from both the Calendar and the Added pane still work (and fills the appropriate data). 

## Issues
Closes #700 and Closes #720

## Future Follow-up
1. Refactor `Import.tsx` and properly subscribe to AppStore for the status of SkeletonMode: #771
2. Utilize react-hook-form to clean up `CustomEventDialog.tsx`


apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -156,6 +156,7 @@ class CustomEventDialog extends PureComponent<CustomEventDialogProps, CustomEven
                             variant="outlined"
                             size="small"
                             startIcon={<Add fontSize="small" />}
+                            disabled={AppStore.getSkeletonMode()}
```
KevinWu098: Creating or Editing while in skeleton mode should not be allowed IMO. Seems like it would be a damn easy place to create bugs and problems.

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -159,6 +160,111 @@ function CopyScheduleButton() {
...
+interface CustomEventsBoxProps {
+    customEventsFromProps: RepeatingCustomEvent[];
+    isSkeletonMode: boolean;
+    scheduleNames: string[];
+}
+
+function CustomEventsBox(props: CustomEventsBoxProps) {
+    const { customEventsFromProps, scheduleNames, isSkeletonMode } = props;
+
+    const [customEvents, setCustomEvents] = useState(customEventsFromProps);
```
KevinWu098, line 172: Here I've used a bit of an odd prop, `customEventsFromProps` for two reasons.

1. I can't use customEvents twice (for the prop and the useState)
2. skeletonSchedule doesn't have the luxury of calling the AppStore

I'm open to refactoring the name or the code here if it's too confusing or non-idiomatic

apps/antalmanac/src/components/Calendar/CalendarToolbar.tsx
```diff
@@ -178,6 +178,7 @@ function SelectSchedulePopover(props: { scheduleNames: string[] }) {
                 variant="outlined"
                 onClick={handleClick}
                 sx={{ minWidth, maxWidth, justifyContent: 'space-between' }}
+                disabled={AppStore.getSkeletonMode()}
```
KevinWu098: Allowing for schedules to be created, edited, and deleted while in skeleton mode creates new bugs and is generally just a bad time.

We don't currently support any aspect of editing schedule names or creating new schedules, so it would be best to disable it when in skeleton mode.

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -159,6 +160,112 @@ function CopyScheduleButton() {
...
+interface CustomEventsBoxProps {
+    customEvents: RepeatingCustomEvent[];
+    isSkeletonMode: boolean;
+    scheduleNames: string[];
+}
+
+function CustomEventsBox(props: CustomEventsBoxProps) {
+    const { scheduleNames, isSkeletonMode } = props;
+
+    const [customEvents, setCustomEvents] = useState(props.customEvents);
```
KevinWu098: This was a stylistic choice to use `props.customEvents` instead of another prop name due to the overlap in naming with useState. 

Open to suggestions if this is too confusing / non-idiomatic.

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -159,6 +160,112 @@ function CopyScheduleButton() {
...
+
+interface ScheduleNoteBoxProps {
+    scheduleNote: string;
+    isSkeletonMode: boolean;
+}
+
+function ScheduleNoteBox(props: ScheduleNoteBoxProps) {
+    const { isSkeletonMode } = props;
+
+    const [scheduleNote, setScheduleNote] = useState(props.scheduleNote);
```
KevinWu098: Same choice as above; `props.scheduleNote` is used due to an overlap in names

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -159,6 +160,112 @@ function CopyScheduleButton() {
...
+                type="text"
+                variant="filled"
+                label="Click here to start typing!"
+                onChange={handleNoteChange}
+                value={scheduleNote}
+                inputProps={{ maxLength: NOTE_MAX_LEN }}
+                InputProps={{ disableUnderline: true }}
+                fullWidth
+                multiline
+                disabled={isSkeletonMode}
```
KevinWu098: When the box is disabled, it's not clear in dark mode that it is disabled. One choice could be to add a `not-allowed` cursor CSS styling, but that would be inconsistent with the rest of the site.

An alternative choice would be to use a ternary in the label, with `"Editing is not allowed in skeleton mode"`. Thoughts?

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -214,6 +321,11 @@ function SkeletonSchedule() {
                     </Box>
                 ))
             }
+
+            <CustomEventsBox customEvents={skeletonSchedule.customEvents} scheduleNames={[]} isSkeletonMode={true} />
```
KevinWu098: `scheduleNames={[]}` is a magic array here; skeletonSchedule doesn't have a `scheduleNames` prop, but `CustomEventDetailView` is disabled in skeleton mode so the lack of this prop doesn't matter.

apps/antalmanac/src/components/Calendar/CalendarToolbar.tsx
ap0nia: This won't stay up to date with the current store's value. In order to properly capture the reactivity, you have to declare a `useState` at the top of the function, and then subscribe to the store accordingly.

```ts
function Component() {
  const [skeleton, setSkeleton] = useState(AppStore.getSkeletonMode())

  useEffect(() => {
    const handleSkeletonChange = () => {
     setSkeleton(AppStore.getSkeletonMode())
    }

    // Idk the actual event name
    AppStore.on('skeleton', handleSkeletonChange)

    return () => {
      AppStore.off('skeleton', handleSkeletonChange)
    }
  }, [])
}
```

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
ap0nia: See comment above concerning AppStore's reactivity.

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -159,6 +160,112 @@ function CopyScheduleButton() {
...
+                                        scheduleNames={scheduleNames}
+                                        isSkeletonMode={isSkeletonMode}
+                                    />
+                                </Grid>
+                            );
+                        })}
+                    </Grid>
+                </Box>
+            )}
+        </>
```
ap0nia: Because there's a top-level conditional, you can just have two different returns.

```suggestion
    if (customEvents.length <= 0) {
      return <Box />
    }
    
    return (
              <Box>
                  <Typography variant="h6">Custom Events</Typography>
                  <Grid container spacing={1} padding={0}>
                      {customEvents.map((customEvent) => {
                          return (
                              <Grid item md={12} xs={12} key={customEvent.title}>
                                  <CustomEventDetailView
                                      customEvent={customEvent}
                                      scheduleNames={scheduleNames}
                                      isSkeletonMode={isSkeletonMode}
                                  />
                              </Grid>
                          );
                      })}
                  </Grid>
              </Box>
```

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -159,6 +160,112 @@ function CopyScheduleButton() {
     );
 }
 
+interface CustomEventsBoxProps {
+    customEvents: RepeatingCustomEvent[];
+    isSkeletonMode: boolean;
+    scheduleNames: string[];
```
ap0nia: Don't pass these as props, you literally subscribe to the AppStore to get the values; there's not much point to passing around data that doesn't need to be passed around, and it's more reliable to use a single point of truth.

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -159,6 +160,112 @@ function CopyScheduleButton() {
...
+                    </Grid>
+                </Box>
+            )}
+        </>
+    );
+}
+
+interface ScheduleNoteBoxProps {
+    scheduleNote: string;
+    isSkeletonMode: boolean;
```
ap0nia: Don't pass these in as props.

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -304,7 +393,7 @@ function AddedSectionsGrid() {
 
             <Box>
                 <Typography variant="h6">{`${scheduleName} (${scheduleUnits} Units)`}</Typography>
-                <Grid container spacing={2} padding={0}>
+                <Box marginBottom={-1}>
```
ap0nia: I don't like this negative margin. It feels like something that can cause lots of CSS edge case pains in the future. Could you try to figure out a different way to accomplish this? (i.e. the typography in MUI generally has a lot of vertical spacing, so if you wanted the text to be squished vertically, you could look at modifying the typography directly)

apps/antalmanac/src/components/RightPane/AddedCourses/CustomEventDetailView.tsx
```diff
@@ -55,30 +37,33 @@ const CustomEventDetailView = (props: CustomEventDetailViewProps) => {
...
+                            deleteCustomEvent(customEvent.customEventID);
+                        }}
+                        size="large"
+                    >
+                        <Delete fontSize="small" />
+                    </IconButton>
+                    <CustomEventDialog customEvent={customEvent} scheduleNames={scheduleNames} />
+                </CardActions>
+            )}
         </Card>
```
ap0nia: Honestly, just for simplicity, it might be easier to have two different returns (or two components) depending on whether it's the skeleton schedule or not.

ap0nia (approved):
Good improvements. Left some suggestions :rocket:

ap0nia (changes requested):
Oh, I meant to request changes. Oops lol

apps/antalmanac/src/components/Calendar/CalendarToolbar.tsx
KevinWu098: Well, I'm curious if updating the state matters? If Websoc is down, is AntAlmanac going to react to when it's up again?

My current working assumption is that the only way to leave skeleton mode is to refresh the page?

ap0nia: That's probably the visible behavior. But because it's part of an observable store, it's best to subscribe to the changes for consistency. i.e. the skeleton mode would otherwise just be a const value that's determined on page load that we know would never change.

KevinWu098: Aye aye 🫡

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
KevinWu098, line 1: So... rewriting this entire component into a functional component might've been a bit out of scope, but to properly subscribe to the AppStore for `skeletonMode`, I think it was worth it.

There should be no major changes as I ported it over pretty faithfully, but I'll comment areas where I did make some changes.

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -46,185 +39,209 @@ interface CustomEventDialogProps {
...
-        this.setState({ title: event.target.value });
-    };
+    const handleClose = useCallback(() => {
+        setOpen(false);
+    }, [setOpen]);
 
-    handleEndTimeChange: React.ChangeEventHandler<HTMLInputElement | HTMLTextAreaElement> = (event) => {
-        this.setState({ end: event.target.value });
-    };
+    const handleSubmit = () => {
```
KevinWu098: This is a new function. Previously `handleClose` had a conditional inside of it telling it whether or not to add the custom event to calendar.

That doesn't make a ton of sense as `handleClose` should close the component, so I've made a new function to handle the closing **and** adding.

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -46,185 +39,209 @@ interface CustomEventDialogProps {
...
+        setStart(defaultCustomEventValues.start);
+        setEnd(defaultCustomEventValues.end);
+        setTitle(defaultCustomEventValues.title);
+        setDays(defaultCustomEventValues.days);
     };
 
-    handleAddToCalendar = () => {
-        if (!this.state.days.some((day) => day) || this.state.scheduleIndices.length === 0) return;
+    const handleAddToCalendar = () => {
+        if (!days.some((day) => day) || scheduleIndices.length === 0) return;
```
KevinWu098, line 111:
```suggestion
        if (!days.includes(true) || scheduleIndices.length === 0) return;
```
Would this be make more sense?

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -46,185 +39,209 @@ interface CustomEventDialogProps {
...
-                            })
-                        }
+    const handleSelectScheduleIndices = useCallback(
+        (scheduleIndices: number[]) => {
+            setScheduleIndices(scheduleIndices);
+        },
+        [setScheduleIndices]
+    );
+
+    const disabled = !(scheduleIndices.length && days.some(Boolean));
```
KevinWu098:
```suggestion
    const disabled = !(scheduleIndices.length && days.includes(true));
```
Same as above

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -46,185 +39,209 @@ interface CustomEventDialogProps {
...
-                            })
-                        }
+    const handleSelectScheduleIndices = useCallback(
+        (scheduleIndices: number[]) => {
+            setScheduleIndices(scheduleIndices);
+        },
+        [setScheduleIndices]
+    );
+
+    const disabled = !(scheduleIndices.length && days.includes(true));
```
KevinWu098: This was redone for readability

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -46,185 +39,212 @@ interface CustomEventDialogProps {
...
+                <DialogActions>
+                    <Button onClick={() => handleClose()} color={isDarkMode() ? 'secondary' : 'primary'}>
+                        Cancel
+                    </Button>
+                    <Button onClick={() => handleSubmit()} variant="contained" color="primary" disabled={disabled}>
+                        {disabled
+                            ? 'Schedule and day must be checked'
+                            : props.customEvent
+                            ? 'Save Changes'
+                            : 'Add Event'}
```
KevinWu098, line 229: I don't love how Prettify handles a double ternary, but I can live with it

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -46,185 +39,212 @@ interface CustomEventDialogProps {
...
+                            }}
+                            style={{ marginRight: 5, marginTop: 5 }}
                         />
-                    </DialogContent>
-
-                    <DialogActions>
-                        <Button onClick={() => this.handleClose(true)} color={isDarkMode() ? 'secondary' : 'primary'}>
-                            Cancel
-                        </Button>
-                        <Tooltip title="Schedule and day must be checked" disableHoverListener={!this.isAddDisabled()}>
```
KevinWu098, line 211: This didn't work because when children are disabled, tooltips are as well.

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -122,15 +122,16 @@ export const loadSchedule = async (userId: string, rememberMe: boolean) => {
 
                 if (scheduleSaveState === undefined) {
                     openSnackbar('error', `Couldn't find schedules for username "${userId}".`);
-                } else if (await AppStore.loadSchedule(scheduleSaveState)) {
-                    openSnackbar('success', `Schedule for username "${userId}" loaded.`);
```
KevinWu098: This check fails if no courses are in the schedule. This results in `queryWebsoc` never being called, and never erroring out. Skeleton mode would not end up being triggered.

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -122,15 +122,16 @@ export const loadSchedule = async (userId: string, rememberMe: boolean) => {
...
-                } else {
+                } else if (!(await pingWebsoc())) {
```
ap0nia: Could you try applying guard clauses here? It's my personal preference because it makes it easier to distinguish between different cases. For example:

```ts
if (scheduleSaveState == null) {
  openSnackbar('error', `Couldn't find schedules for username "${userId}".`);
  return
}

const isWebsocAlive = await pingWebsoc()

if (!isWebsocAlive) {
  // ... simple logic here
  return
}

// ... can put lots of logic here, outside of any if-blocks.
```

Of course, it depends a lot on the circumstance, but I generally prefer straightforward case handling to if/else-if blocks because it reduces nesting and gives me more flexibility over the size of each block (i.e. I'll prefer to handle the biggest blocks at the very end, outside of any if statement, and simpler cases inside their own isolated if statements)

PS: Also, prefer loose comparison with null whenever possible because it matches null and undefined, which are usually synonymous.

apps/antalmanac/src/lib/course-helpers.ts
```diff
@@ -35,6 +35,28 @@ export function getCourseInfo(SOCObject: WebsocAPIResponse) {
...
+            .then((r) => r.json())
+            .then((r) => r.payload);
+
+        // If the ping returns an empty payload, something is wrong
+        // If there's an error, there will not be a payload, triggering the catch statement
+        return response.schools.length > 0 ? true : false;
+    } catch (error) {
+        return false;
+    }
+}
```
ap0nia: - Define the ping URL at the top level
- I like to distinguish between raw responses, processed responses, data, etc.
- Name the function something a bit more specific (alternative below)
- Handle the try/catch directly in the promise chain so there's less overall nesting

```suggestion
// An arbitrary URL used to check if Peterportal and/or Websoc are live
const PING_PETERPORTAL_WEBSOC_URL =
        'https://api-next.peterportal.org/v1/rest/websoc?year=2023&quarter=Spring&sectionCodes=34271';
export async function isWebsocWorking() {

   const isWebsocAlive = await fetch(PING_PETERPORTAL_WEBSOC_URL, {
            headers: {
                Referer: 'https://antalmanac.com/',
            },
        })
            .then((response) => response.json())
            .then((data) => data.payload.schools.length > 0)
            .catch((error) => {
              // console.log(error) --> can handle error if desired
              return false
            })
            
     return isWebsocAlive
}
```

Alternatively, we can also keep it as "ping", but I think a more appropriate interface would be to throw an error if pinging websoc fails; i.e. a ping doesn't inherently "return" any info, it just tries to do something and either succeeds or fails.

```ts
/**
 * Pings websoc to verify that it is active and functional.
 * @throws Error if not active.
 */
export async function pingWebsoc(): void {
  const isWebsocAlive = false // logic for determining this
  if (!isWebsocAlive) {
    throw new Error('Websoc has encountered SKILL ISSUE')
  }
}

// Example usage of this version of the interface
function doSomething() {
  const isWebsocWorking = await pingWebsoc.then(() => true).catch(() => false)
}
```

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -304,7 +403,7 @@ function AddedSectionsGrid() {
 
             <Box>
                 <Typography variant="h6">{`${scheduleName} (${scheduleUnits} Units)`}</Typography>
-                <Grid container spacing={2} padding={0}>
+                <Box display="flex" flexDirection="column" gap={1}>
                     {courses.map((course) => {
                         return (
                             <Grid item md={12} xs={12} key={course.deptCode + course.courseNumber}>
```
ap0nia: Don't use the grid component inside a box; either commit to grid container parent and grid item children, or just use boxes for everything.

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -43,188 +36,214 @@ export interface RepeatingCustomEvent {
...
+    const handleOpen = useCallback(() => {
+        setOpen(true);
+        setScheduleIndices([AppStore.schedule.getCurrentScheduleIndex()]);
 
         logAnalytics({
             category: analyticsEnum.calendar.title,
             action: analyticsEnum.calendar.actions.CLICK_CUSTOM_EVENT,
         });
-    };
+    }, [setOpen]);
```
ap0nia: You can remove this since `setOpen` should never change, and similar idea with other dependency arrays where applicable.

```suggestion
    }, []);
```

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
ap0nia: Very good. You're smarter than the average smooth brain React copium inhaler :sunglasses:

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -43,188 +36,214 @@ export interface RepeatingCustomEvent {
...
+        handleClose();
+        handleAddToCalendar();
 
-    handleStartTimeChange: React.ChangeEventHandler<HTMLInputElement | HTMLTextAreaElement> = (event) => {
-        this.setState({ start: event.target.value });
+        logAnalytics({
+            category: analyticsEnum.calendar.title,
+            action: analyticsEnum.calendar.actions.ADD_CUSTOM_EVENT,
+        });
     };
```
ap0nia: You can memoize this.

One thing you might notice is that `handleClose` technically never reruns because its dependency array should be empty. But for consistency in React, you have to include it since it in a dependency array anyways because any sort of memoization makes the variable become considered as a "stateful" variable that technically "could" change. (imagine if React worked the way it looked like it would work).

```ts
   const handleSubmit = useCallback(() => {
        handleClose();
        handleAddToCalendar();

        logAnalytics({
            category: analyticsEnum.calendar.title,
            action: analyticsEnum.calendar.actions.ADD_CUSTOM_EVENT,
        });
    }, [handleClose, handleAddToCalendar])
```

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
ap0nia, line 1: Ok screw it, I was gonna give a bunch of recommendations for memoization, but it's making this component way too hard to read. Just don't memoize anything lol. Remove all useCallbacks and useMemos.

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -43,188 +36,214 @@ export interface RepeatingCustomEvent {
...
+        },
+        [setTitle]
+    );
+
+    const handleStartTimeChange: React.ChangeEventHandler<HTMLInputElement | HTMLTextAreaElement> = useCallback(
+        (event) => {
+            setStart(event.target.value);
+        },
+        [setStart]
+    );
```
ap0nia: Instead of typing the function, could you type the event? It sorta depends on the scenario, but for simple callbacks, typing the event is easier than typing the entire function.
```suggestion
    const handleStartTimeChange = useCallback(
        (event: React.ChangeEvent<HTMLInputElement>) => {
            setStart(event.target.value);
        },
        [setStart]
    );
```

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
ap0nia: Could you move this closer to the top? In my experience, I prefer declarations in roughly this hierarchy
- useState 
- Custom hooks that provide stateful variables, i.e. `useRouter`, `useParams`, etc.
- values (i.e. values memoized with useMemo, const declarations, etc.)
- functions (i.e. handlers, useCallback, etc.)
- useEffect

The main rationale being that stateful variables are generally sources of truth for other operations in the component, so they should be at the top (useState, custom hooks, and values). Functions and callbacks will typically use these values to do their operations. And then useEffect just introduces arbitrary side effects into the component, sometimes referring to either of the previous two or not.

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -43,188 +36,214 @@ export interface RepeatingCustomEvent {
...
+                            label="Start Time"
+                            type="time"
+                            defaultValue={start}
+                            InputLabelProps={{
+                                shrink: true,
+                            }}
+                            inputProps={{
+                                step: 300,
+                            }}
+                            style={{ marginRight: 5, marginTop: 5 }}
```
ap0nia, line 194: It might be more succinct to flex the container and then apply gap to it. Then group the fields into boxes organized by how you want to spacing to look.

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -43,188 +36,214 @@ export interface RepeatingCustomEvent {
...
+    );
+
+    const handleEndTimeChange: React.ChangeEventHandler<HTMLInputElement | HTMLTextAreaElement> = useCallback(
+        (event) => {
+            setEnd(event.target.value);
+        },
+        [setEnd]
+    );
+
+    const handleDayChange = useCallback(
```
ap0nia: Ok, these are way too many callbacks. Do you wanna plug in react-hook-form for this yourself? Or would you want me to take care of it?

ap0nia (dismissed):

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
KevinWu098: It was causing weird styling prior with a grid parent; I'll look into fixing the grid in box thing

apps/antalmanac/src/lib/course-helpers.ts
```diff
@@ -35,6 +35,26 @@ export function getCourseInfo(SOCObject: WebsocAPIResponse) {
     return courseInfo;
 }
 
+// An arbitrary URL used to check if Peterportal and/or Websoc are live
+const PING_PETERPORTAL_WEBSOC_URL =
```
MinhxNguyen7: This URL should be in api/endpoints.ts

apps/antalmanac/src/lib/course-helpers.ts
```diff
@@ -35,6 +35,26 @@ export function getCourseInfo(SOCObject: WebsocAPIResponse) {
...
+    'https://api-next.peterportal.org/v1/rest/websoc?year=2023&quarter=Spring&sectionCodes=34271';
+
+export async function isWebsocWorking() {
```
MinhxNguyen7: I don't love this way of checking whether PP/WebSOC is working because we're essentially making a random extra request for them to have to handle before we make the bulk of our requests. It's not a ping. It's extra load for PP (although I don't know if it's non-negligible), and it slows down our schedule load times, especially after #713.

If it really simplifies the process, then it's justified, but I'm not sure if it does.

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -120,18 +120,26 @@ export const loadSchedule = async (userId: string, rememberMe: boolean) => {
...
                     openSnackbar('error', `Couldn't find schedules for username "${userId}".`);
-                } else if (await AppStore.loadSchedule(scheduleSaveState)) {
-                    openSnackbar('success', `Schedule for username "${userId}" loaded.`);
-                } else {
+                    return;
+                }
+
+                const isWebsocAlive = await isWebsocWorking();
+
+                if (!isWebsocAlive) {
```
MinhxNguyen7: This doesn't account for when the initial "ping" succeeds but one of the subsequent requests fail. It's not that likely, but the previous code addressed that. I'm not sure why you wanted to change it.

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -120,18 +120,26 @@ export const loadSchedule = async (userId: string, rememberMe: boolean) => {
...
                     AppStore.loadSkeletonSchedule(scheduleSaveState);
                     openSnackbar(
                         'error',
                         `Network error loading course information for "${userId}". 
                         If this continues to happen, please submit a feedback form.`
                     );
                 }
+
+                if (isWebsocAlive) {
```
MinhxNguyen7:
```suggestion
                else {
```

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -43,188 +36,197 @@ export interface RepeatingCustomEvent {
 interface CustomEventDialogProps {
     customEvent?: RepeatingCustomEvent;
     onDialogClose?: () => void;
-    scheduleNames: string[];
```
MinhxNguyen7, line 37: Not sure if this was intentional, but this is causing an error in CalendarToolbar.tsx

MinhxNguyen7 (changes requested):

apps/antalmanac/src/actions/AppStoreActions.ts
KevinWu098: https://github.com/icssc/AntAlmanac/pull/702/files#r1332386420

> This check fails if no courses are in the schedule. This results in queryWebsoc never being called, and never erroring out. Skeleton mode would not end up being triggered.

MinhxNguyen7 (changes requested):
Also, it'd be helpful to have a more comprehensive test plan because you're changing a few things here. 

For example:
1) Change PP endpoint to trigger skeleton mode.
2) Load a schedule with notes in the most recent schedule.
3) Check that the notes are there.
4) Check that buttons x, y, z are greyed out and disabled.
etc.

FYI, the way that you said to test this actually wouldn't work because you hardcoded the "ping" URL in `course-helpers.ts`.

I know that the more detailed test plan is a bit "do as I say, not as I do", but I do think that it matters more here because you changed a decent number of things.

MinhxNguyen7 (commented):
If it's not too much work, I don't see any reason why you can't edit the notes and save them when PP is not reachable, since AA is still reachable. That might be more work than it's worth, so it's up to you to decide

KevinWu098:
> If it's not too much work, I don't see any reason why you can't edit the notes and save them when PP is not reachable, since AA is still reachable. That might be more work than it's worth, so it's up to you to decide

This will giga-break existing courses since AA saves based on what's loaded in, apparently.

KevinWu098:
> If it's not too much work, I don't see any reason why you can't edit the notes and save them when PP is not reachable, since AA is still reachable. That might be more work than it's worth, so it's up to you to decide

This will giga-break existing courses since AA saves based on what's loaded in, apparently.

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
KevinWu098: _To be implemented in the future_

MinhxNguyen7 (approved):
Works and LGTM!

MinhxNguyen7 (approved):
Ship it

KevinWu098:
> Ship it

This PR is cursed to be forever be locked in merge conflict hell 😭 -> I merged in #726 's changes and re-tested; hopefully everything's running ok still

apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
```diff
@@ -294,16 +294,18 @@ const CourseCalendarEvent = (props: CourseCalendarEventProps) => {
...
+                        Location:&nbsp;
+                        <Link
+                            className={classes.clickableLocation}
+                            to={`/map?location=${building ?? 0}`}
+                            onClick={focusMap}
+                        >
+                            {buildingCatalogue[+building].name}
+                        </Link>
+                    </div>
+                )}
```
KevinWu098, line 308: I took the liberty of adjusting this piece of code, since it would often display `Location: `, if there was no selected building.

MinhxNguyen7:
There are more merge conflicts 💀

KevinWu098:
> There are more merge conflicts 💀

i love coding :)

MinhxNguyen7 (approved):
Idk, seems alright

PR #709, opened 2023-09-27
Fix Undefined Trim Bug

### LINKED ISSUE #708: Undefined Schedule Trim Bug
Screenshot 2023-09-27 at 2 39 41 PM

.trim() is bugging out, likely due to undefined scheduleName

KevinWu098:
## Summary
Following fixes in #695, a new bug was unearthed in `RenameSchedule.tsx` because of how it was accessing (and utilizing) the array `scheduleNames.` 

```typescript

    const [name, setName] = useState(scheduleNames[index]);
```

`name` can end up being set to undefined, meaning `name.trim()` crashes and burns. Adding optional chaining should resolve the bug.

Additional inspection of `CalendarToolbar.tsx` and `DeleteSchedule.tsx` could be useful for sniffing out other bugs, but from my own testing, the bug seems to be resolved. (The other two files don't really utilize `scheduleNames` like how `RenameSchedule.tsx` did.

## Test Plan
Make sure that clicking the schedule dropdown before schedules have loaded doesn't crash the site. Testing is somewhat difficult in my experience, but reloading then spam-clicking the dropdown was what I used.

## Issues
Closes #708


apps/antalmanac/src/components/dialogs/RenameSchedule.tsx
```diff
@@ -96,7 +96,7 @@ function RenameScheduleDialog(props: ScheduleNameDialogProps) {
                 <Button onClick={handleCancel} color={isDarkMode() ? 'secondary' : 'primary'}>
                     Cancel
                 </Button>
-                <Button onClick={submitName} variant="contained" color="primary" disabled={name.trim() === ''}>
+                <Button onClick={submitName} variant="contained" color="primary" disabled={name?.trim() === ''}>
```
ap0nia: I think there's two (alternative) ways of approaching this problem.

1. Use a memoized variable that safely calculates the disabled state. (It's basically the same thing, but more optimized and marginally easier to read since the operation is moved to the function body instead of the returned UI.)
```ts
const disabled = useMemo(() => {
  return name?.trim() === ''
}, [name])
```

2. Initialize the state with a default value of empty string. (Unless it's possible for the name in the schedule store to be set to null, which should honestly be prevented altogether)
```ts
const [name, setName] = useState(scheduleNames[index] ?? '')
```

The former is more robust, i.e. for dealing with edge cases more, but the latter is simpler.

ap0nia (changes requested):
This logic is good; we can consider delegating it to a slightly different part of the component.

ap0nia (changes requested):
Ok this is good. Sorry for not bringing this up in the first review, but since this is supposed to be fixing a particular rendering error, could you look into using testing the component [rendering](https://testing-library.com/docs/react-testing-library/api#render-result)?

A test might look something like this:

```tsx
import { describe, test, expect } from 'vitest'
import { render } from '@testing-library/react'

describe('RenameScheduleDialog', () => {
  test('does not error if schedule name is not defined', () => {
    // Reset the schedule store somehow.
    ScheduleStore.setSchedules([])

    // Use a schedule index that doesn't have a schedule, and thus no schedule name
    // Also, it would be good to know if this previously threw an error prior to this fix.
    expect(render(<RenameScheduleDialog index={100} />)).not.toThrowError()
  })
})
```

apps/antalmanac/tests/schedule.test.tsx
```diff
@@ -0,0 +1,16 @@
+import { describe, test, expect } from 'vitest';
+import { render } from '@testing-library/react';
+import { Schedules } from '../src/stores/Schedules';
```
ap0nia: Just because we have aliases, they could be used for shorter import paths.

```suggestion
import { Schedules } from '$stores/Schedules';
```

apps/antalmanac/tests/schedule.test.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+describe('schedule logic', () => {
+    const scheduleStore = new Schedules();
+
+    test('no error when loading undefined schedule', () => {
+        expect(() => scheduleStore.getScheduleName(100)).not.toThrowError();
+    });
+
+    test('does not error if schedule name is undefined', () => {
+        expect(() => render(<RenameScheduleDialog index={100} open={true} />)).not.toThrowError();
```
ap0nia: Boolean props can be passed with just the prop name for brevity.

```suggestion
        expect(() => render(<RenameScheduleDialog index={100} open />)).not.toThrowError();
```

ap0nia (approved):
Beautiful work :rocket:

EricPedley (approved):
just looked at the suggestions you merged in from Brian and LGTM as far as that goes, I'll trust Brian's judgement on the rest.

PR #711, opened 2023-09-29
Fix Added Pane Course Grouping

KevinWu098:
## Summary
Screenshot 2023-09-28 at 8 04 25 PM

This piece of filtering was too general for the courses that shared the same courseNumber and deptCode (EECS 298, BIO SCI 199, etc). Because they all have different course titles, however, the fix is simply to narrow down the filter even more with `.courseTitle`.

Old Code ->
```typescript
        let formattedCourse = formattedCourses.find(
            (needleCourse) =>
                needleCourse.courseNumber === course.courseNumber && needleCourse.deptCode === course.deptCode
```

## Test Plan
Make sure courses like BIO SCI 199 are properly grouped, as well as that no other functionality has been broken.

## Issues
Closes #


EricPedley (approved):
🚢🚢🚢🚢🚢🚢

PR #713, opened 2023-09-29
Parallelize Schedule Load Websoc Requests

### LINKED ISSUE #712: Load schedules faster
For users with lots of old schedules across different terms, loading is hella slow because we fetch _all_ the class info before finishing the load. I think we can speed this up by lazy loading the course info for schedules that aren't the default one, and maybe make requests to PPAPI concurrently.

EDIT: actually lazy-loading is pointless when we're doing everything concurrently since all requests seem to take roughly the same amount of time so all we really need is parallelization (correct me if I'm wrong here pls).
image

EricPedley:
## Summary
Saves a bunch of time on schedule loading by making PPAPI requests for different terms concurrently.

Before:
image

After:
image

## Test Plan
Verify with network inspector that requests are being done concurrently, and also that nothing breaks.

## Issues
Closes #712

### DIFF
--- a/apps/antalmanac/src/stores/Schedules.ts
+++ b/apps/antalmanac/src/stores/Schedules.ts
@@ -463,13 +463,8 @@ export class Schedules {
             const courseInfoDict = new Map<string, { [sectionCode: string]: CourseInfo }>();
 
             const websocRequests = Object.entries(courseDict).map(async ([term, courseSet]) => {
-                const sectionCodes = Array.from(courseSet);
-                const courseInfo = getCourseInfo(
-                    await queryWebsoc({
-                        term: term,
-                        sectionCodes: sectionCodes.join(','),
-                    })
-                );
+                const sectionCodes = Array.from(courseSet).join(',');
+                const courseInfo = getCourseInfo(await queryWebsoc({ term, sectionCodes }));
                 courseInfoDict.set(term, courseInfo);
             });
 
--- a/apps/antalmanac/src/stores/Schedules.ts
+++ b/apps/antalmanac/src/stores/Schedules.ts
@@ -461,9 +461,9 @@ export class Schedules {
 
             // Get the course info for each course
             const courseInfoDict = new Map<string, { [sectionCode: string]: CourseInfo }>();
-            for (const [term, courseSet] of Object.entries(courseDict)) {
+
+            const websocRequests = Object.entries(courseDict).map(async ([term, courseSet]) => {
                 const sectionCodes = Array.from(courseSet);
-                // Code from ImportStudyList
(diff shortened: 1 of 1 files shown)


apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -461,20 +461,23 @@ export class Schedules {
...
                 );
                 courseInfoDict.set(term, courseInfo);
-            }
+            });
+
+            await Promise.all(websocRequests);
 
             // Map course info to courses and transform shortened schedule to normal schedule
             for (const shortCourseSchedule of saveState.schedules) {
+                console.log(shortCourseSchedule);
```
MinhxNguyen7: Is this an accident?

Otherwise, code looks good. I haven't checked if it works though

MinhxNguyen7 (changes requested):

apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -461,20 +461,23 @@ export class Schedules {
 
             // Get the course info for each course
             const courseInfoDict = new Map<string, { [sectionCode: string]: CourseInfo }>();
-            for (const [term, courseSet] of Object.entries(courseDict)) {
+
+            const websocRequests = Object.entries(courseDict).map(async ([term, courseSet]) => {
                 const sectionCodes = Array.from(courseSet);
```
ap0nia: If this variable name is `sectionCodes`, then I think it should correspond directly with the parameter passed to `queryWebsoc`. Suggestion is continued below.

```suggestion
                const sectionCodes = Array.from(courseSet).join(',');
```

apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -461,20 +461,23 @@ export class Schedules {
...
-                // Code from ImportStudyList
                 const courseInfo = getCourseInfo(
                     await queryWebsoc({
                         term: term,
                         sectionCodes: sectionCodes.join(','),
                     })
                 );
```
ap0nia: Since both `term` and `sectionCodes` are calculated correctly prior to this function call, it makes the provided object argument cleaner.

```suggestion
                const courseInfo = getCourseInfo(await queryWebsoc({ term, sectionCodes }));
```

ap0nia (approved):
Good optimization :rocket:

apps/antalmanac/src/stores/Schedules.ts
EricPedley: good catch, will remove

MinhxNguyen7 (approved):
LGTM!

MinhxNguyen7:
Blazing

Issue #558, opened 2023-05-01
Make every class search parameter sharable


EricPedley:
Whenever you search, it should update the URL with query params for all the options in your search query and you should be able to copy/paste it and load the search again, just from the URL. Also, when you load the URL it should fire the search query upon load instead of waiting for the user to press enter or search.

calejvaldez:
Hi. I'm trying to work on this, but I'm trying to figure out what part of the code "triggers" the search to begin. I have the parameters on the search bar already, now I'm just trying to trigger a search with the parameters.

These are the changes I've made so far: https://github.com/calejvaldez/AntAlmanac/commit/e172c1d3baa4d5c76b31fc0525ac7b03a80d2b1f

EricPedley:
It's a bit complicated: one of either `FuzzySearch` or `LegacySearch` updates the form values in the `RightPaneStore`, then toggles the view in `CoursePaneRoot` from `SearchForm` to `CourseRenderPane`. `CourseRenderPane` then gets the form values from `RightPaneStore` when it is mounted, queries PeterPortal API's websoc endpoint to do the search, then writes the query results to its internal state, and uses that to render the results.

To answer precisely, the `SearchForm` component gets passed a function as a prop, `toggleSearch`, that it runs as a callback when the user submits the form. Before that function is called, all the form values need to be set in the `RightPaneStore`

Lmk if you have more questions, I'm glad someone is tackling this feature.

MinhxNguyen7:
@ap0nia might want to comment on this since he implemented the router. Otherwise, I can take a look and suggest how you can do this.

calejvaldez:
Hi all. I think I've gotten a bit too busy, and I don't see myself finishing this anytime soon. I'm going to unassign myself for now.

PR #714, opened 2023-09-30
refactor: made course pane components functional

ap0nia:
## Summary
If we want to support more sophisticated searching, then we'd want to integrate better with react-router so that we can manage state in the URL. This is in regards to #558 and the work that's being done on a [fork by Carlos](https://github.com/calejvaldez/AntAlmanac/commit/e172c1d3baa4d5c76b31fc0525ac7b03a80d2b1f)

The core component logic has been mostly preserved and optimized for functional components.

## Test Plan
I can write tests, if anybody wants me to. Some tests I can think of at the top of my head are:
- Displays the search results only when the conditions are met (i.e. selected a course, specified the proper advanced search parameters, etc.)
- Returns to search when a key is pressed and conditions are met

### DIFF
--- a/apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneRoot.tsx
+++ b/apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneRoot.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useState } from 'react';
+import { useCallback, useEffect, useReducer } from 'react';
 
 import RightPaneStore from '../RightPaneStore';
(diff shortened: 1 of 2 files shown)


MinhxNguyen7 (approved):
LGTM!

EricPedley (approved):
Only reading the last commit that replaces the state hook with a reducer hook, and that looks good to me. Trusting Minh's approval for the rest. @ap0nia if you're ready to merge can you press the button? Not sure if you wanted to throw in anything else before merging.

PR #719, opened 2023-10-03
fix: Color Changing Bug w/ Duplicate Courses

### LINKED ISSUE #716: Duplicate Courses in Multiple Schedules Breaks Color Picker
https://github.com/icssc/AntAlmanac/assets/100006999/4c79f0aa-112f-4563-a8fe-a83e3eb2f283

To reproduce the bug, add two of the same course into different schedules. Attempting to edit the color of the course in the schedule of the higher index will fail and only change the index of the first.

[screen-recorder-sat-sep-30-2023-15-39-24 (1).webm](https://github.com/icssc/AntAlmanac/assets/100006999/f8469893-8b38-4c75-a83e-c8621cee04d7)

KevinWu098:
## Summary
[screen-recorder-sat-sep-30-2023-15-39-24 (1).webm](https://github.com/icssc/AntAlmanac/assets/100006999/f8469893-8b38-4c75-a83e-c8621cee04d7)

As shown above (and in #716), changing colors of a course event fails if the same course event exists in a schedule that is lower in index than the one currently being changed. 

**The reason for this is that `getExistingCourse` (shown below) only returned the reference of _a singular_ course that matched the given params.** This meant that courses (which matched the params) in other schedules would not have their colors changed.

```typescript
    /**
     * @return Reference of a course that matches the params
     */
    getExistingCourse(sectionCode: string, term: string) {
        for (const course of this.getAllCourses()) {
            if (course.section.sectionCode === sectionCode && term === course.term) {
                return course;   // <----- the problem
            }
        }
        return undefined;
    }
```

TL;DR Color Picker and Courses will now correctly update their colors (without affecting other schedules and courses)

## Test Plan
1. Create at least two schedules which contain the same course
2. Attempt to change the color on each. Success will be defined by steps 3-6
3. The color of the course in the Calendar should update
4. The color of the course in the Search pane should update
5. The color of the course in the Added pane should update
6. The updated colors should save and be loaded in correctly

## Issues
Closes #716

### DIFF
--- a/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
+++ b/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
@@ -254,7 +254,6 @@ class ScheduleCalendar extends PureComponent<ScheduleCalendarProps, ScheduleCale
         return (
             <div className={classes.container} style={isMobile ? { height: 'calc(100% - 50px)' } : undefined}>
                 <CalendarToolbar
-                    onTakeScreenshot={this.handleTakeScreenshot}
                     currentScheduleIndex={this.state.currentScheduleIndex}
                     toggleDisplayFinalsSchedule={this.toggleDisplayFinalsSchedule}
                     showFinalsSchedule={this.state.showFinalsSchedule}
--- a/apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
+++ b/apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
@@ -141,6 +141,14 @@ const SectionTableWrapped = (
     return <div>{component}</div>;
 };
 
+const LoadingMessage = () => {
+    return (
+        <Box sx={{ height: '100%', display: 'flex', justifyContent: 'center', alignItems: 'center' }}>
+            <img src={isDarkMode() ? darkModeLoadingGif : loadingGif} alt="Loading courses" />
+        </Box>
+    );
+};
+
 const ErrorMessage = () => {
     return (
         <Box sx={{ height: '100%', display: 'flex', justifyContent: 'center', alignItems: 'center' }}>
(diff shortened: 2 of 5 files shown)


apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
KevinWu098, line 1: Somewhat unrelated to this PR tbh, but this bug has been around for a sec, and might as well clean it up rn (xd)

EricPedley: Why was there a bug here and why does removing this line fix it?

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
```diff
@@ -162,134 +180,87 @@ export function CourseRenderPane() {
...
-            ge: formData.ge as GE,
-        };
-
         try {
-            // Query websoc for course information and populate gradescache
-            const [websocJsonResp, _] = await Promise.all([
-                websocQueryParams.units.includes(',')
-                    ? queryWebsocMultiple(websocQueryParams, 'units')
-                    : queryWebsoc(websocQueryParams),
-                Grades.populateGradesCache(gradesQueryParams),
```
EricPedley: Why is this line to populate the grades cache getting removed? Is this relevant to the color picker thing or a miscellaneous bug?

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
```diff
@@ -162,134 +180,87 @@ export function CourseRenderPane() {
...
+            }
+            setLoading(false);
             setError(false);
-            setCourseData(flattenSOCObject(websocJsonResp));
+            setWebsocResp(jsonResp);
+            setCourseData(flattenSOCObject(jsonResp));
         } catch (error) {
-            setError(true);
-        } finally {
             setLoading(false);
```
EricPedley: Why did you remove the `finally` clause? Does the order of `setLoading`, `setWebsocResp`, and `setCourseData` matter?

EricPedley (changes requested):
Results look good, just a few comments about the implementation

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
KevinWu098: Thanks for the catch, I think this was the result of a poor merge conflict resolution on my part 😅

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
KevinWu098: Thanks for the catch, I think this was the result of a poor merge conflict resolution on my part 😅 x2

EricPedley (approved):
🚢🚢🚢

PR #724, opened 2023-10-06
Add workflow to mark issues and PRs as stale

EricPedley:
## Summary
With this workflow, if an issue or PR has had no activity for 60 days, it will be marked as stale, which is basically a reminder for us to look at it again so nothing gets forgotten. It's 60 days because that's the default, if anyone has a compelling reason to change that I'd be down.
## Test Plan
Just wait for it to mark something as stale 🤷‍♂️

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,24 @@
+# This workflow warns and then closes issues and PRs that have had no activity for a specified amount of time.
+#
+# You can adjust the behavior by modifying this file.
+# For more information, see:
+# https://github.com/actions/stale
+name: Mark stale issues and pull requests
+
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


KevinWu098:
30 days wouldn't hurt tbh, would also probably be a good indicator since that's two "sprints" w/o any activity?

MinhxNguyen7:
> 30 days wouldn't hurt tbh, would also probably be a good indicator since that's two "sprints" w/o any activity?

I think that's not a bad idea for PRs because you shouldn't be idle on a PR for more than a month. Issues aren't necessarily stale after 30 days if no one's gotten to them, though. We don't move that quickly.

MinhxNguyen7 (approved):
LGTM!

PR #727, opened 2023-10-11
Zotistics graph pops up on hover

### LINKED ISSUE #672: RFC: Display Zotistics graph on hover
The other blue buttons lead to an external page, so it would make sense if this one does as well. The graph should display on hover, which would be more convenient as well IMO. We'd have to think about how it would work on mobile though.

Douglas-Hong:
## Summary
- The Zotistics graph now pops up when the user hovers over the Zotistics course info button. The popup closes when the mouse moves away from the popup or button.
- The `Popover` component has some weird interactions with `OnMouseLeave` and `OnMouseEnter` events. When the mouse leaves the area, sometimes the popup doesn't close. To avoid this, we can use the `Popper` component, which is basically a lower-level `Popover`.
- Some minor styling adjustments were made. The popup is a bit smaller so that it doesn't take much space, and so that the "View on Zotistics" link doesn't get cut off.

Demo:

https://github.com/icssc/AntAlmanac/assets/78244965/3235ea48-e42a-4c5a-827a-3091f155e3fd

hover

## Test Plan
- Verify that the graph pops up in the Search and Added tabs.
- Verify that the graph looks okay on mobile devices.
- Verify that the "View on Zotistics" link still works.

## Issues

Closes #672


Douglas-Hong:
@MinhxNguyen7 do you still think we should redirect to Zotistics when the button is clicked? If we do that, mobile users would not be able to see the graph popup, which they might prefer to see instead of having to go to Zotistics. Feel free to give other suggestions too.

EricPedley:
Never noticed this before but the x axis on the bar graph is unlabeled. Probably a good idea to do that at some point but it was also this way before this PR so I'll open it as a separate issue.

EricPedley (changes requested):
Code looks pretty good. Two requests form me here:
1. On mobile, when you tap the Zotistics button again, can it close the popup? I think it would be a better experience, feel free to push back here.
2. Can you update the patch notes in this PR?

EricPedley (approved):
🚀

PR #731, opened 2023-10-13
Updated the logic of parsing for letterGrades

### LINKED ISSUE #728: Zotistics graph has labels on the x axis.
Here's what it looks like rn:
image

And here's what I'm imagining it should look like:
image
There's questions marks on the last two because I don't even know what they are. My guess is pass and no pass but idk for sure.

JackyZzZz:
## Summary
Added Lettergrades to the x-axis of zotistics preview grade table.
image

## Test Plan
Make sure each Letter grade displays correctly under its associated bar. 
## Issues
Closes #728

## Future Followup
We might want to update the parsing logic once a while. I believe the parsing logic was correct before, but the data format from Zotistics might changed after that.

### DIFF
--- a/apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
+++ b/apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
@@ -21,6 +21,7 @@ import { SectionTableProps } from './SectionTable.types';
 import SectionTableBody from './SectionTableBody';
 import useColumnStore, { SECTION_TABLE_COLUMNS, type SectionTableColumn } from '$stores/ColumnStore';
 import analyticsEnum from '$lib/analytics';
+import { GlobalStyles } from '@mui/material';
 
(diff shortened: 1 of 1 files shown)


EricPedley:
@stevenguyukai this is a pretty simple one so I'm gonna leave the review completely to you. Message me on Discord if you have any questions.

Douglas-Hong:
It looks like #671, which was merged in very recently, added the labels for the x-axis in the graph:

image

@EricPedley maybe close this PR?

EricPedley:
I think we can salvage this one. @JackyZzZz can you just modify this PR so the only change is pulling the value for `name` out into `letterGrade` but preserving the string parsing logic as it currently is on `main`? Or just close this if you don't feel like doing that, since that basically reduces this PR to a two-line refactor.

EricPedley (changes requested):
See my previous comment

PR #742, opened 2023-10-18
Refactor functions used for ExportCalendar

### LINKED ISSUE #147: Download ICS: refactor to use calendarize helper

JacE070:
## Summary
This change does refactoring to have ExportCalendar use `getEventsInCalendar` instead of `getCurrentCourses`

## Test Plan
Tests passed. I also added some courses and custom events to test manully. Works fine.

## Issues

Closes https://github.com/icssc/AntAlmanac/issues/147

## Future Followup

1. Add `location` to customEvent when https://github.com/icssc/AntAlmanac/issues/249 is done.
2. Find a way to get the term for the corresponding custom event. Now we are using the default term.

### DIFF
--- a/apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
+++ b/apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
@@ -102,7 +102,7 @@ export interface Location {
     /**
      * If the location only applies on specific days, this is non-null.
      */
-    days?: string[];
+    days: string | null;
 }
 
 export interface CourseEvent extends CommonCalendarEvent {
@@ -138,6 +138,7 @@ export interface CourseEvent extends CommonCalendarEvent {
 export interface CustomEvent extends CommonCalendarEvent {
     customEventID: number;
     isCustomEvent: true;
+    days: string[];
 }
 
 export type CalendarEvent = CourseEvent | CustomEvent;
--- a/apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
+++ b/apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
@@ -102,27 +102,29 @@ export interface Location {
     /**
      * If the location only applies on specific days, this is non-null.
      */
-    days: string | null;
+    days?: string;
+}
+
+export interface FinalExam {
+    examStatus: 'NO_FINAL' | 'TBA_FINAL' | 'SCHEDULED_FINAL';
+    dayOfWeek: 'Sun' | 'Mon' | 'Tue' | 'Wed' | 'Thu' | 'Fri' | 'Sat' | null;
+    month: number | null;
+    day: number | null;
+    startTime: {
+        hour: number;
+        minute: number;
+    } | null;
+    endTime: {
+        hour: number;
+        minute: number;
+    } | null;
+    locations: Location[] | null;
 }
 
 export interface CourseEvent extends CommonCalendarEvent {
     locations: Location[];
     showLocationInfo: boolean;
-    finalExam: {
-        examStatus: 'NO_FINAL' | 'TBA_FINAL' | 'SCHEDULED_FINAL';
-        dayOfWeek: 'Sun' | 'Mon' | 'Tue' | 'Wed' | 'Thu' | 'Fri' | 'Sat' | null;
-        month: number | null;
-        day: number | null;
-        startTime: {
-            hour: number;
-            minute: number;
-        } | null;
-        endTime: {
-            hour: number;
-            minute: number;
-        } | null;
-        locations: Location[] | null;
-    };
+    finalExam: FinalExam;
     courseTitle: string;
     instructors: string[];
     isCustomEvent: false;
--- a/apps/antalmanac/src/lib/download.ts
+++ b/apps/antalmanac/src/lib/download.ts
@@ -110,16 +110,6 @@ export function dateToIcs(date: Date) {
     ] as YearMonthDay;
 }
 
-export function dateToDateTimeArray(date: Date) {
-    return [
-        date.getFullYear(),
-        date.getMonth() + 1, // Add 1 month since it is 0-indexed
-        date.getDate(),
-        date.getHours(),
-        date.getMinutes(),
-    ] as DateTimeArray;
-}
-
 /**
  * Get the start and end datetime of the first class.
  *
--- a/apps/antalmanac/src/lib/download.ts
+++ b/apps/antalmanac/src/lib/download.ts
@@ -1,10 +1,12 @@
-import type { HourMinute, WebsocSectionFinalExam } from 'peterportal-api-next-types';
+import { title } from 'node:process';
+import type { Course, HourMinute, WebsocSectionFinalExam } from 'peterportal-api-next-types';
 import { saveAs } from 'file-saver';
 import { createEvents, type EventAttributes } from 'ics';
 import { notNull } from './utils';
 import { openSnackbar } from '$actions/AppStoreActions';
 import analyticsEnum, { logAnalytics } from '$lib/analytics';
-import { termData } from '$lib/termData';
+import { getDefaultTerm, termData } from '$lib/termData';
+import { CourseEvent, CustomEvent } from '$components/Calendar/CourseCalendarEvent';
 import AppStore from '$stores/AppStore';
 
 export const quarterStartDates = Object.fromEntries(
@@ -108,6 +110,16 @@ export function dateToIcs(date: Date) {
     ] as YearMonthDay;
 }
 
+export function dateToDateTimeArray(date: Date) {
+    return [
+        date.getFullYear(),
+        date.getMonth() + 1, // Add 1 month since it is 0-indexed
+        date.getDate(),
+        date.getHours(),
+        date.getMinutes(),
+    ] as DateTimeArray;
+}
+
 /**
  * Get the start and end datetime of the first class.
  *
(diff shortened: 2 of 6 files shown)


apps/antalmanac/src/lib/download.ts
```diff
@@ -1,10 +1,12 @@
-import type { HourMinute, WebsocSectionFinalExam } from 'peterportal-api-next-types';
+import { title } from 'node:process';
```
EricPedley: Why is this here?

apps/antalmanac/tests/download-ics.test.ts
```diff
@@ -83,13 +66,25 @@ describe('download-ics', () => {
...
+                // TODO: Add location to custom events, waiting for https://github.com/icssc/AntAlmanac/issues/249
+                // location: `${location.building} ${location.room}`,
+                start: [2023, 9, 29, 1, 2],
+                end: [2023, 9, 29, 3, 4],
+                recurrenceRule: 'FREQ=WEEKLY;BYDAY=MO,WE,FR;INTERVAL=1;COUNT=31',
+            },
         ];
 
         const result = getEventsFromCourses(courses);
+        console.log(result);
```
EricPedley: If we're asserting this is equal to the expected one, I don't think we need to print it out here.

EricPedley (changes requested):
mostly looks good, just two minor comments

apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
```diff
@@ -102,7 +102,7 @@ export interface Location {
     /**
      * If the location only applies on specific days, this is non-null.
      */
-    days?: string[];
+    days: string | null;
```
ap0nia:
```suggestion
    days?: string;
```
I'm under the impression that this is artificially nullable; I'm pretty sure it should be safe to make it optional, i.e. it exists or it's undefined. The only way it would be null is if we ever set it as null ourselves (which I noticed below, which we shouldn't do if we want to keep the interface simpler)

apps/antalmanac/src/lib/download.ts
```diff
@@ -246,74 +248,84 @@ export function getRRule(bydays: string[], quarter: string) {
...
-            section: { sectionType, instructors, meetings, finalExam },
-        } = course;
-
-        const courseEvents: EventAttributes[] = meetings
-            .map((meeting) => {
-                if (meeting.timeIsTBA) {
-                    return;
+export function getEventsFromCourses(events = AppStore.getEventsInCalendar()): EventAttributes[] {
+    const calendarEvents = events.flatMap((event) => {
+        if (!event.isCustomEvent) {
```
ap0nia: Could you swap the two blocks of conditional logic? It can be easier to read "positive" conditions.

For instance, given the two conditions: "if today is not raining" vs. "if today is sunny", I'd prefer the latter. It depends on the scenario, but I think it makes sense here. So the pseudocode would be "if it's a custom event, process the item as a custom event, otherwise process it as a course event".

```suggestion
        if (event.isCustomEvent) {
```

apps/antalmanac/src/stores/calendarizeHelpers.ts
```diff
@@ -10,7 +10,7 @@ const FINALS_WEEK_DAYS = ['Sat', 'Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri'];
 
 export function getLocation(location: string): Location {
     const [building = '', room = ''] = location.split(' ');
-    return { building, room };
+    return { building, room, days: null };
```
ap0nia: As mentioned above, days should just be optional and it can be omitted from the object altogether (which is synonymous with being undefined).

```suggestion
    return { building, room };
```

apps/antalmanac/tests/download-ics.test.ts
ap0nia:
```suggestion
```

ap0nia: I'll make it a code suggestion 👍

apps/antalmanac/tests/download-ics.test.ts
ap0nia, line 1: Something to note about testing is that we can also leverage [Vitests's snapshots](https://vitest.dev/guide/snapshot.html). Essentially, you can generate and save the expected output to version control, and the test will always compare its output with the saved one. This is basically what we're trying to accomplish because the output of `calendarize` should remain consistent into the future.

You don't have to implement this, but just an idea (maybe it could be made into an issue? @EricPedley)

apps/antalmanac/src/stores/calendarizeHelpers.ts
```diff
@@ -135,6 +137,7 @@ export function calendarizeCustomEvents(currentCustomEvents: RepeatingCustomEven
                 isCustomEvent: true,
                 end: new Date(2018, 0, dayIndex, endHour, endMin),
                 title: customEvent.title,
+                days: days,
```
ap0nia: JavaScript shorthand for defining properties with the same name as the variable name.

```suggestion
                days
```

apps/antalmanac/src/lib/download.ts
```diff
@@ -153,7 +155,7 @@ export function getFirstClass(
  * // examEnd = [2019, 6, 7, 12, 30]
  * ```
  */
-export function getExamTime(exam: WebsocSectionFinalExam, year: number): [DateTimeArray, DateTimeArray] | [] {
+export function getExamTime(exam: CourseEvent['finalExam'], year: number): [DateTimeArray, DateTimeArray] | [] {
```
ap0nia: Could you go back into the `CourseCalendarEvent` file and extract the type of `CourseEvent['finalExam']` into it's own type? It's convenient to have direct access to the types you need instead of accessing properties of a parent type.

```suggestion
export function getExamTime(exam: FinalExam, year: number): [DateTimeArray, DateTimeArray] | [] {
```

ap0nia (changes requested):
Looks solid! Suggested some improvements mostly concerning TS stuff.

apps/antalmanac/src/lib/download.ts
JacE070: Oops, just unintentional auto-import. Removed.

apps/antalmanac/src/stores/calendarizeHelpers.ts
JacE070: First time, and good to know this!

apps/antalmanac/tests/download-ics.test.ts
JacE070: Nice idea, I will take a look into this.

apps/antalmanac/tests/download-ics.test.ts
```diff
@@ -2,68 +2,51 @@ import { EventAttributes } from 'ics';
...
+                start: new Date(2023, 9, 29, 1, 2),
+                end: new Date(2023, 9, 29, 3, 4),
+                title: 'placeholderCustomEventTitle',
+                customEventID: 123,
+                isCustomEvent: true,
+                days: ['M', 'W', 'F'],
+            },
         ];
 
         const expectedResult: EventAttributes[] = [
```
ap0nia: You don't need this if you're using a snapshot. It should be safe to remove.

apps/antalmanac/tests/download-ics.test.ts
```diff
@@ -2,68 +2,51 @@ import { EventAttributes } from 'ics';
 import type { Schedule } from '@packages/antalmanac-types';
 import { describe, test, expect } from 'vitest';
 import { getEventsFromCourses } from '$lib/download';
+import { CourseEvent, CustomEvent } from '$components/Calendar/CourseCalendarEvent';
 
 describe('download-ics', () => {
     test('converts schedule courses to events for the ics library', () => {
-        const courses: Schedule['courses'] = [
+        const courses: (CourseEvent | CustomEvent)[] = [
```
ap0nia: We have a type that represents a union of all our event types.

```suggestion
        const courses: CalendarEvent[] = [
```

apps/antalmanac/src/stores/calendarizeHelpers.ts
```diff
@@ -121,7 +123,7 @@ export function calendarizeFinals(currentCourses: ScheduleCourse[] = []): Course
 export function calendarizeCustomEvents(currentCustomEvents: RepeatingCustomEvent[] = []): CustomEvent[] {
     return currentCustomEvents.flatMap((customEvent) => {
         const dayIndiciesOcurring = customEvent.days.map((day, index) => (day ? index : undefined)).filter(notNull);
-
+        const days = dayIndiciesOcurring.map((dayIndex) => COURSE_WEEK_DAYS[dayIndex]);
```
ap0nia, line 131: In the other helper functions, I included comments describing how these data structures look, e.g. here https://github.com/icssc/AntAlmanac/blob/85ff4a87e04507373e163ee65559caa67a8eaaaa/apps/antalmanac/src/stores/calendarizeHelpers.ts#L93

Personally, I think it's useful to see examples of the data being generated nearby the variable it's assigned to. I think it would be good to include some documentation here as well (you can copy/paste the previous comments too, since the data should be the same/similar). Let me know if you or @EricPedley have alternative opinions on this.

For example:
```ts
 /**
  * Only include the day indices that the final is occurring.
  *
  * @example [false, false, false, true, false, true, false] -> [3, 5]
  */
const dayIndicesOcurring = weekdaysOccurring.map((day, index) => (day ? index : undefined)).filter(notNull);
```

ap0nia (changes requested):
Solid improvements 👏 
It's basically ready to approve, but I'd appreciate some final touches.

TLDR:
- Add some comments
- Use a more succinct type
- Remove an unused variable

apps/antalmanac/src/stores/calendarizeHelpers.ts
JacE070: Sounds solid! Yeah the comment about the data structure will be very helpful.

ap0nia (approved):
Good work 💯

EricPedley (approved):

PR #749, opened 2023-10-19
Open error snackbar when we get an error during search

EricPedley:
## Summary
Open an error snackbar and log to console when we get an error during search. This shouldn't trigger when we legitimately can't find 
## Test Plan
Search up a class not offered this quarter, and it shouldn't open the error snackbar. Rn the API is flaky so if you search enough times you'll be able to see an error, and verify on the network tab the request didn't work.


ap0nia (approved):
Epic :rocket:

PR #752, opened 2023-10-21
fix: CSS Cleanup

### LINKED ISSUE #743: Uneven margin around center bar
Once you see it, you can't unsee it 😢  
Also, on the left it's rounded but it's not rounded on the right, but the rounding on the left kinda looks funny when it hits the component below it, so really they should both probably only have rounding on the top corners.
image

KevinWu098:
## Summary
Did minor CSS cleanup around the Calendar and Right Pane margins/paddings to (hopefully) remove styling inconsistencies.

## Test Plan
1. As this change is entirely visual, please confirm that the UX hasn't changed across a variety of screen sizes.

## Issues
Closes #743


EricPedley (approved):
Nice job!

PR #754, opened 2023-10-22
fix: Column Toggle Select Element Positioning

### LINKED ISSUE #753: Column Toggle Select is Positioned Wrong
`visibility: 'hidden'` FormControl Element:
Screenshot 2023-10-21 at 7 31 21 PM

Select Element:
Screenshot 2023-10-21 at 7 32 30 PM

A Popover can be used here to fix this!

KevinWu098:
## Summary
The Select element was positioned to a `visibility: 'hidden'` FormControl, not the IconButton as it should have. This creates an off-centered visual experience.

Before:

After:
Screenshot 2023-10-21 at 7 35 03 PM

## Test Plan
1. This change is primarily visual in nature, so confirm that the positioning is correct on mobile and on desktop
2. Additionally, there was some bugginess while adding the Popover and Select, so make sure that selecting and deselecting elements doesn't reposition the element (it shouldn't)

## Issues
Closes #753

### DIFF
--- a/apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
+++ b/apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
@@ -67,7 +67,7 @@ export function ColumnToggleButton() {
         store.setSelectedColumns,
     ]);
     const [anchorEl, setAnchorEl] = useState<HTMLElement>();
-    const open = useMemo(() => Boolean(anchorEl), [anchorEl]);
+    const open = Boolean(anchorEl);
 
     const handleChange = useCallback(
         (e: SelectChangeEvent<SectionTableColumn[]>) => {
--- a/apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
+++ b/apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useMemo, useState } from 'react';
+import { useCallback, useMemo, useState } from 'react';
 import {
     Box,
     Checkbox,
@@ -10,6 +10,7 @@ import {
     Tooltip,
     type SelectChangeEvent,
     type SxProps,
+    Popover,
 } from '@mui/material';
 import { ArrowBack, Visibility, Refresh } from '@mui/icons-material';
 import { useColumnStore, SECTION_TABLE_COLUMNS, type SectionTableColumn } from '$stores/ColumnStore';
@@ -65,20 +66,24 @@ export function ColumnToggleButton() {
         store.selectedColumns,
         store.setSelectedColumns,
     ]);
-    const [open, setOpen] = useState(false);
+    const [anchorEl, setAnchorEl] = useState<HTMLElement>();
+    const open = useMemo(() => Boolean(anchorEl), [anchorEl]);
 
-    const handleChange = useCallback((e: SelectChangeEvent<SectionTableColumn[]>) => {
-        if (typeof e.target.value !== 'string') {
-            setSelectedColumns(e.target.value);
-        }
+    const handleChange = useCallback(
+        (e: SelectChangeEvent<SectionTableColumn[]>) => {
+            if (typeof e.target.value !== 'string') {
+                setSelectedColumns(e.target.value);
+            }
+        },
+        [setSelectedColumns]
+    );
+
+    const handleClick = useCallback((event: React.MouseEvent<HTMLButtonElement>) => {
+        setAnchorEl(event.currentTarget);
     }, []);
 
-    const handleOpen = useCallback(() => {
-        setOpen(true);
-    }, [setOpen]);
-
     const handleClose = useCallback(() => {
-        setOpen(false);
+        setAnchorEl(undefined);
     }, []);
 
     const selectedColumnNames = useMemo(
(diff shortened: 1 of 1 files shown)


apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
```diff
@@ -89,28 +94,31 @@ export function ColumnToggleButton() {
...
+            <Popover open={open} anchorEl={anchorEl} onClose={handleClose} sx={{ visibility: 'hidden' }}>
+                <FormControl>
+                    <Select
+                        multiple
+                        value={selectedColumnNames}
+                        open={open}
+                        onChange={handleChange}
+                        onClose={handleClose}
+                        renderValue={renderEmptySelectValue}
+                        MenuProps={{ anchorEl }}
```
KevinWu098, line 111: `MenuProps={{ anchorEl }}` resolves a Select bug, loosely documented and described in [a chain of duplicate/related Github Issues](https://github.com/mui/material-ui/issues/19245)...

TL;DR anchorEl stops it from repositioning

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
```diff
@@ -65,20 +66,24 @@ export function ColumnToggleButton() {
         store.selectedColumns,
         store.setSelectedColumns,
     ]);
-    const [open, setOpen] = useState(false);
+    const [anchorEl, setAnchorEl] = useState<HTMLElement>();
+    const open = useMemo(() => Boolean(anchorEl), [anchorEl]);
```
ap0nia:
```suggestion
    const open = Boolean(anchorEl)
```
This isn't wrong, but I don't think it's worth memo-izing. Good job for being aware of this, you're becoming a true React developer! 💯 

Another fun :tm: part of React is having incurring mental overhead when deciding whether or not to memo-ize things because you don't know how many times it'll be recalculated and realizing that you'll still get a poorly optimized app anyways.

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
```diff
@@ -65,20 +66,24 @@ export function ColumnToggleButton() {
...
 
-    const handleChange = useCallback((e: SelectChangeEvent<SectionTableColumn[]>) => {
-        if (typeof e.target.value !== 'string') {
-            setSelectedColumns(e.target.value);
-        }
-    }, []);
+    const handleChange = useCallback(
+        (e: SelectChangeEvent<SectionTableColumn[]>) => {
```
ap0nia, line 73: If you check the type signature for MUI's select `onChange` prop, the callback function should receive the value itself as the second argument. It's easier to work with the value itself if that's all you need, rather than extract it from the event.

```suggestion
        (e: unknown, value: SectionTableColumn[]) => {
```

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
```diff
@@ -65,20 +66,24 @@ export function ColumnToggleButton() {
...
+                setSelectedColumns(e.target.value);
+            }
+        },
+        [setSelectedColumns]
+    );
 
-    const handleOpen = useCallback(() => {
-        setOpen(true);
-    }, [setOpen]);
+    const handleClick = useCallback((event: React.MouseEvent<HTMLButtonElement>) => {
```
ap0nia, line 81: You can generalize this type so it's more succinct, since we don't really need accuracy in this function. A similiar, maybe synonymous type might be `React.SyntheticEvent`

```suggestion
    const handleClick = useCallback((event: React.MouseEvent) => {
```

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
ap0nia: I love the part when React (and its ecosystem) works the way it looks like it should.

image

ap0nia (dismissed):
Epic work 👏  Just some minor adjustments and we should be good to go.

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
KevinWu098: Argument of type 'EventTarget & Element' is not assignable to parameter of type 'SetStateAction<HTMLElement | undefined>'.
  Type 'EventTarget & Element' is missing the following properties from type 'HTMLElement': accessKey, accessKeyLabel, autocapitalize, dir, and 122 more.ts(2345)

```typescript
    const handleClick = useCallback((event: React.MouseEvent) => {
        setAnchorEl(event.currentTarget);
    }, []);
```

I think it's probably okay to specify HTMLButtonElement here >.<

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
KevinWu098: 😭 i can't make this work so feel free to take a crack at it; I'm getting spit nasty errors (and also it just doesn't work idk)

MinhxNguyen7 (approved):
LGTM!

PR #766, opened 2023-10-27
Grades error does not fail search

### LINKED ISSUE #745: Error message when class search fails
Instead of just saying there are no classes

### LINKED ISSUE #746: Display results even when grades endpoint fails

MinhxNguyen7:
## Summary
Display an error message but do not fail search when the grades API fails

## Test Plan
1) Block the grades API using FireFox devtools or change the graphql API endpoint.
2) Search and see that course information (except for GPA) still show up as normal

## Issues

Closes #746


ap0nia (approved):
I think it looks perfectly fine, but I'm not sure if this review is intended to only be reviewed by @stevenguyukai

PR #770, opened 2023-10-30
Fix weird columns behavior

### LINKED ISSUE #737: Strange behavior for show/hide columns dropdown
https://github.com/icssc/AntAlmanac/assets/8922227/85a4e707-2a5c-4aea-bb92-92aafa0d944a

MinhxNguyen7:
## Summary
Columns show/hide didn't work properly, per #737. 

CoPilot wrote the code, and I missed the fact that it was reading `prevState` instead of the current value. All my fault, of course, but we should've caught this in review.

## Test Plan
Check that the column toggles work properly in the two panes.

## Issues

Closes #737


KevinWu098 (approved):
LGTM!

PR #773, opened 2023-11-02
chore: move header's files around

ap0nia:
# Summary
The component directory's structure kinda sucks, and the header is the easiest to alleviate.


apps/antalmanac/src/components/Header/index.tsx
```diff
@@ -0,0 +1 @@
+export { default } from './Header';
```
ap0nia, line 1: Barrel export from an `index` file allows you to import from the directory, i.e.
```ts
import Header from '$components/Header'

// vs

import Header from '$components/Header/index'
import Header from '$components/Header/Header'
```

KevinWu098 (approved):
LGTM

PR #775, opened 2023-11-02
feat: Add 24 Hour Time Format

KevinWu098:
## Summary
1. Refactor the Calendar Component
2. Update `translate24To12HourTime` to `formatTimes`, adding a `timeFormat` param so it knows which format to return the times in. Applies to Right Pane (Search/Added)
3. Add additional props to the RBC component to enable 24 hour time for both gutter and event labels
4. Refactor Theme/Settings Menu
5. Add Theme and Time to the (new) Settings Menu

## Screenshots

Light Mode, Full App:
24 Hour Format
Dark Mode, Calendar:
Screenshot 2023-11-09 at 2 18 13 AM

chrome-capture-2023-10-9

## Test Plan
1. Light/Dark Mode
2. Existing Calendar functionality should be unchanged (swapping schedules, clicking events, creating events, etc)
3. Toggling between time formats should be seamless

## Issues
Closes #637

## Future Followup
1. Implement a newly redesigned settings menu!

PR #776, opened 2023-11-02
feat: Add Dates to Finals Calendar Header

### LINKED ISSUE #758: RFC: Finals date
It would be useful to see finals' dates in the finals tab in the calendar (next to the day), at least in the desktop version. I haven't looked at whether that's part of the API/WebSOC, but I think it should be pretty easy if it is. 

Maybe we can count back from the end dates of the quarters. Note: pay attention to how Summer session works.

Displaying the date and day with React Big Calendar *should* be as simple as passing flags in.

KevinWu098:
# This PR is branched from #775, which is kind of dumb of me, but I'll fix that later or something...

## Summary
1. When Finals mode is toggled on, the calendar will display the dates
2. Currently, this is based on a new field in `termData`, finalsStartDate + 5 days. I'm not entirely sure why this requires an additional 5 days, but it does, and it's currently hard coded in.

chrome-capture-2023-10-2

## Test Plan
1. Check functionality for finals on weekends... will ask Eddy for a test course (if it exists)

## Issues
Closes #758

### DIFF
--- a/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
+++ b/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
@@ -2,10 +2,8 @@ import 'react-big-calendar/lib/css/react-big-calendar.css';
 import './calendar.css';
 
 import { Box, ClickAwayListener, Popper } from '@material-ui/core';
-import { Theme, withStyles } from '@material-ui/core/styles';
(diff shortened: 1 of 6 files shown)


KevinWu098:
Closing because it'll be easier to move one line of code changes to a proper branch

PR #779, opened 2023-11-02
fix: remove duplicate buildings from the catalogue

ap0nia:
## Summary
Remove duplicate buildings


EricPedley (approved):
I manually confirmed that all the locations deleted still work on the map search. LGTM!

PR #783, opened 2023-11-03
Hotfix: swap mapbox tiles for free ones

EricPedley:
## Summary
We ran past the mapbox free tier and it's still early in the month.
## Test Plan
Make sure the map still loads


MinhxNguyen7 (approved):
LGTM!

Issue #784, opened 2023-11-03
Some waypoints don't show up in map


MinhxNguyen7:
The second waypoint is missing in this schedule. See schedule "map-error"

<img width="717" alt="image" src="https://github.com/icssc/AntAlmanac/assets/64875104/baf685b1-eac9-43cd-a5ad-01661a7c36d2">

KevinWu098:
```typescript
    const validBuildingCodes = [...uniqueBuildingCodes].filter(
        (buildingCode) => buildingCatalogue[locationIds[buildingCode]] != null
    );
```

This is the reason why (from `Map.tsx`) ^^

The changes in #779, although confirmed to still work for all of the locations in map search, didn't take into account that `location_ids.ts` are the apparent source of truth for which ID in the list is the one we use. There may be more discrepancies still, but this is likely the source of the issue here

KevinWu098:
Should we revert #779 in this instance? Or just push a patch for it?

MinhxNguyen7:
Fixed by #785

PR #785, opened 2023-11-03
Revert "fix: remove duplicate buildings from the catalogue"

KevinWu098:
Reverts icssc/AntAlmanac#779

# Reasoning:
```typescript
    const validBuildingCodes = [...uniqueBuildingCodes].filter(
        (buildingCode) => buildingCatalogue[locationIds[buildingCode]] != null
    );
```

This is the reason why (from `Map.tsx`) ^^

The changes in #779, although confirmed to still work for all of the locations in map search, didn't take into account that `location_ids.ts` are the apparent source of truth for which ID in the list is the one we use. There may be more discrepancies still, but this is likely the source of the issue here

[Originally noted in #784](https://github.com/icssc/AntAlmanac/issues/784#issuecomment-1792998596) ^^

Screenshot 2023-11-03 at 1 47 57 PM


MinhxNguyen7 (approved):
👍

PR #786, opened 2023-11-04
Optimized the logic of focus

### LINKED ISSUE #782: Auto-focusing Search on Tablets
#767 added an auto-focus to the Search Bar every time Search pane is opened; However, on mobile tablets, that means the keyboard will constantly pop up, preventing convenient views of the Calendar

https://github.com/icssc/AntAlmanac/assets/100006999/5778ce9f-7bac-4b89-8b12-2d4f4c91394c

No one has complained about it yet, but I'd imagine this wouldn't be a great experience regardless...

JackyZzZz:
## Summary
This change optimized the auto-focus feature for mobile devices through the react-device-detect library.
## Test Plan
Simulated the behavior of auto-focus on mobile devices in Chrome Dev Tools, works as expected.
## Issues
Closes #782


KevinWu098:
I think this is a PR worth pinging @EricPedley and @ap0nia on! We currently "detect" devices based on view width, but that's kind of a poor measuring stick...

Would this more properly standardize our `isMobile` (or whatever the helper function is called) across the app? I'm not super sure about the procedure for adding in new dependencies, so I'd love y'alls thoughts.

At first glance (haven't tested yet), it seems like it would be useful not just here, but essentially anywhere in the app we use view width as a proxy for device (like with the recent Zotistics graph hover)

JackyZzZz:
Yeah bruh. I was planning on using width to distinguish between mobile and computers since it's the most straightforward way, but I did some research and it seems like that's very unstable across devices and platforms. And I figured out that react actually has this detect feature in their library. So I thought this might be a better approach lol

EricPedley:
new library sounds good, we'll just do another PR to change it everywhere else to not use screen size.

ap0nia:
No, we don't need to do the latter step. The other times we use it are just regular media queries. This is a specific instance when we actually care about the device. If it's possible, could we look into using ua-parser-js? https://www.npmjs.com/package/ua-parser-js

This is the actual library that does the parsing, and the react version is just a shallow wrapper around.

EricPedley:
Idk man don't we like having a wrapper?

ap0nia:
We want a "react-specific" version to do this?

```ts
new UAParser().getDevice().type === 'mobile'
```

EricPedley:
Oh yeah ur right it is kinda dumb that would need to be react specific

JackyZzZz:
Okay so I digged into it a little bit. The ua-parser-js works in most cases as it returns a type of mobile or tablets which we should not auto-focus. But when it comes to certain tablets model like Ipad Air or Ipad Pro, the type returned by the library is actually undefined... So that we wouldn't be able to distinguish them.  I think when I test the isMobile from react library it works fine but it is indeed a little dumb to stick with a certain version. However, the ua-parser-js also have some constraints dealing with certain models. Any suggestions maybe guys? At least I think it works with most mobile types, but not for certain model like ipads

ap0nia:
This is the full logic that they use: https://github.com/duskload/react-device-detect/blob/master/src/lib/selectors.js#L11

They check that it's either a mobile device, tablet, or ipad.

The full logic would look something like this.
```ts
const ua = new UAParser()
const device = ua.getDevice()
const isMobile = device.type === 'mobile' || device.type === 'tablet' || isPad()

function isIpad() {
  if (typeof window === 'undefined') {
    return false
  }
return   (/iPad|iPhone|iPod/.test(window.navigator.platform) ||
      (window.navigator.platform === 'MacIntel' && window.navigator.maxTouchPoints > 1)) &&
    !window.MSStream
}
```

JackyZzZz:
Okay just pushed the ua-parser way. However, idk I simulated the react library way but it didn't really workout lol. I still implemented the ua-parser since it detects most devices okay. For the ipad case it's annoying because right now it's OS becomes Mac too. The way I am currently checking is if it supports touch capabiity, if it is then we will identify it as mobile.

KevinWu098:
I just attended a fantastic web accessibility workshop by Alex and Eddy! TLDR; auto-focus is not accessibility-friendly, so I think we should strongly consider reverting the original auto-focus PR.

(It also saves us the work of handling tablets)

EricPedley:
Can u elaborate on how it isn't accessible?

KevinWu098:
> Can u elaborate on how it isn't accessible?

The issue mainly stems from the fact that autofocus creates a jarring effect for screen readers -- users end up in a location they didn't expect/navigate to

ap0nia:
I agree with Kevin. From a developer perspective, I'd generally consider it a goal to make the app act in a predictable manner. This is basically already an impossible challenge because of React's consistently unpredictable rendering. Adding logic that arbitrarily abducts control from the user will invariable end up with a edge cases because of the inherently jarring behavior.

JackyZzZz:
@EricPedley What do you think boss

EricPedley (approved):
I saw we just merge this and worry about whether or not to remove autofocus later.

KevinWu098:
> I saw we just merge this and worry about whether or not to remove autofocus later.

I'm still of the opinion that this PR shouldn't be merged; rather, simply reverting the autofocus feels more straightforward.

We added a feature -> it's not accessible -> we patch a fix to fix a semi-related issue to the auto-focus -> if we end up removing auto-focus, we've wasted time and energy coming up with a solution for a problem that shouldn't have existed

EricPedley:
ok how about we just add a setting to enable autofocus and make it disabled by default? IMO this feature is hella nice so I don't want to see it removed.

KevinWu098:
> ok how about we just add a setting to enable autofocus and make it disabled by default? IMO this feature is hella nice so I don't want to see it removed.

Smart compromise; I'm onboard with that. Let's merge this in then :)

PR #787, opened 2023-11-04
Add Winter 2024 to Term Data

KevinWu098:
## Summary
1. Add winter 2024 to term data so we can search for it (API already has the data 🤩) 

## Test Plan
1. It's one line diff xD, just make sure [I didn't typo or something](https://www.reg.uci.edu/calendars/quarterly/2023-2024/quarterly23-24.html)
2. Actually make sure that courses are addable and searchable correctly

## Issues
Closes a very quick user request from the Discord


MinhxNguyen7 (approved):
What could go wrong?

PR #788, opened 2023-11-04
Winter 2024

MinhxNguyen7:
## Summary
Term data for winter 2024

## Test Plan
Check that the term shows up and it fetches the right term's classes.

## Future Followup
Use PP API to get data instead of hardcoding


KevinWu098:
#787 Dupe!

KevinWu098:
#787 dupe

PR #794, opened 2023-11-08
fix: update the snapshot because the default term for custom events c…

### LINKED ISSUE #793: Download ICS test failing
The event that is getting produced for a fall quarter course has the winter quarter start date for some reason, and this started after we updated the term list to include winter.
image

ap0nia:
# Summary
Update the snapshot because `getDefaultTerm` returns a different term.

# Technical Details
In #787, a new term was added to the `termData` list in `$lib/termData`. This results in the [getDefaultTerm](https://github.com/icssc/AntAlmanac/blob/main/apps/antalmanac/src/lib/termData.ts#L80) function returning a different term (the new one). All custom events are assigned this default term by [this line of code](https://github.com/icssc/AntAlmanac/blob/main/apps/antalmanac/src/lib/download.ts#L257), so the snapshot should be updated since the calculations are different depending on the term returned by the function.


KevinWu098 (approved):
All tests run clean, so I'll approve and merge, but we should probably discuss Calendar endpoint / less hard-coding during this week's meeting

PR #806, opened 2023-11-16
AANTS Redesign

Douglas-Hong:
## Summary
- This is just a WIP. This currently sends an email from the AntAlmanac account to itself. To test sending an email, click on the "SES Test" button next to the "Past Enrollment" button.

## Tasks
- [x] Set up backend endpoint.
- [ ] Resolve Access Denied error. I had to add some IAM permissions for this staging instance in order to access SES, so we need to make sure everything else has SES access.
- [ ] Add ability to send email to any user, not just any verified SES email address. This requires an application for a production SES account.

## Issues

Closes #


EricPedley:
gonna close this cause we're pivoting to web push notifications. If we ever want to add SMS later we can reference this.

PR #808, opened 2023-11-23
fix: Swap Popover to Tooltip for Restrictions Cell

### LINKED ISSUE #798: Restrictions Column Stops Scrolling
If you hover over a cell in the restrictions column, it stops the user from scrolling. In order to continue scrolling, the user has to click away from the cell and then scroll. Ideally, the user can continue scrolling without stopping. If the user scrolls away, make sure the popup from the cell in the restrictions column disappears.

KevinWu098:
## Summary
Swapped out a Popover component for a Tooltip

## Test Plan
1. Make sure that functionality is now as expected; Scrolling, non-focusing, and other expected behaviors should be working

## Issues
Closes #798


Douglas-Hong (approved):
LGTM! Thanks for fixing this!

PR #810, opened 2023-11-23
feat: Update Settings Menu

### LINKED ISSUE #805: Create new Settings Menu
Screenshot 2023-11-16 at 10 45 04 AM

Do that ^^

KevinWu098:
## Summary
1. Implemented a new settings menu using the `Drawer` component
2. `ButtonGroup` replaces the previous implementation which used the `RadioGroup` component

Settings Menu Gif

Screenshot 2023-11-26 at 10 34 40 PM

## Test Plan
1. Mobile and Desktop Responsiveness
3. The underlying functions haven't changed, but check that the buttons do work as intended

## Issues
Closes #805


MinhxNguyen7 (changes requested):
Looks awesome, I love it. 

While you're at it, can you  move the settings out to a Zustand store? We want to be removing, not adding things to AppStore.

Also, if you're in the mood, you could make the settings persist with localstorage.

MinhxNguyen7 (changes requested):
Looks awesome, I love it. 

While you're at it, can you move the settings out to a Zustand store? We want to be removing, not adding things to AppStore.

Also, if you're in the mood, you could make the settings persist with localstorage.

KevinWu098:
> Looks awesome, I love it.
> 
> While you're at it, can you move the settings out to a Zustand store? We want to be removing, not adding things to AppStore.
> 
> Also, if you're in the mood, you could make the settings persist with localstorage.

Will do :), ✨ Zustand ✨.Also, I believe the settings are stored in localstorage already!

apps/antalmanac/src/providers/Theme.tsx
```diff
@@ -12,25 +11,20 @@ interface Props {
...
-            setDarkMode(isDarkMode());
-        });
         window.matchMedia('(prefers-color-scheme: dark)').addEventListener('change', (e) => {
-            if (AppStore.getTheme() === 'auto') {
-                setDarkMode(e.matches);
-            }
+            setTheme(e.matches ? 'dark' : 'light');
         });
-    }, []);
+    }, [setTheme, theme]);
```
ap0nia:
```suggestion
    }, [theme]);
```

apps/antalmanac/src/providers/Theme.tsx
```diff
@@ -12,25 +11,20 @@ interface Props {
...
 
     useEffect(() => {
-        AppStore.on('themeToggle', () => {
-            setDarkMode(isDarkMode());
-        });
         window.matchMedia('(prefers-color-scheme: dark)').addEventListener('change', (e) => {
-            if (AppStore.getTheme() === 'auto') {
-                setDarkMode(e.matches);
-            }
+            setTheme(e.matches ? 'dark' : 'light');
```
ap0nia: Could you fix this? You have to remove the event listener once the component unmounts. I don't have some of the TypeScript definitions on hand, so you'll have to find them.

```tsx
const mediaQueryList =  window.matchMedia('(prefers-color-scheme: dark)')

const onChange = (e) => {
  setTheme(e.matches ? 'dark' : 'light');
}

mediaQueryList.addEventListener('change', onChange)

return () => {
  mediaQueryList.removeEventListener('change', onChange)
}
```

apps/antalmanac/src/components/Header/SettingsMenu.tsx
```diff
@@ -1,96 +1,142 @@
...
-import AppStore from '$stores/AppStore';
 import { useTimeFormatStore } from '$stores/TimeStore';
+import { useThemeStore } from '$stores/ThemeStore';
 
-function SettingsMenu() {
-    const [anchorEl, setAnchorEl] = useState<HTMLElement>();
-    const [theme, setTheme] = useState(AppStore.getTheme());
+function ThemeMenu() {
+    const handleThemeChange = (event: React.MouseEvent<HTMLButtonElement>) => {
+        useThemeStore.getState().setTheme(event.currentTarget.value);
```
ap0nia: Wait, that's illegal. Never invoke hooks outside of the top level of a function. It might work because Zustand is "hook-like" and not a real hook? But don't do that regardless.

Apply this to the other relevant handlers in the file.

```tsx
function ThemeMenu() {
  const setTheme = useThemeStore((store) => store.setTheme)

   const handleThemeChange = useCallback((event: React.MouseEvent<HTMLButtonElement>) => {
     setTheme(event.currentTarget.value);
  };
}, [])
```

P.S.
I didn't come up with that rule, some brain-rotted, React-andy troglodyte soydev did.

apps/antalmanac/src/components/Header/SettingsMenu.tsx
```diff
@@ -1,96 +1,142 @@
...
+                </Button>
+            </ButtonGroup>
+        </Box>
+    );
+}
 
-    const handleTimeChange = (event: ChangeEvent<HTMLInputElement>) => {
-        useTimeFormatStore.getState().setTimeFormat(event.target.value == 'true');
+function TimeMenu() {
+    const handleTimeChange = (event: React.MouseEvent<HTMLButtonElement>) => {
```
ap0nia: Same advice I had above

apps/antalmanac/src/components/Header/SettingsMenu.tsx
```diff
@@ -1,96 +1,142 @@
...
-    }, []);
+            <ButtonGroup style={{ display: 'flex', placeContent: 'center', width: '100%' }}>
+                <Button
+                    style={{
+                        padding: '1rem 2rem',
+                        borderRadius: '12px 0px 0px 12px',
+                        width: '100%',
+                        fontSize: '12px',
+                    }}
+                    value="false"
```
ap0nia, line 114: If you have true/false, why not use a [switch](https://mui.com/material-ui/react-switch/)?
Or a single checkbox, or a single radio.

apps/antalmanac/src/components/Header/SettingsMenu.tsx
```diff
@@ -1,96 +1,142 @@
...
+function SettingsMenu() {
+    const [drawerOpen, setDrawerOpen] = useState(false);
 
     return (
         <>
             <Button
-                onClick={(event) => {
-                    setAnchorEl(event.currentTarget);
+                onClick={() => {
+                    setDrawerOpen(true);
```
ap0nia: You could memo-ize this callback. But it's not a big deal I guess. Same difference

apps/antalmanac/src/providers/Themev5.tsx
```diff
@@ -37,32 +36,27 @@ interface Props {
  * sets and provides the MUI theme for the app
  */
 export default function AppThemev5Provider(props: Props) {
-    const [darkMode, setDarkMode] = useState(isDarkMode());
+    const { theme, setTheme } = useThemeStore();
```
ap0nia: This is fine because that's all the properties that exist. Typically, you want to slice the store to reduce the number of subscriptions that the component makes. For example:

```tsx
const theme = useThemeStore(s => s.theme)
```

You could even make it function more like `useState`

```tsx
const [theme, setTheme] = useThemeStore(store => [store.theme, store.setTheme])
```

apps/antalmanac/src/providers/Themev5.tsx
```diff
@@ -37,32 +36,27 @@ interface Props {
...
 
     useEffect(() => {
-        AppStore.on('themeToggle', () => {
-            setDarkMode(isDarkMode());
-        });
         window.matchMedia('(prefers-color-scheme: dark)').addEventListener('change', (e) => {
-            if (AppStore.getTheme() === 'auto') {
-                setDarkMode(e.matches);
-            }
+            setTheme(e.matches ? 'dark' : 'light');
```
ap0nia: Same advice as the other about removing the event handlers when the component unmounts.

ap0nia (changes requested):
The main content is solid. React is a garbage framework so now you get to deal with the consequences of using it 😔

apps/antalmanac/src/components/Header/SettingsMenu.tsx
KevinWu098: Buttons are clearly hotter and more appealing to UCI Anteaters 🤓; It's mainly for the design choice, and MUI themselves even use buttons for they left-right / right-left setting

Screenshot 2023-11-24 at 5 36 03 AM

ap0nia (approved):
This looks good 🎉

KevinWu098:
Actually no, PR isn't done yet 🫠; i forgot to add in indicators for which setting is selected 😭

KevinWu098:
I think it would also be worth considering moving the Schedule Conflict feature into the Settings Menu. Because of how apparent it makes itself, it feels like a feature users could choose to enable / disable; Thoughts?

_To be implemented in another PR*_

apps/antalmanac/src/stores/ThemeStore.ts
```diff
@@ -0,0 +1,43 @@
+import { create } from 'zustand';
+import analyticsEnum, { logAnalytics } from '$lib/analytics';
+
+export interface ThemeStore {
+    /**
+     * The 'raw' theme (either 'light', 'dark', or 'system')
+     */
+    value: string;
```
MinhxNguyen7: I'd make this an enum for type safety.

apps/antalmanac/src/stores/ThemeStore.ts
```diff
@@ -0,0 +1,43 @@
...
+    /**
+     * The 'derived' theme (either 'light' or 'dark', based on user settings and device preferences)
+     */
+    theme: string;
```
MinhxNguyen7: I think that `themeSetting` vs. `computedTheme` or `derivedTheme` makes better distinction than `theme` and the generic `value`. It can also be an enum.

MinhxNguyen7: Looking at the interface alone, I wouldn't know if `setTheme` accepts `"system"`.

apps/antalmanac/src/stores/ThemeStore.ts
```diff
@@ -0,0 +1,43 @@
+import { create } from 'zustand';
```
MinhxNguyen7: I'm of the opinion that there should be a `settingsStore.ts`, where `themeStore` and whatever the time preference store is contained. Splitting it like this makes it a little fractured, but I don't feel so strongly about it.

apps/antalmanac/src/lib/helpers.ts
```diff
@@ -42,7 +42,7 @@ export async function clickToCopy(event: React.MouseEvent<HTMLElement, MouseEven
 }
 
 export function isDarkMode() {
```
MinhxNguyen7, line 44: This shouldn't exist anymore, and it's consumers should refactor to use the store directly. Up to you to include it in this PR.

apps/antalmanac/src/components/Header/SettingsMenu.tsx
```diff
@@ -1,39 +1,72 @@
...
-            <ButtonGroup style={{ display: 'flex', placeContent: 'center', width: '100%' }}>
+            <ButtonGroup style={{ display: 'flex', placeContent: 'center', width: '100%', borderColor: 'unset' }}>
                 <Button
                     startIcon={<LightMode fontSize="small" />}
-                    style={{ padding: '1rem 2rem', borderRadius: '12px 0px 0px 12px', width: '100%' }}
+                    style={{
+                        padding: '1rem 2rem',
+                        borderRadius: '12px 0px 0px 12px',
+                        width: '100%',
+                        ...(value === 'light' ? (theme == 'dark' ? darkSelectedStyle : lightSelectedStyle) : {}),
```
MinhxNguyen7: Please extract this. This is too complex to just leave in the HTML.

MinhxNguyen7 (changes requested):
Looks great. Just a few code things.

apps/antalmanac/src/lib/helpers.ts
KevinWu098: I think there's like 15 something files which call isDarkMode(), so I'd tentatively prefer to punt that down the line in a separate PR; I'll see how I feel about it once I address the other comments

apps/antalmanac/src/stores/ThemeStore.ts
KevinWu098: I agree; implemented

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
```diff
@@ -239,6 +239,9 @@ export default function ScheduleCalendar(props: ScheduleCalendarProps) {
                     views={[Views.WEEK, Views.WORK_WEEK]}
                     defaultView={Views.WORK_WEEK}
                     view={hasWeekendCourse ? Views.WEEK : Views.WORK_WEEK}
+                    onView={() => {
+                        return;
+                    }}
```
KevinWu098, line 244: Since we're now in this file, I'm taking another liberty to clean up this console error

apps/antalmanac/src/lib/helpers.ts
KevinWu098: Yeah; I'll move this to another PR

MinhxNguyen7 (approved):
LGTM!

PR #811, opened 2023-11-25
feat: Create Snackbar When Copying Schedules

KevinWu098:
## Summary
1. Open snackbar when copying (appending) schedules; happens on success and on failure

Snackbar opens on success

## Test Plan
1. Check that the snackbar opens appropriately on success and on failure

## Issues
Closes #

### DIFF
--- a/apps/antalmanac/src/actions/AppStoreActions.ts
+++ b/apps/antalmanac/src/actions/AppStoreActions.ts
@@ -187,7 +187,7 @@ export const changeCourseColor = (sectionCode: string, term: string, newColor: s
     AppStore.changeCourseColor(sectionCode, term, newColor);
 };
 
-export const copySchedule = (name: string, to: number, options?: CallbackOptions) => {
+export const copySchedule = (to: number, options?: CallbackOptions) => {
     logAnalytics({
         category: analyticsEnum.addedClasses.title,
         action: analyticsEnum.addedClasses.actions.COPY_SCHEDULE,
--- a/apps/antalmanac/src/actions/AppStoreActions.ts
+++ b/apps/antalmanac/src/actions/AppStoreActions.ts
@@ -190,11 +190,10 @@ export const copySchedule = (name: string, to: number) => {
 
     try {
         AppStore.copySchedule(to);
+        openSnackbar('success', `Schedule copied to ${name}.`);
     } catch (error) {
         openSnackbar('error', `Could not copy schedule to ${name}.`);
     }
-
-    openSnackbar('success', `Schedule copied to ${name}.`);
 };
 
 export const toggleTheme = (radioGroupEvent: React.ChangeEvent<HTMLInputElement>) => {
--- a/apps/antalmanac/src/actions/AppStoreActions.ts
+++ b/apps/antalmanac/src/actions/AppStoreActions.ts
@@ -195,9 +195,9 @@ export const copySchedule = (name: string, to: number, options?: CallbackOptions
 
     try {
         AppStore.copySchedule(to);
-        options?.onSuccess(name);
+        options?.onSuccess(to);
     } catch (error) {
-        options?.onError(name);
+        options?.onError(to);
     }
 };
 
--- a/apps/antalmanac/src/actions/AppStoreActions.ts
+++ b/apps/antalmanac/src/actions/AppStoreActions.ts
@@ -12,7 +12,7 @@ import AppStore from '$stores/AppStore';
 import trpc from '$lib/api/trpc';
 import { courseNumAsDecimal } from '$lib/analytics';
 
-export interface CallbackOptions {
+export interface CopyScheduleOptions {
     onSuccess: (index: number) => unknown;
     onError: (index: number) => unknown;
 }
--- a/apps/antalmanac/src/actions/AppStoreActions.ts
+++ b/apps/antalmanac/src/actions/AppStoreActions.ts
@@ -12,6 +12,11 @@ import AppStore from '$stores/AppStore';
 import trpc from '$lib/api/trpc';
 import { courseNumAsDecimal } from '$lib/analytics';
 
+export interface CallbackOptions {
+    onSuccess: (name?: string) => unknown;
+    onError: (name?: string) => unknown;
+}
+
 export const addCourse = (
     section: WebsocSection,
     courseDetails: CourseDetails,
@@ -182,7 +187,7 @@ export const changeCourseColor = (sectionCode: string, term: string, newColor: s
     AppStore.changeCourseColor(sectionCode, term, newColor);
 };
 
-export const copySchedule = (name: string, to: number) => {
+export const copySchedule = (name: string, to: number, options?: CallbackOptions) => {
     logAnalytics({
         category: analyticsEnum.addedClasses.title,
         action: analyticsEnum.addedClasses.actions.COPY_SCHEDULE,
@@ -190,9 +195,9 @@ export const copySchedule = (name: string, to: number) => {
 
     try {
         AppStore.copySchedule(to);
-        openSnackbar('success', `Schedule copied to ${name}.`);
+        options?.onSuccess(name);
     } catch (error) {
-        openSnackbar('error', `Could not copy schedule to ${name}.`);
+        options?.onError(name);
     }
 };
 
--- a/apps/antalmanac/src/actions/AppStoreActions.ts
+++ b/apps/antalmanac/src/actions/AppStoreActions.ts
@@ -13,8 +13,8 @@ import trpc from '$lib/api/trpc';
 import { courseNumAsDecimal } from '$lib/analytics';
 
 export interface CallbackOptions {
-    onSuccess: (name?: string) => unknown;
-    onError: (name?: string) => unknown;
+    onSuccess: (index: number) => unknown;
+    onError: (index: number) => unknown;
 }
 
 export const addCourse = (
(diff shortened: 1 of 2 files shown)


MinhxNguyen7 (approved):
LGTM!

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -182,13 +182,18 @@ export const changeCourseColor = (sectionCode: string, term: string, newColor: s
...
+export const copySchedule = (name: string, to: number) => {
     logAnalytics({
         category: analyticsEnum.addedClasses.title,
         action: analyticsEnum.addedClasses.actions.COPY_SCHEDULE,
     });
 
-    AppStore.copySchedule(to);
+    try {
+        AppStore.copySchedule(to);
+        openSnackbar('success', `Schedule copied to ${name}.`);
```
ap0nia: Don't use this function.

ap0nia (changes requested):

apps/antalmanac/src/actions/AppStoreActions.ts
KevinWu098: Hi, this looks like it's out of scope for the PR 🤓. It's scoped to just the snack bar, so let's move this to a separate PR 

🤓🤓🤓

ap0nia: No, fix this now! 

Don't use the `openSnackbar` function because it's technically non-deterministic. The way it works is by extracting local component context into a global scope, which is bad because components have their own lifecycles and their local state doesn't inherently belong in global scope (this is different when using stores to share global state between components).

Prefer implementing functions with `onSuccess` and `onFailure` callbacks if you need to refer to the component's local stuff (in this case, `enqueueSnackbar` is a hook that grabs the snackbar context from the app, and it should remain local to components)

```tsx
export interface CallbackOptions {
  onSuccess: () => unknown
  onError: () => unknown
}

export const copySchedule = (name: string, to: number, options?: CallbackOptions) => {
    logAnalytics({
        category: analyticsEnum.addedClasses.title,
        action: analyticsEnum.addedClasses.actions.COPY_SCHEDULE,
    });

    // do logic

    if (success) {
      options?.onSuccess()
    } else {
      options?.onError()
    }
}
```

Any additional information needed by the component, i.e. the new schedules, messages, etc. can also be passed into the callback. It would be consumed as follows:

```tsx
function createCopyHandler(name, index, options) {
  return () => copySchedule(name, index,options)
}
function Component() {
  const { enqueueSnackbar } = useSnackbar()

  const options = useMemo(() => {
    return {
      onSuccess: () => enqueueSnackbar( ... )
    }
  }, [enqueueSnackbar])

  return (
    <MenuItem onClick={createCopyHandler(name, index, options} />
  )
}
```

ap0nia: It's fairly verbose, but significantly more "idiomatic" because the responsibility of functions internal to components and external functions are kept separate. External functions just have to handle input and output, making them simpler to test and verify behavior. Component functions maintain their own separate logic to react to outputs generated by the external functions.

If it's too complicated, then I could take a look at it. I just kinda wanted to get contributors in a better mindset for using these tools in an "idiomatic" manner, rather than the scuffed way that's been pre-established.

KevinWu098: Oh that's my bad -- I thought you were talking about `.copySchedule(to)` and got mixed up. I'll take a closer look at what you wrote up later today, but at first pass it makes sense

Will indeed fix it ~~now~~ _later_

KevinWu098:
I'm not sure if my type definition for the CallbackOptions is proper, so just lmk if there's a better way to do them @ap0nia 

```ts
export interface CallbackOptions {
    onSuccess: (name?: string) => unknown;
    onError: (name?: string) => unknown;
}
```

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -12,6 +12,11 @@ import AppStore from '$stores/AppStore';
 import trpc from '$lib/api/trpc';
 import { courseNumAsDecimal } from '$lib/analytics';
 
+export interface CallbackOptions {
+    onSuccess: (name?: string) => unknown;
+    onError: (name?: string) => unknown;
```
ap0nia: Could you just pass the numeric index? I don't think there's any need to modify the function signature in this way for this particular case.

```suggestion
    onSuccess: (index: number) => unknown;
    onError: (index: number) => unknown;
```

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -12,6 +12,11 @@ import AppStore from '$stores/AppStore';
 import trpc from '$lib/api/trpc';
 import { courseNumAsDecimal } from '$lib/analytics';
 
+export interface CallbackOptions {
```
ap0nia: It depends on whether or not we want to reuse these callback options for other functions, i.e. all the helper functions that trigger the snackbar, but I think it's fine to specify that these options are specific to the `copySchedule` function.

```suggestion
export interface CopyScheduleOptions {
```

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -182,13 +187,18 @@ export const changeCourseColor = (sectionCode: string, term: string, newColor: s
...
         action: analyticsEnum.addedClasses.actions.COPY_SCHEDULE,
     });
 
-    AppStore.copySchedule(to);
+    try {
+        AppStore.copySchedule(to);
+        options?.onSuccess(name);
+    } catch (error) {
+        options?.onError(name);
+    }
```
ap0nia:
```suggestion
    try {
        AppStore.copySchedule(to);
        options?.onSuccess(to);
    } catch (error) {
        options?.onError(to);
    }
```

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -118,6 +119,14 @@ function ClearScheduleButton() {
 
 function CopyScheduleButton() {
     const [scheduleNames, setScheduleNames] = useState(AppStore.getScheduleNames());
+    const { enqueueSnackbar } = useSnackbar();
+
+    const options = useMemo(() => {
+        return {
+            onSuccess: (name?: string) => enqueueSnackbar(`Schedule copied to ${name}.`, { variant: 'success' }),
+            onError: (name?: string) => enqueueSnackbar(`Could not copy schedule to ${name}.`, { variant: 'error' }),
```
ap0nia:
```suggestion
            onSuccess: (index: number) => enqueueSnackbar(`Schedule copied to ${scheduleNames[index]}.`, { variant: 'success' }),
            onError: (index: number) => enqueueSnackbar(`Could not copy schedule to ${scheduleNames[index]}.`, { variant: 'error' }),
```

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -148,12 +157,14 @@ function CopyScheduleButton() {
                             <MenuItem
                                 key={index}
                                 disabled={AppStore.getCurrentScheduleIndex() === index}
-                                onClick={createCopyHandler(index)}
+                                onClick={createCopyHandler(name, index, options)}
```
ap0nia:
```suggestion
                                onClick={createCopyHandler(index, options)}
```

ap0nia (changes requested):
Thanks for implementing the strategy 🙏 

I think the original function signature should be completely preservable, i.e. it should accept just the index and possibly options. The options will be callback functions that will be called with the index after a successful/unsuccessful attempt to copy the schedule.

The callback functions can figure out the schedule name by indexing the array of schedule names with the index they were called with.

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -182,13 +187,18 @@ export const changeCourseColor = (sectionCode: string, term: string, newColor: s
     AppStore.changeCourseColor(sectionCode, term, newColor);
 };
 
-export const copySchedule = (to: number) => {
+export const copySchedule = (name: string, to: number, options?: CallbackOptions) => {
```
ap0nia:
```suggestion
export const copySchedule = (to: number, options?: CallbackOptions) => {
```

ap0nia:
I kinda wanna write tests for it, but I don't wanna write tests for it lol.

What do you wanna do Kevin? If you don't wanna test, then I'll just approve and merge 🤓

KevinWu098:
> I kinda wanna write tests for it, but I don't wanna write tests for it lol.
> 
> What do you wanna do Kevin? If you don't wanna test, then I'll just approve and merge 🤓

If it breaks it breaks, and somebody gets extra contribution history 😋

KevinWu098:
TLDR let's merge 😅, do you need me to do a noop?

ap0nia (approved):
Good work refactoring. This was more involved than it could've been otherwise, but I have to slow down Kevin from moving too fast 💯

ap0nia:
Yeah I think I need an extra commit from someone aside from me lol. Unless you're able to merge it yourself.

ap0nia (approved):

ap0nia:
image
lol it's not working

EricPedley (approved):

PR #814, opened 2023-11-27
fix: Update Arktype for Buildings

KevinWu098:
## Summary
The Arktype definition for undefined is a little more specific than what we currently have:
```ts
const schema = type({ 'building?': 'string' })
```
It should instead be:
```ts
const schema = type({ 'building?': 'string | undefined' })
```

## Test Plan
1. Can schedules with custom events that don't have selected buildings be saved?
2. Check if this change broke anything else; Good places to look would be export, creating, editing, copying, etc.
 
## Issues
Bug report from feedback

### DIFF
--- a/apps/antalmanac/tests/buildings-type.test.ts
+++ b/apps/antalmanac/tests/buildings-type.test.ts
@@ -1,10 +1,7 @@
 import { describe, test, expect } from 'vitest';
-import { RepeatingCustomEvent } from '@packages/antalmanac-types';
-import AppStore from '$stores/AppStore';
-import trpc from '$lib/api/trpc';
+import { RepeatingCustomEvent, RepeatingCustomEventSchema } from '@packages/antalmanac-types';
 
 describe('building type', () => {
-    const scheduleSaveState = AppStore.schedule.getScheduleAsSaveState();
     const customEvent: RepeatingCustomEvent = {
         title: 'placeHoldertitle',
         start: '10:30',
@@ -15,11 +12,7 @@ describe('building type', () => {
         building: undefined,
     };
 
-    AppStore.addCustomEvent(customEvent, [0]);
-
     test('schema does not throw error when building property exists and is undefined', async () => {
-        expect(
-            await trpc.users.saveUserData.mutate({ id: 'testUser', userData: scheduleSaveState })
-        ).not.toThrowError();
+        expect(() => RepeatingCustomEventSchema.assert(customEvent)).not.toThrowError();
     });
 });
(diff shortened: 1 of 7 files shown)


ap0nia:
Force this to deploy the backend. It's the only relevant part that needs to be updated.

KevinWu098:
If you look at the deployment, it will error out, but there's no schema issue so it should (the word should is doing heavy lifting here) work on prod

KevinWu098:
@ap0nia Could you take a cursory glance and approve?

Worst comes to worst it fails and we go from fail -> fail

ap0nia:
Could you write the test case I mentioned? I think it would be helpful for documenting the history of buildings. i.e. "It must be optional **and** string | undefined because of these reasons, and this test ensures that"

KevinWu098:
> Could you write the test case I mentioned? I think it would be helpful for documenting the history of buildings. i.e. "It must be optional **and** string | undefined because of these reasons, and this test ensures that"

Sure @ap0nia; Do you think that should be a new test file or should that be in one of the existing?

ap0nia:
You can make a new file because this is a somewhat miscellaneous test.

apps/backend/src/db/ddb.ts
```diff
@@ -10,7 +10,11 @@ const client = new DynamoDB({
 });
 
 // Create DynamoDB DocumentClient
-const documentClient = DynamoDBDocument.from(client);
+const documentClient = DynamoDBDocument.from(client, {
+    marshallOptions: {
+        removeUndefinedValues: true
```
ap0nia: Maybe we can add a quick snippet detailing why this setting was flipped.

```suggestion
        /**
         * Some JSON properties may exist and be undefined; DDB will throw an error unless this is true.
         * @example { "property": undefined }
         */
        removeUndefinedValues: true
```

apps/backend/src/index.ts
ap0nia, line 1: I'm not entirely sure why this file was changed, is it possible to remove it from the diff?

apps/antalmanac/tests/buildings-type.test.ts
```diff
@@ -0,0 +1,18 @@
+import { describe, test, expect } from 'vitest';
+import { RepeatingCustomEvent, RepeatingCustomEventSchema } from '@packages/antalmanac-types';
+
+describe('building type', () => {
```
ap0nia: Sorry I didn't specify, but can we make this file `tests/custom-events.test.ts`? And the description for the test suite would be `CustomEvents` or something similar.

```suggestion
describe('CustomEvents', () => {
```

apps/backend/src/index.ts
KevinWu098: Sure, I'll clean it up

apps/antalmanac/tests/buildings-type.test.ts
KevinWu098: Gotcha, will do

ap0nia (approved):
Solid work tracking down the bug. Looks good :rocket:

Feel free to merge after you've addressed the comments!

ap0nia (approved):

PR #818, opened 2023-11-30
Fix the duplicate calendar events bug

### LINKED ISSUE #816: Download ICS: Duplicate calendar events are added
As from Eric, 

> when we run events through the calenderize helper functions, it returns an event for every day the course is present, and the locations also contains info on which day each location applies. But, then we loop through all the locations and add an event for each day there

This should be fixed or it will mess up users' calendars.

Related PR: https://github.com/icssc/AntAlmanac/pull/742

JacE070:
## Summary
This change fixes the bug caused by looping over events' location.days when the calendarizeHelper has already returned events for each day.
For example, class A occurs every MWF, the calendarizeHelper returns **(event for Monday)** -> {..., location: {..., days: "MWF"}}, also events for **Wednesday** and **Friday** with the same location field.
Therefore, we will finally get 3*3=9 calendar events.
Finals are also added repetitively.
## Test Plan
Passed the vite test. Also tested manually.
## Issues

Closes https://github.com/icssc/AntAlmanac/issues/816

### DIFF
--- a/apps/antalmanac/src/lib/download.ts
+++ b/apps/antalmanac/src/lib/download.ts
@@ -247,7 +247,9 @@ export function getRRule(bydays: string[], quarter: string) {
     return `FREQ=WEEKLY;BYDAY=${bydays.toString()};INTERVAL=1;COUNT=${count}`;
 }
 
-export function getEventsFromCourses(events = AppStore.getEventsInCalendar()): EventAttributes[] {
+export function getEventsFromCourses(
+    events = [...AppStore.getEventsInCalendar(), ...AppStore.getFinalEventsInCalendar()]
+): EventAttributes[] {
     const calendarEvents = events.flatMap((event) => {
         if (event.isCustomEvent) {
             // FIXME: We don't have a way to get the term for custom events,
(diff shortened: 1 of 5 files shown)


apps/antalmanac/src/lib/download.ts
```diff
@@ -247,7 +247,9 @@ export function getRRule(bydays: string[], quarter: string) {
     return `FREQ=WEEKLY;BYDAY=${bydays.toString()};INTERVAL=1;COUNT=${count}`;
 }
 
-export function getEventsFromCourses(events = AppStore.getEventsInCalendar()): EventAttributes[] {
+export function getEventsFromCourses(
+    events = [...AppStore.getEventsInCalendar(), ...AppStore.getFinalEventsInCalendar()]
```
ap0nia: Can you extract this to a dedicated function? I think it might even exist. Shorter function signatures is easier to read.

apps/antalmanac/src/lib/download.ts
```diff
@@ -281,45 +283,42 @@ export function getEventsFromCourses(events = AppStore.getEventsInCalendar()): E
                     if (location.days === undefined) return null;
```
ap0nia: This should be enforced by the linter, so this change is purely personal preference. After reading through lots of code, I really like it when things are consistent, even if they're marginally more verbose. I get throw off sometimes when reading through a bunch of if-statements with newlines and curly braces and then coming across one if-statement that's compacted into one line.

```suggestion
                    if (location.days === undefined) {
                      return null;
                    }
```

apps/antalmanac/src/lib/download.ts
```diff
@@ -281,45 +283,42 @@ export function getEventsFromCourses(events = AppStore.getEventsInCalendar()): E
...
-                        location: `${location.building} ${location.room}`,
-                        start: firstClassStart,
-                        end: firstClassEnd,
-                        recurrenceRule: rrule,
-                    };
+                    if (sectionType === 'Fin') {
+                        return {
+                            productId: 'antalmanac/ics',
+                            startOutputType: 'local' as const,
+                            endOutputType: 'local' as const,
```
ap0nia: An improvement would be to capture these strings as their own constants or enums.

```ts
const CALENDAR_ID = 'antalamanc/ics'
const CALENDAR_OUTPUT = 'local' as const
```

apps/antalmanac/src/lib/download.ts
```diff
@@ -281,45 +283,42 @@ export function getEventsFromCourses(events = AppStore.getEventsInCalendar()): E
...
+                            title: `${title} Final Exam`,
+                            description: `Final Exam for ${courseTitle}`,
+                            start: getExamTime(finalExam, getYear(term))[0]!,
+                            end: getExamTime(finalExam, getYear(term))[1]!,
```
ap0nia: This is a blunder, don't call the same function twice for the same info. You can correctly extract all the data with the following snippet:

```ts
const [ start = '', end = '' ] = getExamTime(finalExam, getYear(term))

return {
  // ...
  end,
  start
}
```

Use array destructuring to immediately assign meaningful names to the variables, and then give them a default value of empty string if they're missing.

apps/antalmanac/src/stores/calendarizeHelpers.ts
```diff
@@ -49,7 +49,10 @@ export function calendarizeCourseEvents(currentCourses: ScheduleCourse[] = []):
                         title: `${course.deptCode} ${course.courseNumber}`,
                         courseTitle: course.courseTitle,
                         locations: meeting.bldg.map(getLocation).map((location: Location) => {
-                            return { ...location, days: meeting.days === null ? undefined : meeting.days };
+                            return {
+                                ...location,
+                                days: meeting.days === null ? undefined : COURSE_WEEK_DAYS[dayIndex],
```
ap0nia: This seems a bit misleading, because `meeting.days` shouldn't be set to null manually? You didn't change this code, but it would be more appropriately captured as 
```suggestion
                                ...(meeting.days && { days: COURSE_WEEK_DAYS[dayIndex] })
```

Logic: 
The statement will either return something falsy (null, undefined, false, etc.) or an object with the `days` property.
When you destructure a falsy object, it doesn't return anything. When you destructure an actual object, its properties get inherited by the parent object

```ts
const myObject = { hello: 'world', ...(false) }
// myObject = { hello: 'world' }

const myOtherObject = { hello: 'world', ...({ goodbye: 'earth' }) }
// myOtherObject = { hello: 'world', goodbye: 'earth' }
```

ap0nia (changes requested):
Everything looks solid 👍 

I left some suggestions for minor improvements. Once you've taken a look at everything, this should be totally fine to merge. :rocket:

apps/antalmanac/src/stores/calendarizeHelpers.ts
JacE070: Makes sense. Glad to know this!

EricPedley (approved):
since this is a bugfix, imma just merge it real quick. @ap0nia if u want anything else to be refactored we can do it in a separate PR.

PR #820, opened 2023-11-30
Add Past Enrollment Popup

### LINKED ISSUE #182: Once PeterPortal API implements past enrollment data, uncomment the graph modal
AntAlmanac used to store past enrollment data and display it in a modal in the EnrollmentGraph component, rendered inside SectionTable. We removed this feature from our API and requested that PeterPortal API implements it (https://github.com/icssc/peterportal-public-api/issues/111). Once they finish, we should put the modal back but still keep the link to ZotTracker.
Reasoning behind moving the data to PeterPortal:
- Provides this information for all student developers to use
- Logically it makes sense for PeterPortal API to collect and serve all UCI related data. 
- It will fit nicely into their larger data pipeline plans.

### LINKED ISSUE #341: Past Enrollment Modal
- Create a drop down enrollment graph. This can be a simple line graph.
- At the bottom of the bar graph, show a button that says "View on ZotTracker" that links to more info.

Here's a suggested design from Design at UCI. No need to follow it exactly but feel free to use it for inspiration. 
image

Douglas-Hong:
## Summary
- Now that the enrollment history endpoint is implemented, we can bring back and improve the Past Enrollment popup.
- Currently, the popup just displays the enrollment history of the most recent lecture section offered.

Demo:

https://github.com/icssc/AntAlmanac/assets/78244965/1b6bcee9-cc37-4057-a85f-2fea05f3b124

## TODO

- [X] Make a request to PeterPortal to get the enrollment history, and parse the response
- [X] Create the popup and display an enrollment history graph
- [x] Cache responses
- [x] Add ZotTracker link

## Test Plan
- Verify this works for classes in the most recent quarter (e.g. CS 132).
- Verify this works for classes that haven't been offered very recently (e.g. CS 122D from Spring 2023).
- Verify this works for classes that have never been offered before (force `queryEnrollmentHistory` to return null).
- Make sure it looks good on dark and light mode, as well as mobile.

## Issues

Closes #341 
Closes #182 

## Future Followup
- We should allow the user to see the enrollment history of any lecture that was previously offered: #830

### DIFF
--- a/apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
+++ b/apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
@@ -1,7 +1,7 @@
 import { useState, useEffect } from 'react';
 import { LineChart, Line, CartesianGrid, ResponsiveContainer, XAxis, YAxis, Tooltip, Legend } from 'recharts';
 import { Box, Typography, Skeleton } from '@mui/material';
-import { queryEnrollmentHistory, EnrollmentHistory } from '$lib/enrollmentHistory';
+import EnrollmentHistoryHelper, { EnrollmentHistory } from '$lib/enrollmentHistory';
 import { isDarkMode } from '$lib/helpers';
 
 export interface EnrollmentHistoryProps {
@@ -32,15 +32,15 @@ const EnrollmentHistoryPopup = (props: EnrollmentHistoryProps) => {
             return;
         }
 
-        queryEnrollmentHistory(department, courseNumber, sectionType).then((enrollmentHistory) => {
-            if (enrollmentHistory) {
-                setEnrollmentHistory(enrollmentHistory);
-                console.log(enrollmentHistory);
-                console.log(enrollmentHistory[enrollmentHistory.length - 1].days);
-            }
+        EnrollmentHistoryHelper.queryEnrollmentHistory(department, courseNumber, sectionType).then(
+            (enrollmentHistory) => {
+                if (enrollmentHistory) {
+                    setEnrollmentHistory(enrollmentHistory);
+                }
 
-            setLoading(false);
-        });
+                setLoading(false);
+            }
+        );
     }, [loading, department, courseNumber, sectionType]);
 
     if (loading) {
--- a/apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
+++ b/apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
@@ -10,7 +10,7 @@ export interface EnrollmentHistoryPopupProps {
     courseNumber: string;
 }
 
-const EnrollmentHistoryPopup = ({ department, courseNumber }: EnrollmentHistoryPopupProps) => {
+export function EnrollmentHistoryPopup({ department, courseNumber }: EnrollmentHistoryPopupProps) {
     const [loading, setLoading] = useState(true);
     const [enrollmentHistory, setEnrollmentHistory] = useState<EnrollmentHistory>();
     const isMobileScreen = useMediaQuery(`(max-width: ${MOBILE_BREAKPOINT})`);
@@ -20,7 +20,7 @@ const EnrollmentHistoryPopup = ({ department, courseNumber }: EnrollmentHistoryP
     const graphWidth = useMemo(() => (isMobileScreen ? 250 : 450), [isMobileScreen]);
     const graphHeight = useMemo(() => (isMobileScreen ? 175 : 250), [isMobileScreen]);
     const popupTitle = useMemo(() => {
-        if (!enrollmentHistory) {
+        if (enrollmentHistory == null) {
             return 'No past enrollment data found for this course';
         }
 
@@ -54,7 +54,7 @@ const EnrollmentHistoryPopup = ({ department, courseNumber }: EnrollmentHistoryP
         );
     }
 
-    if (!enrollmentHistory) {
+    if (enrollmentHistory == null) {
         return (
             <Box padding={1}>
                 <Typography variant="body1" align="center">
@@ -99,6 +99,4 @@ const EnrollmentHistoryPopup = ({ department, courseNumber }: EnrollmentHistoryP
             </Link>
         </Box>
     );
-};
-
-export default EnrollmentHistoryPopup;
+}
--- a/apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
+++ b/apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
@@ -1,7 +1,7 @@
 import { useState, useEffect, useMemo } from 'react';
 import { LineChart, Line, CartesianGrid, ResponsiveContainer, XAxis, YAxis, Tooltip, Legend } from 'recharts';
 import { Box, Link, Typography, Skeleton } from '@mui/material';
-import EnrollmentHistoryHelper, { EnrollmentHistory } from '$lib/enrollmentHistory';
+import enrollmentHistoryCache, { EnrollmentHistory } from '$lib/enrollmentHistory';
 import { isDarkMode } from '$lib/helpers';
 
 export interface EnrollmentHistoryPopupProps {
@@ -35,9 +35,9 @@ const EnrollmentHistoryPopup = ({ department, courseNumber, isMobileScreen }: En
             return;
         }
 
-        EnrollmentHistoryHelper.queryEnrollmentHistory(department, courseNumber).then((enrollmentHistory) => {
-            if (enrollmentHistory) {
-                setEnrollmentHistory(enrollmentHistory);
+        enrollmentHistoryCache.queryEnrollmentHistory(department, courseNumber).then((data) => {
+            if (data) {
+                setEnrollmentHistory(data);
             }
             setLoading(false);
         });
--- a/apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
+++ b/apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
@@ -10,19 +10,20 @@ export interface EnrollmentHistoryPopupProps {
     isMobileScreen: boolean;
 }
 
-const EnrollmentHistoryPopup = (props: EnrollmentHistoryPopupProps) => {
-    const { department, courseNumber, isMobileScreen } = props;
+const EnrollmentHistoryPopup = ({ department, courseNumber, isMobileScreen }: EnrollmentHistoryPopupProps) => {
     const [loading, setLoading] = useState(true);
     const [enrollmentHistory, setEnrollmentHistory] = useState<EnrollmentHistory>();
 
     const graphWidth = useMemo(() => (isMobileScreen ? 250 : 450), [isMobileScreen]);
     const graphHeight = useMemo(() => (isMobileScreen ? 175 : 250), [isMobileScreen]);
     const popupTitle = useMemo(() => {
-        return enrollmentHistory
-            ? `${department} ${courseNumber} | ${enrollmentHistory.year} ${
-                  enrollmentHistory.quarter
-              } | ${enrollmentHistory.instructors.join(', ')}`
-            : 'No past enrollment data found for this course';
+        if (!enrollmentHistory) {
+            return 'No past enrollment data found for this course';
+        }
+
+        return `${department} ${courseNumber} | ${enrollmentHistory.year} ${
+            enrollmentHistory.quarter
+        } | ${enrollmentHistory.instructors.join(', ')}`;
     }, [courseNumber, department, enrollmentHistory]);
 
     const encodedDept = useMemo(() => encodeURIComponent(department), [department]);
--- a/apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
+++ b/apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
@@ -7,18 +7,17 @@ import { isDarkMode } from '$lib/helpers';
 export interface EnrollmentHistoryPopupProps {
     department: string;
     courseNumber: string;
-    sectionType: string;
     isMobileScreen: boolean;
 }
 
 const EnrollmentHistoryPopup = (props: EnrollmentHistoryPopupProps) => {
-    const { department, courseNumber, sectionType, isMobileScreen } = props;
+    const { department, courseNumber, isMobileScreen } = props;
     const [loading, setLoading] = useState(true);
     const [enrollmentHistory, setEnrollmentHistory] = useState<EnrollmentHistory>();
 
     const graphWidth = useMemo(() => (isMobileScreen ? 250 : 450), [isMobileScreen]);
     const graphHeight = useMemo(() => (isMobileScreen ? 175 : 250), [isMobileScreen]);
-    const graphTitle = useMemo(() => {
+    const popupTitle = useMemo(() => {
         return enrollmentHistory
             ? `${department} ${courseNumber} | ${enrollmentHistory.year} ${
                   enrollmentHistory.quarter
@@ -35,15 +34,13 @@ const EnrollmentHistoryPopup = (props: EnrollmentHistoryPopupProps) => {
             return;
         }
 
-        EnrollmentHistoryHelper.queryEnrollmentHistory(department, courseNumber, sectionType).then(
-            (enrollmentHistory) => {
-                if (enrollmentHistory) {
-                    setEnrollmentHistory(enrollmentHistory);
-                }
-                setLoading(false);
+        EnrollmentHistoryHelper.queryEnrollmentHistory(department, courseNumber).then((enrollmentHistory) => {
+            if (enrollmentHistory) {
+                setEnrollmentHistory(enrollmentHistory);
             }
-        );
-    }, [loading, department, courseNumber, sectionType]);
+            setLoading(false);
+        });
+    }, [loading, department, courseNumber]);
 
     if (loading) {
         return (
@@ -57,7 +54,7 @@ const EnrollmentHistoryPopup = (props: EnrollmentHistoryPopupProps) => {
         return (
             <Box padding={1}>
                 <Typography variant="body1" align="center">
-                    {graphTitle}
+                    {popupTitle}
                 </Typography>
             </Box>
         );
@@ -75,7 +72,7 @@ const EnrollmentHistoryPopup = (props: EnrollmentHistoryPopupProps) => {
                     marginBottom: '.5rem',
                 }}
             >
-                {graphTitle}
+                {popupTitle}
             </Typography>
             <Link
                 href={`https://zot-tracker.herokuapp.com/?dept=${encodedDept}&number=${courseNumber}&courseType=all`}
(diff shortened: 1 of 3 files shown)


KevinWu098:
Screenshot 2023-11-29 at 5 18 35 PM

That's seriously hot 💯

apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
```diff
@@ -0,0 +1,100 @@
...
+    const encodedDept = useMemo(() => encodeURIComponent(department), [department]);
+    const axisColor = isDarkMode() ? '#fff' : '#111';
+    const tooltipDateColor = '#111';
+
+    useEffect(() => {
+        if (!loading) {
+            return;
+        }
+
+        EnrollmentHistoryHelper.queryEnrollmentHistory(department, courseNumber).then((enrollmentHistory) => {
```
MinhxNguyen7: I hate that this is called `EnrollmenthistoryHelper`, and I hate that it's pascal-case even though it's actually a singleton object. The former, I have little suggestion for. Maybe it should be called `enrollmentHistoryCache`. The latter should be easily fixable once the former is.

apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
```diff
@@ -0,0 +1,100 @@
...
+            if (enrollmentHistory) {
+                setEnrollmentHistory(enrollmentHistory);
```
MinhxNguyen7: This just looks so confusing when browsing through. I think you should rename `enrollmentHistory` here to `data` or `enrollmentHistory_`, which is not elegant but at least doesn't mirror an outside declaration

apps/antalmanac/src/lib/enrollmentHistory.ts
```diff
@@ -0,0 +1,142 @@
...
+                courseNumber: enrollmentHistory.courseNumber,
+                days: enrollmentDays,
+                instructors: enrollmentHistory.instructors,
+            });
+        }
+
+        return parsedEnrollmentHistory;
+    };
+
+    sortEnrollmentHistory = (enrollmentHistory: EnrollmentHistory[]) => {
```
MinhxNguyen7: Please add a short docstring that clarifies the sorting order instead of relying on the comment on in invokation.

MinhxNguyen7 (dismissed):
Great feature, looks great. Only have a few readability suggestions.

apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
Douglas-Hong: Thanks for catching that, I renamed it to `data` just to keep it short

apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
Douglas-Hong: Yeah the naming is definitely awkward, I think `enrollmentHistoryCache` is a good name for the singleton

apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
```diff
@@ -168,7 +169,13 @@ function SectionTable(props: SectionTableProps) {
                     analyticsAction={analyticsEnum.classSearch.actions.CLICK_PAST_ENROLLMENT}
                     text="Past Enrollment"
                     icon={<ShowChartIcon />}
-                    redirectLink={`https://zot-tracker.herokuapp.com/?dept=${encodedDept}&number=${courseDetails.courseNumber}&courseType=all`}
+                    popupContent={
+                        <EnrollmentHistoryPopup
+                            department={courseDetails.deptCode}
+                            courseNumber={courseDetails.courseNumber}
+                            isMobileScreen={isMobileScreen}
```
ap0nia: Since `isMobileScreen` is a hook usable by any component, you can just use it directly in the popup component. It's easier to manage functions with less inputs.
```suggestion
```

apps/antalmanac/src/lib/enrollmentHistory.ts
```diff
@@ -0,0 +1,154 @@
...
+    constructor() {
+        this.enrollmentHistoryCache = {};
+        this.termShortNames = termData.map((term) => term.shortName);
+    }
+
+    queryEnrollmentHistory = async (department: string, courseNumber: string): Promise<EnrollmentHistory | null> => {
+        const cacheKey = department + courseNumber;
+        if (cacheKey in this.enrollmentHistoryCache) {
+            return this.enrollmentHistoryCache[cacheKey];
+        }
```
ap0nia: A code density optimization might be to implement the class like this:
```suggestion
        return this.enrollmentHistoryCache[cacheKey] ??= this.findEnrollmentHistory(...)
```

You could delegate the search to a separate method, and apply nullish coalescing. The result of an assignment is the value that's assigned. If the cache entry exists, it'll be returned; otherwise the method will run to evaluate value, assign it to the cache, and get returned.

apps/antalmanac/src/lib/enrollmentHistory.ts
```diff
@@ -0,0 +1,154 @@
...
+                quarter
+                department
+                courseNumber
+                dates
+                totalEnrolledHistory
+                maxCapacityHistory
+                waitlistHistory
+                instructors
+            }
+        }`;
```
ap0nia: I wonder how efficient it would be store the template string in a more global scope, and then apply immutable string manipulation. This is extremely minute; it's just worth considering what scope is appropriate for variables that don't need to be initialized multiple times.

```tsx
const QUERY_TEMPLATE = `{
        enrollmentHistory(department: "$$DEPARTMENT$$", courseNumber: "$$COURSE$$", sectionType: Lec) {
            year
            quarter
            department
            courseNumber
            dates
            totalEnrolledHistory
            maxCapacityHistory
            waitlistHistory
            instructors
        }
    }`;

// ...
const queryString = QUERY_TEMPLATE.replace('$$DEPARTMENT', ...).replace('$$COURSE$$', ...)
```

apps/antalmanac/src/lib/enrollmentHistory.ts
```diff
@@ -0,0 +1,154 @@
...
+
+        const res =
+            (await queryGraphQL<EnrollmentHistoryGraphQLResponse>(queryString))?.data?.enrollmentHistory ?? null;
+
+        if (res && res.length > 0) {
```
ap0nia:
```suggestion
        if (res?.length) {
```

apps/antalmanac/src/lib/enrollmentHistory.ts
```diff
@@ -0,0 +1,154 @@
...
+                dates
+                totalEnrolledHistory
+                maxCapacityHistory
+                waitlistHistory
+                instructors
+            }
+        }`;
+
+        const res =
+            (await queryGraphQL<EnrollmentHistoryGraphQLResponse>(queryString))?.data?.enrollmentHistory ?? null;
```
ap0nia: I don't think there's a need to nullish coalesce it with null.

```suggestion
            (await queryGraphQL<EnrollmentHistoryGraphQLResponse>(queryString))?.data?.enrollmentHistory;
```

apps/antalmanac/src/lib/enrollmentHistory.ts
```diff
@@ -0,0 +1,154 @@
...
+     * This function parses enrollment history data from PeterPortal so that
+     * we can pass the data into a recharts graph. For each element in the given
+     * array, merge the dates, totalEnrolledHistory, maxCapacityHistory,
+     * and waitlistHistory arrays into one array that contains the enrollment data
+     * for each day.
+     *
+     * @param res an array of enrollment histories from PeterPortal
+     * @returns an array of enrollment histories that we can use for the graph
+     */
+    parseEnrollmentHistoryResponse = (res: EnrollmentHistoryGraphQL[]): EnrollmentHistory[] => {
```
ap0nia: This should be a static method or global function.
```suggestion
    static parseEnrollmentHistoryResponse (res: EnrollmentHistoryGraphQL[]): EnrollmentHistory[] {
```

apps/antalmanac/src/lib/enrollmentHistory.ts
```diff
@@ -0,0 +1,154 @@
...
+    // Each key in the cache will be the department and courseNumber concatenated
+    enrollmentHistoryCache: Record<string, EnrollmentHistory>;
+    termShortNames: string[];
+
+    constructor() {
+        this.enrollmentHistoryCache = {};
+        this.termShortNames = termData.map((term) => term.shortName);
+    }
+
+    queryEnrollmentHistory = async (department: string, courseNumber: string): Promise<EnrollmentHistory | null> => {
```
ap0nia: Please use the regular function syntax instead of arrow syntax. The arrow syntax typically represents unbound functions, but class methods are inherently bound to an instance.

```suggestion
    async queryEnrollmentHistory async (department: string, courseNumber: string): Promise<EnrollmentHistory | null> {
```

apps/antalmanac/src/lib/enrollmentHistory.ts
```diff
@@ -0,0 +1,154 @@
...
+            const bTerm = `${b.year} ${b.quarter}`;
+            // If the term for a appears earlier than the term for b in the list of
+            // term short names, then a must be the enrollment history for a later quarter
+            return this.termShortNames.indexOf(aTerm) - this.termShortNames.indexOf(bTerm);
+        });
+    };
+}
+
+const enrollmentHistoryCache = new _EnrollmentHistory();
+export default enrollmentHistoryCache;
```
ap0nia: Personally, I don't really see the value in this pattern. It's essentially a singleton class that could be effectively captured within a module scope and functions. But this pattern already exists in the codebase, so the precedent has already been established. 

I think it would be more appropriate to mark most of the properties static.
For example, most of the class could be static -- including the cache, and you could initialize special instances that will write to the static cache.

```tsx
class EnrollmentHistory {
  static cache: Record<string, any> = {}
  department: string
  partialQueryString: string
  constructor(department: string) {
    this.department = department
    this.partialQueryString = `...`
  }

  find(course: string) {
    // (not real, just example)
    const fullQueryString = this.partialQueryString + course
    return EnrollmentHistory.cache[fullQueryString] ?? this.findAndCache(course)
  }

  findAndCache(course: string) {
    // ...
  }
}

export const CsEnrollmentHistory = EnrollmentHistory("CS")
```
This makes sense to me, because individual instances of the classes serve the distinct purpose of searching and caching for a specific department. And they all write to a shared static cache. This is at your discretion as the implementer.

apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
```diff
@@ -0,0 +1,101 @@
...
+            <Box padding={1}>
+                <Typography variant="body1" align="center">
+                    {popupTitle}
+                </Typography>
+            </Box>
+        );
+    }
+
+    return (
+        <Box sx={{ padding: '4px' }}>
```
ap0nia: When assigning values, try to use raw numbers instead of specifying units. It's because those units will be used by MUI's theme awareness https://mui.com/system/getting-started/the-sx-prop/#spacing

```suggestion
        <Box sx={{ padding: 4}}>
```

ap0nia (changes requested):
Looks great 🚀 

Most of the improvements are regarding JS syntax. The comments regarding lower level implementation details can be applied at your discretion.

apps/antalmanac/src/lib/enrollmentHistory.ts
Douglas-Hong: Still not sure about the efficiency, but I do think putting it in a more visible scope is cleaner. For now I just have it as a static variable in `DepartmentEnrollmentHistory`, but I also see the case to put it globally.

apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
```diff
@@ -0,0 +1,104 @@
...
+    department: string;
+    courseNumber: string;
+}
+
+const EnrollmentHistoryPopup = ({ department, courseNumber }: EnrollmentHistoryPopupProps) => {
+    const [loading, setLoading] = useState(true);
+    const [enrollmentHistory, setEnrollmentHistory] = useState<EnrollmentHistory>();
+    const isMobileScreen = useMediaQuery(`(max-width: ${MOBILE_BREAKPOINT})`);
+
+    const deptEnrollmentHistory = useMemo(() => new DepartmentEnrollmentHistory(department), [department]);
```
ap0nia, line 18: I think this should be ok, since instances of this class don't really manage their own state and just use a shared static cache. 

From my experience, it's generally paramount to use `useRef` with any sort of external class in React, since `useMemo` can't preserve the class in some instances (like when the component is unrendered, but it still exists?? or something stupid like that in React land ™️ )

One possible usage of classes could be like

```ts
class MyClass {}

function MyComponent() {
  const [open, setOpen] = useState(false);
  const myClass = useRef(new MyClass());

  // e.g. update the class instance as needed
  useEffect(() => {
    myClass.updateState(open)
  }, [open])
}
```
where the class is only created once, and it just updates itself whenever it's necessary based on the component's state. This preserves the external class's state in a way closest to vanilla JS.

TLDR: it's fine, but just be careful using classes in React because it's a garbage technology.

apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+    const graphWidth = useMemo(() => (isMobileScreen ? 250 : 450), [isMobileScreen]);
+    const graphHeight = useMemo(() => (isMobileScreen ? 175 : 250), [isMobileScreen]);
+    const popupTitle = useMemo(() => {
+        if (!enrollmentHistory) {
```
ap0nia: Purely stylistic preference, and I can see both ways.
Clarifying that the condition is a null check instead of a boolean check can be helpful, but that adds extra characters to read.

```suggestion
        if (enrollmentHistory == null) {
```

apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { MOBILE_BREAKPOINT } from '../../../globals';
+import { DepartmentEnrollmentHistory, EnrollmentHistory } from '$lib/enrollmentHistory';
+import { isDarkMode } from '$lib/helpers';
+
+export interface EnrollmentHistoryPopupProps {
+    department: string;
+    courseNumber: string;
+}
+
+const EnrollmentHistoryPopup = ({ department, courseNumber }: EnrollmentHistoryPopupProps) => {
```
ap0nia: Purely stylistic preference; I like regular function syntax because it's similar to syntax in other languages and easier to read for me. I usually reserve arrow functions for callbacks that benefit from a short syntax or need to be unbound.

```suggestion
export function EnrollmentHistoryPopup({ department, courseNumber }: EnrollmentHistoryPopupProps) {
```

Also, another personal preference I have is using named exports over default exports because my LSP finds named exports better than default exports lol.

apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        return `${department} ${courseNumber} | ${enrollmentHistory.year} ${
+            enrollmentHistory.quarter
+        } | ${enrollmentHistory.instructors.join(', ')}`;
+    }, [courseNumber, department, enrollmentHistory]);
+
+    const encodedDept = useMemo(() => encodeURIComponent(department), [department]);
+    const axisColor = isDarkMode() ? '#fff' : '#111';
+    const tooltipDateColor = '#111';
```
ap0nia, line 34: You could technically put this at module scope. It's not a big deal because it's a negligible calculation and contained to the component.

apps/antalmanac/src/lib/enrollmentHistory.ts
```diff
@@ -0,0 +1,158 @@
...
+export interface EnrollmentHistoryGraphQLResponse {
+    data: {
+        enrollmentHistory: EnrollmentHistoryGraphQL[];
+    };
+}
+
+// To organize the data and make it easier to graph the enrollment
+// data, we can merge the dates, totalEnrolledHistory, maxCapacityHistory,
+// and waitlistHistory arrays into one array that contains the enrollment data
+// for each day
```
ap0nia: I prefer JSDoc for multiline comments.

```suggestion
/**
  * To organize the data and make it easier to graph the enrollment
  * data, we can merge the dates, totalEnrolledHistory, maxCapacityHistory,
  * and waitlistHistory arrays into one array that contains the enrollment data
  * for each day
 */
```

apps/antalmanac/src/lib/enrollmentHistory.ts
```diff
@@ -0,0 +1,158 @@
...
+    maxCapacity: number;
+    waitlist: number | null;
+}
+
+export class DepartmentEnrollmentHistory {
+    // Each key in the cache will be the department and courseNumber concatenated
+    static enrollmentHistoryCache: Record<string, EnrollmentHistory | null> = {};
+    static termShortNames: string[] = termData.map((term) => term.shortName);
+    static QUERY_TEMPLATE = `{
+        enrollmentHistory(department: "$$DEPARTMENT$$", courseNumber: "$$COURSE_NUMBER$$", sectionType: Lec) {
```
ap0nia, line 50: I'm biased and I think this is a cool application of static properties 😎

apps/antalmanac/src/lib/enrollmentHistory.ts
```diff
@@ -0,0 +1,158 @@
...
+        return parsedEnrollmentHistory;
+    }
+
+    /**
+     * This function sorts the given array of enrollment histories so that
+     * the most recent quarters are in the beginning of the array.
+     *
+     * @param enrollmentHistory an array where each element represents the enrollment
+     * history of a course section during one quarter
+     */
```
ap0nia: How I'd write comments. Definitely important since AA is a writing workshop 🤓 
```suggestion
    /**
     * Sort the given array of enrollment histories so that
     * the most recent quarters are in the beginning of the array.
     *
     * @param enrollmentHistory Array where each element represents the enrollment
     * history of a course section during one quarter.
     */
```

apps/antalmanac/src/lib/enrollmentHistory.ts
```diff
@@ -0,0 +1,158 @@
...
+    courseNumber: string;
+    days: EnrollmentHistoryDay[];
+    instructors: string[];
+}
+
+export interface EnrollmentHistoryDay {
+    date: string;
+    totalEnrolled: number;
+    maxCapacity: number;
+    waitlist: number | null;
```
ap0nia, line 42: I prefer optional over nullable unless `null` means a different thing than `undefined`. e.g. sometimes a `NULL` value in a database might mean that the user selected "No" or something equivalent, and other times it might mean that the user just didn't select anything.

```suggestion
    waitlist?: number;
```

ap0nia (approved):
Good work 🚀 

All the comments I left were personal opinions. Feel free to peruse them before merging!

PR #823, opened 2023-12-01
fix: typescript errors

ap0nia:
## Summary
Fix errors that appear when `tsc` is executed.

- There's one last error due to the schema for custom events, which will be rectified in #814

## Test Plan
Run `tsc`


EricPedley (changes requested):
when I run TSC I get a bajillion errors still. Is this a problem with the code or is it just me? I noticed that some of the errors are stupid like "can't find (package that is already installed)", so maybe my `tsc` is broken?
image

ap0nia:
There should only be one error for the custom events. Could you try removing all instances of `node_modules` and running `pnpm install` from the project root before running `pnpm tsc` in the `apps/antalamanc` directory again?

EricPedley:
> There should only be one error for the custom events. Could you try removing all instances of `node_modules` and running `pnpm install` from the project root before running `pnpm tsc` in the `apps/antalamanc` directory again?

Ok that worked. Do u think its worth throwing this troubleshooting step in the docs somewhere?

ap0nia:
Sure, I actually had one last update from the `update-config` branch that had a list of the frontend libraries we use and possible reasons why. I could re-open that branch and add these troubleshooting steps.

EricPedley:
why reopen the branch instead of just creating a new one?

ap0nia:
Misinput, yeah new branch.

EricPedley (approved):

PR #826, opened 2023-12-01
Add Christmas Logo and move logos to assets folder

EricPedley:
## Summary
Added the christmas logo and moved all logos to an assets folder
## Test Plan
Open staging and look at it
## Issues
N/A


ap0nia (changes requested):
Could you try creating a `src/assets` directorfy and import the png files from there?

I think the simplest strategy is to put everything in the `public` directory and just reference it like ``, but there is one difference in that importing a `png` file directly results in [Vite inlining it](https://vitejs.dev/guide/assets.html), potentially optimizing it during the build process.

So a valid compromise to get the inlining optimization without cluttering the `components` directory would be to make an `assets` directory to put all of the miscellaneous static files that we want inlined.

EricPedley:
Yeah thats a good idea, will do

ap0nia (approved):
LGTM 🚢

PR #831, opened 2023-12-04
test: Add setup file, handle matchMedia

### LINKED ISSUE #827: Schedule test failing
Basically, `window.matchMedia` is a real function but `jsdom` doesn't have it. I found an official way to fix this in `jest`, now we just have to translate it to `vitest`: https://stackoverflow.com/questions/39830580/jest-test-fails-typeerror-window-matchmedia-is-not-a-function
image

KevinWu098:
## Summary
Added in a `setup.ts` file to mock the window.matchMedia function. JSDOM does not support it, resulting in our tests breaking.

## Test Plan
Tests should run clean

## Issues
Closes #827

### DIFF
--- a/apps/antalmanac/tests/setup/setup.ts
+++ b/apps/antalmanac/tests/setup/setup.ts
@@ -1,0 +1,15 @@
+import { vi } from 'vitest';
+
+Object.defineProperty(window, 'matchMedia', {
+    writable: true,
+    value: vi.fn().mockImplementation((query) => ({
+        matches: false,
+        media: query,
+        onchange: null,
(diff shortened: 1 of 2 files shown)


KevinWu098:
@ap0nia I think you mentioned using a different config than `vitest/config`. Is that still how you think it should be implemented?

KevinWu098:
Also, is there any way to change the devops to exclude the author from the random reviewer selection? Or, alternatively, automatically rerun it if it fails?

EricPedley (approved):
LGTM :rocket:

PR #834, opened 2023-12-05
Change colors in manifest

EricPedley:
## Summary
The goal here is to make the top status bar on mobile the same color as our app bar. Right now it's just black.
## Test Plan
Download the staging instance as a PWA on your phone and see if the colors look fine.


EricPedley:
image
Before

EricPedley:
Ok I actually looked at the staging instance on my phone and this didn't work. I don't really want to dig through to figure out why this isn't working, but if anyone else wants to pick this up feel free.

PR #838, opened 2023-12-07
feat: MVP authentication user flow

ap0nia:
## Summary
- Login button allows you to login either as a legacy user with the username or with Google OAuth
- New workflow requires you to login to save a schedule (i.e. "remember me" is always on)
- Sample popover with relevant content depending on the user's current authentication status

## TODO
- Remove the old save/load forms and finalize the authentication flow
- Make an endpoint for uploading schedules
  - The endpoint will differentiate between users logged in with Google and users logged in with the legacy code; the frontend shouldn't have to care 

## Test Plan
The changes can be previewed in staging and merged into the feature branch without disrupting the production website.


ap0nia:
Closing, but will leave branch open for reference.

PR #844, opened 2023-12-11
feat: refactor Import

### LINKED ISSUE #771: Disable Import in Skeleton Mode, Subscribe to Skeleton Mode
Followup from #702

Currently, Import is still able to be clicked and used while in Skeleton Mode. The import will fail, however, so this isn't a particularly pressing issue.

The solution would probably be to refactor Import and have it properly subscribe to the Store to detect Skeleton Mode. Examples can be seen in other AppBar components.

KevinWu098:
## Summary
1. Refactored Import
2. Refactored Term Selector
3. Import now subscribes to AppStore for SkeletonMode, and disables appropriately

## Test Plan
1. Terms should be correctly set from the url, when loading in
2. Terms should update correctly when using the Select element
3. Import should be disabled when in Skeleton Mode (use firefox dev tools, or change the api endpoint)

## Issues
Closes #771


apps/antalmanac/src/components/Header/Import.tsx
```diff
@@ -0,0 +1,248 @@
...
+    const handleOpen = useCallback(() => {
+        setOpen(true);
+    }, []);
+
+    const handleClose = useCallback(() => {
+        setOpen(false);
+    }, []);
+
+    const handleSubmit = async () => {
+        console.log(term);
```
Douglas-Hong: No need for this console log?

Douglas-Hong (approved):
LGTM! Just remove one little line and this is good to go!

Douglas-Hong (approved):

PR #846, opened 2023-12-23
feat: Add warning text to save modal

### LINKED ISSUE #845: Add warning text to save modal
image

Maybe not as long, and maybe not as red, but some users aren't aware that usernames should be unique, otherwise they're at a higher risk of being overwritten

Also, instead of "username", we can say "unique user ID", or something along those lines (don't forget to change the default text too)

tig-github:
## Summary
Added warning for users to use a unique user ID in red text
modalfix

## Test Plan
1. Check on default modal
2. Check if responsive on mobile

Note: using span for red text, probably a better convention

## Issues

Closes #845

### DIFF
--- a/apps/antalmanac/src/components/Header/LoadSaveFunctionality.tsx
+++ b/apps/antalmanac/src/components/Header/LoadSaveFunctionality.tsx
@@ -101,10 +101,8 @@ class LoadSaveButtonBase extends PureComponent<LoadSaveButtonBaseProps, LoadSave
                         <DialogContentText>
                             Enter your unique user ID here to {this.props.actionName.toLowerCase()} your schedule.
                         </DialogContentText>
-                        <DialogContentText>
-                            <span style={{ color: 'red' }}>
-                                Make sure the user ID is unique and secret, or someone else can overwrite your schedule.
-                            </span>
+                        <DialogContentText style={{ color: 'red' }}>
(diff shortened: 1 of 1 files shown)


apps/antalmanac/src/components/Header/LoadSaveFunctionality.tsx
```diff
@@ -99,13 +99,18 @@ class LoadSaveButtonBase extends PureComponent<LoadSaveButtonBaseProps, LoadSave
                     <DialogTitle>{this.props.actionName}</DialogTitle>
                     <DialogContent>
                         <DialogContentText>
-                            Enter your username here to {this.props.actionName.toLowerCase()} your schedule.
+                            Enter your unique user ID here to {this.props.actionName.toLowerCase()} your schedule.
+                        </DialogContentText>
+                        <DialogContentText>
+                            <span style={{ color: 'red' }}>
+                                Make sure the user ID is unique and secret, or someone else can overwrite your schedule.
+                            </span>
```
KevinWu098: You can move the styling into the `DialogContextText`, no need for the span

KevinWu098 (changes requested):
LGTM, good work. Just one small change and we can merge!

KevinWu098 (approved):
🚢

PR #847, opened 2023-12-25
Deprecate isDarkMode in favor of useThemeStore

### LINKED ISSUE #812: isDarkMode() should be deprecated in favor of ThemeStore
## Issue
              This shouldn't exist anymore, and it's consumers should refactor to use the store directly. Up to you to include it in this PR.
_Originally posted by @MinhxNguyen7 in https://github.com/icssc/AntAlmanac/pull/810#discussion_r1405330346_

```typescript
export function isDarkMode() {
    switch (useThemeStore.getState().appTheme) {
        case 'light':
            return false;
        case 'dark':
            return true;
        default:
            return window.matchMedia('(prefers-color-scheme: dark)').matches;
    }
}
```

`isDarkMode`, found in `helpers.ts` should be removed in favor of having components directly subscribe to the ThemeStore.

## Good First Taskers
For any good first task doers, relevant code can be found in `helpers.ts` as noted above. If you're using VSCode, you can find all instances of `isDarkMode()` by right-clicking and looking for "Find All References." 

In each of those files, you'll want to import the ThemeStore's value, and use that instead!
```typescript
    const appTheme = useThemeStore((store) => store.appTheme);
```

### LINKED ISSUE #854: refactor: themeStore
Because I'm a pedant, I'll say that the the better implementation of `appTheme == "dark"` would be `appTheme.isDark()` so that "dark" and "light" aren't magic values. Intellisense tells you, but you ideally shouldn't need Intellisence. I say it for the sake of future code.

_Originally posted by @MinhxNguyen7 in https://github.com/icssc/AntAlmanac/pull/847#pullrequestreview-1795945578_

srukelman:
## Summary
Refactored implementations of isDarkMode() (from helpers) in favor of useThemeStore

Closes #812 #854


apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
```diff
@@ -121,8 +109,20 @@ interface CourseCodeCellProps {
 }
 
 const CourseCodeCell = withStyles(styles)((props: CourseCodeCellProps) => {
+    const appTheme = useThemeStore((store) => store.appTheme);
+
     const { classes, sectionCode } = props;
 
+    const [isHovered, setIsHovered] = useState(false);
+
+    const handleMouseEnter = () => {
```
MinhxNguyen7, line 118: I think this slipped into your commit. It doesn't have to do with deprecating dark mode. 

You can delete the violating code and use `git commit--amend` to change the most recent commit.

MinhxNguyen7 (changes requested):
This mostly looks good other than a different change that slipped into the commit. Also, instead of deprecating `isDarkMode()` and, therefore, leaving it there, just delete it since nothing uses it.

Because I'm a pedant, I'll say that the the better implementation of `appTheme == "dark"` would be `appTheme.isDark()` so that "dark" and "light" aren't magic values. Intellisense tells you, but you ideally shouldn't need Intellisence. I know you didn't write the interface to begin with, so I won't make you do the search-and-replace (but it should be pretty easy). I say it for the sake of future code.

Also, for the future, please fix the merge conflicts so that there is at least one deployment so the reviewer doesn't need to spin up a local instance.

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
srukelman: I added that code because the style on the Section Code Chip does not support the hover pseudo class, so I had to add my own hover event handling to implement the hover styles on the Section Code Chip. Is there a different way I can do this?

apps/antalmanac/src/stores/SettingsStore.ts
```diff
@@ -23,9 +25,19 @@ export const useThemeStore = create<ThemeStore>((set) => {
...
             : 'light';
 
+    const isDark =
+        themeSetting !== 'system'
+            ? themeSetting === 'dark'
+                ? true
+                : false
+            : window.matchMedia('(prefers-color-scheme: dark)').matches
+            ? true
+            : false;
```
MinhxNguyen7: This code chunk is repeated below, so it should be extracted to a function in this file named something like `computeTheme()` or `themeShouldBeDark()`.

There are also serious problems about the way your wrote the logic here.
1) This does not tell you how the condition works if you don't already know. In a function, you can name each condition and combine them to make readable logical sentences. In a different format, you can watch 20 seconds of [this video](https://youtu.be/Bf7vDBBOBUA?si=p_G2cEH9-xhZpmRj&t=40), which I'd highly recommend.
2) The two innermost ternary operations are completely redundant. `themeSetting === 'dark' ? true : false` is equivalent to `themeSetting`. Same with the `matchMedia`. Simplifying this could justify this inline condition *if it only appears once*. Since it doesn't it should be extracted in any case.

I know this seems trivial, but it's critical to write clean, readable code to avoid logical mistakes and save our time. (As trivia, it's at least important enough for Professor Klefstad to mark you down.)

MinhxNguyen7 (changes requested):
Sorry for holding your PR up in review. Everything looks good except for the logic I mentioned. I can't let that slide.

MinhxNguyen7 (approved):
LGTM!

PR #853, opened 2024-01-16
Add custom event to the route

### LINKED ISSUE #819: Making Custom Event into the Map route
For now, the custom event is an independent marker on the map. It should better be in the route that connects between classes and other events.

Image

JacE070:
## Summary
Add custom events to the route

## Test Plan
Tested manually

## Issues

Closes https://github.com/icssc/AntAlmanac/issues/819

### DIFF
--- a/apps/antalmanac/src/components/Map/Map.tsx
+++ b/apps/antalmanac/src/components/Map/Map.tsx
@@ -148,7 +148,7 @@ export default function CourseMap() {
     const [searchParams] = useSearchParams();
     const [selectedDayIndex, setSelectedDay] = useState(0);
     const [markers, setMarkers] = useState(getCoursesPerBuilding());
-    const [customEventMarkers] = useState(getCustomEventPerBuilding());
+    const [customEventMarkers, setCustomEventMarkers] = useState(getCustomEventPerBuilding());
     const [calendarEvents, setCalendarEvents] = useState(AppStore.getCourseEventsInCalendar());
 
     useEffect(() => {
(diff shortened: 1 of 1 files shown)


apps/antalmanac/src/components/Map/Map.tsx
```diff
@@ -274,14 +274,18 @@ export default function CourseMap() {
...
     const startDestPairs = useMemo(() => {
-        return markersToDisplay.reduce((acc, cur, index) => {
-            acc.push([cur]);
-            if (index > 0) {
-                acc[index - 1].push(cur);
-            }
-            return acc;
-        }, [] as (typeof markersToDisplay)[]);
-    }, [markersToDisplay]);
+        const allEvents = [...markersToDisplay, ...customEventMarkersToDisplay];
```
KevinWu098, line 291: `customEventsMarkersToDisplay` doesn't update when new events are made/updated because there's no listener and no code in useEffect that would update it.

KevinWu098 (changes requested):
LGTM aside from the comment on reactivity.

apps/antalmanac/src/components/Map/Map.tsx
JacE070: Makes sense! But will it be better if I add listeners in useEffect so changes are able to be reflected?

KevinWu098: Oh yes, that's the proper way to do it; sorry for not communicating my feedback more clearly! Didn't want to write the comment on code outside of the diff, so I just chose 277 instead.

JacE070: Looks like a duplicate to https://github.com/icssc/AntAlmanac/pull/855 0.0 Can I leave the event listener undeclared right now and let that PR (and possibly one following) do the work?

KevinWu098: Sure, why not.

KevinWu098 (approved):
LGTM; will be finished up by #855

PR #855, opened 2024-01-17
Fixed Custom Events Loading Onto Map

### LINKED ISSUE #813: Custom Events don't load initially onto Map
## Issue
Updates to Custom Events will not properly reflect onto the Map until it's rerendered. Additionally, Custom Events won't load initially onto Map if you visit https://antalmanac.com/map

chrome-capture-2023-10-26

## Good First Taskers
For any good first task doers, the solution shouldn't take more than ~13 lines of code by my estimation, and should be entirely contained to one file!

I'll do less hand-holding in this Issue, but I'll hint that the relevant file will be in the `Map` folder. Once you've found the relevant file, look at the surrounding code and try to see where there's an inconsistency in how Custom Events are handled versus Course Events! 

<details>
  <summary>If you need an extra nudge:</summary>

  take a look at `Map.tsx`
</details>

<details>
  <summary>And as a final nudge:</summary>

  the inconsistency I bring up above is related to `useEffect()`
</details>

_P.S: The relevant PR which implemented/caused the bug would be #726, so you can take a peek if you're curious about the source of the bug._

adcockdalton:
## Summary
1. Add proper listeners for custom events updating in a manner analogous to that of classes in the Map directory

## Test Plan
1. View map
2. Add new custom event with location (try RH, it's one of the acceptable and nearby ones)
3. See if change is reflected
4. See if change reflects when saved and you visit antalmanac.com/map

## Issues

Closes #813 

## Future Followup
1. Not all locations are recognized for custom events on the map even though they are for classes (BS3, for example) and some don't even read on the calendar
2. Routes do not work for custom events (it is difficult to order them correctly with the classes due to how the indexes work)

### DIFF
--- a/apps/antalmanac/src/components/Map/Map.tsx
+++ b/apps/antalmanac/src/components/Map/Map.tsx
@@ -148,7 +148,7 @@ export default function CourseMap() {
     const [searchParams] = useSearchParams();
     const [selectedDayIndex, setSelectedDay] = useState(0);
     const [markers, setMarkers] = useState(getCoursesPerBuilding());
-    const [customEventMarkers] = useState(getCustomEventPerBuilding());
+    const [customEventMarkers, setCustomEventMarkers] = useState(getCustomEventPerBuilding());
     const [calendarEvents, setCalendarEvents] = useState(AppStore.getCourseEventsInCalendar());
 
     useEffect(() => {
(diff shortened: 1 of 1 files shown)


apps/antalmanac/src/components/Map/Map.tsx
```diff
@@ -167,6 +167,22 @@ export default function CourseMap() {
...
+        AppStore.on('currentScheduleIndexChange', updateCustomEventMarkers);
+        AppStore.on('colorChange', updateCustomEventMarkers);
+
+        return () => {
+            AppStore.removeListener('customEventsChange', updateCustomEventMarkers);
+            AppStore.removeListener('currentScheduleIndexChange', updateCustomEventMarkers);
+            AppStore.removeListener('colorChange', updateCustomEventMarkers);
+        };
+    }, []);
+
```
KevinWu098: There's a good amount of code duplication between this `useEffect()` and the one above for regular markers. Instead of adding three new listeners, you can change the original `updateMarkers()` function to _also_ set custom event markers.

Whether you refactor the name of `updateMarkers()` is up to you

KevinWu098 (changes requested):
Thanks for the PR. Functionality is good, bug appears fixed. Just a comment on code quality/code duplication!

KevinWu098:
For the future followup, I've opened #856 to address point 1 (good catch!), and for point 2, that should be added in #853

KevinWu098 (approved):
LGTM 💖

PR #864, opened 2024-01-22
testing pr: auth devops

### LINKED ISSUE #857: prepare devops for auth setup
# Summary
Revamp the devops and add setup whatever's necessary to facilitate the new auth system.

## Why
A new DDB table has to be created for backend on staging instances. One DDB table for users logged in with a username and another for users logged in via Google OAuth.

# Project
Will open a PR by Friday.

ap0nia:
## Summary
Overhaul the devops and add a new "auth-user" table in DynamoDB for development and production environments to be used for storing schedules authenticated with Google separately from usernames.

### DevOPs Scenarios
- [x] Staging frontend deploy refers to `dev.api.antalmanac.com` when neither the CDK or backend changed.
- [x] Staging frontend refers to staging-# when either the CDK or backend changed.
- [x] Backend staging deploy runs when either the CDK or backend changed.
- [x] Backend staging deploy is skipped when neither the CDK or backend changed.
- [x] Staging instances are successfully destroyed when the PR is closed.
- [x] Deployment environments are deactivated when the PR is closed.
- [ ] Production application on GitHub Pages gets updated properly during production deploy.
- [ ] New development backend gets deployed properly during production deploy.
- [x] Creates user DynamoDB database - for users using the legacy username system.
- [x] Creates auth user DynamoDB database - for users logging in with Google.

## Usage Scenarios
- [x] Can load/save schedules to staging-# api endpoint.
- [x] Can view the map.

### Issues I ran into
- Apparently I can't split variables onto multiple lines.
- Whenever an output returns a boolean value, GitHub stores it as a string, i.e `true` becomes `"true"`.
- In light of the former, I have to compare boolean values to "true".
- Alternatively, I could use `fromJson` to convert them to boolean values 🤷.

## Test Plan
Open and close PRs with this branch until the staging deploy and destroy behaviors are verified.

## Issues

Closes #857


ap0nia:
Will close to verify whether the staging instances are destroyed and deployment environments are deactivated.

ap0nia:
Oops, I forgot that the apps do have to be built or CDK will get mad 😔

ap0nia:
Ok, let's try closing again.

ap0nia:
It didn't properly skip the yes/no prompt. I need to add "--force" to the destroy commend.

ap0nia:
Closing again, hopefully it works this time 🤞

ap0nia:
Reopening to test deploy development backend.

ap0nia:
Last time closing. This is just to confirm that the staging deployment environments are **destroyed**, not just deactivated.

ap0nia:
More testing.

ap0nia:
Ok, devops hype time

ap0nia:
Wasn't hype the last time. Will be hype now 🚀

ap0nia:
Ok, it was close enough 😔

.github/actions/create-deployment/action.yml
ap0nia, line 1: GitHub's [built-in support for deployments + environments](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment) is too limited.

A simple environment name like "staging" can only have one deployment active at a time. A specific environment name like "staging-frontend-684" isn't trivially deleted after the fact (the permissions required to delete an environment is inanely high and outside the scope of GitHub Actions)


## Review of Deployments and Environments
- An environment describes a deployment target, like "production", "staging", "development".
- A deployment is like an "attempt" to deploy to that envrionment. e.g. every time you push a commit to a PR, a new attempt is made to deploy to the staging environment.
- Deployment statuses are associated with one deployment and indicate its status 😱 
- So a workflow might go as follows:
  1. Create a new deployment that's marked as "in_progress" and save this ID as an output.
  2. Download Git repository, build stuff, etc.
  3. Create a deployment status for the previously made deployment by using the saved ID, indicating "success", "failure", etc.

.github/actions/create-deployment/src/index.js
ap0nia, line 1: I don't think there's too much benefit to understanding this file outside of devops, so you can mostly ignore it until something breaks, then you can let me know 🤓 

## Why JS?
No compilation step, and the action can be run after downloading the repository and installing dependencies without any extra steps at any other point in time. Downside is that the deployment doesn't get created until after those two steps, which might take a while.

.github/actions/create-deployment/tsconfig.json
ap0nia, line 1: These are my personally preferred tsconfig settings, you can ignore them because their scope has been restricted to the CDK and GitHub Actions JS scripts I wrote for devops; nothing has changed for the frontend/backend apps.

.github/actions/delete-deployment/src/index.js
ap0nia, line 1: There will be two staging environments, `staging-frontend` and `staging-backend`. Each environment can house multiple deployments, which will be identified by their payload (a JSON field that can store custom info). Because of this, deployments shouldn't be manually deactivated, so a separate action needs to take care of that (i.e. by filtering for the correct payload and deleting the related deployments.

GitHub only allows one active deployment per environment by default, but we want deployments based on open PRs, which can mean multiple active at once.

.github/actions/setup-node-and-pnpm/action.yml
ap0nia, line 1: Convenience reusable workflow for setting up Node.js and PNPM cache; basically used everywhere.

.github/actions/stabilize_stack/action.yaml
ap0nia, line 1: The following deleted files are reusable workflows. I dislike splitting up the main workflow too much because it makes the deployment process unclear. The primary steps of the deployment should be readily visible, e.g. 
1. checkout repository
2. install dependencies
3. build
4. deploy to CDK
5. done

This is literally the main benefit of YAML in the first place -- viewing processes in a linear fashion.

And lower level details like "install correct Node.js version and setup cache before installing dependencies" are valid abstraction targets IMO.

.github/workflows/deploy_staging.yml
```diff
@@ -0,0 +1,190 @@
...
+    AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+    AWS_DEFAULT_REGION: us-east-1
+
+    # Information used by AWS CDK stacks.
+    HOSTED_ZONE_ID: ${{ secrets.HOSTED_ZONE_ID }}
+    CERTIFICATE_ARN: ${{ secrets.CERTIFICATE_ARN }}
+    PR_NUM: ${{ github.event.pull_request.number }}
+
+jobs:
+    get_staging_configuration:
```
ap0nia, line 48: This is a somewhat sophisticated job that can be reusable, except it's only relevant to deploying staging, so it's fine as-is.

.github/workflows/deploy_staging.yml
```diff
@@ -0,0 +1,190 @@
...
+
+concurrency:
+    group: ${{ github.workflow }}-${{ github.ref }}
+    cancel-in-progress: true
+
+permissions:
+    contents: read
+    deployments: write
+
+env:
```
ap0nia, line 20: A nice improvement would be to consolidate the env somewhere, but this isn't the worst thing just because it's pretty visible/clear what the env should look like for each workflow.

.github/workflows/deploy_staging.yml
ap0nia, line 1: # Summary of Workflow
1. Checkout repository.
2. Install dependencies.
3. Create "in_progress" deployment.
4. Build frontend or backend app.
5. Deploy built app via CDK.
6. Create "success" or "failure" deployment status.

Afterwards, a new event should be posted on the pull request, similar to the URL commenting, but via GitHub's deployment API. For example:
image

apps/cdk/package.json
```diff
@@ -1,34 +1,40 @@
...
+    "repository": {
+        "type": "git",
+        "url": "https://github.com/icssc/AntAlmanac",
+        "directory": "apps/cdk"
+    },
+    "scripts": {
+        "backend-development": "cdk --app 'npx tsx src/app/backend-development.ts' --require-approval never",
+        "backend-staging": "cdk --app 'npx tsx src/app/backend-staging.ts' --require-approval never",
+        "backend-production": "cdk --app 'npx tsx src/app/backend-production.ts' --require-approval never",
+        "frontend-staging": "cdk --app 'npx tsx src/app/frontend-staging.ts' --require-approval never"
```
ap0nia, line 19: These are CDK scripts with pre-bound apps. It allows you to suffix the command with "deploy" or "destroy". e.g.

`pnpm run backend-development deploy`
`pnpm run frontend-staging destroy`

apps/cdk/src/lib/wait-for-stack-idle.ts
ap0nia, line 1: This deprecates the YAML code from the reusable workflow, [stabilize stack](https://github.com/icssc/AntAlmanac/blob/main/.github/actions/stabilize_stack/action.yaml)

apps/cdk/src/stacks/backend.ts
ap0nia, line 1: Equivalent to the current [BackendStack](https://github.com/icssc/AntAlmanac/blob/main/apps/cdk/lib/backend.ts)

apps/cdk/src/stacks/frontend.ts
ap0nia, line 1: Equivalent to the current [FrontendStack](https://github.com/icssc/AntAlmanac/blob/main/apps/cdk/lib/frontend.ts)

apps/cdk/src/app/backend-development.ts
ap0nia, line 1: Separate apps are used to handle things like checking for `PR_NUM` in staging stacks, or using the appropriate stack name for production stacks. It also facilitates the previously mentioned pre-bound app CDK scripts with very little overhead.

apps/cdk/lib/app.ts
ap0nia, line 1: The single app, single CDK command approach has been deprecated in favor of multiple apps and multiple CDK commands. It allows each stack to be named accordingly, handle specific environment variables individually, etc. and reuse the same stack in a more modular manner.

ap0nia (commented):
Hopefully these comments are helpful! 🙏

.github/actions/create-deployment/src/index.js
```diff
@@ -0,0 +1,176 @@
...
+        return branchMatch[1] ?? '';
+    }
+
+    const tagMatch = REGEX.TAG.exec(ref);
+    if (tagMatch) {
+        return tagMatch[1] ?? '';
+    }
+
+    return process.env['GITHUB_HEAD_REF'] ?? '';
+}
```
Voark, line 42: Why is this done with regex instead of the default [Github context object](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context)? If I'm not mistaken you can just use `github.ref_name` and `github.ref_type`.

ap0nia: Oh yeah I should mention that. The ref for all PRs always looks like `refs/pull/<PR NUM>/merge`. If you use this as the ref for a deployment, it gives you absolutely no information on the deployment dashboard. It will always just say "merging PR# into main" and that's it. Literally GitHub doesn't use this ref in their own Actions deployment/environmwnt API either; you have to use the literal short branch name or `refs/heads/<branch name>` for the commit title to show up on the deployment info, and most importantly for the PR to receive the "deployed" buttons.

I spent like an hour agonizing over this lol

Voark: Oh I see yea I forgot about that. We love Github actions!

.github/workflows/destroy_staging.yml
```diff
@@ -0,0 +1,100 @@
...
+              uses: actions/checkout@v3
+
+            - name: Setup Node.js and pnpm
+              uses: ./.github/actions/setup-node-and-pnpm
+
+            - name: Install dependencies
+              run: pnpm install --frozen-lockfile
+
+            - name: Build frontend
+              run: pnpm --filter "antalmanac" build
```
EricPedley, line 62: Why does this need to build the frontend? Shouldn't it just run the cdk destroy command? Also, can we just pnpm install only the dependencies for the cdk code?

EricPedley (changes requested):
One comment to address, nothing huge. I think our best bet here is to just pick a time to merge this when we can monitor prod closely to rollback or hotfix anything that goes wrong, cause these changes are pretty massive so I don't trust myself to have caught any tricky failure modes.

.github/workflows/destroy_staging.yml
ap0nia: Re: https://github.com/icssc/AntAlmanac/commit/54032fd0161015d829a74fc75bcbab6266b77b86

The CDK requires that all referenced assets actually exist and are in the right place. When deploying, we indicate that we want to upload the `apps/antalmanac/build` directory to S3, and CDK checks that it exists prior to actually doing anything. Likewise, it does the same when destroying the stack. So you should basically do the same steps as deploying, but just invoke "destroy" at the end.

I think it's pretty dumb, and it's something we had to deal with in PPA too lol

EricPedley: Understandable. Good thing Actions cpu time is free for us lol

EricPedley (approved):
ok just merge this whenever you're ready to monitor the prod deployment. YOLO!

PR #865, opened 2024-01-23
Auto-Save Feature Complete

### LINKED ISSUE #817: AutoSave Feature (Overtaken by #891)
My friends and I have all experienced closing the browser and finding no savings. Especially for freshmen, they may not know they need to save when they first use this tool. So, I think adding an autosave feature will improve usability. 
 
1. The creation or login UserID is displayed if the user does not automatically log in to the previous UserId when adding a course. 
2. Autosave can call every minute, or every time after adding, deleting, or making any changes (but this may give the server a lot of pressure?).

stevenguyukai:
# Summary
### 1.  Key feature 1 - Adding Auto Save Setting 

image

The default setting is `OFF`. This is considering the workload that the server may experience. 
When a user sets it to `ON`, every action from the user will be saved. (For example, `AddCourse` `ChangeCustomEventColor` etc ) 

asd

### 2.  Key feature 2 - Alert User of unsaved action

image

When the user didn't set the autosave to `ON` and didn't manually save. Next time users enter the website, they will receive an alert of unsaved action. If the user clicks `OK` (Despite the picture, the button is in Chinese; it changes as system settings), the actions users made previously will be loaded to the calendar. 

## Test Plan

Check for every action, like `DeleteCourse,` `UpdateCustomEvent,` etc., to work as intended.


apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -86,10 +86,14 @@ export const saveSchedule = async (userID: string, rememberMe: boolean) => {
...
-                    'success',
-                    `Schedule saved under username "${userID}". Don't forget to sign up for classes on WebReg!`
-                );
+                if (!autoSave) {
+                    openSnackbar(
+                        'success',
+                        `Schedule saved under username "${userID}". Don't forget to sign up for classes on WebReg!`
+                    );
+                } else {
+                    openSnackbar('success', `Updated`);
```
KevinWu098: I think this could probably be a little less terse, but I'm not too sure what the message *should be*. Open to suggestions!

apps/antalmanac/src/stores/AppStore.ts
```diff
@@ -299,6 +341,29 @@ class AppStore extends EventEmitter {
...
     termsInSchedule = (term: string) =>
         new Set([term, ...this.schedule.getCurrentCourses().map((course) => course.term)]);
+
+    autoSaveSchedule = (action: any[] = [null, null, null]) => {
+        const autoSave = typeof Storage !== 'undefined' && window.localStorage.getItem('autoSave') == 'true';
+        if (autoSave) {
+            const savedUserID = window.localStorage.getItem('userID');
+
+            if (savedUserID != null) {
+                saveSchedule(savedUserID, true, true).then((r) => void r);
```
KevinWu098: What is this `.then void func` for?

```suggestion
                saveSchedule(savedUserID, true, true);
```

apps/antalmanac/src/stores/AppStore.ts
```diff
@@ -299,6 +341,29 @@ class AppStore extends EventEmitter {
...
+        } else {
+            if (window.localStorage.getItem('unsavedAction') == null) {
+                const unsavedAction = [action];
+                localStorage.setItem('unsavedAction', JSON.stringify(unsavedAction));
+            } else {
+                // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+                // @ts-ignore
+                const unsavedAction = JSON.parse(window.localStorage.getItem('unsavedAction'));
+                unsavedAction.push(action);
+                localStorage.setItem('unsavedAction', JSON.stringify(unsavedAction));
```
KevinWu098: I'd rather not use `@ts-ignore` here; Would prefer to rewrite this to handle the `null` more gracefully, like you do above on line 350.

apps/antalmanac/src/stores/AppStore.ts
```diff
@@ -299,6 +341,29 @@ class AppStore extends EventEmitter {
 
     termsInSchedule = (term: string) =>
         new Set([term, ...this.schedule.getCurrentCourses().map((course) => course.term)]);
+
+    autoSaveSchedule = (action: any[] = [null, null, null]) => {
+        const autoSave = typeof Storage !== 'undefined' && window.localStorage.getItem('autoSave') == 'true';
+        if (autoSave) {
+            const savedUserID = window.localStorage.getItem('userID');
+
+            if (savedUserID != null) {
```
KevinWu098: Can this be simplified?
```suggestion
            if (savedUserID) {
```

apps/antalmanac/src/stores/AppStore.ts
```diff
@@ -223,6 +232,37 @@ class AppStore extends EventEmitter {
             return false;
         }
         this.unsavedChanges = false;
+
+        if (window.localStorage.getItem('unsavedAction') != null) {
+            if (confirm('You have unsaved changes. Would you like to load them?')) {
+                // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+                // @ts-ignore
```
KevinWu098: Would prefer not to `ts-ignore` the localStorage

apps/antalmanac/src/stores/AppStore.ts
```diff
@@ -223,6 +232,37 @@ class AppStore extends EventEmitter {
...
+                        this.schedule.editCustomEvent(action[1], action[2]);
+                    } else if (action[0] == 'changeCustomEventColor') {
+                        this.schedule.changeCustomEventColor(action[1], action[2]);
+                    } else if (action[0] == 'changeCourseColor') {
+                        this.schedule.changeCourseColor(action[1], action[2], action[3]);
+                    } else if (action[0] == 'clearSchedule') {
+                        this.schedule.clearCurrentSchedule();
+                    } else if (action[0] == 'copySchedule') {
+                        this.schedule.copySchedule(action[1]);
+                    }
```
KevinWu098: I think a `switch` statement is better here. Alternatively, use a giga-mega-uber nested ternary /joke

apps/antalmanac/src/stores/AppStore.ts
```diff
@@ -299,6 +341,29 @@ class AppStore extends EventEmitter {
 
     termsInSchedule = (term: string) =>
         new Set([term, ...this.schedule.getCurrentCourses().map((course) => course.term)]);
+
+    autoSaveSchedule = (action: any[] = [null, null, null]) => {
```
KevinWu098: I'd like to explicitly define potential actions, rather than use an `any`.

apps/antalmanac/src/stores/AppStore.ts
```diff
@@ -223,6 +232,37 @@ class AppStore extends EventEmitter {
             return false;
         }
         this.unsavedChanges = false;
+
+        if (window.localStorage.getItem('unsavedAction') != null) {
+            if (confirm('You have unsaved changes. Would you like to load them?')) {
```
KevinWu098: Will discuss more in the main review comment, but I'm feeling more and more that this feature should be a separate PR and be fleshed out a bit more.

KevinWu098 (changes requested):
## Summary
Very exciting PR, looking forward to using this feature myself. Still, gotta dig into a few considerations.

First, this feature is both an **auto-save** and a **save unsaved actions** feature. I don't think it violates atomic commits *too* much, but I'd rather split them up (plus, you get an additional commit ;)).

Edit: oh also the merge conflicts :')

## Auto-Save
Good feature, works well. Not much to consider here other than a little code cleanup. 

The main consideration I want to bring up is that the constant deluge of "Updated" popups is a tad loud. Would love to investigate a way of "buffering" or changing how often update messages are sent.

## Unsaved Actions
Also a great feature, but needs a little more TLC. 

I think an Enum or some other explicit type definition for the unsavedActions is probably best. Additionally, extracting some of the logic from `loadSchedule` would also be preferred.

--------------
Can discuss this more at tomorrow's (actually today's) meeting!

apps/antalmanac/src/actions/AppStoreActions.ts
stevenguyukai: Or maybe not using the snack bar to alert the user at all?
Use the following style for `SAVE` to show that there is a saving going on.

image

stevenguyukai:
# Todo

## 1. Make Auto Saving animation circle
## 2. Change the style for the experimental feature `AutoSave` in the setting

KevinWu098 (changes requested):
Looks like a bit of code overwriting due to the merge, rerequest when that's handled 😅

stevenguyukai:
This pull request is been overtaken by #891

PR #867, opened 2024-01-28
Show all past enrollment graphs

### LINKED ISSUE #830: Allow past enrollment popup to show the enrollment history of all previous quarters
In #820, we only display the enrollment history of one lecture in the most recent quarter a course is offered. To expand upon this, we can make it so that the user can view the enrollment history of any lecture that was previously offered. We can refer to the Figma design from #341:

image

Douglas-Hong:
## Summary
- The user can navigate back and forth between all the past enrollment graphs for a specific course

image

## TODO
- [x] Show all past enrollment graphs
- [x] Improve the design for mobile

## Test Plan
- Verify this works for classes in the most recent quarter (e.g. CS 143A).
- Verify this works for classes that haven't been offered very recently (e.g. CS 122D from Spring 2023).
- Make sure the modal looks good on mobile and light/dark mode.

## Issues

Closes #830 

## Future Follow-up
- I tried to add an instructor filter, but Material UI has problems when we add a select menu to a popper. We can tackle this later. Or, we can just not implement it because it would clutter the popper, and the user can go to ZotTracker for more advanced searching.


apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
```diff
@@ -21,10 +21,13 @@ import { openSnackbar } from '$actions/AppStoreActions';
...
-    const courseColors = AppStore.schedule.getCurrentCourses().reduce((accumulator, { section }) => {
-        accumulator[section.sectionCode] = section.color;
-        return accumulator;
-    }, {} as { [key: string]: string });
+    const courseColors = AppStore.schedule.getCurrentCourses().reduce(
+        (accumulator, { section }) => {
+            accumulator[section.sectionCode] = section.color;
+            return accumulator;
+        },
+        {} as { [key: string]: string }
```
ap0nia:
```suggestion
        {} as Record<string, string>
```

apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
```diff
@@ -64,32 +78,55 @@ export function EnrollmentHistoryPopup({ department, courseNumber }: EnrollmentH
         );
     }
 
+    const handleBack = () => {
+        setGraphIndex((prev) => prev - 1);
+    };
+
+    const handleForward = () => {
+        setGraphIndex((prev) => prev + 1);
+    };
```
ap0nia: You could memoize these callbacks if you want.

```suggestion
    const handleBack = useCallback(() => {
        setGraphIndex((prev) => prev - 1);
    }, []);

    const handleForward = useCallback(() => {
        setGraphIndex((prev) => prev + 1);
    }, []);
```

apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
```diff
@@ -1,6 +1,16 @@
...
+    LineChart,
+    Line,
+    CartesianGrid,
+    ResponsiveContainer,
+    XAxis,
+    YAxis,
+    Tooltip as RechartsTooltip,
+    Legend,
+} from 'recharts';
+import { ArrowBack, ArrowForward } from '@material-ui/icons';
```
ap0nia:
```suggestion
import { ArrowBack, ArrowForward } from '@mui/icons-material';
```

apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
```diff
@@ -64,32 +78,55 @@ export function EnrollmentHistoryPopup({ department, courseNumber }: EnrollmentH
...
+                    display: 'flex',
+                    justifyContent: 'space-between',
+                    alignItems: 'center',
+                    width: graphWidth,
                 }}
             >
-                {popupTitle}
-            </Typography>
+                <Tooltip title="Newer Graph">
+                    {/* In order for a tooltip to work properly with disabled buttons, we need to wrap the button in a span */}
```
ap0nia: Good comment 👍

apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
```diff
@@ -64,32 +78,55 @@ export function EnrollmentHistoryPopup({ department, courseNumber }: EnrollmentH
...
+                <Tooltip title="Older Graph">
+                    <span>
+                        <IconButton onClick={handleForward} disabled={graphIndex === enrollmentHistory.length - 1}>
+                            <ArrowForward />
+                        </IconButton>
+                    </span>
+                </Tooltip>
+            </Box>
             <Link
                 href={`https://zot-tracker.herokuapp.com/?dept=${encodedDept}&number=${courseNumber}&courseType=all`}
```
ap0nia, line 154: You could optionally define this as a variable prior to the return statement (optionally also memoized), so that the readability of the JSX is simpler.

apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
```diff
@@ -64,32 +78,55 @@ export function EnrollmentHistoryPopup({ department, courseNumber }: EnrollmentH
...
                 target="_blank"
                 rel="noopener noreferrer"
                 sx={{ display: 'flex', height: graphHeight, width: graphWidth }}
             >
                 <ResponsiveContainer width="95%" height="95%">
-                    <LineChart data={enrollmentHistory.days} style={{ cursor: 'pointer' }}>
+                    <LineChart data={enrollmentHistory[graphIndex].days} style={{ cursor: 'pointer' }}>
```
ap0nia: I think this would be a good variable to include outside of the return statement. e.g.
```tsx
const lineChartData = enrollmentHistory[graphIndex].days
```

apps/antalmanac/src/lib/enrollmentHistory.ts
```diff
@@ -84,11 +83,7 @@ export class DepartmentEnrollmentHistory {
         if (res?.length) {
```
ap0nia: For a more succinct guard clause, you can reverse the condition
```ts
if (!res?.length) {
  return null
}

// parse enrollment history here
```

apps/antalmanac/src/lib/enrollmentHistory.ts
```diff
@@ -68,14 +68,13 @@ export class DepartmentEnrollmentHistory {
         this.partialQueryString = DepartmentEnrollmentHistory.QUERY_TEMPLATE.replace('$$DEPARTMENT$$', department);
     }
 
-    async find(courseNumber: string): Promise<EnrollmentHistory | null> {
+    async find(courseNumber: string): Promise<EnrollmentHistory[] | null> {
         const cacheKey = this.department + courseNumber;
```
ap0nia, line 72: Optional, but this might be useful as a static method, so that creating and decoding cache keys are centralized. 
This is pretty insubstantial in this scenario tbh, but might be useful thinking in some other case.
e.g.
```tsx
// Example: splitting with '/'
createKey(courseNumber: string) {
  return this.department + '/' + courseNumber;
}

// For example, this behavior could be enabled
decodeKey(key: string) {
  const [department, courseNumber] = key.split('/')
}
```

ap0nia (changes requested):
Everything looks solid 💯 

Most concerns are just style and preference.

apps/antalmanac/src/lib/enrollmentHistory.ts
```diff
@@ -68,30 +68,33 @@ export class DepartmentEnrollmentHistory {
...
-        return (DepartmentEnrollmentHistory.enrollmentHistoryCache[cacheKey] ??= await this.queryEnrollmentHistory(
-            courseNumber
-        ));
+    createCacheKey(courseNumber: string) {
+        return this.department + '/' + courseNumber;
+    }
+
+    decodeCacheKey(cacheKey: string) {
+        return cacheKey.split('/');
     }
```
ap0nia: Could you remove this, I think this is a bad suggestion lol

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -226,8 +226,8 @@ function CustomEventDialogs(props: CustomEventDialogProps) {
                         {disabled
                             ? 'Schedule and day must be checked'
                             : props.customEvent
-                            ? 'Save Changes'
-                            : 'Add Event'}
+                              ? 'Save Changes'
+                              : 'Add Event'}
```
ap0nia: Could you remove this diff?
You might need to commit with `git commit --no-verify` to prevent the linter from acting up. Or `git commit --empty` to allow empty changes.

apps/antalmanac/src/components/Map/Map.tsx
```diff
@@ -277,13 +277,16 @@ export default function CourseMap() {
...
+        return allEvents.reduce(
+            (acc, cur, index) => {
+                acc.push([cur]);
+                if (index > 0) {
+                    acc[index - 1].push(cur);
+                }
+                return acc;
+            },
+            [] as (typeof allEvents)[]
+        );
```
ap0nia: Could you remove this change?

apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
```diff
@@ -1,18 +1,80 @@
...
 import { MOBILE_BREAKPOINT } from '../../../globals';
 import { DepartmentEnrollmentHistory, EnrollmentHistory } from '$lib/enrollmentHistory';
 import { useThemeStore } from '$stores/SettingsStore';
 
-export interface EnrollmentHistoryPopupProps {
+interface PopupHeaderProps {
+    graphWidth: number;
+    graphIndex: number;
+    handleForward: () => void;
+    handleBack: () => void;
```
ap0nia: It might be helpful to generalize this callback function. e.g.

```tsx
type PopupHeaderCallback = () => void
```

ap0nia (approved):
Looks good, thanks for the enhancement 💯

ap0nia (approved):
Good changes 👍

KevinWu098:
Great feature, but I'm very confused about the arrow placement?

Why are *past* enrollments navigated to with the right arrow and vice versa? It seems very unintuitive, and after some user testing, 5/5 users immediately went towards the lefthand side when asked to show me a prior quarter's courses.

Screenshot 2024-02-23 at 12 45 01 PM

@Douglas-Hong Can we swap it around?

Douglas-Hong:
@KevinWu098 that's a good point. I also thought about this a lot when I was implementing the feature. Here is my thought process: first, I want to show the most recent quarter's enrollment data when the popup is opened since the most recent quarters are probably more relevant. Then, I thought it would be awkward to have the right arrow instead of the left arrow be disabled when the popup first appears because normally that implies we have reached the end of a list. Therefore, instead of having the list be from oldest quarter to most recent quarter, I thought it would be better to go from most recent quarter to oldest quarter. I'm not sure if that makes sense, lmk if you need any clarification.

However, based on the user testing, I think it's a good idea to swap the arrows. @EricPedley @ap0nia do you guys have any opinions about this?

KevinWu098:
> @KevinWu098 that's a good point. I also thought about this a lot when I was implementing the feature. Here is my thought process: first, I want to show the most recent quarter's enrollment data since the most recent quarters are probably more relevant. Then, I thought it would be awkward to have the right arrow instead of the left arrow be disabled when the popup first appears because normally that implies we have reached the end of a list. Therefore, instead of having the list be from oldest quarter to most recent quarter, I thought it would be better to go from most recent quarter to oldest quarter. I'm not sure if that makes sense, lmk if you need any clarification.
> 
> 
> 
> However, based on the user testing, I think it's a good idea to swap the arrows. @EricPedley @ap0nia do you guys have any opinions about this?

Your reasoning makes sense, but I think we have to defer to the "default user behavior" of "older being lefter"

MinhxNguyen7:
I agree that older should be left.

PR #868, opened 2024-01-29
feat: swap to search on mobile

### LINKED ISSUE #858: Switch to search automatically on mobile when not logged in
So that it's not confusing for a new user.

KevinWu098:
## Summary
1. On mobile, if a user isn't signed in, we'll default to the Classes Tab

Demo

## Test Plan
1. On mobile, check functionality both when signed in and signed out
3. Should swap to classes tab when signed out
4. Shouldn't swap / stay on calendar tab when signed in

## Issues
Closes #858


ap0nia (approved):
LGTM 🚀

PR #869, opened 2024-01-29
feat: copy custom events

### LINKED ISSUE #725: Copy Schedule doesn't copy Custom Events
As the title describes, the current implementation of Copy Schedule does not copy over custom events. Should it?

I'm paraphrasing here, but Fred said that the lack of Custom Event copying was neither intended nor unintended (but probably could/should be added), moreso just an oversight on his part when working on the codebase there. So, it's up to us to choose whether or not to copy over Custom Events as well.

KevinWu098:
## Summary
1. `copySchedule` now also copies custom events

## Test Plan
1. Check that custom events are indeed copied over, for both specific and all schedules

## Issues
Closes #725

### DIFF
--- a/apps/antalmanac/src/stores/Schedules.ts
+++ b/apps/antalmanac/src/stores/Schedules.ts
@@ -126,6 +126,14 @@ export class Schedules {
                 this.addCourse(course, to, false);
             }
         }
+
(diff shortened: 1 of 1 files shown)


stevenguyukai (approved):
Looks good!

PR #870, opened 2024-01-31
add text to header

EricPedley:
## Summary
Opening this to test if staging is working


EricPedley:
now let's see if the stack annihilates itself

PR #874, opened 2024-02-01
fix: tutorial followup

KevinWu098:
## Summary
Noticed two small bugs in the tutorial:

1. `localStorage.getItem()` was using `userId` instead of `userID`
2. tourHasRun isn't set to true if `esc` or close button is used on the first page

## Issues

Closes #


MinhxNguyen7 (approved):
LGTM!

PR #875, opened 2024-02-01
Refactor dark mode logic

MinhxNguyen7:
## Summary
Implementing what I was trying to explain in https://github.com/icssc/AntAlmanac/pull/847#discussion_r1459873553. Only the second commit is relevant.

I believe that the code written here is much more readable and should be fully comprehensible within a minute of anyone reading. It's clear, minimally-nested, and well-named.

I know that this seems like a matter of pedantry, but it genuinely took me more than 5 minutes to understand what the logic was there. Someone might come along in the future to work on this code and be even more confused. Since you already understood how the logic works, it would've only taken you a few minutes to save us much more time (and perhaps even bugs) in the future. 

If you don't understand what I'm getting at and why this is important, @ap0nia can explain it to you in person.


srukelman (approved):
I see how this logic is so much clearer than what I wrote. Thank you.

Issue #778, opened 2023-11-02
Use Calendar endpoint instead of hardcoding term dates.


MinhxNguyen7:
There's an endpoint on PP to get term dates, but we're still hardcoding it like monkeys.

https://docs.api-next.peterportal.org/docs/developers-guide/rest-api/reference/calendar

### Related Issues
#758

KevinWu098:
Terms endpoint is kinda dope >:), but it's not documented ;(

KevinWu098:
Passing this onto other AntAlmanac folk 🫡

MinhxNguyen7:
It's too much of a PITA.

PR #881, opened 2024-02-05
Display Finals Date From Default Term

### LINKED ISSUE #758: RFC: Finals date

adcockdalton:
## Summary

1. Added new optional field to TermData.ts "finalsStartDate"
2. Changed calendarizeFinals to correctly assign finals to their respective dates instead of the arbitrary (2018, 0, 1)
3. Refactored CalendarRoot to use date instead of defaultDate
4. Dynamically update date

Regular classes still populate to the arbitrary 2018 date, which is simpler since those are displayed dateless. All other data relies on the finalsStartDate field specified by defaultTerm.

## Test Plan

1. Make a schedule with classes that have different finals, toggle back and forth.
2. Try with only Saturday finals classes (Math 2b) or a mix of each (any class and Math 2b), etc.

## Issues

Closes #758 

## Future Followup
Make an alternate to hardcoding future dates (#778) including for finals, or at least hardcode the next few years.

### DIFF
--- a/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
+++ b/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
@@ -157,12 +157,15 @@ export default function ScheduleCalendar(props: ScheduleCalendarProps) {
 
     const events = getEventsForCalendar();
     const hasWeekendCourse = events.some((event) => event.start.getDay() === 0 || event.start.getDay() === 6);
+
     const calendarStyling = isMobile ? { height: `calc(100% - 55px)` } : { height: `calc(100vh - 104px)` };
+
     const calendarTimeFormat = isMilitaryTime ? 'HH:mm' : 'h:mm A';
     const calendarGutterTimeFormat = isMilitaryTime ? 'HH:mm' : 'h A';
+
     const defaultFinals = getDefaultFinalsStartDate();
     const finalsDateFormat = defaultFinals ? 'ddd MM/DD' : 'ddd';
-    const date = showFinalsSchedule ? defaultFinals : new Date(2018, 0, 1);
+    const date = showFinalsSchedule && defaultFinals ? defaultFinals : new Date(2018, 0, 1);
 
     // If a final is on a Saturday or Sunday, let the calendar start on Saturday
     moment.updateLocale('es-us', {
--- a/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
+++ b/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
@@ -160,6 +160,9 @@ export default function ScheduleCalendar(props: ScheduleCalendarProps) {
     const calendarStyling = isMobile ? { height: `calc(100% - 55px)` } : { height: `calc(100vh - 104px)` };
     const calendarTimeFormat = isMilitaryTime ? 'HH:mm' : 'h:mm A';
     const calendarGutterTimeFormat = isMilitaryTime ? 'HH:mm' : 'h A';
+    const defaultFinals = getDefaultFinalsStartDate();
+    const finalsDateFormat = defaultFinals ? 'ddd MM/DD' : 'ddd';
+    const date = showFinalsSchedule ? defaultFinals : new Date(2018, 0, 1);
 
     // If a final is on a Saturday or Sunday, let the calendar start on Saturday
     moment.updateLocale('es-us', {
@@ -247,7 +250,7 @@ export default function ScheduleCalendar(props: ScheduleCalendarProps) {
                             date.getMinutes() > 0 || !localizer
                                 ? ''
                                 : localizer.format(date, calendarGutterTimeFormat, culture),
-                        dayFormat: `${showFinalsSchedule ? 'ddd MM/DD' : 'ddd'}`,
+                        dayFormat: `${showFinalsSchedule ? finalsDateFormat : 'ddd'}`,
                         eventTimeRangeFormat: (
                             range: { start: Date; end: Date },
                             culture?: string,
@@ -267,7 +270,7 @@ export default function ScheduleCalendar(props: ScheduleCalendarProps) {
                     }}
                     step={15}
                     timeslots={2}
-                    date={showFinalsSchedule ? getDefaultFinalsStartDate() : new Date(2018, 0, 1)}
+                    date={date}
                     min={getStartTime()}
                     max={new Date(2018, 0, 1, 23)}
                     events={events}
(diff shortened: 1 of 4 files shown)


apps/antalmanac/src/stores/calendarizeHelpers.ts
KevinWu098, line 1: Why is this file updated to use the `defaultFinalsStart()` values?

apps/antalmanac/src/lib/termData.ts
```diff
@@ -81,4 +90,19 @@ function getDefaultTerm() {
...
-export { defaultTerm, getDefaultTerm, termData };
+function getDefaultFinalsStart() {
+    return termData[defaultTerm].finalsStartDate;
+}
+
+/**
+ * Defaults to finals for Winter 2024
+ * Days offset by 1 to accomodate toggling with Saturday finals
+ */
+function getDefaultFinalsStartDate() {
```
KevinWu098, line 102: If there's no `finalsStartDate`, I think it would be preferable to return `undefined`, rather that the hardcoded values

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
```diff
@@ -159,6 +160,9 @@ export default function ScheduleCalendar(props: ScheduleCalendarProps) {
     const calendarStyling = isMobile ? { height: `calc(100% - 55px)` } : { height: `calc(100vh - 104px)` };
     const calendarTimeFormat = isMilitaryTime ? 'HH:mm' : 'h:mm A';
     const calendarGutterTimeFormat = isMilitaryTime ? 'HH:mm' : 'h A';
+    const defaultFinals = getDefaultFinalsStartDate();
+    const finalsDateFormat = defaultFinals ? 'ddd MM/DD' : 'ddd';
+    const date = showFinalsSchedule ? defaultFinals : new Date(2018, 0, 1);
```
KevinWu098:
```suggestion
    const date = defaultFinals ?? new Date(2018, 0, 1);
```
If finals aren't being shown, the value of date doesn't matter, so this can be simplified. 

_The code above is also written assuming you choose to return `undefined` from `getDefaultFinalsStartDate()` like described below. The unneeded `showFinalsSchedule` is still true no matter what, unless I've missed something (very possible!)._

apps/antalmanac/src/lib/termData.ts
```diff
@@ -81,4 +90,19 @@ function getDefaultTerm() {
...
+    const defaultFinals = termData[defaultTerm].finalsStartDate;
+    return defaultFinals
+        ? new Date(defaultFinals[0], defaultFinals[1], defaultFinals[2] + 1)
+        : new Date(2024, 2, 16 + 1);
+}
```
KevinWu098:
```suggestion
function getDefaultFinalsStartDate(): Date | undefined {
    const [year, month, day] = termData[defaultTerm].finalsStartDate || [];

    return year && month && day ? new Date(year, month, day + 1) : undefined;
}
```
Additionally, it might be better to refactor so that you're not accessing a "magic" array -- all values should (ideally) tell you what they are

KevinWu098 (changes requested):
Great PR, looking forward to using this feature.

Some minor refactoring suggestions and code questions, but otherwise looks great!

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
KevinWu098:
```suggestion

    const calendarStyling = isMobile ? { height: `calc(100% - 55px)` } : { height: `calc(100vh - 104px)` };
    
    const calendarTimeFormat = isMilitaryTime ? 'HH:mm' : 'h:mm A';
    const calendarGutterTimeFormat = isMilitaryTime ? 'HH:mm' : 'h A';
    
    const defaultFinals = getDefaultFinalsStartDate();
    const finalsDateFormat = defaultFinals ? 'ddd MM/DD' : 'ddd';
    const date = showFinalsSchedule ? defaultFinals : new Date(2018, 0, 1);
```

Honestly, this is a total nitpick (not your fault/not related to this PR), and could be a code smell, but I'd like to break up some of these values so it's more immediately clear what is related to what.

apps/antalmanac/src/stores/calendarizeHelpers.ts
adcockdalton: Apologies if I was not clear enough on point 2; I can explain more on that: 

RBC, as we are implementing it now, populates the calendar with events that are assigned specific days. Prior to this PR, every event, class, final etc. populates on the first week of January 2018 for whatever reason. However, to get the finals to display with the correct actual dates, we have to set the date (by date, I mean the value assigned to the date variable passed into the Calendar object) to the start date of finals (adjusted by 1 to show the correct week regardless of a Saturday final) so that the current viewed week is the correct finals week when toggled rather than the first week of January 2018. That's just how RBC works unfortunately... I added a couple comments to the code for posterity.

So what I did is, if the defaultFinalsStart exists, is assign the finals to the actual week they occur on. Then if showFinalsSchedule is on you will actually see the dates.

This is also why I can't simplify the logic with showFinalsSchedule like you suggested in the other comment: if we do so, then you will not see any classes because the classes only show up in 2018.

lmk if there are any more questions on that, it was annoying figuring it all out haha

adcockdalton: making a push with the other changes you suggested in about 20 minutes once I finish a quiz 👍

adcockdalton: If it helps to demo, observe the default behavior by removing the hardcoded finals date from the winter 2024 term in termdata.ts

KevinWu098: Gotcha, thanks for explaining it to me

KevinWu098 (approved):
LGTM, thanks for working on this! 🚀

There's probably some more discussion to be had about how to handle other terms than the current one, but that's out of scope for this PR (and is likely gated behind Google Auth).

I'll approve but not merge for now, as I want to confirm the final implementation at the sprint this Wednesday

KevinWu098 (changes requested):
Whoops! Sike, not LGTM.

@adcockdalton, you get to now enjoy writing tests. The `calendarize-helpers` test is failing (should be quick fix), but that'll need to get done before we can merge.

KevinWu098 (approved):
LGTM; bit loathe to use a hardcoded test for a changing value but I can live with it (we'll just deal with it later TM)

PR #882, opened 2024-02-06
fix(cdk): only enable deletion protection if the ddb table is retained

ap0nia:
## Summary
If deletion protection is on, then the DDB table can't get deleted after a staging deployment closes.


KevinWu098 (approved):
LGTM, but I have no clue what's going on

ap0nia:
It's very simple! I can explain...

EricPedley (approved):
I kinda get what's going on, and LGTM

PR #888, opened 2024-02-07
feat: login button

ap0nia:
## Summary
Add login button and necessary dependencies. Does not add it to the live website.

## Followup (each will be a PR)
1. Add new backend handlers for auth
2. Replace current login system with new login button


ap0nia:
jk, I'll do the backend portion first.

PR #891, opened 2024-02-07
Auto-Save Feature (Solving Conflict)

stevenguyukai:
# Summary
## This Pull request is the same as #865 but solving the conflict with main

Closes #865


apps/antalmanac/src/components/Header/LoadSaveFunctionality.tsx
```diff
@@ -178,14 +185,27 @@ const LoadSaveScheduleFunctionality = () => {
...
     return (
         <div id="load-save-container">
             <LoadSaveButtonBase
                 id="save-button"
                 actionName={'Save'}
-                action={saveSchedule}
+                action={saveScheduleAndSetLoading}
                 disabled={loading}
-                loading={false}
+                loading={saving}
```
KevinWu098, line 225: Let's set the loading spinner to just the icon, not the whole component

apps/antalmanac/src/components/Header/SettingsMenu.tsx
```diff
@@ -159,6 +164,18 @@ function ExperimentalMenu() {
                 </Box>
                 <Switch color="primary" value={previewMode} checked={previewMode} onChange={handlePreviewChange} />
             </Box>
+
+            <Box display="flex" justifyContent="space-between" width={1}>
+                <Box display="flex" alignItems="center" style={{ gap: 4 }}>
+                    <Typography variant="h6" style={{ display: 'flex', alignItems: 'center', alignContent: 'center' }}>
+                        Auto Save
+                    </Typography>
+                    <Tooltip title={<Typography>Auto Saving your process when you modified the calendar</Typography>}>
```
KevinWu098:
```suggestion
                    <Tooltip title={<Typography>Auto Save your schedule!</Typography>}>
```

apps/antalmanac/src/stores/AppStore.ts
```diff
@@ -305,6 +362,30 @@ class AppStore extends EventEmitter {
 
     termsInSchedule = (term: string) =>
         new Set([term, ...this.schedule.getCurrentCourses().map((course) => course.term)]);
+
+    autoSaveSchedule = async (action: ActionType = [null, null, null, null]) => {
```
ap0nia: Possible strat, but not ideal:
```tsx
    autoSaveSchedule = async (userData: any) => { };
    autoSaveSchedule = async (userData: any, id: any) => { };
    autoSaveSchedule = async (userData: any, id: any, somethingelse: any) => { };
    autoSaveSchedule = async (anyArgs: any[]) => {
        type ActionType = [
            null | string,
            null | number | string | RepeatingCustomEvent | ScheduleCourse,
            null | number | string | number[],
            null | number | string | RepeatingCustomEvent
        ];
    };
```

apps/antalmanac/src/stores/AppStore.ts
```diff
@@ -6,6 +6,14 @@ import { Schedules } from './Schedules';
...
 import { CalendarEvent, CourseEvent } from '$components/Calendar/CourseCalendarEvent';
 import { useTabStore } from '$stores/TabStore';
+import { saveSchedule } from '$actions/AppStoreActions';
+
+type ActionType = [
+    null | string,
+    null | number | string | RepeatingCustomEvent | ScheduleCourse,
+    null | number | string | number[],
+    null | number | string | RepeatingCustomEvent,
+];
```
KevinWu098: As we discussed in the sprint, it's best to handle this separately and be more explicit

apps/antalmanac/src/stores/AppStore.ts
```diff
@@ -229,6 +245,45 @@ class AppStore extends EventEmitter {
             return false;
         }
         this.unsavedChanges = false;
+
+        const unsavedActionString = window.localStorage.getItem('unsavedAction');
+        if (unsavedActionString !== null && confirm('You have unsaved changes. Would you like to load them?')) {
+            for (const action of JSON.parse(unsavedActionString)) {
```
KevinWu098: This is an `any`. Handle `ActionType` differently, as discussed

KevinWu098 (changes requested):

apps/antalmanac/src/stores/AppStore.ts
```diff
@@ -305,6 +362,30 @@ class AppStore extends EventEmitter {
...
+            if (savedUserID) {
+                this.emit('autoSaveStart');
+                await saveSchedule(savedUserID, true, true);
+                this.unsavedChanges = false;
+                this.emit('autoSaveEnd');
+            }
+        } else {
+            const unsavedActionString = window.localStorage.getItem('unsavedAction');
+            if (unsavedActionString == null) {
+                const unsavedAction = [action];
```
KevinWu098, line 380: Can we rename this to `unsavedActions` (plural)

apps/antalmanac/src/stores/AppStore.ts
```diff
@@ -305,6 +362,30 @@ class AppStore extends EventEmitter {
...
+
+            if (savedUserID) {
+                this.emit('autoSaveStart');
+                await saveSchedule(savedUserID, true, true);
+                this.unsavedChanges = false;
+                this.emit('autoSaveEnd');
+            }
+        } else {
+            const unsavedActionString = window.localStorage.getItem('unsavedAction');
+            if (unsavedActionString == null) {
```
KevinWu098: Can we rename this to `unsavedActions` (plural)

KevinWu098 (changes requested):

stevenguyukai:
For only spinning the icon when saving and loading, it was not supported by the library. 

So I made a custom `SaveLoadIcon`, which will spin when `loading` is `True`. I set the default `LoadingButton` component's `loading` to `False` so the text will not spin.

> Just to document the change

stevenguyukai:
There is a new file for the action type: `ActionTypes.ts`.

KevinWu098 (approved):
Great work on a fantastic feature, I think users will be really happy with this.

Code and feature looks really solid to me, so I'm going to approve it, but I'd like to discuss two (non-blocking) points at next Wednesday's meeting before merging.

1. Are all these additional DB writes a potential concern? (My guess is no, but I'll CC @EricPedley @ap0nia on this one)
2. The unsaved courses is great. However, when I load them in, and go through the `confirm` flow, it will clear: `window.localStorage.removeItem('unsavedActions');`, regardless of whether or not I ended up saving it.

KevinWu098 (approved):
LGTM Banger feature, merge when you get a chance 🚀🚀

apps/antalmanac/src/components/Header/LoadSaveFunctionality.tsx
```diff
@@ -32,6 +33,17 @@ interface LoadSaveButtonBaseState {
...
+    render() {
+        const { loading, actionName } = this.props;
+        if (loading) {
+            return <CircularProgress size={20} color={'inherit'} />;
+        } else {
+            return actionName === 'Save' ? <Save /> : <CloudDownload />;
+        }
+    }
+}
+
```
ap0nia: Is there a reason you're using a legacy class-based component instead of a functional one? Prefer functional components, and leave a comment documenting why it needs to be a class component if that's the case (i.e. maybe the button base is a class component that's incompatible with function components as its start icon?)

apps/antalmanac/src/components/Header/LoadSaveFunctionality.tsx
```diff
@@ -32,6 +33,17 @@ interface LoadSaveButtonBaseState {
     rememberMe: boolean;
 }
 
+class SaveLoadIcon extends PureComponent<{ loading: boolean; actionName: 'Save' | 'Load' }> {
+    render() {
+        const { loading, actionName } = this.props;
+        if (loading) {
+            return <CircularProgress size={20} color={'inherit'} />;
```
ap0nia: If it's a string prop, you don't need extra braces around it.

```suggestion
            return <CircularProgress size={20} color='inherit' />;
```

apps/antalmanac/src/components/Header/LoadSaveFunctionality.tsx
```diff
@@ -6,6 +6,7 @@ import {
     DialogContentText,
     DialogTitle,
     TextField,
+    CircularProgress,
 } from '@material-ui/core';
 import Checkbox from '@material-ui/core/Checkbox';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
```
ap0nia: Could you move the last two imports from `@material-ui/core` to the destructured import statement? You didn't change any code there, but it's just weird that they're not already included. i.e.

```tsx
import {
  // ...previous imports
  Checkbox,
  FormControlLabel
} from '@material-ui/core'
```

apps/antalmanac/src/stores/AppStore.ts
ap0nia, line 1: My first impression of this is that you shouldn't be modifying the AppStore at all. You wrote a completely new store from scratch, so you can synchronize to that store instead of this legacy one.

apps/antalmanac/src/components/Header/LoadSaveFunctionality.tsx
```diff
@@ -178,14 +197,27 @@ const LoadSaveScheduleFunctionality = () => {
...
+        const handleAutoSaveStart = () => setSaving(true);
+        const handleAutoSaveEnd = () => setSaving(false);
+
+        AppStore.on('autoSaveStart', handleAutoSaveStart);
+        AppStore.on('autoSaveEnd', handleAutoSaveEnd);
+
+        return () => {
+            AppStore.off('autoSaveStart', handleAutoSaveStart);
+            AppStore.off('autoSaveEnd', handleAutoSaveEnd);
+        };
```
ap0nia: I think you should remove dependency on the legacy app store.

ap0nia (changes requested):
Most the implementation is sensible and seems to be architected properly 👍 .
I really dislike having the dependency on the legacy appstore since there's a new store dedicated to managing state for the autosave feature. I don't have concrete implementation examples atm, but I really think it would be best to relocate all of the additional functionality from the appstore to the autosave store.

apps/antalmanac/src/components/Header/LoadSaveFunctionality.tsx
stevenguyukai: no, there is no specific reason. I just want to maintain consistency like how `LoadSaveButtonBase` class did that already in the file. I can change it to the functional components if that's preferred in React.

apps/antalmanac/src/actions/ActionTypesStore.ts
```diff
@@ -0,0 +1,151 @@
+import { EventEmitter } from 'events';
+import { RepeatingCustomEvent, ScheduleCourse } from '@packages/antalmanac-types';
+import { saveSchedule } from '$actions/AppStoreActions';
+import AppStore from '$stores/AppStore';
+
+export interface BaseAction {
+    type: string;
+}
```
ap0nia:
```suggestion
```

There's no reason for the base action. The type discrimination will work without it, and the reason this type of interface would exist is to ensure classes adhere to a basic set of functionality via the `implements` (TypeScript) directive. For example

```tsx
interface BaseClass {
  doSomething: () => any
}

class MyExtendedClass implements BaseClass {
  // TS will display an error if you don't have a `doSomething` method. 
}
```

apps/antalmanac/src/actions/ActionTypesStore.ts
```diff
@@ -0,0 +1,151 @@
...
+            }
+        }
+    }
+
+    async loadScheduleFromLocalSave() {
+        const unsavedActionsString = window.localStorage.getItem('unsavedActions');
+        if (unsavedActionsString !== null) {
+            if (confirm('You have unsaved changes. Would you like to load them?')) {
+                for (const action of JSON.parse(unsavedActionsString)) {
+                    switch (action.type) {
```
ap0nia: Inverse the conditional logic to utilize guard statements and reduce nesting. Also, use a type-safe function for accessing the raw string data, so that way you can easily factor in schema validation later if that's desired.

```tsx

// Somewhere else in the file
/**
  * (Somewhere else in the file)
  * This isn't type-safe within the function, 
  * but the parent scope using the function will have defined types.
  */
function parseUnsavedActionsString(unsavedActionsString: string): ActionType[] {
  try {
    return JSON.parse(unsavedActionsString)
  } catch {
    return []
  }
}

if (unsavedActionsString == null) return

if (!confirm('You have unsaved changes...')) return

const actions = parseActionsString(unsavedActionsString)

actions.forEach(action => {
  switch (action.type) {
    // ...
  }
})
```

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -86,10 +86,12 @@ export const saveSchedule = async (userID: string, rememberMe: boolean) => {
...
-                openSnackbar(
-                    'success',
-                    `Schedule saved under username "${userID}". Don't forget to sign up for classes on WebReg!`
-                );
+                if (!autoSave) {
+                    openSnackbar(
+                        'success',
+                        `Schedule saved under username "${userID}". Don't forget to sign up for classes on WebReg!`
+                    );
+                }
```
ap0nia: This is an anti-pattern. Passing in booleans to mutate a function's behavior is like a half-baked builder pattern implementation. If you have specific functionality that you want to perform in a certain context, then take care of that in the same context instead of delegating it to another module that doesn't inherently understand it.

I'd be willing to go as far as just creating an entirely new function for auto saving, even if it means duplicating some code, just to keep these interfaces well defined. For example

```ts
export async function autosaveSchedule(userID: string, rememberMe: boolean) {
  // ...
}
```

apps/antalmanac/src/components/Header/LoadSaveFunctionality.tsx
```diff
@@ -32,6 +34,21 @@ interface LoadSaveButtonBaseState {
     rememberMe: boolean;
 }
 
+interface SaveLoadIconProps {
+    loading: boolean;
+    actionName: 'Save' | 'Load';
+}
+
+function SaveLoadIcon(SaveLoadIconProps: SaveLoadIconProps) {
```
ap0nia: Generally, prefer to name props literally as `props`. It's easy to understand, and there shouldn't be naming conflicts since functions are distinct entities with different namespaces. The interfaces for the props can be unique to make it clear what component the props are for.

```suggestion
function SaveLoadIcon(props: SaveLoadIconProps) {
```

apps/antalmanac/src/components/Header/SettingsMenu.tsx
```diff
@@ -141,11 +144,26 @@ function TimeMenu() {
...
+        if (event.target.checked) {
+            const savedUserID = window.localStorage.getItem('userID');
+
+            if (savedUserID) {
+                actionTypesStore.emit('autoSaveStart');
+                await saveSchedule(savedUserID, true, true);
+                appStore.unsavedChanges = false;
+                actionTypesStore.emit('autoSaveEnd');
+            }
+        }
```
ap0nia: Use guard clauses to reduce nesting.
```suggestion
        if (!event.target.checked) return
        
        const savedUserId = window.localStorage.getItem('userID');
        
        if (!savedUserId) return

        actionTypesStore.emit('autoSaveStart');
        await saveSchedule(savedUserID, true, true);
        appStore.unsavedChanges = false;
        actionTypesStore.emit('autoSaveEnd');
```

apps/antalmanac/src/components/Header/SettingsMenu.tsx
```diff
@@ -159,6 +177,18 @@ function ExperimentalMenu() {
                 </Box>
                 <Switch color="primary" value={previewMode} checked={previewMode} onChange={handlePreviewChange} />
             </Box>
+
+            <Box display="flex" justifyContent="space-between" width={1}>
+                <Box display="flex" alignItems="center" style={{ gap: 4 }}>
```
ap0nia: Boxes can be styled either with the `style` prop, `sx` prop (MUIv5), or via the props directly. Generally, I prefer the former two (`style` and `sx`), since it's a consistent strategy for all components. For boxes, prefer not mixing them. i.e. You can use either, but just not at the same time.

```suggestion
                <Box style={{ display: 'flex', alignItems: 'center', gap: 4 }}>
```
Here, since gap was captured in a `style` prop, just move all the other styles to the `style` prop as well. Or vice versa if that's possible and preferrable.

apps/antalmanac/src/stores/AppStore.ts
```diff
@@ -6,6 +6,19 @@ import { Schedules } from './Schedules';
 import { SnackbarPosition } from '$components/NotificationSnackbar';
 import { CalendarEvent, CourseEvent } from '$components/Calendar/CourseCalendarEvent';
 import { useTabStore } from '$stores/TabStore';
+import actionTypesStore from '$actions/ActionTypesStore';
+import {
```
ap0nia: This isn't enforced by the tsconfig, but prefer to indicate when things are being imported specifically for their type information, vs. as actual JS entities.

```suggestion
import type {
```

apps/antalmanac/src/actions/ActionTypesStore.ts
```diff
@@ -0,0 +1,151 @@
...
+                this.emit('autoSaveEnd');
+            }
+        } else {
+            const unsavedActionsString = window.localStorage.getItem('unsavedActions');
+            if (unsavedActionsString == null) {
+                const unsavedActions = [action];
+                localStorage.setItem('unsavedActions', JSON.stringify(unsavedActions));
+            } else {
+                const unsavedActions = JSON.parse(unsavedActionsString);
+                unsavedActions.push(action);
```
ap0nia: You can also use a type-safe parsing function here, like I mentioned in a different comment. Also, could you add an upper limit to the length of actions? It can be arbitrary for now, but I think it should conceptually exist. For example, a max of 1000 unsaved actions can be set.

```ts
// At the top of the module.
const MAX_UNSAVED_ACTIONS = 1_000

const unsavedActions = parseUnsavedActionsString(unsavedActionsString)
if (unsavedActions.length > MAX_UNSAVED_ACTIONS) {
  // Do something, e.g. slice of the earliest 100?
}

// If max was reached, and resolved by removing some items, 
// then you could proceed as normal
unsavedActions.push(action)
```

ap0nia (changes requested):
Just some final touch ups before the final merge, keep up the good work 🚀 

Functionality seems to work.
- When autosave is off, all unsaved changes get pushed to local storage.
- When refreshing the page with autosave off and unsaved changes in local storage, I get an alert asking whether to load them.
- When autosave is on, no unsaved changes get pushed to local storage.
- When autosave is on, all actions get saved.

apps/antalmanac/src/actions/AppStoreActions.ts
stevenguyukai: I added a specific function `autosaveSchedule` for auto save, and I remove the `rememberMe` in the function because autosave will not change the state of current user id in local storage.

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -102,6 +102,33 @@ export const saveSchedule = async (userID: string, rememberMe: boolean) => {
...
+    logAnalytics({
+        category: analyticsEnum.nav.title,
+        action: analyticsEnum.nav.actions.SAVE_SCHEDULE,
+        label: userID,
+    });
+    if (userID != null) {
+        userID = userID.replace(/\s+/g, '');
+
+        if (userID.length > 0) {
+            const scheduleSaveState = AppStore.schedule.getScheduleAsSaveState();
```
ap0nia: You can apply guard clauses here to reduce nesting.

```suggestion
    if (userID == null) return
    userID = userID.replace(/\s+/g, '');

    if (userID.length < 0) return
    const scheduleSaveState = AppStore.schedule.getScheduleAsSaveState();
```

ap0nia (approved):
Looks pretty good, thanks for the work 🚀 

Feel free to merge it whenever you're ready. If you make any minor adjustments based on my latest feedback, just re-request review and I'll merge it.

stevenguyukai:
just a quick add of the guard condition, you can just merge it if there is no problem

ap0nia (approved):
👍

PR #901, opened 2024-02-11
fix: remove hovered course on unmount

### LINKED ISSUE #900: Hovered course previews remain if you escape out of search results
If you hover over a course with hover to preview on and press escape to go back from the current search results, the course preview will remain in the calendar.

https://github.com/icssc/AntAlmanac/assets/8922227/1fc123e8-386d-4ec6-9518-7733c96968de

KevinWu098:
## Summary
1. Set hovered course to `undefined` whenever `SectionTableWrapped` unmounts

chrome-capture-2024-1-10

## Test Plan
1. Attempt to replicate #900, by hovering a course (it should appear in the calendar), then hitting `esc` (or any other action that removes the course results). The hovered course should disappear.

## Issues
Closes #900

### DIFF
--- a/apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
+++ b/apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
@@ -259,7 +259,11 @@ export default function CourseRenderPane(props: { id?: number }) {
         };
     }, [loadCourses, props.id]);
 
-    /* Removes hovered course when component unmounts */
+    /**
+     * Removes hovered course when component unmounts
+     * Handles edge cases where the Section Table is removed, rather than the mouse
+     * ex: Swapping to the Added tab, clicking the LocationCell link
+     */
     useEffect(() => {
         return () => {
             setHoveredCourseEvents(undefined);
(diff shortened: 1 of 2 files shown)


js0mmer (commented):
Looks good, although just for fun, I found another way to break it using tab/shift + tab and pressing enter on the back button :)

https://github.com/icssc/AntAlmanac/assets/8922227/b937fa40-5457-4174-9b4c-5e28df6ccff9

js0mmer:
and another (clicking the room)

https://github.com/icssc/AntAlmanac/assets/8922227/65f0a03f-cb40-4fdc-a072-51cf10278f21

js0mmer:
Should also mention that bypassing via tabbing is not just limited to the back button but also added and map tabs.

https://github.com/icssc/AntAlmanac/assets/8922227/24e923bc-5ec5-4080-b5b2-0b8b21295d4a

js0mmer (commented):
Nice, that seems to be every edge case handled 😂

I'm not super familiar with the codebase, but is this the highest place that this useEffect can be placed? Just want to make sure that it's not called more times than necessary.

Also I personally would update the comment to include a little bit of information why the useEffect is needed. It explains what it does which I think is clear enough to a developer but a future developer may not understand why it's necessary. Should mention the possibility of switching away with esc, clicking room, etc. or reference back to this PR.

js0mmer (approved):
LGTM!

PR #904, opened 2024-02-12
fix: reduce theme swap time

### LINKED ISSUE #694: Swapping Themes take very Long
When you have a lot of stuff loaded in on the page, changing the theme takes very very long. I'm not sure if that's something that we can fix, but what I do think can be done is some sort of loading indicator while the theme is preparing to swap over.

chrome-capture-2023-8-11

With all of ICS loaded, it took ~7 seconds to change themes.

KevinWu098:
## Summary
1. Refactored `forceUpdate`, the mechanism which reloaded the search results, into `CoursePaneStore`
2. Immediately prior to theme swaps, we hook onto `forceUpdate`, derendering the components while maintaining the current search. 

Massive, massive "performance" boost. No more crashing sites. Much credit to @js0mmer for the idea.

chrome-capture-2024-1-12

## Test Plan
1. Scroll to the bottom of COMPSCI and test out theme swaps!

## Issues
Closes #694


EricPedley (approved):
lgtm, nj

Issue #890, opened 2024-02-07
RFC: Login system


ap0nia:
Central login location -> either login with username or with Google
Once a user is logged in, the website is synchronized and identifies them to one account

If user is logged in with Google, a "import from legacy account" button will exist.
They hit the button to open up a dialogue, enter their legacy username, and submitting copies that schedule from legacy schedules.
Google authenticated users can make their schedule private, public, or open
* private: only the owner can view
* public readonly: other users can view, but can't edit
* open: anybody can edit

table
columns
* user ID / username
* google ID
* schedule
* visibility

Given a google ID:
find the corresponding google ID and load the schedule

Given a username:
* find the user name
* if they have a visiblity field, then verify that before loading the schedule

Legacy username and custom username mapped to Google ID supported

Login UI
* Login button
* opens up a modal with two options
* login with username (i.e. either legacy or Google)
* login with Google OAuth (redirect link)

After user logs in
They can remap a custom, unique username to their Google ID
- If this unique username is taken, action fails
- If it is available, succeed

Import Button
* import schedule from username

Viewing schedules
* If user views different account, don't automatically save if not same credentials as saved

Userflow: TODO

RFC: Only put account related buttons in the header, and put schedule related buttons, e.g. exporting, somewhere lower in the app

PR #906, opened 2024-02-14
feat: initial auth backend

ap0nia:
# Features (based on #890)

## Correct columns in schema
- id  (required) -> user ID / username
- googleId (optional) -> google ID
- userData (required) -> schedule
- visibility (optional) -> visibility

The `id` column is required and the primary key for identification.
The `googleId` is optional, and populated once the user has logged in with Google OAuth.
Users logging in via Google will be checked using the `googleId` column instead of `id` column.

### Loading user data
No visibility checks, just loads the schedule as asked. This may be adjusted to be more like the "viewing" functionality, which isn't currently active on the frontend.

### Viewing user data
Assumes a user ID requesting another user ID flow.
Checks that the requested schedule is visible.


packages/types/src/user.ts
```diff
@@ -6,4 +6,13 @@ export const UserSchema = type({
     userData: ScheduleSaveStateSchema,
 });
 
+export const AuthUserSchema = type({
+    id: 'string',
+    'userData?': ScheduleSaveStateSchema,
+    name: 'string',
```
EricPedley: Is this `name` their name on google or the name of their userData object? Can we document that somewhere?

packages/types/src/user.ts
```diff
@@ -6,4 +6,13 @@ export const UserSchema = type({
...
+    email: 'string',
+    picture: 'string',
```
EricPedley: Thoughts on calling this `pictureURL` instead of just `picture`?

EricPedley (dismissed):
Are we still going to go with the idea of sharing the userid namespace between both schedules that are owned by an auth user, and schedules that aren't owned? I thought we were just going to use one table for this. IK that currently the cdk props up two tables in the backend, but that was before we had the discussion about login at the meeting, so I thought maybe we were going to change it. We can discuss this point at this week's meeting if you want.

ap0nia:
Oops, yeah I forgot about the discussion. Re: #890

packages/types/src/user.ts
ap0nia: When using the combined schema, the user's selected name should be under "id", so this field shouldn't be necessary.

packages/types/src/user.ts
ap0nia: What about `image`? Maybe same problem, but I generally like simple names even if they're a bit ambiguous, just because there's not many different things it could be IMO.

EricPedley: I guess simple could be fine here cause it's already defined as a string and normally you never see raw image data being passed around in web dev. We can just leave this as-is.

MinhxNguyen7:
Seems about right ¯\_(ツ)_/¯

MinhxNguyen7 (dismissed):

MinhxNguyen7:
I'm not entirely sure what's going on, and I don't know what the risks are to merging this, so you can merge it in when you feel fit @ap0nia

ap0nia (commented):
# Manual Testing Results

## Saving
1. Select `2023 Spring` term.
2. Filter for `CS` department.
3. Add the first course (CS 111, code: `34010`).
4. Click `Save` button in header.
5. Save under `abc` username.
6. Request is completed successfully.

## Loading
1. Click `Load` button in header.
2. Type username from `Saving` procedure for username (i.e. `abc`) and submit.
3. Schedule from `Saving` procedure should be loaded successfully.

# Conclusion
LGTM

PR #915, opened 2024-02-19
fix: cleanup calendar warning

KevinWu098:
## Summary
1. One line diff adding an `onNavigate` prop, removes console warning ⬇️ 

Screenshot 2024-02-19 at 1 59 48 PM

## Test Plan
1. Check if any unexpected behavior arises from adding the `onNavigate` prop.


stevenguyukai (approved):
I was not able to reproduce the console error message from https://antalmanac.com. 

But it looks like it has no problem, since we are not changing the month or date of the calendar, the `onNavigate` should not be triggered anyway.

PR #917, opened 2024-02-21
feat: create local storage abstraction

### LINKED ISSUE #889: Local storage abstraction
Instead of pulling up `localstorage` with a magic value every time, which is annoying and error-prone, we should abstract everything behind a layer, either in one class or in a store for each group of local storage values.

KevinWu098:
## Summary
1. Create `localStorage` abstractions for getting, setting, and removing. Not all getters/setters/removers are utilized, but I've chosen to keep/include them for completeness-sake.

## Test Plan
1. Confirm that every abstraction works as intended
2. Or, assuming that the abstractions are all written correctly, you can read over the PR and confirm that they were abstracted one-to-one

## Issues
Closes #889


MinhxNguyen7 (approved):
Seems about right. I tested a few things out, and they didn't break. 

The real UAT happens in prod amiright?

PR #919, opened 2024-02-21
Extending the Hover Feature for Finals

adcockdalton:
## Summary

Since we've been diversifying the information available about finals, I decided to generalize the functionality of the hover feature.
- Modified HoveredStore.ts to have separate data fields for calendarized finals, calendarized courses, and raw ScheduleCourse data
- Correspondingly updated the variable names in courseRenderPane and sectionTableBody to decentralize it from "courses" to events in general
- Cleaned up the logic in CalendarRoot for incorporating the hovered courses, and added finals to the logic
- Changed the date view to override any selected finals week to the week of the hovered final

I also made an extremely small one line refactor in TermData for consistency's sake since it is related to this PR and was discussed.

## Test Plan
- Attempt hover on courses as well as finals for an arbitrary course on an empty schedule
- Attempt hover on both after adding a course
- Attempt hover on both after adding a course from a different, non-default quarter
- Attempt hover on both after adding a course from a different but default quarter

## Future Followup
Is there a reason hover to preview is still an experimental toggle feature?

### DIFF
--- a/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
+++ b/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
@@ -80,14 +80,15 @@ export default function ScheduleCalendar(props: ScheduleCalendarProps) {
     const [scheduleNames, setScheduleNames] = useState(AppStore.getScheduleNames());
 
     const { isMilitaryTime } = useTimeFormatStore();
-    const { hoveredCourseEvents } = useHoveredStore();
+    const { hoveredCalendarizedCourses, hoveredCalendarizedFinal } = useHoveredStore();
 
     const getEventsForCalendar = (): CalendarEvent[] => {
-        return showFinalsSchedule
-            ? finalsEventsInCalendar
-            : hoveredCourseEvents
-              ? [...eventsInCalendar, ...hoveredCourseEvents]
-              : eventsInCalendar;
+        if (showFinalsSchedule)
+            return hoveredCalendarizedFinal
+                ? [...finalsEventsInCalendar, hoveredCalendarizedFinal]
+                : finalsEventsInCalendar;
+        else
+            return hoveredCalendarizedCourses ? [...eventsInCalendar, ...hoveredCalendarizedCourses] : eventsInCalendar;
     };
 
     const handleClosePopover = () => {
@@ -165,8 +166,11 @@ export default function ScheduleCalendar(props: ScheduleCalendarProps) {
 
     const onlyCourseEvents = eventsInCalendar.filter((e) => !e.isCustomEvent) as CourseEvent[];
 
-    const finalsDate =
-        onlyCourseEvents.length > 0 ? getFinalsStartDateForTerm(onlyCourseEvents[0].term) : getDefaultFinalsStartDate();
+    const finalsDate = hoveredCalendarizedFinal
+        ? getFinalsStartDateForTerm(hoveredCalendarizedFinal.term)
+        : onlyCourseEvents.length > 0
+          ? getFinalsStartDateForTerm(onlyCourseEvents[0].term)
+          : getDefaultFinalsStartDate();
 
     const finalsDateFormat = finalsDate ? 'ddd MM/DD' : 'ddd';
     const date = showFinalsSchedule && finalsDate ? finalsDate : new Date(2018, 0, 1);
--- a/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
+++ b/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
@@ -80,7 +80,10 @@ export default function ScheduleCalendar(props: ScheduleCalendarProps) {
     const [scheduleNames, setScheduleNames] = useState(AppStore.getScheduleNames());
 
     const { isMilitaryTime } = useTimeFormatStore();
-    const { hoveredCalendarizedCourses, hoveredCalendarizedFinal } = useHoveredStore();
+    const [hoveredCalendarizedCourses, hoveredCalendarizedFinal] = useHoveredStore((store) => [
+        store.hoveredCalendarizedCourses,
+        store.hoveredCalendarizedFinal,
+    ]);
 
     const getEventsForCalendar = (): CalendarEvent[] => {
         if (showFinalsSchedule)
--- a/apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
+++ b/apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
@@ -268,7 +268,7 @@ export default function CourseRenderPane(props: { id?: number }) {
         return () => {
             setHoveredEvents(undefined);
         };
-    }, [setHoveredEvents]);
+    }, []);
 
     return (
         <>
--- a/apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
+++ b/apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
@@ -178,7 +178,7 @@ export default function CourseRenderPane(props: { id?: number }) {
     const [error, setError] = useState(false);
     const [scheduleNames, setScheduleNames] = useState(AppStore.getScheduleNames());
 
-    const setHoveredCourseEvents = useHoveredStore((store) => store.setHoveredCourseEvents);
+    const setHoveredEvents = useHoveredStore((store) => store.setHoveredEvents);
 
     const loadCourses = useCallback(async () => {
         setLoading(true);
@@ -266,9 +266,9 @@ export default function CourseRenderPane(props: { id?: number }) {
      */
     useEffect(() => {
         return () => {
-            setHoveredCourseEvents(undefined);
+            setHoveredEvents(undefined);
         };
-    }, [setHoveredCourseEvents]);
+    }, [setHoveredEvents]);
 
     return (
         <>
--- a/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
+++ b/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
@@ -533,15 +533,17 @@ const SectionTableBody = withStyles(styles)((props: SectionTableBodyProps) => {
 
     const [hoveredEvents, setHoveredEvents] = useHoveredStore((store) => [store.hoveredEvents, store.setHoveredEvents]);
 
-    const handleHover = useCallback(() => {
-        const alreadyHovered =
-            hoveredEvents &&
-            hoveredEvents.some((scheduleCourse) => scheduleCourse.section.sectionCode == section.sectionCode);
+    const alreadyHovered = useMemo(() => {
+        return hoveredEvents?.some((scheduleCourse) => scheduleCourse.section.sectionCode == section.sectionCode);
+    }, [hoveredEvents, section.sectionCode]);
 
-        !previewMode || alreadyHovered || addedCourse
-            ? setHoveredEvents(undefined)
-            : setHoveredEvents(section, courseDetails, term);
-    }, [addedCourse, courseDetails, hoveredEvents, previewMode, section, setHoveredEvents, term]);
+    const handleHover = useCallback(() => {
+        if (!previewMode || alreadyHovered || addedCourse) {
+            setHoveredEvents(undefined);
+        } else {
+            setHoveredEvents(section, courseDetails, term);
+        }
+    }, [alreadyHovered, section, courseDetails, term]);
 
     // Attach event listeners to the store.
     useEffect(() => {
--- a/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
+++ b/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
@@ -531,20 +531,17 @@ const SectionTableBody = withStyles(styles)((props: SectionTableBodyProps) => {
         setCalendarEvents(AppStore.getCourseEventsInCalendar());
     }, [setCalendarEvents]);
 
-    const [hoveredCourseEvents, setHoveredCourseEvents] = useHoveredStore((store) => [
-        store.hoveredCourseEvents,
-        store.setHoveredCourseEvents,
-    ]);
+    const [hoveredEvents, setHoveredEvents] = useHoveredStore((store) => [store.hoveredEvents, store.setHoveredEvents]);
 
     const handleHover = useCallback(() => {
         const alreadyHovered =
-            hoveredCourseEvents &&
-            hoveredCourseEvents.some((courseEvent) => courseEvent.sectionCode == section.sectionCode);
+            hoveredEvents &&
+            hoveredEvents.some((scheduleCourse) => scheduleCourse.section.sectionCode == section.sectionCode);
 
         !previewMode || alreadyHovered || addedCourse
-            ? setHoveredCourseEvents(undefined)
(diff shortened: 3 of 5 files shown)


apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
```diff
@@ -531,20 +531,17 @@ const SectionTableBody = withStyles(styles)((props: SectionTableBodyProps) => {
...
-            hoveredCourseEvents.some((courseEvent) => courseEvent.sectionCode == section.sectionCode);
+            hoveredEvents &&
+            hoveredEvents.some((scheduleCourse) => scheduleCourse.section.sectionCode == section.sectionCode);
 
         !previewMode || alreadyHovered || addedCourse
-            ? setHoveredCourseEvents(undefined)
-            : setHoveredCourseEvents(section, courseDetails, term);
-    }, [addedCourse, courseDetails, hoveredCourseEvents, previewMode, section, setHoveredCourseEvents, term]);
+            ? setHoveredEvents(undefined)
+            : setHoveredEvents(section, courseDetails, term);
```
ap0nia: The value returned by the ternary isn't being used for anything, so it's misleading. Prefer if/else blocks for representing alternative code flows.
```tsx
if (!previewMode || alreadyHovered || addedCourse) {
  setHoveredEvents(undefined)
} else {
  setHoveredEvents(section, courseDetails, term)
}
```

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
```diff
@@ -531,20 +531,17 @@ const SectionTableBody = withStyles(styles)((props: SectionTableBodyProps) => {
...
-        store.setHoveredCourseEvents,
-    ]);
+    const [hoveredEvents, setHoveredEvents] = useHoveredStore((store) => [store.hoveredEvents, store.setHoveredEvents]);
 
     const handleHover = useCallback(() => {
         const alreadyHovered =
-            hoveredCourseEvents &&
-            hoveredCourseEvents.some((courseEvent) => courseEvent.sectionCode == section.sectionCode);
+            hoveredEvents &&
+            hoveredEvents.some((scheduleCourse) => scheduleCourse.section.sectionCode == section.sectionCode);
```
ap0nia: You can extract this to a separate value, which will make the memoization of this callback simpler.
```tsx
const alreadyHovered = useMemo(() => {
 return hoveredEvents?.some((scheduleCourse) => scheduleCourse.section.sectionCode == section.sectionCode);
}, [hoveredEvents, section.sectionCode])
```

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
```diff
@@ -531,20 +531,17 @@ const SectionTableBody = withStyles(styles)((props: SectionTableBodyProps) => {
...
 
         !previewMode || alreadyHovered || addedCourse
-            ? setHoveredCourseEvents(undefined)
-            : setHoveredCourseEvents(section, courseDetails, term);
-    }, [addedCourse, courseDetails, hoveredCourseEvents, previewMode, section, setHoveredCourseEvents, term]);
+            ? setHoveredEvents(undefined)
+            : setHoveredEvents(section, courseDetails, term);
+    }, [addedCourse, courseDetails, hoveredEvents, previewMode, section, setHoveredEvents, term]);
```
ap0nia: Extension of previous suggestion for memoization; simplify the dependency array.

```suggestion
    }, [alreadyHovered, section, courseDetails, term]);
```

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
```diff
@@ -266,9 +266,9 @@ export default function CourseRenderPane(props: { id?: number }) {
      */
     useEffect(() => {
         return () => {
-            setHoveredCourseEvents(undefined);
+            setHoveredEvents(undefined);
         };
-    }, [setHoveredCourseEvents]);
+    }, [setHoveredEvents]);
```
ap0nia: This shouldn't be in the dependency array since it's a setter, but ESLint might complain about that, so I guess it's alright to leave it if that's the case unfortunately.

```suggestion
    }, []);
```

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
```diff
@@ -80,14 +80,15 @@ export default function ScheduleCalendar(props: ScheduleCalendarProps) {
     const [scheduleNames, setScheduleNames] = useState(AppStore.getScheduleNames());
 
     const { isMilitaryTime } = useTimeFormatStore();
-    const { hoveredCourseEvents } = useHoveredStore();
+    const { hoveredCalendarizedCourses, hoveredCalendarizedFinal } = useHoveredStore();
```
ap0nia: Slice the store to reduce the number of dependencies this component has (i.e. how often it needs to re-render due to variables in the store updating).

There are different ways to slice it. Here's an example of returning only the desired properties in an array, and destructuring it.
 
```suggestion
    const [ hoveredCalendarizedCourses, hoveredCalendarizedFinal ] = useHoveredStore(store => [store.hoveredCalendarizedCourses, store.hoveredCalendarizedFinal]);
```

apps/antalmanac/src/stores/HoveredStore.ts
```diff
@@ -1,20 +1,37 @@
...
+                              {
+                                  ...courseDetails,
+                                  section: {
+                                      ...section,
+                                      color: '#80808080',
+                                  },
+                                  term,
+                              },
+                          ]
+                        : undefined,
```
ap0nia: Could you refactor the conditional to make the code flow simpler?

```tsx

const DEFAULT_HOVERED_STORE = {
  hoveredEvents: undefined,
  hoveredCalendarizedCourses: undefined,
  hoveredCalendarizedFinal: undefined,
}

// spread operator to use the default store.
...DEFAULT_HOVERED_STORE,
setHoveredEvents: (section, courseDetails, term) => {
  if (section == null || courseDetails == null || term == null) {
    set({ ...DEFAULT_HOVER_STORE })
    return
  }

  // Get the correct data and set appropriately.
}
```

The main points are:
* Use a guard clause to exit early, which makes sense when the logic for the guard clause is comparatively simple to the logic outside of it.
* Create an object that represents an "unhovered" state, and use the spread operator to copy its contents whenever you want to initialize a new "unhovered" state.
* Whenever you create any object representing a default state, ensure that you deep-copy it correctly to prevent accidentally mutating the default state. If it's a one-layer object, spread operator is fine; if it's a nested object, use `structuredClone`
* You can use the default "unhovered" state both when initializing the store, and when setting it back to this state.

apps/antalmanac/src/stores/HoveredStore.ts
```diff
@@ -27,6 +44,19 @@ export const useHoveredStore = create<HoveredStore>((set) => {
...
                           ])
                         : undefined,
+                hoveredCalendarizedFinal:
+                    section && courseDetails && term
+                        ? calendarizeFinals([
+                              {
+                                  ...courseDetails,
+                                  section: {
+                                      ...section,
+                                      color: '#80808080',
```
ap0nia: I think this color could be stored somewhere descriptive, maybe at the top of this file is fine for now. For example,

```tsx
const HOVERED_SECTION_COLOR = #80808080;
```

(Example future followup)
If we decide to refactor the hovered section color to depend on the theme at some point, then the const variable might be converted to a hook. For instance, 

```tsx
const useHoveredSectionColor = () => {
  const theme = useTheme()
  return theme.colors.gray
}
```

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
```diff
@@ -80,14 +80,15 @@ export default function ScheduleCalendar(props: ScheduleCalendarProps) {
...
-            ? finalsEventsInCalendar
-            : hoveredCourseEvents
-              ? [...eventsInCalendar, ...hoveredCourseEvents]
-              : eventsInCalendar;
+        if (showFinalsSchedule)
+            return hoveredCalendarizedFinal
+                ? [...finalsEventsInCalendar, hoveredCalendarizedFinal]
+                : finalsEventsInCalendar;
+        else
+            return hoveredCalendarizedCourses ? [...eventsInCalendar, ...hoveredCalendarizedCourses] : eventsInCalendar;
```
ap0nia, line 94: This would be larger scale refactor that might be out of scope of this PR, but one thing that would be nice is a derived store, `useEvents`, that returns the most recent array of events. Its implementation might look something like this:

```tsx
export interface UseEventsProps {
  showFinalsSchedule?: boolean
}

const useEvents = (props: UseEventsProps) => {
  const [ hoveredCalendarizedCourses, hoveredCalendarizedFinal ] = useHoveredStore(store => [store.hoveredCalendarizedCourses, store.hoveredCalendarizedFinal]);
  const [eventsInCalendar, setEventsInCalendar] = useState(AppStore.getEventsInCalendar());
  const [finalsEventsInCalendar, setFinalEventsInCalendar] = useState(AppStore.getFinalEventsInCalendar());

 useEffect(() => {
        const updateEventsInCalendar = () => {
            setEventsInCalendar(AppStore.getEventsInCalendar());
            setFinalEventsInCalendar(AppStore.getFinalEventsInCalendar());
        };


        AppStore.on('addedCoursesChange', updateEventsInCalendar);
        AppStore.on('customEventsChange', updateEventsInCalendar);
        AppStore.on('colorChange', updateEventsInCalendar);
        AppStore.on('currentScheduleIndexChange', updateEventsInCalendar);

        return () => {
            AppStore.off('addedCoursesChange', updateEventsInCalendar);
            AppStore.off('customEventsChange', updateEventsInCalendar);
            AppStore.off('colorChange', updateEventsInCalendar);
            AppStore.off('currentScheduleIndexChange', updateEventsInCalendar);
        };
    }, []);

   if (showFinalsSchedule)
            return hoveredCalendarizedFinal
                ? [...finalsEventsInCalendar, hoveredCalendarizedFinal]
                : finalsEventsInCalendar;
        else
            return hoveredCalendarizedCourses ? [...eventsInCalendar, ...hoveredCalendarizedCourses] : eventsInCalendar;
}
```

The main ideas are:
* Hooks are basically like function components because they follow the same rendering logic, i.e. it updates when its `showFinalsSchedule` prop is updated, just like a component.
* Hooks can proxy over complicated component logic by hiding multiple hooks inside of itself and exposing a simpler interface for the component. Here, the component would do something like `const events = useEvents(showFinalsSchedule)`, and then just render the events without any additional logic.

This is pretty large-scale, so don't worry about it for this PR, but it's good consideration for React architecture (and to be clear, React is an abysmal technology with the shittiest architectural patterns conceivable).

ap0nia (changes requested):
It looks like it works correctly for the most part, and the overall scope of the PR is appropriate 💯 . The main improvement I can see is just handling conditional flows more elegantly. Good work 👍 

This was my testing procedure:
- Open the settings menu and toggle "hover to preview".
- Go to the search tab and search "COMPSCI".
- Hover over different course offerings and see if their course day showed up on the calendar. This works ✅ 
- Toggle the finals button above the calendar to preview only finals.
- Hover over courses again, and the final previews show up on the calendar. ✅ 

Both features seemed to work correctly, so only code improvements seem relevant.

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
adcockdalton: I have incorporated this change and tried to study it, but my ability in React is a bit shaky in understanding this usage regardless. Could you perhaps briefly explain your use context of the arrow function here (for the sake of my learning, haha)?

Thank you!

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
adcockdalton: Oh wow! This last bit was really interesting. Thank you for the writeup. Maybe down the line I'll look into that refactor.

Thank you!

adcockdalton: Testing was successful again with all code improvements incorporated. That was a good exercise for me, I spent almost an hour examining those. Thanks!

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
ap0nia: Here's [Zustand's documentation on slicing](https://docs.pmnd.rs/zustand/guides/slices-pattern)

I visualize it like this:

```tsx
const useMyStore = () => {
  const [state1, setState1] = useState()
  const [state2, setState2] = useState()
  const [state3, setState3] = useState()
  
  return {
    state1, state2, state3, setState1, setState2, setState3
  }
}
```

Here, a custom hook provides you with 3 different stateful values. Any component that uses this hook, will re-render whenever any of those three state variables change (i.e. when their corresponding setter is called). Zustand's slicing pattern allows you to select which state variables to actually subscribe to via a callback.

```tsx
  const [state1, setState1] = useState()
  const [state2, setState2] = useState()
  const [state3, setState3] = useState()
  
  return sliceFunction({ state1, state2, state3, setState1, setState2, setState3 })
``` 

Whenever you provide a custom slicing function, it allows you to select a sub-set of stateful values to actually use in the component, so you can reduce the number of renders. This visualization isn't entirely accurate to reality, but it's how I interpret it 🤓

apps/antalmanac/src/stores/HoveredStore.ts
```diff
@@ -1,33 +1,64 @@
...
-                              },
-                          ])
-                        : undefined,
-            });
+        ...DEFAULT_HOVERED_STORE,
+        setHoveredEvents: (section, courseDetails, term) => {
+            if (section == null || courseDetails == null || term == null) {
+                set({ ...DEFAULT_HOVERED_STORE });
+                return;
+            } else {
```
ap0nia:
```suggestion
            }
```
If you have a guard clause, you can remove the explicit `else`, since there's no way for the code to continue after reaching a return statement.

apps/antalmanac/src/lib/termData.ts
```diff
@@ -112,7 +112,7 @@ function getFinalsStartForTerm(term: string) {
  */
 function getDefaultFinalsStartDate() {
     // FIXME: Un-offset once Spring starts, or figure out a proper fix
-    const [year, month, day] = termData[defaultTerm + 1].finalsStartDate || [];
+    const [year, month, day] = getDefaultFinalsStart() || [];
```
ap0nia:
```suggestion
    const [year, month, day] = getDefaultFinalsStart() ?? [];
```
Prefer nullish coalescing to an `or`, since nullish coalescing only falls back to the right value if the left is explicitly null or undefined, while `or` will fallback whenever it's falsy. Falsy values include empty strings, `0`, etc. which might be valid values and not the same as null/undefined.

ap0nia (approved):
Looks good, thanks for the changes ✨ 

Left some last comments; if you apply any changes, just re-request review and I'll merge it. Otherwise, you can merge it yourself whenever you're ready.

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
adcockdalton: Thank you! Very helpful!!

adcockdalton:
Apologies for the delay, I seem to have forgot that we didn't merge this one yet. Made the slight change. LGTM!

ap0nia (approved):
🚀

PR #920, opened 2024-02-23
Make dark mode look like how it looks with dark reader

### LINKED ISSUE #842: Make dark mode look like how it looks with dark reader
Dark Reader is a browser extension that turns any website dark mode, and IMO our site looks better with that on. Mainly it makes the white lines on the site less white.
Dark reader + our dark mode:
image
image

Our dark mode:
image

JacE070:
## Summary
Changed the color to make the dark mode UI more like with dark reader

## Test Plan
Bare eyes testing

## Issues

Closes https://github.com/icssc/AntAlmanac/issues/842

### DIFF
--- a/apps/antalmanac/src/components/Calendar/CalendarToolbar.tsx
+++ b/apps/antalmanac/src/components/Calendar/CalendarToolbar.tsx
@@ -301,7 +301,7 @@ function CalendarPaneToolbar(props: CalendarPaneToolbarProps) {
                 alignItems: 'center',
                 padding: 1,
                 borderRadius: '4px 4px 0 0',
-                backgroundColor: isDark ? darkModePalette.DARK_SECTION_HEADER_BACKGROUND : '',
+                backgroundColor: isDark ? darkModePalette.DARK_SECTION_HEADER_BACKGROUND : undefined,
             }}
         >
             <Box gap={1} display="flex" alignItems="center">
--- a/apps/antalmanac/src/components/Calendar/CalendarToolbar.tsx
+++ b/apps/antalmanac/src/components/Calendar/CalendarToolbar.tsx
@@ -16,6 +16,7 @@ import RenameScheduleDialog from '$components/dialogs/RenameSchedule';
 import DeleteScheduleDialog from '$components/dialogs/DeleteSchedule';
 import analyticsEnum, { logAnalytics } from '$lib/analytics';
 import AppStore from '$stores/AppStore';
+import { isDarkMode } from '$lib/helpers';
 
 function handleScheduleChange(index: number) {
     logAnalytics({
@@ -283,7 +284,6 @@ function CalendarPaneToolbar(props: CalendarPaneToolbarProps) {
 
     useEffect(() => {
         AppStore.on('scheduleNamesChange', handleScheduleNamesChange);
-
         return () => {
             AppStore.off('scheduleNamesChange', handleScheduleNamesChange);
         };
@@ -300,6 +300,7 @@ function CalendarPaneToolbar(props: CalendarPaneToolbarProps) {
                 alignItems: 'center',
                 padding: 1,
                 borderRadius: '4px 4px 0 0',
+                backgroundColor: isDarkMode() ? 'rgb(24, 26, 27)' : '',
             }}
         >
             <Box gap={1} display="flex" alignItems="center">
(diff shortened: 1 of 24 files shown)


apps/antalmanac/src/components/Calendar/calendar.css
```diff
@@ -40,6 +40,26 @@
...
+}
+
+.dark-mode .rbc-time-header-content {
+    border-left-color: rgb(58, 62, 65);
+}
+
+.dark-mode .rbc-time-header.rbc-overflowing {
+    border-right-color: rgb(58, 62, 65);
+}
+
```
ap0nia: If possible, you can see if these CSS styles can be embedded in a `styles` prop, so that way you can refer to the theme if possible. For example:
```tsx
function Calendar() {
  const theme = useTheme()

  return (
    <div style={{ '.dark-mode .rbc-time-view': { backgroundColor: theme.palette.main }}}>
    </div>
  )
}
```

JacE070: Hi Brian, I didn't figure out how to directly apply CSS selectors within the style attribute.
Can you help me out? @ap0nia

EricPedley (changes requested):
I'm just gonna comment on the new colors. If you agree with my opinions here you can go ahead and implement changes, but everything's pretty subjective so feel free to push back.
- There is too little contrast between the colors of the section headers and bodies (e.g. calendar grid vs toolbar and search results vs search/added classes/map tab). This is the only feedback I feel strongly about, the others are more like my personal preferences.
- I think the new background is too dark
- I think there is too much contrast on the white borders of the calendar and right panes, and the calendar grid lines. This was also a problem with our old colors.


For reference:
Current staging:
image
Old prod:
image
Old prod with my dark reader settings:
image

JacE070:
Hi Eric, I think your feedback is solid, and can you share your dark reader config with me? I was using the default config (everything off) @EricPedley

apps/antalmanac/src/components/Calendar/CalendarToolbar.tsx
```diff
@@ -300,6 +301,7 @@ function CalendarPaneToolbar(props: CalendarPaneToolbarProps) {
                 alignItems: 'center',
                 padding: 1,
                 borderRadius: '4px 4px 0 0',
+                backgroundColor: isDark ? 'rgb(24, 26, 27)' : '',
```
ap0nia: Could you store this as a variable somewhere? I think it would be helpful to keep theme colors centralized somewhere visible. The best place would be on the theme object where, and the next best class is just a top-level constant. e.g.

```tsx
const DARK_CALENDAR_BACKGROUND = 'rgb(25, 26, 27)'
```

Also, maybe hex color is nicer? I think HSL > Hex > RGB is a good, arbitrary preference to have. OKLCH is interesting and apparently displays colors more accurately, but I don't have experience with it.

apps/antalmanac/src/components/Header/Header.tsx
```diff
@@ -51,6 +52,9 @@ const components = [
 
 const Header = ({ classes }: CustomAppBarProps) => {
     const isMobileScreen = useMediaQuery('(max-width:750px)');
+    const isDark = useThemeStore((store) => store.isDark);
+
+    const backgroundColor = isDark ? '#264a92' : undefined;
```
ap0nia: Similar to the previous comment, prefer to store this variable somewhere. e.g.

```tsx
const DARK_HEADER_BACKGROUND = '#264a92'
```

Also, it looks like you opt for Hex and RGB interchangeably? I think it would be good to stick to one, and I think HSL would be a good option.

ap0nia (dismissed):
This is my initial review, but I'll take a closer look at the theme colors and get back to you on next steps.

apps/antalmanac/src/providers/Themev5.tsx
```diff
@@ -24,7 +24,7 @@ const darkTheme: PaletteOptions = {
         main: '#ffffff',
     },
     background: {
-        default: '#303030',
+        default: darkModePalette.DARK_BODY_BACKGROUND,
```
JacE070, line 27: I'm concerned about this... what is a better way?

apps/antalmanac/src/components/Calendar/calendar.css
```diff
@@ -40,6 +40,42 @@
     font-size: 0.9rem;
 }
 
+.dark-mode .rbc-time-content {
```
JacE070, line 43: Didn't find a way to conditionally apply the styles with CSS selectors orz

ap0nia:
This is too hard. I'd rather just overhaul to MUIv5 before dealing with any sort of theming. I'll defer to Eric for this 🤓

apps/antalmanac/src/providers/Themev5.tsx
MinhxNguyen7: Could be worse

MinhxNguyen7 (changes requested):
Mostly looks good. My only gripe is the border and grid colors. They stand out too much now that the background color are darker.

PR #921, opened 2024-02-24
feat: one tab mobile layer

### LINKED ISSUE #795: RFC: One tab layer on mobile
Right now, You need two tab layers to navigate on mobile. That's a pretty bad user experience because it makes it takes an extra click to get to places, and it makes it more daunting to learn the app. It also takes up valuable screen real estate.

Therefore, there should only be one tab layer with tabs, "Calendar, Search, Added, Map." Better yet, take up less space by using icons instead of text. We already have icons for everything except for the calendar.

There's a technical reason for there to be two layers, since the RightPane and the CourseCalendarPane are two separate things, but I can't imagine it being that difficult to consolidate them

KevinWu098:
## Summary
1. Simplify our tab layout by removing the Calendar / Classes split in favor of Calendar / Search / Added / Map

To-do:
- [x] Refactor to share code, reduce duplication
- [x] Consider using icons (?), if screen width allows

## Test Plan
1. UI testing

## Issues
Closes #795


KevinWu098:
Reminder: update the tutorial so that it hooks onto something other than the desktop tabs

MinhxNguyen7:
@KevinWu098 How this looking? Also, the tutorial shouldn't show on mobile, so I'm not sure what you're referring to.

KevinWu098:
> @KevinWu098 How this looking? Also, the tutorial shouldn't show on mobile, so I'm not sure what you're referring to.

I'll start back up on this PR this week; the comment above is referencing how the Tutorial is positioned to the RightPane, which the current implementation on this branch is reusing for mobile (and thus causing it to show up on mobile).

apps/antalmanac/src/components/Map/Map.tsx
```diff
@@ -175,17 +174,14 @@ export default function CourseMap() {
...
         };
 
         AppStore.on('addedCoursesChange', updateCalendarEvents);
-        AppStore.on('customEventsChange', updateCalendarEvents);
         AppStore.on('currentScheduleIndexChange', updateCalendarEvents);
 
         return () => {
             AppStore.removeListener('addedCoursesChange', updateCalendarEvents);
-            AppStore.removeListener('customEventsChange', updateCalendarEvents);
-            AppStore.removeListener('currentScheduleIndexChange', updateCalendarEvents);
```
MinhxNguyen7: Why is this listener not removed anymore when it's still being added above?

apps/antalmanac/src/components/Map/Map.tsx
```diff
@@ -175,17 +174,14 @@ export default function CourseMap() {
 
     useEffect(() => {
         const updateCalendarEvents = () => {
-            setCalendarEvents(AppStore.getEventsInCalendar());
+            setCalendarEvents(AppStore.getCourseEventsInCalendar());
         };
 
         AppStore.on('addedCoursesChange', updateCalendarEvents);
-        AppStore.on('customEventsChange', updateCalendarEvents);
```
MinhxNguyen7: Why is this being removed? Everything seems to work fine, so I guess it wasn't necessary but why is that the case?

MinhxNguyen7 (changes requested):
Very nice. Looks great. I do think that the bar should be at the bottom, but it's up to you if you want it to be.

I only have the concern about the listener. Also, while you're at it, can you resolve the merge conflict?

KevinWu098:
> Very nice. Looks great. I do think that the bar should be at the bottom, but it's up to you if you want it to be.
> 
> I only have the concern about the listener. Also, while you're at it, can you resolve the merge conflict?

Issues are resolved.

I also agree that the button should be on the bottom, but that change carries a host of styling fixes and adjustments and I'd rather do those in a follow-up PR. We can discuss @ sprint

MinhxNguyen7 (approved):
LGTM!

PR #912, opened 2024-02-16
Fix location ids and building catalogue

srukelman:
Created 1:1 parity between location ids and buildings catalogue

closes #856

PR #928, opened 2024-02-27
fix: handle undefined buildings

KevinWu098:
## Summary
1. The de-duplication / cleanup in #912 breaks schedules that used the removed locations, or more specifically, their IDs.
2. This PR is a hotfix to handle those undefined values more gracefully as empty strings, but it's not a full fix / undoing of the regression.

## Test Plan
1. Create a custom event with a location
2. Comment out location in `buildingCatalogue.ts`
3. See if clicking the Calendar Event errors out
4. See if opening Added errors out

## Issues
Closes [bug report](https://discord.com/channels/772739905981644850/780316580819107860/1211856788614291476):

> Feedback Type
Issue / Bug
Feedback
It says:
"Unexpected Application Error! Cannot read properties of undefined (reading 'name')
TypeError: Cannot read properties of undefined (reading 'name')


EricPedley (approved):
lgtm, clean fix.

PR #932, opened 2024-03-01
Swap past enrollment arrows and order

### LINKED ISSUE #931: Swap Past Enrollment Arrows
As described in https://github.com/icssc/AntAlmanac/pull/867#issuecomment-1961966081, it would be better to swap the arrows in the past enrollment popup such that the left arrow navigates to an older quarter's graph and the right arrow navigates to a more recent quarter's graph.

When the user first opens the popup, the most recent quarter's graph should still be displayed.

Douglas-Hong:
## Summary
- Swapped the arrows in the past enrollment popup such that the left arrow navigates to an older graph and the right arrow navigates to a newer graph.
- Instead of sorting the graphs from most recent quarter to oldest quarter, we will sort them from oldest quarter to most recent quarter.
- When the user first opens the popup, the most recent quarter's past enrollment graph should still be displayed.

## Test Plan
- Open the past enrollment popup and see if the arrows work properly and are disabled at the right times.

## Issues

Closes #931

### DIFF
--- a/apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
+++ b/apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
@@ -45,7 +45,7 @@ function PopupHeader({
                 width: graphWidth,
             }}
         >
-            <Tooltip title="Newer Graph">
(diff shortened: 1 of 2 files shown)


KevinWu098 (approved):
LGTM

PR #935, opened 2024-03-02
Make area around top bar of mobile devices the same color as app bar (PWA)

### LINKED ISSUE #840: Make the area around the notch on the PWA the same color as app bar
See the ZotMeal PWA for an example of this. It makes that area light blue. Also see #834.

pinapelz:
## Summary
This PR builds off of #834 . 
The issue is that you need to explicitly tell iOS that your website is PWA capable and that you want the status bar to match the theme color in your meta tags.

Assuming you guys also want something similar on Android, you'll need to specify what color as a `theme-color` meta tag.

I've applied the previously proposed colors to the manifest files and made the necessary changes to the meta tags. 

### Deployment Note
The PWA currently does not work off of the dev server with port forwarding, so we need to figure that out or find another way to test it.

### Before and After
<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>
      
    </td>
    <td>
      
    </td>
  </tr>
<tr>
    <td>
      
    </td>
    <td>
      
    </td>
</tr>
</table>

## Test Plan
Run locally and test visually

## Issues

Closes #840

### DIFF
--- a/apps/antalmanac/index.html
+++ b/apps/antalmanac/index.html
@@ -20,7 +20,9 @@
             name="viewport"
             content="width=device-width, initial-scale=1, shrink-to-fit=no, user-scalable=no, viewport-fit=cover"
         />
-        <meta name="theme-color" content="#000000" />
+        <meta name="theme-color" content="#305db7" />
+        <meta name="apple-mobile-web-app-capable" content="yes" />
+        <meta name="apple-mobile-web-app-status-bar-style " content="black-translucent" />
         <meta
             name="keywords"
             content="UCI,Anteater,search,classes,Calendar,development,software,add courses,antalmanac,course,course planner"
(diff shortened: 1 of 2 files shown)


MinhxNguyen7:
How did you test this locally as a PWA? 

The server seems to get completely borked when I try to access it as a PWA. I'm running Chrome on Android. @ap0nia any ideas why?

image

pinapelz:
Are you running the dev server and forwarding the port? I tried that and ran into an issue for that where the PWA was just a white screen (even when I just cloned directly from the main repo)

I ended up just deploying it to Cloudflare pages via `pnpm run build` and that seemed to allow me to use the PWA.  I have no idea why directly using the dev server doesn't work though.

MinhxNguyen7 (approved):
What can go wrong?

PR #940, opened 2024-03-06
chore: update lint configs

ap0nia:
The config is a bit outdated and doesn't enforce import order that well.

### DIFF
--- a/apps/antalmanac/.eslintrc.cjs
+++ b/apps/antalmanac/.eslintrc.cjs
@@ -1,43 +1,27 @@
-/** @type {import('eslint').Linter.Config} */
-const config = {
-  root: true,
-  parser: '@typescript-eslint/parser',
-  parserOptions: {
-    project: 'tsconfig.json',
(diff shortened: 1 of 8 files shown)


apps/antalmanac/.eslintrc.cjs
```diff
@@ -1,43 +1,27 @@
...
-  settings: {
-    react: {
-      version: 'detect',
-    }
-  },
-  ignorePatterns: ['vite.config.ts'],
-}
+    parserOptions: {
+        project: ['tsconfig.json'],
+    },
```
ap0nia, line 9: Since this config doesn't have the `"root": true` property, it means that it also inherits the settings from the eslint config from the project root, i.e. which contains the general TypeScript setup. That means that this eslint config only needs to specify react-specific eslint settings.

ap0nia:
Using the power of admin, I will merge since this shouldn't cause any major conflicts with open PRs.

PR #941, opened 2024-03-06
refactor: action cell

ap0nia:
## Summary
Begin migrating section table body to MUIv5.

Section table body cells will be encapsulated in `apps/antalmanac/src/components/RightPane/SectionTable/cells`. This can be refactored later, but it will provide visual clarity on the current progress of refactoring and reduce the size of the main body file.

## Test Plan
Visually confirm that the component works the same as before.


apps/antalmanac/src/components/RightPane/SectionTable/cells/action.tsx
```diff
@@ -0,0 +1,188 @@
+import { Add, ArrowDropDown, Delete } from '@mui/icons-material';
```
MinhxNguyen7, line 1: I'd prefer to call the file `ActionCell` instead of just `action`, especially because it sounds like it's a module of form actions.

apps/antalmanac/src/components/RightPane/SectionTable/cells/action.tsx
```diff
@@ -0,0 +1,188 @@
...
+import ColorPicker from '$components/ColorPicker';
+import analyticsEnum, { logAnalytics } from '$lib/analytics';
+import { CourseDetails } from '$lib/course_data.types';
+import AppStore from '$stores/AppStore';
+
+/**
+ * Copying a specific class's link will only copy its course code.
+ * If there is random value let in the url, it will interfere with the generated url.
+ */
+const fieldsToReset = ['courseCode', 'courseNumber', 'deptLabel', 'deptValue', 'GE', 'term'];
```
MinhxNguyen7: Considering this is only used once in `CopyAndAlert`, I think putting this down there would make it more readable instead of extracting it far away, especially because its name isn't strongly indicative of its *reason to-be*.

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
```diff
@@ -486,6 +487,7 @@ interface SectionTableBodyProps {
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 const tableBodyCells: Record<SectionTableColumn, React.ComponentType<any>> = {
+    action: SectionActionCell,
```
MinhxNguyen7, line 490: There's a type error here because the action cell is not currently a `SectionTableColumn`.

MinhxNguyen7 (changes requested):
The action cell does not show up. Otherwise, only nitpicks.

image

ap0nia:
My mistake, the columns not showing and the type errors were caused by me forgetting to commit the changes I made to the column store.

apps/antalmanac/src/components/RightPane/SectionTable/cells/action.tsx
ap0nia: I disagree because the `action.tsx` file is in the `cells` directory, and accessing the file `cells/action-cell.tsx` seems redundant for the same way that you wouldn't create component files like `components/TutorialComponent.tsx`.

MinhxNguyen7: Alright

MinhxNguyen7 (approved):
LGTM, but @KevinWu098 thinks we should wait until after the fee deadline to merge.

PR #944, opened 2024-03-08
Fix past enrollment date

### LINKED ISSUE #943: Past enrollment popup does not parse dates correctly
2/30/2024 is not a real date 😭 
image

Douglas-Hong:
## Summary
- We don't need to add 1 to the return value of `getDate()` since it is 1-indexed. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getDate

## Test Plan
- Go to a course (e.g. CS 142A) in Spring 2024 and verify that the last date of February is 2/29/2024 and the first date of March is 3/1/2024.
- Go to a course (e.g. CS 122D) in Spring 2023 and verify that the last date of February is 2/28/2023 and the first date of March is 3/1/2023.

## Issues

Closes #943

### DIFF
--- a/apps/antalmanac/src/lib/enrollmentHistory.ts
+++ b/apps/antalmanac/src/lib/enrollmentHistory.ts
@@ -107,7 +107,7 @@ export class DepartmentEnrollmentHistory {
 
             for (const [i, date] of enrollmentHistory.dates.entries()) {
                 const d = new Date(date);
-                const formattedDate = `${d.getMonth() + 1}/${d.getDate() + 1}/${d.getFullYear()}`;
+                const formattedDate = `${d.getMonth() + 1}/${d.getDate()}/${d.getFullYear()}`;
 
                 enrollmentDays.push({
                     date: formattedDate,


KevinWu098 (commented):
Briefly looked at the deployment on mobile, and the enrollment graphs seem to be NaNing.

imageimage
image

Happens on the examples above, as well as others I glanced at.

Douglas-Hong:
@KevinWu098 Sorry for the late reply, but I can't seem to reproduce the NaN dates on desktop or mobile on Chrome and Firefox. I'm guessing you're using it on Safari, but I don't have any Apple device.

If possible, can you try to reproduce it and see PeterPortal's response from `queryEnrollmentHistory` in `enrollmentHistory.ts`? Also, can you try to see if `parsedEnrollmentHistory` from `queryEnrollmentHistory` looks okay? It's weird how the dates are messed up but all the other data (number enrolled, max number, waitlist number) is still fine.

KevinWu098 (approved):
LGTM; The x-axis labels are NaNing on Safari (but is unrelated to this PR), will follow up in a subsequent PR

PR #945, opened 2024-03-12
Default Schedule name to recent terms

### LINKED ISSUE #907: RFC: Schedule bind with specific quarter
# The Default name for the schedule should be a current or upcoming quarter
Image

## For example:
Image
When May 6, the default schedule name should be `23' Fall`
Nov 4 --> `24' Winter`
Feb 10 --> `24' Spring`

# Bind the quarter with the schedule
The new schedule will default to bind with a current or upcoming quarter. Users can manually choose a term to bind with the schedule.
## Key features
1. When the user edits the schedule name, they can choose quarter info
Image

2. By choosing a different `schedule`, the `term` in search will change correspondingly. Same as changing `term` in search, the schedule will automatically be changed to the most recent schedule with the same term
Image

3. When user is looking at a schedule that is not current or upcoming quarter, there will be a clear warning to user.
Image

## Benefits of doing so
1. Clearer schedule that will not mess up with different term's course
2. Easy to use and save times

stevenguyukai:
## Summary
Default schedule name to the recent term
image

## Issues

Closes first half of #907


apps/antalmanac/src/components/dialogs/AddSchedule.tsx
```diff
@@ -46,7 +53,14 @@ function AddScheduleDialog(props: ScheduleNameDialogProps) {
...
     const submitName = useCallback(() => {
         addSchedule(name);
-        setName(`Schedule ${AppStore.getScheduleNames().length + 1}`);
+        setName(
+            `${
+                termName +
+                (AppStore.getScheduleNames().filter((name) => name.includes(termName)).length == 0
+                    ? ''
+                    : '(' + AppStore.getScheduleNames().filter((name) => name.includes(termName)).length + ')')
+            }`
```
ap0nia: Please refactor this into a dedicated function. The implementation would be something like this:

```ts
function getDefaultScheduleName() {
  const termName = termData[0].shortName.replaceAll(' ', '-');
  const countSameScheduleNames = scheduleNames.filter((name) => name.includes(termName)).length;
  return `${termName + (countSameScheduleNames == 0 ? '' : '(' + countSameScheduleNames + ')')}`
}
```

You can put it in the `AppStore` so it has access to the current state of the schedule. Inlining the code pollutes the function and can lead to inconsistencies between name calculations

apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -25,7 +28,12 @@ export class Schedules {
     constructor() {
         const scheduleNoteId = Math.random();
         this.schedules = [
-            { scheduleName: 'Schedule 1', courses: [], customEvents: [], scheduleNoteId: scheduleNoteId },
+            {
+                scheduleName: `${termData[0].shortName.replaceAll(' ', '-')}`,
```
ap0nia, line 41: After you write the new function for calculating the default schedule name, use that here.

ap0nia (changes requested):
Good simple change, just needs a small refactor

apps/antalmanac/src/stores/Schedules.ts
stevenguyukai: For this, there is a problem: The construction of Schedule is inside the AppStore, so Schedule will not be able to get the function from AppStore since AppStore has not been initialized yet.

MinhxNguyen7:
@ap0nia can you review this real quick?

apps/antalmanac/src/components/dialogs/AddSchedule.tsx
```diff
@@ -34,7 +36,12 @@ function AddScheduleDialog(props: ScheduleNameDialogProps) {
 
     const [scheduleNames, setScheduleNames] = useState(AppStore.getScheduleNames());
 
-    const [name, setName] = useState(`Schedule ${scheduleNames.length + 1}`);
+    const termName = termData[0].shortName.replaceAll(' ', '-');
+    const countSameScheduleNames = scheduleNames.filter((name) => name.includes(termName)).length;
+
+    const [name, setName] = useState(
+        `${termName + (countSameScheduleNames == 0 ? '' : '(' + countSameScheduleNames + ')')}`
+    );
```
ap0nia:
```suggestion
    const [name, setName] = useState(AppStore.getDefaultScheduleName());
```

Add a new method to AppStore as well:

```ts
class AppStore {
  // It just forwards the call to its schedule instance.
  getDefaultScheduleName() {
    return this.schedules.getDefaultScheduleName()
  }
}

Comment (ap0nia) [inline apps/antalmanac/src/components/dialogs/AddSchedule.tsx]:
```suggestion
    const [name, setName] = useState(AppStore.getDefaultScheduleName());
```

ap0nia (changes requested):
Ok really close, good work!

PR #949, opened 2024-03-28
Delineate Day and Schedule Selector in Custom Event Dialog

### LINKED ISSUE #520: Delineate day and schedule name in custom events modal
Right now, the days and schedules look like they're part of the same list. I think that there should be labels/headings to indicate the two. I think it would look better as well in addition to being quicker to parse.

image

srukelman:
closes #520


MinhxNguyen7 (approved):
LGTM!

PR #951, opened 2024-03-31
chore: add summer terms

KevinWu098:
## Summary
1. Adds summer terms

## Test Plan
1. Click around, add some courses

## Issues
Closes #950

PR #952, opened 2024-04-01
fix: custom event term

ap0nia:
## Summary

#951 Did in fact break something lol. Fixes the term assumed for custom events.

### DIFF
--- a/apps/antalmanac/src/lib/download.ts
+++ b/apps/antalmanac/src/lib/download.ts
@@ -7,7 +7,7 @@ import { notNull } from './utils';
 import { openSnackbar } from '$actions/AppStoreActions';
 import { CustomEvent, FinalExam } from '$components/Calendar/CourseCalendarEvent';
 import analyticsEnum, { logAnalytics } from '$lib/analytics';
-import { getDefaultTerm, termData } from '$lib/termData';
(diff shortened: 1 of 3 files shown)


apps/antalmanac/src/lib/download.ts
```diff
@@ -255,16 +255,30 @@ export function getRRule(bydays: string[], quarter: string) {
...
+    let customEventTerm = term ?? '';
+
+    // If a designated term has not been determined, assume term of first course event found.
+    if (!customEventTerm) {
+        for (const event of events) {
+            if (!event.isCustomEvent && event.term) {
+                customEventTerm = event.term as typeof customEventTerm;
+            }
+        }
+    }
```
ap0nia: We could also augment the `getDefaultTerm` function to accept an optionally provided list of events, and abstract this calculation there. Thoughts? @KevinWu098

KevinWu098: That makes sense to me -- I think abstracting/extracting it outside of `getEventsFromCourses` is better

KevinWu098 (changes requested):

KevinWu098 (approved):
LGTM; Downloaded a .ics with a regular event and a custom event, and it added correctly.

PR #954, opened 2024-04-02
fix: custom event term (for real)

ap0nia:
## Summary
Bad type assertion, rookie mistake 😭


KevinWu098 (approved):
I’ll approve, but I’ll say for the record I haven’t personally tested yet (am eating din din)

KevinWu098:
Okay, I've now tested w/ Jacob's schedule. LGTM

PR #955, opened 2024-04-08
MapBar updates when weekend custom events are added

### LINKED ISSUE #929: MapBar doesn't update to allow filtering by Sun/Sat for Custom Events
When a weekend course is added, the MapBar will update, like the Calendar does, to allow for filtering by Sun/Sat. 

However, Custom Events which are on the weekends do not trigger the update as it's based on `CalendarEvents` which comes from `AppStore.getCourseEventsInCalendar()`. 

It should instead be `AppStore.getEventsInCalendar()`.

Douglas-Hong:
## Summary
- When a weekend custom event is added, the map bar will be updated with the Sunday and Saturday filters.
- Fixed the ESLint errors in `Map.tsx`, pretty much just rearranging the imports.

Demo:

https://github.com/icssc/AntAlmanac/assets/78244965/9002ce52-7fbe-4244-ab28-58bfb492b6d6

## Test Plan
- Add a weekend custom event, and see if the map bar gets updated with Sun/Sat filters.
- Remove the weekend custom event, and see if the Sun/Sat filters disappear from the map bar.
- Try adding/removing a weekend custom event when the schedule already includes a weekend course (e.g. DRAMA 100 in Spring 2024).

## Issues

Closes #929

### DIFF
--- a/apps/antalmanac/src/components/Map/Map.tsx
+++ b/apps/antalmanac/src/components/Map/Map.tsx
@@ -1,21 +1,24 @@
 import './Map.css';
 
-import { Fragment, useEffect, useRef, useCallback, useState, createRef, useMemo } from 'react';
-import { useSearchParams, useNavigate } from 'react-router-dom';
-import L, { type Map, type LatLngTuple } from 'leaflet';
-import { MapContainer, TileLayer } from 'react-leaflet';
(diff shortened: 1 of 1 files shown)


MinhxNguyen7 (approved):
LGTM!

PR #902, opened 2024-02-11
Fix lag when going back from search results

js0mmer:
## Summary
* This has been annoying me for the past few months so I ran a bisect last night and located the culprit PR (#738)
* Moved the GlobalStyles from the SectionTable to the CourseRenderPane so there should only be one of them in existence now and not a hundred
  * kind of wild that was the issue

before:

https://github.com/icssc/AntAlmanac/assets/8922227/39635b71-d439-4521-a289-9321da250743

after:

https://github.com/icssc/AntAlmanac/assets/8922227/645194cc-e5c1-470e-b380-21a77d024269

## Test Plan
Try on both staging and prod:
1. Load a bunch of search results e.g. i've been searching COMPSCI and scrolling down until I reach the research classes
2. Click back
3. Compare how long it takes

## Issues

Closes #790

PR #738, opened 2023-10-16
Thinner section table scroll bar

MinhxNguyen7:
## Summary
Make horizontal scroll bar thin to match the rest of the app.

image

## Test Plan
Make sure that it shows up properly and that it's usable.

## Issues

Closes #735 

## Future Followup

Maybe have one side scroll for all sections like ZotCourse.

PR #958, opened 2024-04-15
Make Scrollbar a Uniform Thinness for Right Pane

### LINKED ISSUE #938: Horizontal scrollbar is thicker on added courses pane
This might be my bad. I didn't account for usage in the added pane in #902 

added courses pane
image

vs search pane
image

adcockdalton:
## Summary
This ended up being a really simple fix. To avoid creating unnecessary instances of GlobalStyles, I moved the one from CourseRenderPane.tsx into RightPaneRoot.tsx. Now all scrollbars on the right have unified thinness, which is the intended functionality, and are properly affected by one line rather than duplicate lines.

## Test Plan

_Test the styling_
Search CompSci
Add a bunch of random classes
Resize the window to show horizontal scrollbars
Confirm on Search and Added that they as well as their vertical counterparts are all thin

_Test the lag_
Scroll all the way down on a search or CompSci then use the back arrow and confirm no lag was reintroduced

## Issues

Closes #938 without breaking #902 , thereby expanding #738


adcockdalton:
Looking at this PR I'm realizing there might be some version control wonkiness--this was supposed to be a 2 file change and this PR is showing 11. Looking into it

PR #964, opened 2024-04-22
No Unsaved Warning From Tour

### LINKED ISSUE #914: Warning prompts without adding any classes
## Description
It seems that any key press is causing the unsaved warning to show up. This includes either typing in the search bar or into the schedule load modal.

## Related
#865 might have changed the way that save states work, and I do remember seeing some weird code in there. @stevem-zhou do you have any ideas?

MinhxNguyen7:
## Summary
- Fix bug where the unsaved schedule warning would pop up after dismissing tour and typing something in the search bar before closing the page.
- The problem was that the tour removes the sample classes on close, which toggles the warning, even though the user doesn't care about those classes, and they might not have even showed up.

## Test Plan
- Open the page in incognito, type something into the search bar, and close the page. Verify that the warning doesn't pop up.
- Load an existing schedule, remove a class, then close the page. Verify that the warning does still pop up.

## Issues
- Closes #914

### DIFF
--- a/apps/antalmanac/src/lib/tourExampleGeneration.ts
+++ b/apps/antalmanac/src/lib/tourExampleGeneration.ts
@@ -98,7 +98,7 @@ export function addSampleClasses() {
 }
 
 export function removeSampleClasses() {
-    AppStore.deleteCourses(sampleClassesSectionCodes, CURRENT_TERM);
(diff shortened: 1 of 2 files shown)


KevinWu098 (approved):
LGTM

PR #972, opened 2024-04-29
Mobile Drawer instead of Dropdown

### LINKED ISSUE #839: Move the Feedback and About into the drawer
They're not essential to the user experience. 

The settings button (which currently opens the drawer) should become a hamburger menu. The current settings should go on top, and the about and feedback should go at the bottom or stacked below, depending on what looks good.

There is currently a hamburger menu that opens a popover on mobile. The import and export buttons should move back into the app bar. That way, the drawer acts the same way on mobile and desktop.

To reiterate: The settings menu will no longer exist. Settings become a section in the app drawer.

MinhxNguyen7:
## Summary
- Replace the dropdown menu on mobile with a drawer. The fact that we had a drawer for desktop but a dropdown for mobile is just stupid because it should be the other way around, if we were to use a dropdown at all.
- Follows up from the two commits that I accidentally pushed to `main` to put the `About`, `Feedback`, and `Donate` buttons into the drawer (7d4ba66a7307719cfaf3b83cee8739be80af5550 and 2c00c3d393f70eb979ae791432beec602d81ea26).
- Also fixes the fact that the drawer open button was in the dropdown menu (although, before, the settings button was in there and would open the drawer, which is also not good). 

## Test Plan
- Verify that all the features are still in the header or drawer. 
- Validate my UI design choices.

## Issues
Closes #839

### DIFF
--- a/apps/antalmanac/src/components/Header/Header.tsx
+++ b/apps/antalmanac/src/components/Header/Header.tsx
@@ -1,9 +1,6 @@
-import { AppBar, Box, Menu, Toolbar, useMediaQuery } from '@material-ui/core';
-import MenuItem from '@material-ui/core/MenuItem';
+import { AppBar, Toolbar, useMediaQuery } from '@material-ui/core';
 import { withStyles } from '@material-ui/core/styles';
 import { ClassNameMap } from '@material-ui/core/styles/withStyles';
-import MenuIcon from '@material-ui/icons/Menu';
-import { useState, type MouseEventHandler } from 'react';
 
 import Export from './Export';
 import Import from './Import';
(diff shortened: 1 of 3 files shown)


adcockdalton:
ltgm.

edit: refer to following comments

(all features are present and I was able to export a schedule. UI decisions are good as well)

adcockdalton:
Apologies for prematurely closing the issue with a comment, I thought I was in the review text box. Moving my chrome windows around too much looking at the mobile view. Will move the suggestion into a proper code review

adcockdalton (approved):
The only thing I might consider looking at (though it is okay with me as-is so I'll close) is the vertical alignment of the import/export text.
image
image
Optionally the padding could be adjusted to center the text.

lgtm

adcockdalton:
Not deleting the branch in case you want to edit the CSS

Issue #973, opened 2024-04-29
Dynamically Change Logo by Special Event or Timeframe


adcockdalton:
It would be convenient to have the seasonal logos change automatically.

We could write up some general logic and then have an object where we can add these special events or timeframes.

PR #974, opened 2024-04-29
Change Christmas to Normal Logo

### LINKED ISSUE #969: Christmas Logo in mid-Spring
Not to be a scrooge, and this is just a nitpick, but the logos are still the Christmas versions.

Should be a simple change to the default, but I'm also curious if we had a spring design? I don't see it in assets. Did we have a designer for that, and can we ask them for a spring version? Just thinking out loud here.

adcockdalton:
## Summary
Just changed the path from christmas-logo.png to logo.svg. Simply reverted the seasonal logos to regular. Two line change

## Test Plan
Observe the mobile and desktop logos and confirm they are default.

## Issues

Closes #969 

## Future Followup
Do this dynamically (#973)


MinhxNguyen7 (approved):
LGTM!

PR #978, opened 2024-05-04
fix: correct the index of Added in TabStore

KevinWu098:
## Summary
1. #921 changed how tabs were ordered (Calendar is now index 0), so Added went from index 1 -> 2
2. This PR corrects TabStore's Added Tab index

## Test Plan
1. On both mobile and desktop views, the Added pane **should not** display GPA column. The Search pane **should**.

## Issues

Closes #


MinhxNguyen7 (approved):
LGTM!

PR #980, opened 2024-05-05
Always display GPA Column

### LINKED ISSUE #939: Show GPA column on Added Page
From the feedback form:

> It would be great if GPA was also displayed in the "Added" page

MinhxNguyen7:
## Summary
- The GPA column will now be shown on both the search and added panes.
- Remove vestigial logic in the stores which enabled the disabling of the columns in the added pane.
- Fix a bug that broke the column toggle buttons. This was broken before, but no one noticed.

## Test Plan
- Check that the GPA column shows up in both panes.
- Check that the column toggles work as expected.

## Issues
- Closes #939 

## Future Followup
- Would be nice for the column filtering to persist between page loads.

### DIFF
--- a/apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
+++ b/apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
@@ -32,6 +32,7 @@ const buttonSx: SxProps = {
 };
 
 const columnLabels: Record<SectionTableColumn, string> = {
+    action: 'Action',
     sectionCode: 'Code',
     sectionDetails: 'Type',
     instructors: 'Instructors',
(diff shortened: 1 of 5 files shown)


apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
```diff
@@ -91,6 +93,9 @@ export function ColumnToggleButton() {
         [selectedColumns]
     );
 
+    console.log('selectedColumns', selectedColumns);
+    console.log('selectedColumnNames', selectedColumnNames);
```
KevinWu098: chore: remove console.log

KevinWu098 (changes requested):
LGTM! Lookin' forward to having GPA column in the Added section 😋 

Just one chore and we're good to go!

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneButtonRow.tsx
MinhxNguyen7: oops

KevinWu098 (approved):
LGTM

PR #985, opened 2024-05-20
feat: Auth final

ap0nia:
## Summary
- Frontend implementation of auth.
- Uses current DynamoDB without migrating to RDS yet.

## Refinements
- If user is not logged in, they should not be able to change visibility.
- If user is not logged in, they're allowed to pick any user ID to save.
- If user is is logged in, the user ids to save the schedule under is readonly.
- Toggle visibility settings in account dialog.

## Future Follow-Up
- Migrate system to RDS.


MinhxNguyen7:
Is this done? Also, what do you mean by "excludes DB"? Do you mean it's already been implemented?

Also, can you specify what you expect to/not to work so that I can check it?

PR #986, opened 2024-05-20
Auto-update the Logo Based on Season/Holiday

### LINKED ISSUE #973: Dynamically Change Logo by Special Event or Timeframe

adcockdalton:
## Summary
Wrote out a Logo component that consolidates the logic previously used for the logo (such as should it be mobile version) and returns the appropriate logo based on the current date. To add a new seasonal logo, simply create a new Logo object in the Logos array.

Not code related, but the windows I used for the logos can be widened/updated if appropriate. Not sure how long the winter one should extend to, for example. Would love thoughts on that.

## Test Plan
- View the logo as-is. Since we are not in a particular event range it should be default
- Confirm that it still switches to mobile when window is resized
- Edit the "current date" object to be any other possible date. The logo should reflect the season of that date

## Issues

Closes #973 to follow up #974 and similar

## Future Followup

Do we have more designs? The default logo as-is will display for 8 contiguous months.

### DIFF
--- a/apps/antalmanac/src/components/Header/Logo.tsx
+++ b/apps/antalmanac/src/components/Header/Logo.tsx
@@ -1,3 +1,6 @@
+import { useMediaQuery } from '@mui/material';
+import { useMemo } from 'react';
+
 import ChristmasLogo from '$assets/christmas-logo.png';
 import MobileChristmasLogo from '$assets/christmas-mobile-logo.png';
 import HalloweenLogo from '$assets/halloween-logo.png';
@@ -6,6 +9,27 @@ import DefaultLogo from '$assets/logo.svg';
 import MobileDefaultLogo from '$assets/mobile-logo.svg';
 import ThanksgivingLogo from '$assets/thanksgiving-logo.png';
 import MobileThanksgivingLogo from '$assets/thanksgiving-mobile-logo.png';
+
+type Logo = {
+    name: string;
+    mobileLogo: string;
+    desktopLogo: string;
+    startDay: number; // inclusive
+    startMonthIndex: number;
+    endDay: number; // exclusive
+    endMonthIndex: number;
+    attribution?: string;
+};
+
+const defaultLogo: Logo = {
+    name: 'Default',
+    mobileLogo: MobileDefaultLogo,
+    desktopLogo: DefaultLogo,
+    startDay: 0,
+    startMonthIndex: 0,
+    endDay: 31,
+    endMonthIndex: 12,
+};
 
 const logos: Logo[] = [
     {
@@ -16,6 +40,7 @@ const logos: Logo[] = [
         startMonthIndex: 11,
         endDay: 1,
         endMonthIndex: 1,
+        attribution: 'Thanks Aejin for designing this seasonal logo!',
     },
     {
         name: 'Thanksgiving',
@@ -25,6 +50,7 @@ const logos: Logo[] = [
         startMonthIndex: 10,
         endDay: 1,
         endMonthIndex: 11,
+        attribution: 'Thanks Aejin for designing this seasonal logo!',
     },
     {
         name: 'Halloween',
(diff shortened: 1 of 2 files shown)


apps/antalmanac/src/components/Header/Logo.tsx
```diff
@@ -0,0 +1,97 @@
...
+    endMonthIndex: number;
+}
+
+interface LogoProps {
+    isMobileScreen: boolean;
+    logos: Logo[];
+}
+
+const inBounds = (props: Logo): boolean => {
+    // get current date
```
KevinWu098:
```suggestion
```
chore (nitpicky): this comment is a tad redundant, let's remove it

apps/antalmanac/src/components/Header/Logo.tsx
```diff
@@ -0,0 +1,97 @@
...
+    const currentDate = new Date();
+
+    // for testing...
+    // const currentDate = new Date(2024, 8, 30); // within default logo range
+    // const currentDate = new Date(2024, 9, 31); // last day of halloween logo (last day of october)
+    // const currentDate = new Date(2024, 10, 1); // first day of thanksgiving logo (first day of november, etc.)
+    // const currentDate = new Date(2024, 10, 30); // last day of thanksgiving logo
+    // const currentDate = new Date(2024, 0, 15); // should be christmas logo
+    // const currentDate = new Date(2024, 1, 1); // first day back in default range - dalton as of 5/20/24
+
```
KevinWu098:
```suggestion
```
chore: remove testing code

apps/antalmanac/src/components/Header/Logo.tsx
```diff
@@ -0,0 +1,97 @@
...
+    const startDate = new Date(startYear, props.startMonthIndex, props.startDay);
+
+    // check if current date is in bounds of special event/holiday/season
+    return currentDate >= startDate && currentDate < endDate;
+};
+
+const getLogo = (props: LogoProps): string => {
+    for (const logo of props.logos) {
+        if (inBounds(logo)) {
+            // console.log('using the ' + logo.name + ' logo!');
```
KevinWu098:
```suggestion
```
chore: remove console log

apps/antalmanac/src/components/Header/Logo.tsx
```diff
@@ -0,0 +1,97 @@
...
+    // const currentDate = new Date(2024, 1, 1); // first day back in default range - dalton as of 5/20/24
+
+    // get bounds
+    const currentYear = currentDate.getFullYear();
+    const endYear = currentYear;
+    const startYear = props.startMonthIndex > props.endMonthIndex ? endYear - 1 : endYear;
+    const endDate = new Date(endYear, props.endMonthIndex, props.endDay);
+    const startDate = new Date(startYear, props.startMonthIndex, props.startDay);
+
+    // check if current date is in bounds of special event/holiday/season
```
KevinWu098:
```suggestion
    const currentYear = currentDate.getFullYear();
    const endYear = currentYear;
    const startYear = props.startMonthIndex > props.endMonthIndex ? endYear - 1 : endYear;
    const endDate = new Date(endYear, props.endMonthIndex, props.endDay);
    const startDate = new Date(startYear, props.startMonthIndex, props.startDay);

```
chore (nitpicky): these comments are a bit redundant -- too many comments reduces the value of other insightful/clarifying comments.

A good rule of thumb might be that comments should be reserved for more confusing/less common code. Although, do note that confusing code might be better improved by refactoring, rather than depending on comments.

apps/antalmanac/src/components/Header/Logo.tsx
```diff
@@ -0,0 +1,97 @@
...
+    }
+    return props.isMobileScreen ? MobileDefaultLogo : DefaultLogo;
+};
+
+export default function Logo({ isMobileScreen }: { isMobileScreen: boolean }) {
+    return (
+        <img
+            height={32}
+            src={getLogo({ isMobileScreen: isMobileScreen, logos: logos })}
+            title={'Thanks Aejin for designing this seasonal logo!'}
```
KevinWu098: issue: Aejin should not be credited on the default logo

suggestion: because this now requires title to _also_ be aware of the current date range, you might want to extract the bounds logic so that it can be shared between the `src` and `title`. I'll  leave the particulars to you though!

KevinWu098 (changes requested):
Good work -- a handful of nitpicks and one issue to cleanup, but otherwise it's lookin' solid!

apps/antalmanac/src/components/Header/Logo.tsx
adcockdalton: oh yeah, the comments were just so whoever was gonna review the code could uncomment those to test it really quick. I never intended them to stay in the codebase haha

KevinWu098 (approved):
LGTM 🚀 🚀

ap0nia (changes requested):
I'm going to block this until #989 has been resolved.

MinhxNguyen7 (approved):
```
                __/___            
          _____/______|           
  _______/_____\_______\_____     
  \              < < <       |    
```

PR #987, opened 2024-05-24
fix: date formatting logic in enrollmentHistory for Safari

### LINKED ISSUE #947: Past Enrollment Dates are NaN on Safari
In #944, we found that the dates on the x-axis in the past enrollment popup are `NaN/NaN/NaN` in Safari, as seen in the screenshots in https://github.com/icssc/AntAlmanac/pull/944#pullrequestreview-1926077916. The dates are correct on Chrome and Firefox, but Safari has issues.

It might be helpful to see the response we get from PeterPortal in the `queryEnrollmentHistory` function in `enrollmentHistory.ts`. It also might be helpful to look at `parsedEnrollmentHistory` from `queryEnrollmentHistory`.

pinapelz:
## Summary
Safari handles parsing dates differently than other browsers. Here is one way of fixing it by handling the date as a string before making it a Date object

I think you can also maybe using something like `date-fns` to handle the cross platform behavior for parsing dates (although I personally think that might be too overkill for just this one use case)

## Test Plan
Verified that the result is the same as what's on prod. Tested on Mac and iOS
image

## Issues

Closes #947

### DIFF
--- a/apps/antalmanac/src/lib/enrollmentHistory.ts
+++ b/apps/antalmanac/src/lib/enrollmentHistory.ts
@@ -106,8 +106,9 @@ export class DepartmentEnrollmentHistory {
             const enrollmentDays: EnrollmentHistoryDay[] = [];
 
             for (const [i, date] of enrollmentHistory.dates.entries()) {
-                const d = new Date(date);
-                const formattedDate = `${d.getMonth() + 1}/${d.getDate()}/${d.getFullYear()}`;
+                const dateParts = date.split('-');
+                const d = new Date(Number(dateParts[0]), Number(dateParts[1]), Number(dateParts[2]));
+                const formattedDate = `${d.getMonth()}/${d.getDate() - 1}/${d.getFullYear()}`;
 
                 enrollmentDays.push({
                     date: formattedDate,
--- a/apps/antalmanac/src/lib/enrollmentHistory.ts
+++ b/apps/antalmanac/src/lib/enrollmentHistory.ts
@@ -105,10 +105,10 @@ export class DepartmentEnrollmentHistory {
         for (const enrollmentHistory of res) {
             const enrollmentDays: EnrollmentHistoryDay[] = [];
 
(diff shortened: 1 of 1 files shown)


apps/antalmanac/src/lib/enrollmentHistory.ts
```diff
@@ -106,8 +106,9 @@ export class DepartmentEnrollmentHistory {
             const enrollmentDays: EnrollmentHistoryDay[] = [];
 
             for (const [i, date] of enrollmentHistory.dates.entries()) {
-                const d = new Date(date);
-                const formattedDate = `${d.getMonth() + 1}/${d.getDate()}/${d.getFullYear()}`;
+                const dateParts = date.split('-');
+                const d = new Date(Number(dateParts[0]), Number(dateParts[1]), Number(dateParts[2]));
```
ap0nia: Could you make the parts of the date a bit more descriptive?

```suggestion
                const [day='', month='', year='']= date.split('-');
                const date = new Date(Number(day), Number(month), Number(year));
```

apps/antalmanac/src/lib/enrollmentHistory.ts
```diff
@@ -106,8 +106,9 @@ export class DepartmentEnrollmentHistory {
...
+                const formattedDate = `${d.getMonth()}/${d.getDate() - 1}/${d.getFullYear()}`;
```
ap0nia: If there's no naming conflicts, could we use the variable `date`?
```suggestion
                const formattedDate = `${date.getMonth()}/${date.getDate() - 1}/${date.getFullYear()}`;
```

ap0nia (changes requested):
Thanks, I was totally unaware of this fact! I left some suggestions, but I can put in some time to fix it myself later when I have the chance. 🚀

pinapelz:
I've applied the changes. There was a naming conflict with date so I renamed the date data taken from enrollmentHistory to dateString instead.

ap0nia (approved):
Looks good, I'll merge it when I'm able to 🎉

PR #988, opened 2024-05-24
Fixed GPA not showing on Courses Added pane

### LINKED ISSUE #983: GPA column doesn't work in added courses pane
## Description
- The GPA column is there but is always empty.
- When you search for the class in the search pane, the GPA column shows up properly.
- When you come back to the added pane, the number for that course is there.

### Diagnosis
- There is a parameter in the method that handles querying for courses which controls whether to only read from the cache or check upstream from the API.
    - This is so that the search pane can send a batch query to the API instead of the courses individually querying.
- The bug is most likely caused by the fact that the GPA API call is not triggered by a search, then the section table cells search the cache, finds nothing, and the query is not propagated to the API because that parameter is not toggled.
    - This should be fixable by simply by passing `true` into that parameter.
    - However, we should quickly check for query deduplication. I think I implemented something that stops duplicate queries when one is in-flight, but we should check.

adcockdalton:
## Summary
As discussed extensively in #983 , the GPA column on the course added pane would show no data unless it was previously loaded into the cache by a search. There were concerns with query deduplication as well, but I believe those should be dispelled by the tests I conducted:

## Test Plan
- If you encapsulate both of the two calls to query the grades with console.log statements, you can observe that even between courses vs. discussions, different discussions of the same course, etc. they all go to the cache once the class itself has been queried once:
image

To test further,
- Load empty schedule
- Add a couple courses via search pane
- Observe GPA in added courses pane
- Refresh the page and go to added courses pane. GPA should be there
- Add another class and remove one previously added
- Observe added courses pane. Even with refresh, and even if you add the removed course back, added courses pane should always show GPA

## Issues

Closes #983

### DIFF
--- a/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
+++ b/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody.tsx
@@ -214,12 +214,7 @@ async function getGpaData(deptCode: string, courseNumber: string, instructors: s
 
     // Get the GPA of the first instructor of this section where data exists
     for (const instructor of namedInstructors) {
-        const grades = await Grades.queryGrades(
-            deptCode,
-            courseNumber,
-            instructor,
-            useTabStore.getState().activeTab != 1
-        );
+        const grades = await Grades.queryGrades(deptCode, courseNumber, instructor, false);
         if (grades?.averageGPA) {
             return {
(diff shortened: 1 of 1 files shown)


KevinWu098 (approved):
LGTM, tested on the deployment 🚀

ap0nia:
I'm not reviewing this, but I like the detailed PR information 💪

PR #991, opened 2024-06-02
feat: warn when saving empty schedules

### LINKED ISSUE #990: Alert users when saving an empty schedule
Alternatively, we could disallow saving an empty schedule...

The main concern here is that users are deleting their schedules by completely overwriting their data as our backend is fetching data.

KevinWu098:
## Summary
1. When AA or AnteaterAPI is running slow, users may unintentionally save a completely empty schedule and wipe their own schedules.
2. This PR adds a check for completely empty schedules and warns users, as this is an unexpected action for a user to take (saving empty schedules).

## Test Plan
1. Attempt to save empty schedules, a window confirmation should appear
2. After adding courses, custom events, OR a schedule note, the window confirmation should not appear

## Issues
Closes #990


MinhxNguyen7 (approved):
LGTM!

PR #994, opened 2024-06-03
feat: support switching skeleton schedules

### LINKED ISSUE #701: Support Switching Schedules in Skeleton Mode
This would be a follow-up for https://github.com/icssc/AntAlmanac/pull/569

I'm not sure if the current skeleton schedule implementation supports multiple schedules, so that's something worth investigating. 

Skeleton Mode should allow users to switch between different schedules. Currently, it only displays the first schedule the user has, and doesn't behave the same way Save does, in the way that saving a schedule will make it the default when AA is opened again.

This second point about skeleton schedule not behaving like save could be an issue unto itself 👀

KevinWu098:
## Summary
1. Follows up #569 by allowing users to switch their schedule when in skeleton mode

chrome-capture-2024-5-3 (2)

## Test Plan
1. Create at least two schedules (with classes)
3. Block `https://api-next.peterportal.org` in dev tools: https://developer.chrome.com/docs/devtools/network-request-blocking
4. Confirm that schedules can be swapped around

## Issues
Closes #701

## Future Followup
On mobile views, we should make it more explicit that schedules can be changed in skeleton mode

### DIFF
--- a/apps/antalmanac/src/components/Calendar/CalendarToolbar.tsx
+++ b/apps/antalmanac/src/components/Calendar/CalendarToolbar.tsx
@@ -46,7 +46,12 @@ function handleUndo() {
     undoDelete(null);
 }
 
-function EditScheduleButton(props: { index: number }) {
+interface RenameScheduleButtonProps {
+    index: number;
+    disabled?: boolean;
+}
+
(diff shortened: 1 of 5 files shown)


MinhxNguyen7 (approved):
LGTM!

PR #1000, opened 2024-08-20
feat: clean up 98 eslint errors and warnings

KevinWu098:
59 errors and 2 warnings potentially fixable with the `--fix` option.

> ✖ 3 problems (0 errors, 3 warnings)

1. This PR resolves almost 100 eslint problems (the remainder are ones where I didn't feel confident touching on components relating to `Map`) 
2. Broadly, most of these errors and warnings fall into a few buckets:
    - Import Order
    - Default Exports
    - Type errors (`any`)
    - Exhaustive Deps

## Test Plan
1. These changes do span a lot of files, but no business logic has been changed (aside from useEffect/useCallback dependencies). Potential tests would just be clicking around and validating big features haven't regressed

### DIFF
--- a/apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/ScheduleSelector.tsx
+++ b/apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/ScheduleSelector.tsx
@@ -1,9 +1,8 @@
 import Checkbox from '@material-ui/core/Checkbox';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
 import FormGroup from '@material-ui/core/FormGroup';
-import React, { PureComponent } from 'react';
-
 import type { RepeatingCustomEvent } from '@packages/antalmanac-types';
+import { ChangeEvent, PureComponent } from 'react';
 
 interface ScheduleSelectorProps {
     customEvent?: RepeatingCustomEvent;
@@ -21,7 +20,7 @@ class ScheduleSelector extends PureComponent<ScheduleSelectorProps, ScheduleSele
         scheduleIndices: this.props.scheduleIndices,
     };
 
-    handleChange = (dayIndex: number) => (event: React.ChangeEvent<HTMLInputElement>) => {
+    handleChange = (dayIndex: number) => (event: ChangeEvent<HTMLInputElement>) => {
         const checked = event.target.checked;
 
         this.setState(
--- a/apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/ScheduleSelector.tsx
+++ b/apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/ScheduleSelector.tsx
@@ -2,7 +2,7 @@ import Checkbox from '@material-ui/core/Checkbox';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
 import FormGroup from '@material-ui/core/FormGroup';
 import type { RepeatingCustomEvent } from '@packages/antalmanac-types';
-import { ChangeEvent, PureComponent } from 'react';
+import { PureComponent } from 'react';
 
 interface ScheduleSelectorProps {
     customEvent?: RepeatingCustomEvent;
@@ -20,7 +20,7 @@ class ScheduleSelector extends PureComponent<ScheduleSelectorProps, ScheduleSele
         scheduleIndices: this.props.scheduleIndices,
     };
 
-    handleChange = (dayIndex: number) => (event: ChangeEvent<HTMLInputElement>) => {
+    handleChange = (dayIndex: number) => (event: React.ChangeEvent<HTMLInputElement>) => {
         const checked = event.target.checked;
 
         this.setState(
--- a/apps/antalmanac/src/components/Calendar/Toolbar/EditSchedule/ScheduleNameDialog.tsx
+++ b/apps/antalmanac/src/components/Calendar/Toolbar/EditSchedule/ScheduleNameDialog.tsx
@@ -12,7 +12,7 @@ import {
 import { withStyles } from '@material-ui/core/styles';
 import { ClassNameMap } from '@material-ui/core/styles/withStyles';
 import { Add, Edit } from '@material-ui/icons';
-import { forwardRef, useCallback, useState, useMemo, MouseEvent, ChangeEvent, KeyboardEvent } from 'react';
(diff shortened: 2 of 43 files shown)


apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/ScheduleSelector.tsx
```diff
@@ -1,9 +1,8 @@
 import Checkbox from '@material-ui/core/Checkbox';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
 import FormGroup from '@material-ui/core/FormGroup';
-import React, { PureComponent } from 'react';
-
 import type { RepeatingCustomEvent } from '@packages/antalmanac-types';
+import { ChangeEvent, PureComponent } from 'react';
```
ap0nia: Did you intentionally import `ChangeEvent`? The `React` namescope should be available globally in React projects, and it's easier to use `React.ChangeEvent` for types.

```suggestion
import { type ChangeEvent, PureComponent } from 'react';
```

apps/antalmanac/src/components/Calendar/Toolbar/EditSchedule/ScheduleNameDialog.tsx
```diff
@@ -13,6 +12,7 @@ import {
 import { withStyles } from '@material-ui/core/styles';
 import { ClassNameMap } from '@material-ui/core/styles/withStyles';
 import { Add, Edit } from '@material-ui/icons';
+import { forwardRef, useCallback, useState, useMemo, MouseEvent, ChangeEvent, KeyboardEvent } from 'react';
```
ap0nia: I'll recap this in the overall review as well, but please distinguish between type imports and named imports whenever possible.

```suggestion
import { forwardRef, useCallback, useState, useMemo, type MouseEvent, type ChangeEvent, type KeyboardEvent } from 'react';
```

apps/antalmanac/src/components/Calendar/Toolbar/EditSchedule/ScheduleNameDialog.tsx
```diff
@@ -49,7 +49,7 @@ const ScheduleNameDialog = forwardRef((props: ScheduleNameDialogProps, ref) => {
 
     // We need to stop propagation so that the select menu won't close
     const handleOpen = useCallback(
-        (event: React.MouseEvent) => {
+        (event: MouseEvent<HTMLLIElement>) => {
```
ap0nia: Strongly typing the generic when you don't need it is just superfluous.

```suggestion
        (event: MouseEvent) => {
```

apps/antalmanac/src/components/Map/Routes.tsx
ap0nia, line 1: Maps are very finnicky, please double check that it continues to work after these changes. `import L from 'leaflet'` is how imports are done based on all documentation and references to leaflet.
Ideally, they'd be able handle different environments slightly better, but I'm not sure if they do :^)

apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.types.ts
```diff
@@ -1,5 +1,4 @@
 import { ClassNameMap } from '@material-ui/core/styles/withStyles';
-
 import { AACourse } from '@packages/antalmanac-types';
```
ap0nia:
```suggestion
import type { AACourse } from '@packages/antalmanac-types';
```

apps/antalmanac/src/components/inputs/building-select.tsx
```diff
@@ -23,9 +24,12 @@ export type BuildingSelectProps = {
...
 
 export function BuildingSelect(props: BuildingSelectProps) {
-    const handleChange = useCallback(async (_event: React.SyntheticEvent, value: ExtendedBuilding | null) => {
-        await props.onChange?.(value);
-    }, []);
+    const handleChange = useCallback(
+        async (_event: React.SyntheticEvent, value: ExtendedBuilding | null) => {
+            await props.onChange?.(value);
+        },
+        [props]
```
ap0nia: This is bad; don't pass the entire props object because the reference is not stable. Specify the property, via destructuring is best.
I'm pretty sure I'm the one that did this, so this is a note for my past self.

```tsx
function MyComponent(props) {
  const { a, b, c } = props

  const memoizedThing = useMemo(() => {}, [a, b, c])
}
```

*In more depth*
Whenever the parent component re-renders, a new "props" object is constructed and passed to the child. The props object can have the same values as the props object from the previous render, but it's a different entity in memory. If you attempt to use the "props" object as a dependency for memo-ization, it should not be effective because it changes; you want to refer to primitive values within the object explicitly.

apps/antalmanac/src/lib/helpers.ts
```diff
@@ -1,6 +1,7 @@
-import React from 'react';
+import { MouseEvent } from 'react';
 
 import { PETERPORTAL_GRAPHQL_ENDPOINT } from './api/endpoints';
+
 import { openSnackbar } from '$actions/AppStoreActions';
 
 export async function queryGraphQL<PromiseReturnType>(queryString: string): Promise<PromiseReturnType | null> {
```
ap0nia, line 7: On a side note, this is a completely useless use of generics.

```ts
const result = queryGraphQL<string>()

const result: string = queryGraphQL()
```

In both scenarios, you have to explicitly type the result anyways. It's just that when you use generics, the function signature is "more complex" and doesn't provide you any type safety or DX in anyway.

ap0nia (changes requested):
1. Why are we importing types from `react` instead of using the global `React` namespace. If `eslint` is unable to detect this global namespace, then we need to fix the configuration.
2. Please specify imports as types wherever relevant. This should ideally be enforced via `tsconfig`, but it is not. You don't have to apply this rule across the entire codebase, but at least for the changes you're applying in this PR.
3. Please confirm that the map works as expected. Leaflet is generally a very janky library that heavily relies on the use of `window` globals to work.

apps/antalmanac/src/components/Map/Routes.tsx
KevinWu098: Checked on mobile and on desktop (Safari, Chrome, respectively) and the map is still working as expected! I'm happy to revert the changes to the Map though -- don't fix what ain't broke -- if you think it's too risky

MinhxNguyen7 (approved):
Seems alright

PR #1001, opened 2024-08-20
feat: mobile ui and saved scroll position

### LINKED ISSUE #976: Move Mobile Tabs to Bottom
> I do think that the bar should be at the bottom

Screenshot 2024-05-04 at 12 23 30 PM

_Originally posted by @MinhxNguyen7 in https://github.com/icssc/AntAlmanac/pull/921#pullrequestreview-2026748537_

### LINKED ISSUE #1016: download .ics file not working
I cannot download the .ics file to import into my Google Calendar. The button gives visual feedback that it's been clicked but no download begins. I tried disabling adblock, using a different browser, and even using a different computer/phone but it did not work in any case. 

Downloading the screenshot works fine, it's just the .ics file that doesn't work.

ap0nia:
# Summary

Save the scroll position of the previous tab when switching to a new tab, and restore the scroll position if the previous tab is revisited.

Move the tabs to the bottom and accommodate for mobile-specific factors like the bottom edge of the phone, rounded corners, etc.

# Testing

## Mobile Responsiveness
1. Open the website in desktop mode. The calendar and the tabs with search, added courses, and map should be side-by-side.
2. Shrink the viewport to a mobile layout. There should be tabs on the bottom.

Ensure that the calendar and tabs are synchronized with their media queries.
Ensure that the elements are visually balanced at all points during the viewport shift.

## Scroll Save
1. Search for "CS" in the "search courses tab."
2. Scroll down at least two full screen heights, and take note of the courses displayed.
3. Switch to the "map" tab.
4. Switch back to the "search" tab and verify that the same courses are displayed.

# Project
Continuation of #999. I was originally trying to review it and figure out how I would actually implement it so i could provide feedback, but it ended up covering a larger scope.

Resolves #976 
Resolves #1016 by optional chaining the custom event location's name.

### DIFF
--- a/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
+++ b/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
@@ -213,15 +213,7 @@ export default function ScheduleCalendar(props: ScheduleCalendarProps) {
     }, []);
 
     return (
-        <Box
-            style={{
-                height: isMobile ? 'calc(100% - 50px)' : undefined,
-                margin: '0px 4px',
-                borderRadius: '1px',
-                flexGrow: 1,
-            }}
-            id={'calendar-root'}
-        >
+        <Box id="calendar-root" marginY={1} borderRadius={1} flexGrow={1}>
             <CalendarToolbar
                 currentScheduleIndex={currentScheduleIndex}
                 toggleDisplayFinalsSchedule={toggleDisplayFinalsSchedule}
--- a/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
+++ b/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
@@ -218,6 +218,7 @@ export default function ScheduleCalendar(props: ScheduleCalendarProps) {
                 height: isMobile ? 'calc(100% - 50px)' : undefined,
                 margin: '0px 4px',
                 borderRadius: '1px',
+                flexGrow: 1,
             }}
             id={'calendar-root'}
         >
--- a/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
+++ b/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
@@ -213,15 +213,7 @@ export default function ScheduleCalendar(_props?: ScheduleCalendarProps) {
     }, []);
 
     return (
-        <Box
-            id="calendar-root"
-            marginY={1}
-            borderRadius={1}
-            flexGrow={1}
-            height={0}
-            display="flex"
-            flexDirection="column"
-        >
+        <Box id="calendar-root" borderRadius={1} flexGrow={1} height={0} display="flex" flexDirection="column">
             <CalendarToolbar
                 currentScheduleIndex={currentScheduleIndex}
                 toggleDisplayFinalsSchedule={toggleDisplayFinalsSchedule}
--- a/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
+++ b/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
@@ -69,8 +69,6 @@ interface ScheduleCalendarProps {
 }
 
 export default function ScheduleCalendar(_props?: ScheduleCalendarProps) {
-    // const { isMobile } = props;
-
     const [anchorEl, setAnchorEl] = useState<HTMLElement | null>(null);
     const [showFinalsSchedule, setShowFinalsSchedule] = useState(false);
     const [courseInMoreInfo, setCourseInMoreInfo] = useState<CalendarEvent | null>(null);
@@ -162,8 +160,6 @@ export default function ScheduleCalendar(_props?: ScheduleCalendarProps) {
 
     const events = getEventsForCalendar();
     const hasWeekendCourse = events.some((event) => event.start.getDay() === 0 || event.start.getDay() === 6);
-
-    // const calendarStyling = isMobile ? { height: `calc(100% - 55px)` } : { height: `calc(100vh - 104px)` };
 
     const calendarTimeFormat = isMilitaryTime ? 'HH:mm' : 'h:mm A';
     const calendarGutterTimeFormat = isMilitaryTime ? 'HH:mm' : 'h A';
--- a/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
+++ b/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
@@ -65,11 +65,11 @@ const AntAlmanacEvent = ({ event }: { event: CalendarEvent }) => {
 };
 
 interface ScheduleCalendarProps {
-    isMobile: boolean;
+    isMobile?: boolean;
 }
 
-export default function ScheduleCalendar(props: ScheduleCalendarProps) {
-    const { isMobile } = props;
+export default function ScheduleCalendar(_props?: ScheduleCalendarProps) {
+    // const { isMobile } = props;
 
     const [anchorEl, setAnchorEl] = useState<HTMLElement | null>(null);
     const [showFinalsSchedule, setShowFinalsSchedule] = useState(false);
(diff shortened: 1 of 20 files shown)


apps/antalmanac/src/components/SharedRoot.tsx
```diff
@@ -16,53 +15,46 @@ import { useTabStore } from '$stores/TabStore';
...
-    const isDark = useThemeStore((store) => store.isDark);
+/**
+ * Information about the tab navigation buttons.
+ *
+ * Each button should be associated with a different aspect of schedule management.
+ */
+type ScheduleManagementTabInfo = {
+    /**
+     * Label to display on the tab button.
+     */
```
KevinWu098, line 26: nit (non-blocking): this comment on label, href, and icon might be overly descriptive for a more or less self-descriptive property 

```suggestion
```

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
KevinWu098, line 1: *was discussed over Discord*

issue: calendar height is not based on `dvh`, so tabs are overflowed off the screen (inaccessible), and portions of the calendar aren't inaccessible too (without scrolling and holding)

apps/antalmanac/src/components/SharedRoot.tsx
```diff
@@ -1,7 +1,6 @@
-import { Box, Paper, Tab, Tabs, Typography } from '@material-ui/core';
-import { Event, FormatListBulleted, MyLocation, Search } from '@material-ui/icons';
-import { GlobalStyles } from '@mui/material';
-import { Suspense, lazy, useEffect } from 'react';
+import { Event, FormatListBulleted, MyLocation, Search } from '@mui/icons-material';
+import { Box, GlobalStyles, Paper, Stack, Tab, Tabs, Typography, useMediaQuery, useTheme } from '@mui/material';
```
KevinWu098: issue (style): tab text color is now blue when selected, rather the original white (contrast problem)

Screenshot 2024-08-20 at 9 37 18 AM

apps/antalmanac/src/components/SharedRoot.tsx
```diff
@@ -83,97 +75,235 @@ const tabs: Array<TabInfo> = [
...
+            return null;
+    }
 }
 
-const SharedTabs = ({ style, mobile }: SharedTabsProps) => {
+/**
+ * List of interactive tab buttons with their accompanying content.
+ * Each tab's content has functionality for managing the user's schedule.
+ */
+export default function ScheduleManagement() {
```
KevinWu098, line 205: nit: super naming nit, but I feel like `scheduleManagement` is not really descriptive of what's going on here with the tabs

Any reason why we can't prefer `NavigationTabs` or something that's a little closer to what the component *is*

KevinWu098 (changes requested):
Broadly speaking, great PR on a much needed update. Separate even from moving tabs from top to bottom, really love that we're refactoring out our hardcoded styling values (which I'm definitely guilty of perpetuating).

Code-wise, I only have nits as to naming and comments. But since this is a "preview", it definitely has some UI problems (including, but not limited to):
1. Calendar is broken -- tabs don't display right (height is probably not `100dvh`)
2. Search Result is broken -- tabs don't display right (height is probably not `100dvh`)
3. Padding is a little much (maybe `padding={0.5}` would be better than `1`)
4. On iPhones, although `dvh` is fantastic, we may want to add padding to the bottom as the iPhone "navigation bar" does still overlap on PWA

apps/antalmanac/src/components/SharedRoot.tsx
ap0nia: Neither is `SharedRoot` 🤓 

IMO, `NavigationTabs` sounds generic and similar to something you might import from a component library.

I'm not really a big fan of calling them "tabs" because that's not really what this component is. It's multiple screens that can be navigated between via tabs. i.e. It's both the tabs and the tab contents.

In the end, I think the primary purpose of this component is to allow you to manipulate your schedule through its various screens.

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
```diff
@@ -163,7 +163,7 @@ export default function ScheduleCalendar(props: ScheduleCalendarProps) {
     const events = getEventsForCalendar();
     const hasWeekendCourse = events.some((event) => event.start.getDay() === 0 || event.start.getDay() === 6);
 
-    const calendarStyling = isMobile ? { height: `calc(100% - 55px)` } : { height: `calc(100vh - 104px)` };
+    // const calendarStyling = isMobile ? { height: `calc(100% - 55px)` } : { height: `calc(100vh - 104px)` };
```
KevinWu098:
```suggestion
```
nit: remove comment

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
```diff
@@ -65,11 +65,11 @@ const AntAlmanacEvent = ({ event }: { event: CalendarEvent }) => {
...
 
 interface ScheduleCalendarProps {
-    isMobile: boolean;
+    isMobile?: boolean;
 }
 
-export default function ScheduleCalendar(props: ScheduleCalendarProps) {
-    const { isMobile } = props;
+export default function ScheduleCalendar(_props?: ScheduleCalendarProps) {
+    // const { isMobile } = props;
```
KevinWu098: nit: remove comment

apps/antalmanac/src/routes/Home.tsx
```diff
@@ -1,55 +1,79 @@
...
+                sizes={[45, 55]}
+                minSize={100}
+                expandToMin={false}
+                gutterSize={10}
+                gutterAlign="center"
+                snapOffset={30}
+                dragInterval={1}
+                direction="horizontal"
+                cursor="col-resize"
+                style={{ display: 'flex', flexGrow: 1, height: '100dvh' }}
```
KevinWu098:
```suggestion
                style={{ display: 'flex', flexGrow: 1, height: 'calc(100dvh - 50px)' }}
```
issue: we're not accounting for the header height

Screenshot 2024-09-30 at 2 53 06 PM

KevinWu098 (changes requested):
Looks great, left two comments regarding styling:

1. We're not accounting for header height, so 100dvh is too tall
2. Our Mui v5 palette isn't correct, and it makes the tabs hard to read. Without expanding this PR to cover palette updates, I'd suggest using v4's tabs

apps/antalmanac/src/routes/Home.tsx
ap0nia: Suggestion is wrong. Calc is terrible because we're guessing what the height of the header is, instead of relying on the browser's exact knowledge. Solution is to wrap the header and content with a different container that is 100dvh, and then ask the content to fill up the rest of the height after the header.

KevinWu098 (approved):
Beautiful, fabulous, tremendous!

I don't want to hold up this PR over a slight contrast problem (I'm happy to submit a PR following this one fixing it), so I'm going to approve

Feel free to merge or to rerequest review after looking at the tab contrast stuff, both are fine by me at this point since I think this PR is a big win for mobile users that I don't want held up

MinhxNguyen7:
@KevinWu098, I'm going to take your word for it, but @ap0nia, do you know why the tests are failing?

KevinWu098:
> @KevinWu098, I'm going to take your word for it, but @ap0nia, do you know why the tests are failing?

looks like this PR is the source: https://github.com/icssc/AntAlmanac/pull/1004

ap0nia:
Oh, let me try tweaking the test to use real building numbers.

MinhxNguyen7:
| looks like this PR is the source: #1004 
Oops, that's my bad. I must've ignored it because the deploy doesn't work either (for external contributors)

ap0nia:
I handled possibly nullish values and updated the test snapshots.

MinhxNguyen7 (approved):
LGTM!

MinhxNguyen7 (changes requested):
Oops I didn't see that @KevinWu098 had a comment that hasn't been resolved, and it does actually matter.

ap0nia:
Which one? The white color on the tabs? I already fixed that, forgot to resolve it. Any other theming issues are due to the differences in MUI version, which should be resolved separately.

Oh wait, need to fix map issue.

ap0nia:
Wait, is the map broken on live? It looks weird on my side ...

ap0nia:
Ok, I think it was because the CSS from leaflet wasn't imported. Should be sorted out now. For reference, this is what I see on my screen for the live site:

image

MinhxNguyen7:
> The white color on the tabs? I already fixed that, forgot to resolve it. Any other theming issues are due to the differences in MUI version, which should be resolved separately.

If you updated the MUI version, and it changed the color undesirably, than the PR would be atomic if you changed the code such that that does not occur.

MinhxNguyen7 (approved):
LGTM!

PR #1003, opened 2024-09-17
fix(download.ts): set events as PRIVATE by default

### LINKED ISSUE #963: Google Calendar makes `.ics`-imported events public
When I add the events from AntAlmanac to my Google Calendar, it automatically makes the events public. I don't know if this is an AA issue or a GC issue, but I think it raises serious privacy concerns.

calejvaldez:
## Summary

I opened #963 a few months ago when I noticed that Google Calendar events imported from AntAlmanac were public for anybody with my UCI email to see. This concerned me because this meant people would know exactly where I'd be and when.

This fixes the issue by setting the `classification` attribute to `PRIVATE` on the events in the `calendarEvents` array, making the event private.

## Test Plan

I tested this by downloading the `.ics` file using a development environment of AntAlmanac with my changes and importing it to my Google Calendar, and successfully saw that it was a private event by default.

Photo of a private event imported from AntAlmanac.

... no I'm not taking ICS 6B. I can *never* go through that again...

## Issues

Closes #963

### DIFF
--- a/apps/antalmanac/src/lib/download.ts
+++ b/apps/antalmanac/src/lib/download.ts
@@ -279,6 +279,7 @@ export function getEventsFromCourses(
                 // location: `${location.building} ${location.room}`,
                 start: firstClassStart,
                 end: firstClassEnd,
+                classification: 'PRIVATE',
                 recurrenceRule: rrule,
             };
             return customEvent;
@@ -302,6 +303,7 @@ export function getEventsFromCourses(
                             description: `Final Exam for ${courseTitle}`,
                             start: finalStart,
                             end: finalEnd,
+                            classification: 'PRIVATE',
                         };
                     } else {
                         const classStartDate = getClassStartDate(term, days);
@@ -324,6 +326,7 @@ export function getEventsFromCourses(
                             location: `${location.building} ${location.room}`,
                             start: firstClassStart,
                             end: firstClassEnd,
+                            classification: 'PRIVATE',
                             recurrenceRule: rrule,
                         };
                     }


KevinWu098 (approved):
LGTM, thanks for this change! 💪

Before merging though, I'm kind of curious as to what effect this has on a per-user / per-event use case. Just because an event is public visibility, doesn't the calendar which it resides need to be shared? I'm mostly curious as to if there's a specific case where specifying the visibility to private would be important and/or common.

MinhxNguyen7:
@calejvaldez what do you say about @KevinWu098 's question about the utility? If you share your calendar, wouldn't you want the other person to know when you have class by default? It's not like when you have class is any more private than other events on the calendar, right?

calejvaldez:
@MinhxNguyen7 @KevinWu098 That's true, though my biggest concern was that UCI sets our calendars as shared by default without notifying the user. If we don't specify an event's `classification`, then it defaults to the UCI default, which is public. As of now, every event that's imported from AntAlmanac uses UCI's default public setting, making it so that all we need to know where somebody will be, and at what time, is their UCI email address.

Unless users specifically go into their Calendar settings and change the default calendar visibility to private, anything they add to their calendar will be shared. In theory, using the default makes sense, though IMO it makes sense if a user is aware their events are public, which the few people I've pointed this out to said they weren't aware.

MinhxNguyen7:
I think that setting the visibility to the account's default makes sense, regardless of whether that behavior is desired by the user. Put otherwise, it's *not our problem*.

Unless there's substantial opinion otherwise, I'm going to close this for now. Sorry to not take your work, and I appreciate your investigation into this.

PR #1004, opened 2024-09-17
feat(download.ts): add location attribute to custom events

calejvaldez:
## Summary

I noticed a TODO asking to implement a location for custom events once #249 was finished. I noticed that the issue was completed so I went ahead and implemented it.

Though... it looks a bit funky because `CustomEvent.building` saves an ID instead of the actual building name. This happens in `CustomEventDialog.tsx` because it needs a key for the list that's needed in `BuildingSelect`. Though I noticed that `BuildingSelect` gets the name of the building using the format I used here.

## Test Plan

I tested this by creating a custom event and downloading a `.ics`. I then imported it to Google Calendar and saw that the location of the event was successfully added.

Google Calendar event displaying a successfully added location to an event.

## Issues

N/A, but addresses this line in the code:

https://github.com/icssc/AntAlmanac/blob/559b615c2d92ea5c5fa3965c6360abaefe23c830/apps/antalmanac/src/lib/download.ts#L278-L279

## Future Follow-Up
- #1010

### DIFF
--- a/apps/antalmanac/src/lib/download.ts
+++ b/apps/antalmanac/src/lib/download.ts
@@ -7,6 +7,7 @@ import { notNull } from './utils';
 import { openSnackbar } from '$actions/AppStoreActions';
 import { CustomEvent, FinalExam } from '$components/Calendar/CourseCalendarEvent';
 import analyticsEnum, { logAnalytics } from '$lib/analytics';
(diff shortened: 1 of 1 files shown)


MinhxNguyen7 (approved):
LGTM!

Thanks for your contribution!

PR #1006, opened 2024-09-20
fix: default colors deplete too rapidly

### LINKED ISSUE #647: Default Colors deplete too Rapidly
## Issue
After the initial 7 colors are used in a term -- meaning that 7 _unique***_ courses are in a given term (not just a particular schedule, this is done for consistency. See #635 if curious) --, we then select a random color from the palette (See #644).

This is... mildly acceptable at best..., but can create situations where users end up with two same-colored courses back-to-back. Google Chrome colors tab groups linearly (see below), with colors repeating once they're all expended. Linearly here means that the order that colors are used is always the same.

Screenshot 2023-11-30 at 12 31 40 PM

Linear assignment is also preferable to not just random assignment, but also adding new colors:
 
Screenshot 2023-11-30 at 12 25 49 PM

As you can tell, with all 7 colors utilized -- alongside their lighter variants for other course types -- we're already hard-pressed on contrast and differentiability

_***Unique, in this instance means ICS 6B vs ICS 32A. Alternatively, it means that a 6B lecture and a 6B discussion are not unique. Course Name dictates uniqueness_
## Good First Taskers
Relevant code can be found in `scheduleHelpers.ts`, specifically the `getColorForNewSection` function. The color of the 8th course and onwards is set by this block of code:

```ts
    // If there are no existing sections with the same course title, generate a new color. If we run out of unique colors, return a random one that's been used already.
    return (
        defaultColors.find((materialColor) => !usedColors.has(materialColor)) ||
        defaultColors[Math.floor(Math.random() * defaultColors.length)]
    );
```

Your task is to essentially take this from random assignment to linear assignment.

## Old Issue Writeup
_This was the original solution, advocating for additional colors; it's left for documentation purposes (I guess), but it's not relevant to the issue any longer_

Following the merging of #635, #644 and #646, the set of Used Colors for courses is now drawn from courses which share the same term as the newCourse (i.e. the course that is being added). (Prior, used colors only came from the same schedule)

Because of this, the seven base colors are being used up too quickly. New colors should be drawn from the MUI template like the current ones we've got: 

`const defaultColors = [blue[500], pink[500], purple[500], green[500], amber[500], deepPurple[500], deepOrange[500], ];`

matt-franklin225:
## Summary
Adjusted color selection to work by linear assignment from the 8th course onward. This is done by determining the color of the most recent class added and choosing the color directly after it in the defaultColors list (looping back around to the first color when needed).
## Test Plan
- Ensure that the assignment works properly when classes are deleted out of order.
- Ensure assignment works when users select custom colors for courses (this in particular poses an issue).
## Issues

Closes #647 

## Future Followup
There can still be some imbalance in colors if many classes are added and some deleted--if keeping all the colors even is a priority we could auto-update the colors of previously added classes, but that may be unnecessary or overkill. Can also add to the original 7-color palette if desired.

### DIFF
--- a/apps/antalmanac/src/stores/scheduleHelpers.ts
+++ b/apps/antalmanac/src/stores/scheduleHelpers.ts
@@ -167,23 +167,20 @@ export function getColorForNewSection(newSection: ScheduleCourse, sectionsInSche
         (course) => course.section.sectionType === newSection.section.sectionType
     );
 
-    const usedColors = new Set(sectionsInSchedule.map((course) => course.section.color));
+    const usedColors = sectionsInSchedule.map((course) => course.section.color);
+    const lastDefaultColor = usedColors.findLast((materialColor) =>
+        (defaultColors as string[]).includes(materialColor)
+    ) as unknown as (typeof defaultColors)[number];
 
     // If the same sectionType exists, return that color
     if (existingSectionsType.length > 0) return existingSectionsType[0].section.color;
 
     // If the same courseTitle exists, but not the same sectionType, return a close color
-    if (existingSections.length > 0) return generateCloseColor(existingSections[0].section.color, usedColors);
+    if (existingSections.length > 0) return generateCloseColor(existingSections[0].section.color, new Set(usedColors));
 
-    // If there are no existing sections with the same course title, generate a new color. If we run out of unique colors, return the next color after the last one in use, looping after reaching the end.
+    // If there are no existing sections with the same course title, generate a new color. If we run out of unique colors, return the next color up after the last default color in use, looping after reaching the end.
     return (
-        defaultColors.find((materialColor) => !usedColors.has(materialColor)) ||
-        defaultColors[
-            (defaultColors.indexOf(
-                sectionsInSchedule[sectionsInSchedule.length - 1].section.color as (typeof defaultColors)[number]
-            ) +
-                1) %
-                7
-        ]
+        defaultColors.find((materialColor) => !usedColors.includes(materialColor)) ||
+        defaultColors[(defaultColors.indexOf(lastDefaultColor) + 1) % 7]
     );
 }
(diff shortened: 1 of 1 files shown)


matt-franklin225:
Still need to adjust code so that it works properly when users select custom colors for their classes.

apps/antalmanac/src/stores/scheduleHelpers.ts
```diff
@@ -175,9 +175,15 @@ export function getColorForNewSection(newSection: ScheduleCourse, sectionsInSche
...
+    // If there are no existing sections with the same course title, generate a new color. If we run out of unique colors, return the next color after the last one in use, looping after reaching the end.
     return (
         defaultColors.find((materialColor) => !usedColors.has(materialColor)) ||
-        defaultColors[Math.floor(Math.random() * defaultColors.length)]
+        defaultColors[
+            (defaultColors.indexOf(
+                sectionsInSchedule[sectionsInSchedule.length - 1].section.color as (typeof defaultColors)[number]
+            ) +
+                1) %
+                7
```
johwuy: I think that instead of hardcoding the length of `defaultColors`, it is better to use the `defaultColors.length` directly or as a `const`, in case more default colors are added, the code will automatically adapt.
Otherwise, looks good.

apps/antalmanac/src/stores/scheduleHelpers.ts
```diff
@@ -167,23 +167,20 @@ export function getColorForNewSection(newSection: ScheduleCourse, sectionsInSche
...
-        defaultColors.find((materialColor) => !usedColors.has(materialColor)) ||
-        defaultColors[
-            (defaultColors.indexOf(
-                sectionsInSchedule[sectionsInSchedule.length - 1].section.color as (typeof defaultColors)[number]
-            ) +
-                1) %
-                7
-        ]
+        defaultColors.find((materialColor) => !usedColors.includes(materialColor)) ||
+        defaultColors[(defaultColors.indexOf(lastDefaultColor) + 1) % 7]
```
johwuy: I think that instead of hardcoding the length of `defaultColors`, it is better to use the `defaultColors.length` directly or as a `const`, in case more default colors are added, the code will automatically adapt.

matt-franklin225: That's a good point, I'll fix that.

matt-franklin225: Just pushed a commit that fixes this issue.

MinhxNguyen7 (approved):
Seems about right

PR #1008, opened 2024-09-29
Update required pnpm version

### LINKED ISSUE #1007: Project not allowing pnpm 9
## Description
- With pnpm 9's release, the latest version doesn't pass the dependency check (pnpm ^8.0.0) because this doesn't cross major version barriers.
- I've tried to change the requirement to the latest version, which works with no further changes. We just need to put this into a PR.

MinhxNguyen7:
## Summary
- Update manifests to allow for the usage of `pnpm` 9.

## Test Plan
- Update `pnpm` to `9.x.x`. (Check with `pnpm -v`)
- Try to update dependencies with `pnpm install`
- Run app with `pnpm run start:aa`

## Issues

Closes #1007

### DIFF
--- a/apps/cdk/package.json
+++ b/apps/cdk/package.json
@@ -34,7 +34,6 @@
     },
     "packageManager": "pnpm@8.6.12",
     "engines": {
-        "pnpm": "^8.0.0",
         "node": ">=18"
     }
 }
--- a/.github/actions/setup-node-and-pnpm/action.yml
+++ b/.github/actions/setup-node-and-pnpm/action.yml
@@ -16,7 +16,6 @@ runs:
         - name: Install pnpm
           uses: pnpm/action-setup@v2
           with:
-              version: 8
               run_install: false
 
         - name: Get pnpm store directory
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -27,8 +27,6 @@ jobs:
 
             - name: Install pnpm
               uses: pnpm/action-setup@v2
-              with:
-                  version: 8.6.0
 
             - name: Setup Node.js with pnpm cache
               uses: actions/setup-node@v3
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
         "turbo": "latest",
         "vitest": "^0.34.4"
     },
-    "packageManager": "pnpm@8.6.0",
+    "packageManager": "pnpm@9.11.0",
     "engines": {
         "pnpm": "^9.0.0",
         "node": ">=16"
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     },
     "packageManager": "pnpm@8.6.0",
     "engines": {
-        "pnpm": "^8.0.0",
+        "pnpm": "^9.0.0",
         "node": ">=16"
     }
 }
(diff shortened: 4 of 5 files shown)


apps/cdk/package.json
```diff
@@ -34,7 +34,6 @@
     },
     "packageManager": "pnpm@8.6.12",
     "engines": {
-        "pnpm": "^8.0.0",
         "node": ">=18"
     }
```
ap0nia, line 38: This is good; all references to packageManager, engines, etc. should be removed from sub-projects because only the root `package.json` should specify it.

ap0nia:
You should also handle this file in the PR: https://github.com/icssc/AntAlmanac/blob/main/.github/actions/setup-node-and-pnpm/action.yml#L16

Upgrade all the actions to their latest version, and don't specify the pnpm version in the `pnpm/action-setup` because it should pick it up from the root `package.json`.

```yaml
# yaml-language-server: $schema=https://raw.githubusercontent.com/SchemaStore/schemastore/master/src/schemas/json/github-action.json
---
name: Setup Node.js and pnpm

description: Sets up Node.js and pnpm + pnpm cache. Does **not** install dependencies.

runs:
    using: composite

    steps:
        - name: Setup Node.js environment
          uses: actions/setup-node@v4

        - name: Install pnpm
          uses: pnpm/action-setup@v4

        - name: Get pnpm store directory
          id: pnpm-cache
          shell: bash
          run: echo "STORE_PATH=$(pnpm store path)" >> $GITHUB_OUTPUT

        - name: Setup pnpm cache
          uses: actions/cache@v4
          with:
              path: ${{ steps.pnpm-cache.outputs.STORE_PATH }}
              key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
              restore-keys: ${{ runner.os }}-pnpm-store-
```

ap0nia:
Could you also update the [instructions on the README](https://github.com/icssc/AntAlmanac/blob/main/README.md?plain=1#L102) to encourage using `corepack`?

Documentation for the built-in package is [here](https://www.npmjs.com/package/corepack).

The `corepack enable` command modifies the PATH so that way the system can find the package managers that `corepack` installs. Then `corepack install` will read the `packageManger` field in the `package.json` and install the correct version of pnpm. Then `pnpm` should be available from the CLI. The downloaded package manager will be an executable located somewhere in corepack's cache, which it added to the PATH when you enabled it earlier.

````md
2. Install `pnpm`. This is our package manager of choice for this project.
   It's responsible for installing, uninstalling, and keeping track of the app's dependencies.
  ```sh
  corepack enable
  corepack install
  ```
````

The benefit of using corepack is that it can manage multiple versions of package managers, which is useful if you have projects that specify different ones. Additionally, there are no dependencies and it'll automatically install the right version of whatever package manager we specify, so there shouldn't be issues in the future with incorrect versions inhibiting local development.

MinhxNguyen7:
@ap0nia do you want to just commit that on here? I don't use corepack, so I don't know

ap0nia:
I won't be able to approve if I make the most recent commit, but I could push a commit and force merge with admin permissions :^)

MinhxNguyen7:
@ap0nia yeah gp for it. I can override branch protection

MinhxNguyen7:
I'm just going to merge this so that people can contribute to the project. Feel free to throw up a PR that includes the corepack instructions

PR #1012, opened 2024-10-16
feat: changing pop-over x button for hover :) and adding  padding to the popover to make it more visible

### LINKED ISSUE #1011: Making the X button on Reactour bigger with padding and better UX?
image

The button is getting in the way of the text, and is quite small to click on, was wondering if it could be made bigger by adding padding potentially and if it could have an onhover effect to like help the user notice when its being hovered on?

Priyansh4444:
## Summary
Making the UX just a bit better by creating a transition for the hover on the button

## Test Plan
Running the frontend 

## Issues
- Currently the 4th item on the path goes a bit too left on the screen
- Some steps also go a bit off the page

Closes #1011

### DIFF
--- a/apps/antalmanac/src/App.tsx
+++ b/apps/antalmanac/src/App.tsx
@@ -1,9 +1,8 @@
 import './App.css';
 
-import { CloseOutlined } from '@material-ui/icons';
-import { StylesObj, TourProvider } from '@reactour/tour';
+import { TourProvider } from '@reactour/tour';
 import { SnackbarProvider } from 'notistack';
-import { MouseEventHandler, useEffect, useState } from 'react';
+import { useEffect } from 'react';
 import ReactGA4 from 'react-ga4';
 import { createBrowserRouter, RouterProvider } from 'react-router-dom';
 
@@ -14,6 +13,7 @@ import AppThemev5Provider from './providers/Themev5';
 import Feedback from './routes/Feedback';
 import Home from './routes/Home';
 
+import { CloseButton } from '$components/buttons/CloseButton';
 
 const BrowserRouter = createBrowserRouter([
     {
@@ -50,7 +50,7 @@ export default function App() {
                     <TourProvider
                         steps={[] /** Will be populated by Tutorial component */}
                         padding={5}
-                        components={{ Close: (props) => <CloseButton {...props} /> }}
+                        components={{ Close: (props) => <CloseButton {...props} styles={undefined} /> }}
                         styles={{
                             maskArea: (base, _) => ({
                                 // The highlighted area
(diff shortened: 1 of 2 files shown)


MinhxNguyen7 (changes requested):
Thanks for your contribution! 

Code-wise, the button should not be defined with the router; that just doesn't make logical sense. it should go in `components/buttons`. Also, since we drink the MUI Kool-Aid, we shouldn't have to implement such a simple button. Take a look at [this](https://mui.com/material-ui/react-button/). We should be able to just define the button in-line with some customization. Do say though, if that is not fit-for-purpose.

I have a few things to ask in terms of documentation. These are somewhat matters of pedantry, but I think it's important to maintain standards.

Could you please give more information on the test plan? Yes, we run the front end, but what should we expect to see? You also say that there are some problems. Was that introduced by this PR, or was it there before? If the former is the case, why can't it be addressed? If the latter is the case, you should put that in a "Future Follow-Up" section and add it as an issue.

Also, the "Issues" section is for related GitHub issues. The issues that you have there should go in the summary.

Priyansh4444:
Thank you for the feedback I will keep this in mind going on when I contribute to other repositories as well!
I think it is fit to make its own component for the button.
There are two ways to edit the button, either by adding it as a component in styles or making a react functional component for it.

I think the MUI button is the best way to approach the matter since the onHover Effect then can be achieved!
though another solution could potentially be manaully taking the ``.react__tour_close`` class and adding it to App.css

I am still a bit unsure, how to test frontend other than visually, but in terms of expected result having the tourguide is partially going out of bounds of the screen and therefore cannot be seen I am unsure how to control where the tourguide goes. 
The problems I mentioned happens before the PR as well. I will try and fix these issues when I get a bit more time since currently I am a bit busy!


Future Follow Up:
- Fixing the TourGuide So it does not go out of bounds
- Making the 4th step of the tour guide happen to the right of the calendar instead of left (it was on the right before the PR)

Priyansh4444:
I have no idea how to test the front-end other than visually, and visually speaking I think its working and does a good job formatting the X button, any thoughts and guidance would be appreciated!

KevinWu098:
Closing this PR on staleness. If you feel strongly about the PR, please feel free to open the PR again.

My personal recommendation would be to take a simpler approach:

Screenshot 2025-02-01 at 4 50 17 PM

By just adjusting the tourguide styles slightly, we can prevent the textwrap and position the button in a much clearer, overlap-free position. 

```css
// reactour__popover
padding: 24px

// reactour__close-button
top: 8px;
right: 8px;
width: 8px;
height: 8px;
```

This isn't saying that a custom component isn't an equally valid solution, but because this PR regresses the tour in terms of introducing clipping and a somewhat unruly component (e.g. LOC dedicated to styling, see below), we can't accept the PR in the direction that it's currently headed.

```jsx
                                marginLeft: 10,
                                marginRight: 10,
                                marginTop: 10,
                                marginBottom: 10,
                                paddingTop: 20,
                                paddingBottom: 20,
                                paddingLeft: 35,
                                paddingRight: 35,
```

PR #1013, opened 2024-10-24
feat: encode/decode advanced search in URL

### LINKED ISSUE #558: Make every class search parameter sharable

KevinWu098:
## Summary
1. Encode advanced search into the URL
2. Decode params and apply accordingly
    - Open advanced search if params exist
  
chrome-capture-2024-9-24

## Test Plan
1. Test that params are correctly applied, and when used for search, are correct

## Issues
- Closes #558 

## Future Followup
- Have a `/search` route which executes the search immediately.

### DIFF
--- a/apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch.tsx
+++ b/apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch.tsx
@@ -103,24 +103,42 @@ class UnstyledAdvancedSearchTextFields extends PureComponent<
     };
 
     handleChange = (name: string) => (event: ChangeEvent<{ checked?: boolean; name?: string; value: unknown }>) => {
+        const stateObj = { url: 'url' };
+        const url = new URL(window.location.href);
+        const urlParam = new URLSearchParams(url.search);
+
         if (name === 'online') {
             if (event.target.checked) {
                 this.setState({ building: 'ON', room: 'LINE' });
                 RightPaneStore.updateFormValue('building', 'ON');
                 RightPaneStore.updateFormValue('room', 'LINE');
+
+                urlParam.set('building', 'ON');
+                urlParam.set('room', 'LINE');
             } else {
                 this.setState({ building: '', room: '' });
                 RightPaneStore.updateFormValue('building', '');
                 RightPaneStore.updateFormValue('room', '');
+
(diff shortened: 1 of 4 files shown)


MinhxNguyen7:
Technically speaking, this is a valid PR, but it should be a little more. It should close #558. At the very least, all the params, not just the advanced search, should be sharable.

Furthermore, there should be a `/search` route where it actually executes the search, falling to the search screen conditionally if the params are insufficient. Allowing people to share or bookmark the link to a search would be very useful. That would be quite a bit of additional work, but you should do that if you're up for it.

KevinWu098:
> Technically speaking, this is a valid PR, but it should be a little more. It should close #558. At the very least, all the params, not just the advanced search, should be sharable.
> 
> Furthermore, there should be a `/search` route where it actually executes the search, falling to the search screen conditionally if the params are insufficient. Allowing people to share or bookmark the link to a search would be very useful. That would be quite a bit of additional work, but you should do that if you're up for it.

Just pushed a commit that will search if it detects params; also, all the params as of this PR should be sharable -- #546 encodes manual search params

MinhxNguyen7 (changes requested):
Mostly looks good, but it looks like it *sometimes* executes the search, and sometimes incorrectly. See this case: https://staging-1013.antalmanac.com/?GE=GE-3&instructor=ihler

KevinWu098:
> Mostly looks good, but it looks like it _sometimes_ executes the search, and sometimes incorrectly. See this case: https://staging-1013.antalmanac.com/?GE=GE-3&instructor=ihler

Hm, I might not have tried it out enough, but I'm not seeing any unexpected behavior (aside from the Grades toast error, though I feel like that might be more related to us going from F24 -> W25?)

MinhxNguyen7:
That's a GE-3 search, and the results are upper divs, which are not GEs

KevinWu098:
> That's a GE-3 search, and the results are upper divs, which are not GEs

Thanks for the catch -- was reading ge from the url incorrectly :)

MinhxNguyen7 (approved):
I'm not 100% convinced considering we've only tested it a little, but it can't break UX that badly.

PR #1014, opened 2024-10-24
Correctly link internal packages in package.json files

MinhxNguyen7:
## Summary
- Currently, the only reason that our imports keep working for internal packages (`peterportal-schemas` & `antalmanac-types`) is because of the linking in lockfiles. 
- The linking syntax in the lockfiles is incorrect.
- This PR fixes the syntax and re-generates the lockfile  

## Test Plan
- Check that the deployment works.
- Check that `pnpm install` works on a fresh clone of the project.

### DIFF
--- a/apps/backend/package.json
+++ b/apps/backend/package.json
@@ -10,7 +10,7 @@
         "lint": "eslint --fix src"
     },
     "dependencies": {
-        "@packages/antalmanac-types": "*",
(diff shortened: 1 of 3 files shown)


adcockdalton (approved):
I created a fresh clone and installation straight from this branch; behavior is the same on localhost for main as-is and for this branch.

lgtm

PR #1015, opened 2024-10-27
Added handling for custom events download

### LINKED ISSUE #1010: Duplicate custom events when exporting to ics
## Description
- When a custom event with multiple days is exported in an `ics` file, multiple copies of it are created.
- Below is the relevant excerpt from an exported file containing one custom event repeating on Sunday and Monday. 
```
SUMMARY:
DTSTAMP:20241002T004655Z
DTSTART;TZID=America/Los_Angeles:20240929T103000
DTEND;TZID=America/Los_Angeles:20240929T153000
RRULE:FREQ=WEEKLY;BYDAY=SU,MO;INTERVAL=1;COUNT=20
END:VEVENT
BEGIN:VEVENT
UID:5ZX0Zax9m6HBFzZzDHfbZ
SUMMARY:
DTSTAMP:20241002T004655Z
DTSTART;TZID=America/Los_Angeles:20240929T103000
DTEND;TZID=America/Los_Angeles:20240929T153000
RRULE:FREQ=WEEKLY;BYDAY=SU,MO;INTERVAL=1;COUNT=20
```

## Reproduction
- Create a custom event repeating on multiple days.
- Download the `ics` file.
- Inspect it manually or import it into a calendar app.

## Diagnosis
- It's probably to do with how multiple events are created for class sections that meet multiple times.
    - I think this is the case because class sections can meet in different places throughout the week.
- The fix likely involves either creating only one event repeating on multiple days for custom events (preferable) or creating one custom event for each day.

vinnyho:
## Summary
I added `const customEventIDs = new Set();`, where it'll be populated with `const customEventID = event.customEventID;`. Each Custom Event object was created with a specific ID, even if it was on multiple days it'll be the same, so then when it maps to download, it shouldn't repeat the custom events.
This is a custom event created for Tuesday and Wednesday.
~~~
SUMMARY:FIX TEST
DTSTAMP:20241027T205447Z
DTSTART;TZID=America/Los_Angeles:20241001T103000
DTEND;TZID=America/Los_Angeles:20241001T153000
RRULE:FREQ=WEEKLY;BYDAY=TU,WE;INTERVAL=1;COUNT=20
END:VEVENT
END:VCALENDAR       
~~~
This is what it appears like on Google Calendar

Closes #1010

### DIFF
--- a/apps/antalmanac/src/lib/download.ts
+++ b/apps/antalmanac/src/lib/download.ts
@@ -257,6 +257,7 @@ export function getEventsFromCourses(
     events = AppStore.getEventsWithFinalsInCalendar(),
     term = getDefaultTerm(events).shortName
 ): EventAttributes[] {
+    const customEventIDs = new Set();
     const calendarEvents = events.flatMap((event) => {
         if (event.isCustomEvent) {
             // FIXME: We don't have a way to get the term for custom events,
@@ -270,6 +271,11 @@ export function getEventsFromCourses(
                 { hour: start.getHours(), minute: start.getMinutes() },
                 { hour: end.getHours(), minute: end.getMinutes() }
             );
+            const customEventID = event.customEventID;
+            if (customEventIDs.has(customEventID)) {
+                return [];
+            }
+            customEventIDs.add(customEventID);
             const customEvent: EventAttributes = {
                 productId: 'antalmanac/ics',
                 startOutputType: 'local' as const,


adcockdalton (approved):
Tested with/without changes for multi-day custom event with class. Tested also with multiple custom events with identical attributes; behavior still as expected. Code lgtm

PR #1017, opened 2024-11-02
chore: update term data with winter 2025

KevinWu098:
## Summary
1. Add winter 2025
## Test Plan
1. tests don't fail

### DIFF
--- a/apps/antalmanac/src/lib/termData.ts
+++ b/apps/antalmanac/src/lib/termData.ts
@@ -29,6 +29,7 @@ class Term {
  * Months are 0-indexed
  */
 const termData = [
+    new Term('2025 Winter', '2025 Winter Quarter', [2025, 0, 6], [2025, 2, 21]),
     new Term('2024 Fall', '2024 Fall Quarter', [2024, 8, 26], [2024, 11, 7]),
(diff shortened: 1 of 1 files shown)


apps/antalmanac/src/lib/termData.ts
```diff
@@ -29,6 +29,7 @@ class Term {
  * Months are 0-indexed
  */
 const termData = [
+    new Term('2025 Winter', '2025 Winter Quarter', [2025, 0, 6], [2025, 2, 21]),
```
alexanderl19: Not sure what the finals date is used for, but comment on L28 mentions it should be the first day (sat). This would correspond to Mar 15 (`[2025, 2, 15]`).

alexanderl19 (changes requested):

apps/antalmanac/src/lib/termData.ts
alexanderl19: Feel free to dismiss my review to merge this PR if it's not important.

KevinWu098: smh no trust -- good catch xD

alexanderl19 (approved):

alexanderl19:
Not sure what's blocking this PR from being merged. Feel free to ping me if you want me to override and merge anyways.

Edit: Someone was requesting this already, so I went ahead and force merged.

PR #1018, opened 2024-11-04
fix: enable scroll-y

KevinWu098:
## Summary
1. Changes in #1001 broke vertical scrolling on Dekstop
2. This PR fixes those problems 

## Test Plan
1. AA's desktop panels (left, right) should scroll appropriately on smaller view heights

chrome-capture-2024-10-3

### DIFF
--- a/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
+++ b/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
@@ -212,7 +212,7 @@ export default function ScheduleCalendar(_props?: ScheduleCalendarProps) {
     }, []);
 
     return (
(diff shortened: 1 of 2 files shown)


ap0nia:
My bad; I got skill checked 😔

It looks like it has to be anything besides the literal number 0. https://mui.com/system/sizing/#supported-values

adcockdalton:
Scrolling is more laggy on this deployment than on the main page--is that a consequence of staging?

adcockdalton (approved):
otherwise, it lgtm

adcockdalton:
After reviewing I think I'm just going to merge it, we can discuss optimization later we just need the scroll to work atm

PR #1029, opened 2024-11-11
ci: pipeline for RDS migration

ecxyzzy:
## Summary
- Add db URL to Actions workflows.
- Run `drizzle-kit generate` once.
  - We can overwrite the init_db until this hits prod. After that we should always create a new migration.
- Move script for copying data from ddb to pg into separate script.

## Test Plan
Ran migration locally, see if CI/CD pipeline breaks?


MinhxNguyen7 (approved):
Seems alright

Issue #1030, opened 2024-11-12
Remove lint-staged config from backend


ecxyzzy:
The backend's lint-staged config does not cover most of the files in the backend. As a result, things have gone unlinted for a while. (It literally let me commit code that would not build.)

We should probably hold off on doing this until most of the big backend migrations are done, since the blast radius will be huge.

PR #802, opened 2023-11-16
feat: Implement Calendar and Term Endpoints

KevinWu098:
## Issues

Closes #

## To-Dos
- [x] Use calendar and term endpoint
- [ ] Proper types
- [ ] Clean up code
- [ ] Adjust Term class for instruction end, final start, final end
- [ ] Speed up initial load time b‼️

PR #1031, opened 2024-11-12
feat: migrate to Anteater API

ecxyzzy:
This PR shouldn't have been necessary since in an ideal world we would have skipped API-Next, but oh well.

## Summary

- Set up codegen for types using [openapi-typescript](https://github.com/openapi-ts/openapi-typescript), remove ppapi-next-types, and fixup existing usage.
- Move fetching and most processing logic to backend, and set up tRPC routers and procedures.
  - Notably, we now rely only on the API cache layer for WebSoc, since we don't do anything fancy with it and a 5 minute TTL is probably reasonable. Because the backend now makes the requests, everyone fetching the same endpoint with the same params will hit cache, and as a result we can "invalidate" the cache by passing in a new dummy param.
  - Still caching grades on the frontend though since there's some Interesting logic there.
- Move away from GraphQL and nuke dead code.

## Test Plan
Tested locally, server related things need staging for tests.

Also someone else should run through their normal routine and verify that everything still works as expected.

## Things That Still Need Doing
- [x] Migrate fuzzy search.
- [x] Once icssc/anteater-api#30 gets merged we can remove the local declarations for the prerequisite tree.

## Future Follow-Up
- Replace Arktype with bare TS types and Zod where validation is actually necessary, since all our other projects are using Zod. (Sorry Brian.)
- There's a bug where the API tries to return enrollment history data objects for courses we don't have enrollment data for. This needs to be resolved on the API end but for now I've added a shim in the tRPC router.
  - Now tracking this in icssc/anteater-api#31.
- #1030
- #802 take 2


apps/backend/src/routers/websoc.ts
```diff
@@ -0,0 +1,108 @@
...
+        },
+    })
+        .then((data) => data.json())
+        .then((data) => data.data as WebsocAPIResponse);
+
+function combineSOCObjects(SOCObjects: WebsocAPIResponse[]) {
+    const combined = SOCObjects.shift() as WebsocAPIResponse;
+    for (const res of SOCObjects) {
+        for (const school of res.schools) {
+            const schoolIndex = combined.schools.findIndex((s) => s.schoolName === school.schoolName);
```
MinhxNguyen7, line 71: Can you sort out these type errors?
image

MinhxNguyen7 (changes requested):
The backend deployment doesn't work. I should've checked that first

MinhxNguyen7:
Some searches aren't working:
"ICSH32"/"ICS H32" return nothing.
"MAE93" returns nothing, even though it works right now.
"ICS139W" seems to work with fuzzy, but then no results show up

MinhxNguyen7:
Should replace these messages about PeterPortal API with Anteater API
image

MinhxNguyen7 (changes requested):
The finals part of the tour isn't working. I think it's because you changed the sample data factory

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/FuzzySearch.tsx
```diff
@@ -149,16 +124,28 @@ class FuzzySearch extends PureComponent<FuzzySearchProps, FuzzySearchState> {
...
-                            });
-                        } catch (e) {
-                            this.setState({ results: {} });
-                            console.error(e);
-                        }
+                        this.setState({ results: {}, loading: true }, () => {
+                            window.clearTimeout(this.state.pendingRequest);
+                            const pendingRequest = window.setTimeout(
+                                () =>
+                                    trpc.search.doSearch
```
MinhxNguyen7: In theory, couldn't there be a race condition where the older result arrives later and mutates the search results out of turn? Maybe you should look into aborting the call or otherwise invalidating the state update.

apps/backend/src/routers/websoc.ts
```diff
@@ -0,0 +1,127 @@
+import { z } from 'zod';
+import type {WebsocAPIResponse, CourseInfo, WebsocCourse} from '@packages/antalmanac-types';
+import { procedure, router } from '../trpc';
+
+function cleanSearchParams(record: Record<string, string>) {
```
MinhxNguyen7: Should be called `sanitizeSearchParams` so it doesn't sound like a factory method which produces clean search params

apps/backend/src/routers/websoc.ts
```diff
@@ -0,0 +1,127 @@
...
+    }
+    for (const [key, value] of Object.entries(record)) {
+        if (value === '') {
+            delete record[key];
+        }
+    }
+    return record;
+}
+
+function compareCourses(a: WebsocCourse, b: WebsocCourse) {
```
MinhxNguyen7, line 35: Can you comment what the comparison key is?

MinhxNguyen7 (changes requested):

MinhxNguyen7 (approved):
```
              |    |    |                 
             )_)  )_)  )_)              
            )___))___))___)\            
           )____)____)_____)\\
         _____|____|____|____\\\__
---------\                   /---------
  ^^^^^ ^^^^^^^^^^^^^^^^^^^^^
    ^^^^      ^^^^     ^^^    ^^
         ^^^^      ^^^
```

PR #1033, opened 2024-11-14
feat: exclude restriction codes

### LINKED ISSUE #595: Filter out major/school restrictions
This is especially useful when looking at GEs. This can be done on our side, but API should probably do this. It should be a pretty simple change on their side, but I'm not sure how we would implement it UI-wise

KevinWu098:
## Summary
1. Implements Anteater API's [exclude restriction codes](https://github.com/icssc/peterportal-api-next/pull/130) parameter into our Advanced Search

chrome-capture-2024-10-14

## Test Plan
1. Test that a singular code is filtered
2. Test that multiple codes are excluded (OR)
3. Test that the URL encoding works (both read/write)

## Issues
Closes #595


MinhxNguyen7 (changes requested):
The selections should be more highlighted. Otherwise, looks fine

KevinWu098:
> The selections should be more highlighted. Otherwise, looks fine

Screenshot 2024-11-14 at 3 49 46 PM
Screenshot 2024-11-14 at 3 49 37 PM

Updated the look

MinhxNguyen7 (approved):
LGTM!

smalldrew:
LGTM as well

PR #1039, opened 2024-11-18
ci: fix destroy staging workflow

ecxyzzy:
## Summary

Because I suck at devops(tm) I forgot to add the API key to the destroy step which needs to build the backend.

## Test Plan
Not sure if this will deploy the backend? But if so, let the backend build and then add "no deploy" to trigger the destroy workflow


MinhxNguyen7 (approved):
LGTM!

MinhxNguyen7 (approved):

PR #1042, opened 2024-11-20
fix: prevent course info bar overflow

### LINKED ISSUE #729: Prevent Text-Wrap in SectionTableBody and the Department Card Component
Screenshot 2023-10-12 at 11 09 23 AM

):< Long course names make the buttons resize and shifts the UI

The fix should truncate the course names whenever they overflow some predefined min-length of the course info button. This also happens to the School/Dept card.

andrew-wang0:
## Summary

Prevent overflow

## Issues

Closes #729 

## Future Followup

will not be addressing

### DIFF
--- a/apps/antalmanac/src/components/RightPane/SectionTable/CourseInfoButton.tsx
+++ b/apps/antalmanac/src/components/RightPane/SectionTable/CourseInfoButton.tsx
@@ -1,6 +1,7 @@
 import { Button, Paper, Popper } from '@material-ui/core';
 import { withStyles } from '@material-ui/core/styles';
 import { ClassNameMap } from '@material-ui/core/styles/withStyles';
+import { useMediaQuery, useTheme } from '@mui/material';
 import { useEffect, useState } from 'react';
 
 import { logAnalytics } from '$lib/analytics';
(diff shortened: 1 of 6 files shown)


apps/antalmanac/src/components/RightPane/SectionTable/CourseInfoButton.tsx
```diff
@@ -65,11 +63,13 @@ function CourseInfoButton({
         }
     };
 
+    const scheduleManagementWidth = useScheduleManagementStore((state) => state.scheduleManagementWidth);
+    const compact = scheduleManagementWidth && scheduleManagementWidth < 600;
```
KevinWu098: issue: we should keep our breakpoints standard. prefer using a specific breakpoint as defined in our theme (and referencing it as such, not as a magic number)

apps/antalmanac/src/components/RightPane/SectionTable/CourseInfoButton.tsx
```diff
@@ -87,7 +87,20 @@ function CourseInfoButton({
                     }
                 }}
             >
-                {text}
+                <span style={{ display: 'flex', gap: 4 }}>
+                    {icon}
```
KevinWu098, line 96: question/issue: can we continue passing in icon as `startIcon`? Why are we moving this into the children?

KevinWu098 (changes requested):
Thanks for this PR! Really looking forward to the UI improvement via reduction of UI shift. 

Have a couple comments/nits to resolve, but otherwise, looks solid!

andrew-wang0 (commented):
🫡

apps/antalmanac/src/components/RightPane/SectionTable/CourseInfoButton.tsx
andrew-wang0: startIcon auto adds margin right, making it not centered

alexanderl19 (approved):

KevinWu098 (approved):

PR #1043, opened 2024-11-20
Fixed "Error loading grades information" with advanced search

### LINKED ISSUE #1020: "Error loading grades information" with advanced search
## Description
- "Error loading grades information" message shows up even though grades load properly.
- This is not observed in normal search.

## Reproduction
- Search for an instructor with classes this quarter (e.g., Ihler).
- See that the error message shows up but the grades still populate fine.

IsaacNguyen:
## Summary
Stopped the error "Error loading grades information" from displaying when inserting an instructor's name and having a successful search.
image
(The error no longer shows up in the bottom left after a successful advanced search using a professor's name.)

The error was being caused because when trying to populate the grades cache through `populateGradesCache`, the instructor would not be passed in as a parameter.
image
(This is the updated code with instructor being passed as a parameter.)

As a result, even if an instructor was successfully passed into advanced search but a GE and department were _not_ supplied, `populateGradesCache` would throw an error asking to provide a department or GE. 
image
(It would throw the error at the bottom which used to _only_ check for an empty GE or department).

To fix this, as shown above, I fixed the conditions under which the error was thrown, checking if instructor was also empty before throwing.

## Test Plan
Check if any combination of an inputted or not inputted GE, department, and instructor causes an issue.

## Issues

Closes #1020

### DIFF
--- a/apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
+++ b/apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
@@ -212,7 +212,7 @@ export default function CourseRenderPane(props: { id?: number }) {
         const gradesQueryParams = {
             department: formData.deptValue,
             ge: formData.ge as GE,
-            instructorName: formData.instructor,
+            instructor: formData.instructor,
         };
 
         try {
--- a/apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
+++ b/apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
@@ -212,6 +212,7 @@ export default function CourseRenderPane(props: { id?: number }) {
         const gradesQueryParams = {
             department: formData.deptValue,
             ge: formData.ge as GE,
+            instructorName: formData.instructor,
         };
 
         try {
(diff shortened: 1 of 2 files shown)


adcockdalton (commented):
ltgm; tested thoroughly on staging.

Only one tiny code nitpick: why instructorName instead of instructor? Consistency with the naming throughout can help for clarity. If there is a reason for the change please explain and I'll approve the PR

Otherwise great PR! Looks clean

IsaacNguyen:
> ltgm; tested thoroughly on staging.
> 
> Only one tiny code nitpick: why instructorName instead of instructor? Consistency with the naming throughout can help for clarity. If there is a reason for the change please explain and I'll approve the PR
> 
> Otherwise great PR! Looks clean

image
I saw that in the code above, in `websocQueryParams`, `instructorName` was passed as a parameter so I just stuck with it. However, I fully agree that naming convention-wise it should be `instructor`, but I wasn't sure if I should have deviated from it.

MinhxNguyen7:
Just go for `instructor`

MinhxNguyen7:
Also, can you add a brief explanation of what caused the bug?

IsaacNguyen:
> Just go for `instructor`

Got it.

IsaacNguyen:
> Also, can you add a brief explanation of what caused the bug?

Added an explanation!

MinhxNguyen7 (approved):
LGTM!

PR #1045, opened 2024-11-21
fix: handle inconsistent ge param

KevinWu098:
## Summary
1. GE param would be referenced by `ge` or `GE` rather inconsistently in our URL parameter code. This PR resolves inconsistencies by making the GE URL param always `ge`.

## Test Plan
1. Test that searching by GE still works
2. Test that loading a URL with a GE param still works
3. Test that resetting search form clears GE

## Issues

Closes #


adcockdalton (approved):
followed test plan smoothly, works

did a case-sensitive ctrl+f in the AA repo to check for any stragglers and there were none

lgtm

PR #1048, opened 2024-11-22
Duplicate calendar

### LINKED ISSUE #1032: Enforce unique schedule names
## Description
- A user should not be able to save their data with duplicate schedule names.
- They should be prevented from creating a schedule with the same name. If they attempt to save, and there is a duplicate schedule, we can add `(NUMBER)` to the end of the name before sending it to the backend.
- Previously, duplicate names were OK because we stored everything as a JSON object in DynamoDB. Now, each schedule is a row in the table, and they would not be distinguishable if they had the same name.

## Related Issues
- #970 will mangle schedule names as specified above if that's not done on the frontend.

### LINKED ISSUE #1034: Duplicate schedule instead of just copying

jotalis:
## Summary
- Copy to New Schedule functionality 
- Enforce unique schedule names when duplicating, creating, or renaming a schedule
- Fix bug where schedule names aren't updated and thus displayed properly for deletion modals

## Test Plan
- Undo New Schedule Creation: Verify undo functionality removes a newly created schedule.
- Combined Undo for Schedule Creation and Event Copying: Confirm one undo removes both a new schedule and copied events.
- Duplicate Schedule Name on Creation: Validate a duplicate schedule name results in a suffix like (1).
- Duplicate Schedule Name on Renaming: Check renaming to an existing name adds a suffix or prevents duplication.
- Renaming and Duplicating Schedules: Verify renaming a schedule and duplicating it results in proper suffixes.
- Adding Custom Events: Verify custom events are added with all details intact.
- Start with existing schedule 2025-Winter. Create new schedule with same name. Expect 2025-Winter(1). Try renaming schedule to 2025-Winter. Expect 2025-Winter(1)
- Deleting schedule modal shows correct name of schedule to delete; Create 3 schedules; delete the second schedule; open up the delete modal for the third schedule. It should display the name of the third schedule and not the second schedule

## Issues
Closes #1034, #1032

### DIFF
--- a/apps/antalmanac/src/components/dialogs/CopySchedule.tsx
+++ b/apps/antalmanac/src/components/dialogs/CopySchedule.tsx
@@ -51,7 +51,7 @@ function CopyScheduleDialog(props: CopyScheduleDialogProps) {
             <DialogTitle>Copy Schedule</DialogTitle>
             <DialogContent>
                 <Box padding={1}>
-                    <TextField fullWidth label="Name *" onChange={handleNameChange} value={name} />
+                    <TextField fullWidth label="Name" onChange={handleNameChange} value={name} />
                 </Box>
             </DialogContent>
             <DialogActions>
--- a/apps/antalmanac/src/components/dialogs/CopySchedule.tsx
+++ b/apps/antalmanac/src/components/dialogs/CopySchedule.tsx
@@ -5,12 +5,10 @@ import {
     DialogActions,
     DialogContent,
     DialogTitle,
-    MenuItem,
-    Select,
+    TextField,
     type DialogProps,
 } from '@mui/material';
-import { SelectChangeEvent } from '@mui/material';
-import { useState, useEffect, useCallback } from 'react';
+import { useState, useCallback } from 'react';
 
 import { copySchedule, addSchedule } from '$actions/AppStoreActions';
 import AppStore from '$stores/AppStore';
@@ -22,11 +20,11 @@ interface CopyScheduleDialogProps extends DialogProps {
 function CopyScheduleDialog(props: CopyScheduleDialogProps) {
     const { index } = props;
     const { onClose } = props; // destructured separately for memoization.
-    const [scheduleNames, setScheduleNames] = useState(AppStore.getScheduleNames());
-    const [selectedSchedule, setSelectedSchedule] = useState<number>(0);
+    const scheduleNames = AppStore.getScheduleNames();
+    const [name, setName] = useState<string>(`Copy of ${scheduleNames[index]}`);
 
-    const handleScheduleChange = useCallback((event: SelectChangeEvent<number>) => {
-        setSelectedSchedule(event.target.value as number);
+    const handleNameChange = useCallback((event: React.ChangeEvent<HTMLTextAreaElement>) => {
+        setName(event.target.value);
     }, []);
 
     const handleCancel = useCallback(() => {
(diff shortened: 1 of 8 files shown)


MinhxNguyen7:
While you're here, can you replace the dropdown with a list of buttons? There's no reason to select, then having to press a button.

apps/antalmanac/src/components/dialogs/CopySchedule.tsx
```diff
@@ -70,6 +74,7 @@ function CopyScheduleDialog(props: CopyScheduleDialogProps) {
                                 {name}
                             </MenuItem>
                         ))}
+                        <MenuItem value={scheduleNames.length + 1}>Copy to New Schedule</MenuItem>
```
KevinWu098:
```suggestion
                        <MenuItem value={'copyToNew'}>Copy to New Schedule</MenuItem>
```
issue: I think it's confusing for us to declare an item with a value that is greater than the length of the elements. 

suggestion: set the value to something more explicit. Then, you can extract the double if statements upwards and minimize nesting.

KevinWu098 (changes requested):
Congrats on your first PR! Looks solid, barring one issue I've noted below.

jotalis:
Update: discussed with AA team to prioritize a simpler and more familiar duplication functionality (e.g. Make a Copy from Google Docs). The copy schedule modal is now only going to allow the making of duplicate schedules, removing "Copy to All Schedules" and "Copy to Schedule".

jotalis:
Currently when the user tries to duplicate a schedule, a unique schedule name is enforced upon schedule creation, even if what they enter already exists.

For example, if Winter-2025 exists, and a schedule tries to be duplicated with the name Winter-2025, a number (e.g. (1)) will be appended to the schedule name upon creation.

However, users are still able to have schedule with duplicate names in these cases:
1. They have existing schedules with duplicate names already
2. They rename a schedule to an already existing schedule name
3. They create a new schedule using an already existing schedule name

To help migrate towards a system where all schedule names are unique, we should enforce the same naming validation in the duplicate schedule feature for 2 and 3 (renaming and creating schedules).

apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -134,10 +145,11 @@ export class Schedules {
     }
 
     /**
+     * @deprecated This method is no longer used by any features
      * Append all courses from current schedule to the schedule with the target index.
      * @param to Index of the schedule to append courses to. If equal to number of schedules, will append courses to all schedules.
      */
-    copySchedule(to: number) {
+    copyToSchedule(to: number) {
```
KevinWu098: question: If this is no longer being used, can we just remove it?

apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -50,10 +50,19 @@ export class Schedules {
         this.skeletonSchedules = [];
     }
 
+    getNextScheduleName(newScheduleName: string, scheduleIndex: number) {
+        const scheduleNames = this.getScheduleNames().filter((_, index) => index !== scheduleIndex);
```
KevinWu098: question: If I'm understanding this line correctly, we're getting all of the schedule names except the original. Can we splice to remove the given index instead?

KevinWu098 (approved):
LGTM! 🚀 

2 non-blocking questions for you, but otherwise, it looks good to merge

apps/antalmanac/src/stores/Schedules.ts
jotalis: sure; originally i left it in case we ever wanted to implement something that needed schedule copying using index but on second thought nothing else would really need that (other than the feature we took out).

apps/antalmanac/src/stores/Schedules.ts
jotalis: yeah; getScheduleNames() returns a deep copy, so it'll be fine. i'll change it to splice

MinhxNguyen7 (approved):
If this breaks prod, I blame @KevinWu098

PR #1049, opened 2024-11-22
Ensure unique schedule

### LINKED ISSUE #1032: Enforce unique schedule names

vicksey:
## Summary
Ensure unique schedule names when adding and renaming schedules.

## Test Plan
- [ ] Try to add a new schedule with a name that already exists. It shouldn't work.
- [ ] Try to rename a schedule to a name that already exists. It shouldn't work.

## Issues
Unique schedule names.

Closes #1032

### DIFF
--- a/apps/antalmanac/src/components/dialogs/AddSchedule.tsx
+++ b/apps/antalmanac/src/components/dialogs/AddSchedule.tsx
@@ -13,6 +13,7 @@ function AddScheduleDialog({ onClose, onKeyDown, ...props }: DialogProps) {
     const isDark = useThemeStore((store) => store.isDark);
 
     const [name, setName] = useState(AppStore.getDefaultScheduleName());
+    const [errorMessage, setErrorMessage] = useState('');
 
     const handleCancel = () => {
         onClose?.({}, 'escapeKeyDown');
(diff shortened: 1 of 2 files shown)


MinhxNguyen7:
I take that this is ready for review. Can you add the issue that you're closing in the "Issues" section of the PR? Also, please add a brief test plan for how the reviewer would confirm that your change does what it's supposed to do and probably doesn't break anything.

apps/antalmanac/src/components/dialogs/AddSchedule.tsx
```diff
@@ -52,7 +60,14 @@ function AddScheduleDialog({ onClose, onKeyDown, ...props }: DialogProps) {
 
             <DialogContent>
                 <Box padding={1}>
-                    <TextField fullWidth label="Name" onChange={handleNameChange} value={name} />
+                    <TextField
+                        fullWidth
+                        label="Name"
+                        onChange={handleNameChange}
+                        value={name}
+                        error={!!errorMessage}
```
MinhxNguyen7: Please use `Boolean` instead of `!!`. That's less explicit and quite weird to read.

MinhxNguyen7 (changes requested):
Can you extract the schedule name field, along with the logic and error messages so that it can be reused?

Also, fixing the merge conflicts should allow the staging instance to deploy successfully.

KevinWu098:
Closing as completed via #1048

PR #1056, opened 2024-11-23
Fixed "Error loading grades information" with empty course code

IsaacNguyen:
## Summary
Refer to #1043 because it's the same issue just with course code.
Error was propagating because course code was not check in this if statement.
image

## Issues

Closes #


apps/backend/.env.sample
alexanderl19, line 1: Why are we deleting the .env.sample? I think this is unintended.

alexanderl19 (dismissed):
Just the one comment on the deleted file, but otherwise, LGTM

-- Kevin

apps/backend/.env.sample
IsaacNguyen: Oh yeah, my bad; I didn't realize I did that. I just added it back!

KevinWu098 (approved):
LGTM!

PR #1060, opened 2024-11-25
Read from RDS

### LINKED ISSUE #970: Migrate to RDS
## Description
- Migrate the DB from Dynamo to RDS.
- Even though we could make auth work with NoSQL, attaching the notifications system onto it would just not make sense.
- Alternatively, we can make the notification separate from the auth system, and people just sign up for notifications on a per-email/phone number basis.

### Migration Procedures
- [x] Start saving to both DBs. #993 #1064 #1065 #1066 
- [ ] Copy to new DB, while still writing to both. (IN PROGRESS)
    - This needs to be done while few people are using AA. I can set up a chron job to do this at an ungodly hour.
    - Then I'll check in the morning to see if it worked. 
    - Scheduled for 1:40 AM PT, December 6th.
- [ ] Switch read to new DB. #1060 
- [ ] Remove old DB.

MinhxNguyen7:
## Summary
- Switch to read from RDS without changing the schema or deployment.
- Remove legacy MongoDB users. This will only affect users that have not saved a schedule in over a year.
- Still write to DDB without reading from it.
    - This allows us to revert this PR and retrieve all of the saved user data during the intervening period.

### Testing
- The test plan was designed by looking at `ScheduleSaveState` schema and making sure that we test the read and write 
of every attribute.
- I have run through this, and everything looks good, but someone else should go through it as well as suggest more testing steps.

### Test Plan
- [ ] Create a schedule and add a section to it.
- [ ] Create another schedule and add a different section as well as a custom event.
- [ ] Save the user data.
- [ ] Clear the site data and reload the page.
- [ ] Load the previously-saved user.
- [ ] Verify that the previously-added data populates, and the order of schedules is correct.
- [ ] Delete a schedule, and repeat the reloading and verification procedures.

## Issues
- Closes #970 

## Future Follow-Up
- Remove DDB and vestigial scripts once the we're 100% sure that prod is good.


adcockdalton:
Code looks clean. Checking on staging now.

adcockdalton (approved):
Test Plan

- [x]  Create a schedule and add a section to it.
- [x]  Create another schedule and add a different section as well as a custom event.
- [x]  Save the user data.
- [x]  Clear the site data and reload the page.
- [x]  Load the previously-saved user.
- [x]  Verify that the previously-added data populates, and the order of schedules is correct.
- [x]  Delete a schedule, and repeat the reloading and verification procedures.
Additionally:
- [x]  Create an additional user and mimic the prior process
- [x]  Confirm that a 7-day custom event has intended behavior
- [x]  Ensure that all facets of data persist for custom events (color, map location)
- [x]  Ensure that "remember me" has intended functionality when using multiple users

Looks good to me. What an exciting PR! Cheers

packages/types/src/customevent.ts
KevinWu098, line 1: nit (non-blocking): can we rename this to `customEvent.ts`

MinhxNguyen7: No. We can do that later.

PR #1062, opened 2024-12-02
feat: restyle custom events modal

### LINKED ISSUE #1025: Restyle Custom Events Modal
## Description
- This is ugly.
Image

xgraceyan:
## Summary
Restyled the custom events modal:
- Changed schedule select from checkbox to chip multiple select
  - This is so not all the schedules appear in the main modal (the mass of checkboxes for every schedule is what I'm assuming is mainly contributing to the "ugliness" of the modal)
- Added select variant to Building select props to maintain consistency of outlined style
  - Building select in custom events modal is now outlined variant, while in Maps it is the filled variant
image

It should look better with multiple schedules selected now.
image

## Test Plan
Test on different device viewports: Computer (16x9); Tablet (iPad Pro); phone (iPhone 14 pro max); small phone (iPhone SE or any width < 400px)
Test with multiple schedules: 5, 10, 20 schedules selected & make sure the modal doesn't turn ugly.
Test that schedules are selected and the custom event is properly added for all schedules chosen
Test that schedules can be correctly deselected and that they don't have the custom event added to them
Test that the Building selector correctly searches up places in both the custom events modal and Map view, since I made a few changes to it

## Issues

Closes #1025

### DIFF
--- a/apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
+++ b/apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
@@ -171,7 +171,7 @@ function CustomEventDialogs(props: CustomEventDialogProps) {
                 <DialogTitle id="form-dialog-title">
                     {props.customEvent ? 'Edit a Custom Event' : 'Add a Custom Event'}
                 </DialogTitle>
-                <DialogContent>
+                <DialogContent sx={{ display: 'flex', flexDirection: 'column', gap: '12px' }}>
                     <FormControl fullWidth>
                         <TextField
                             id="event-name-input"
@@ -183,10 +183,7 @@ function CustomEventDialogs(props: CustomEventDialogProps) {
                             margin="dense"
                         />
                     </FormControl>
-                    <FormControl
-                        fullWidth
-                        sx={{ display: 'flex', flexDirection: 'row', gap: '12px', marginTop: '12px' }}
-                    >
+                    <FormControl fullWidth sx={{ display: 'flex', flexDirection: 'row', gap: '12px' }}>
                         <TextField
                             onChange={handleStartTimeChange}
                             label="Start Time"
--- a/apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
+++ b/apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
@@ -213,7 +213,7 @@ function CustomEventDialogs(props: CustomEventDialogProps) {
                         />
                     </form>
                     <DaySelector onSelectDay={handleDayChange} days={props.customEvent?.days} />
-                    <BuildingSelect value={building} onChange={handleBuildingChange} />
+                    <BuildingSelect value={building} onChange={handleBuildingChange} variant="outlined" />
                     <ScheduleSelector
                         scheduleIndices={scheduleIndices}
                         onSelectScheduleIndices={handleSelectScheduleIndices}
--- a/apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
+++ b/apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
@@ -167,7 +167,7 @@ function CustomEventDialogs(props: CustomEventDialogProps) {
                 </Tooltip>
             )}
             <Dialog open={open} onClose={handleClose} maxWidth={'xs'}>
-                <DialogTitle id="form-dialog-title" style={{ marginBottom: -10 }}>
+                <DialogTitle id="form-dialog-title">
                     Add a Custom Event
                 </DialogTitle>
                 <DialogContent>
@@ -179,10 +179,9 @@ function CustomEventDialogs(props: CustomEventDialogProps) {
                             required={true}
                             value={title}
                             onChange={handleEventNameChange}
-                            style={{ marginTop: 10, width: '100%' }}
                         />
                     </FormControl>
-                    <form noValidate style={{ display: 'flex', gap: 0, marginTop: 10 }}>
+                    <FormControl fullWidth sx={{ display: 'flex', flexDirection: 'row', gap: '10px' }}>
                         <TextField
                             onChange={handleStartTimeChange}
                             label="Start Time"
@@ -195,7 +194,6 @@ function CustomEventDialogs(props: CustomEventDialogProps) {
                             inputProps={{
                                 step: 300,
                             }}
-                            style={{ marginRight: 5, marginTop: 10 }}
                         />
                         <TextField
                             onChange={handleEndTimeChange}
@@ -209,7 +207,6 @@ function CustomEventDialogs(props: CustomEventDialogProps) {
                             inputProps={{
                                 step: 300,
                             }}
-                            style={{ marginTop: 10 }}
                         />
                     </form>
                     <DaySelector onSelectDay={handleDayChange} days={props.customEvent?.days} />
--- a/apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
+++ b/apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
@@ -4,10 +4,9 @@ import {
     Dialog,
     DialogActions,
     DialogContent,
+    DialogTitle,
     FormControl,
     IconButton,
-    Input,
-    InputLabel,
     TextField,
     Tooltip,
 } from '@mui/material';
@@ -167,38 +166,50 @@ function CustomEventDialogs(props: CustomEventDialogProps) {
                     </IconButton>
                 </Tooltip>
             )}
-            <Dialog open={open} onClose={handleClose} maxWidth={'lg'}>
+            <Dialog open={open} onClose={handleClose} maxWidth={'xs'}>
+                <DialogTitle id="form-dialog-title" style={{ marginBottom: -10 }}>
+                    Add a Custom Event
+                </DialogTitle>
                 <DialogContent>
-                    <FormControl>
-                        <InputLabel htmlFor="EventNameInput">Event Name</InputLabel>
-                        <Input required={true} value={title} onChange={handleEventNameChange} />
+                    <FormControl fullWidth>
+                        <TextField
+                            id="outlined-basic"
+                            label="Event Name"
+                            variant="outlined"
+                            required={true}
+                            value={title}
+                            onChange={handleEventNameChange}
+                            style={{ marginTop: 10, width: '100%' }}
+                        />
                     </FormControl>
-                    <form noValidate style={{ display: 'flex', gap: 5, marginTop: 5 }}>
+                    <form noValidate style={{ display: 'flex', gap: 0, marginTop: 10 }}>
                         <TextField
                             onChange={handleStartTimeChange}
                             label="Start Time"
                             type="time"
                             defaultValue={start}
+                            fullWidth
                             InputLabelProps={{
                                 shrink: true,
                             }}
                             inputProps={{
                                 step: 300,
                             }}
-                            style={{ marginRight: 5, marginTop: 5 }}
+                            style={{ marginRight: 5, marginTop: 10 }}
                         />
                         <TextField
                             onChange={handleEndTimeChange}
                             label="End Time"
                             type="time"
                             defaultValue={end}
+                            fullWidth
                             InputLabelProps={{
                                 shrink: true,
                             }}
                             inputProps={{
                                 step: 300,
                             }}
-                            style={{ marginRight: 5, marginTop: 5 }}
+                            style={{ marginTop: 10 }}
                         />
                     </form>
                     <DaySelector onSelectDay={handleDayChange} days={props.customEvent?.days} />
@@ -211,7 +222,7 @@ function CustomEventDialogs(props: CustomEventDialogProps) {
                     />
                 </DialogContent>
 
-                <DialogActions>
+                <DialogActions style={{ padding: '0px 16px 16px' }}>
                     <Button onClick={handleClose} color={isDark ? 'secondary' : 'primary'}>
                         Cancel
                     </Button>
--- a/apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
+++ b/apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
@@ -139,6 +139,7 @@ function CustomEventDialogs(props: CustomEventDialogProps) {
     }, []);
 
     const isDark = useThemeStore.getState().isDark;
+
     return (
         <>
             {props.customEvent ? (
@@ -168,20 +169,24 @@ function CustomEventDialogs(props: CustomEventDialogProps) {
             )}
             <Dialog open={open} onClose={handleClose} maxWidth={'xs'}>
                 <DialogTitle id="form-dialog-title">
-                    Add a Custom Event
+                    {props.customEvent ? 'Edit a Custom Event' : 'Add a Custom Event'}
                 </DialogTitle>
                 <DialogContent>
                     <FormControl fullWidth>
                         <TextField
-                            id="outlined-basic"
+                            id="event-name-input"
                             label="Event Name"
                             variant="outlined"
                             required={true}
                             value={title}
                             onChange={handleEventNameChange}
+                            margin="dense"
                         />
                     </FormControl>
-                    <FormControl fullWidth sx={{ display: 'flex', flexDirection: 'row', gap: '10px' }}>
+                    <FormControl
+                        fullWidth
+                        sx={{ display: 'flex', flexDirection: 'row', gap: '12px', marginTop: '12px' }}
+                    >
                         <TextField
                             onChange={handleStartTimeChange}
                             label="Start Time"
@@ -208,7 +213,7 @@ function CustomEventDialogs(props: CustomEventDialogProps) {
                                 step: 300,
                             }}
                         />
-                    </form>
+                    </FormControl>
                     <DaySelector onSelectDay={handleDayChange} days={props.customEvent?.days} />
                     <BuildingSelect value={building} onChange={handleBuildingChange} variant="outlined" />
                     <ScheduleSelector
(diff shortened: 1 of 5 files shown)


KevinWu098 (changes requested):
Comments on two aspects (only based on the SS provided):
1. The day of week buttons being blue on an all-gray/all-white modal sticks out to me; could we try using a different variant (secondary should be gray?)
2. The day of week buttons also get unaligned with the rest of the elements on wider modals. Can we style the buttons to take up 100% of horiz. space, by dynamically setting gap or width (I think width makes more sense)?

xgraceyan:
Gotcha! I've made changes according to your comments:
- As far as I've checked, the secondary theme is white, not gray; I've made the buttons be the default color theme instead, which _is_ gray. However, I think maybe the selector would look better if the buttons selected were the primary blue color, instead of also being gray, as it feels slightly bland right now -- what do you think?
image

- I've also fixed the alignment of the day select buttons now. They should resize on larger modals and take up all of the horizontal space.
image

KevinWu098:
The UI tweaks look great -- just updating to say that I'll get around to reviewing this post finals 🫡

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -167,42 +166,54 @@ function CustomEventDialogs(props: CustomEventDialogProps) {
                     </IconButton>
                 </Tooltip>
             )}
-            <Dialog open={open} onClose={handleClose} maxWidth={'lg'}>
+            <Dialog open={open} onClose={handleClose} maxWidth={'xs'}>
+                <DialogTitle id="form-dialog-title" style={{ marginBottom: -10 }}>
+                    Add a Custom Event
+                </DialogTitle>
```
KevinWu098: issue: `CustomEventDialog` is used for both creating and editing custom events. This text should reflect the current usage of the dialog

suggestion: check for `props.customEvent` (`boolean`) and swap between Add end Edit.

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -167,42 +166,54 @@ function CustomEventDialogs(props: CustomEventDialogProps) {
                     </IconButton>
                 </Tooltip>
             )}
-            <Dialog open={open} onClose={handleClose} maxWidth={'lg'}>
+            <Dialog open={open} onClose={handleClose} maxWidth={'xs'}>
+                <DialogTitle id="form-dialog-title" style={{ marginBottom: -10 }}>
```
KevinWu098:
```suggestion
                <DialogTitle id="form-dialog-title">
```
**edit: the issue+suggestion below are still good practice, but read the comment replying to this one as for why we opt to remove the custom styling altogether**

issue: using marginBottom to override another property (in this case `paddingBottom`) is generally bad practice

suggestion: whenever possible, prefer to identify if you can directly target the style that you would otherwise want to (in this case) "cancel out"

----

nit: Also, when working with MUI v5 (denoted by the import `@mui/material`), use `sx` when styling components

KevinWu098: issue: looking at the code below which adds `marginTop` to the element directly below, removing `marginBottom` from `DialogTitle` and `marginTop` from the event name input element achieves the same effect.

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -167,42 +166,54 @@ function CustomEventDialogs(props: CustomEventDialogProps) {
...
-                        <Input required={true} value={title} onChange={handleEventNameChange} />
+                    <FormControl fullWidth>
+                        <TextField
+                            id="outlined-basic"
+                            label="Event Name"
+                            variant="outlined"
+                            required={true}
+                            value={title}
+                            onChange={handleEventNameChange}
+                            style={{ marginTop: 10, width: '100%' }}
```
KevinWu098:
```suggestion
```
issue: the styling is unneeded; `width: '100%'` is applied by `fullWidth` on the `FormControl`. See above comments for more details about `marginTop`.

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -167,42 +166,54 @@ function CustomEventDialogs(props: CustomEventDialogProps) {
...
+                        />
                     </FormControl>
-                    <form noValidate style={{ display: 'flex', gap: 5, marginTop: 5 }}>
+                    <form noValidate style={{ display: 'flex', gap: 0, marginTop: 10 }}>
```
KevinWu098:
```suggestion
                    <FormControl fullWidth sx={{ display: 'flex', flexDirection: 'row', gap: '10px' }}>
```
issue: (I know you didn't add this element) Not sure why we're using an basic `form` element here. Let's replace this with a `FormControl`

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/DaySelector.tsx
```diff
@@ -30,24 +26,29 @@ const DaySelector = ({ days = [false, false, false, false, false, false, false],
             sx={{
                 display: 'flex',
                 flexDirection: 'row',
-                justifyContent: 'center',
-                padding: 5,
+                justifyContent: 'space-between',
+                marginTop: 10,
+                marginBottom: 10,
+                marginLeft: -5,
+                marginRight: -5,
```
KevinWu098:
```suggestion
                marginTop: 10,
                marginBottom: 10,
                marginLeft: -5,
                marginRight: -5,
```
issue: negative margins (although not always bad) are usually a sign of an antipattern -- they introduce more code to "cancel out" other code. Removing this actually makes the component clearer and more maintanable

suggestion: when adding negative margin, ask yourself where the source of the unwanted spacing is coming from

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/DaySelector.tsx
```diff
@@ -30,24 +26,29 @@ const DaySelector = ({ days = [false, false, false, false, false, false, false],
...
             }}
         >
```
KevinWu098, line 32:
```suggestion
                    style={{ gap: '10px' }}
        >
```

suggestion: the answer to the above question (where is this weird spacing on the buttons coming from?) is the margin placed upon each button, which results in spacing on the lefthand and righthand side of the button group. Using `gap` is a more maintainable, cleaner option which only places the spacing between elements

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/DaySelector.tsx
```diff
@@ -30,24 +26,29 @@ const DaySelector = ({ days = [false, false, false, false, false, false, false],
...
             {normal_days.map((day, index) => (
                 <Button
                     key={index}
                     variant={selectedDays[index] ? 'contained' : 'outlined'}
                     size="small"
+                    fullWidth
                     onClick={() => {
                         handleChange(index);
                     }}
```
KevinWu098:
```suggestion
                    onClick={() => handleChange(index)}
```
nit: we can inline this function

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/DaySelector.tsx
```diff
@@ -30,24 +26,29 @@ const DaySelector = ({ days = [false, false, false, false, false, false, false],
...
                     style={{
+                        display: 'block',
+                        aspectRatio: 1 / 1,
                         margin: 5,
-                        maxWidth: 40,
-                        minWidth: 40,
-                        aspectRatio: 1,
+                        minWidth: 20,
+                        minBlockSize: 40,
                     }}
```
KevinWu098:
```jsx
                    style={{
                        aspectRatio: 1 / 1,
                        width: '40px',
                        minWidth: 'unset',
                    }}
```
Together with applying `gap` to the container, we can simplify the button styles dramatically

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -167,42 +166,54 @@ function CustomEventDialogs(props: CustomEventDialogProps) {
...
                             defaultValue={start}
+                            fullWidth
                             InputLabelProps={{
                                 shrink: true,
                             }}
                             inputProps={{
                                 step: 300,
                             }}
-                            style={{ marginRight: 5, marginTop: 5 }}
+                            style={{ marginRight: 5, marginTop: 10 }}
```
KevinWu098:
```suggestion
```
repeat: we can omit these styles using gap

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -167,42 +166,54 @@ function CustomEventDialogs(props: CustomEventDialogProps) {
...
                             defaultValue={end}
+                            fullWidth
                             InputLabelProps={{
                                 shrink: true,
                             }}
                             inputProps={{
                                 step: 300,
                             }}
-                            style={{ marginRight: 5, marginTop: 5 }}
+                            style={{ marginTop: 10 }}
```
KevinWu098:
```suggestion
```
repeat: we can omit these styles using gap

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -211,7 +222,7 @@ function CustomEventDialogs(props: CustomEventDialogProps) {
                     />
                 </DialogContent>
 
-                <DialogActions>
+                <DialogActions style={{ padding: '0px 16px 16px' }}>
```
KevinWu098: question: why are we adding padding to the footer/dialog actions? It's not necessarily incorrect, but I think we would prefer to not add arbitrary styles in favor of maintaining styles across components (i.e. other dialogs). 

To be clear, you can (and should) argue for the styling if you have a compelling reason!

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/ScheduleSelector.tsx
```diff
@@ -1,7 +1,12 @@
...
-import FormGroup from '@material-ui/core/FormGroup';
+import FormControl from '@material-ui/core/FormControl';
+import Box from '@mui/material/Box';
+import Chip from '@mui/material/Chip';
+import InputLabel from '@mui/material/InputLabel';
+import MenuItem from '@mui/material/MenuItem';
+import OutlinedInput from '@mui/material/OutlinedInput';
+import Select, { SelectChangeEvent } from '@mui/material/Select';
 import type { RepeatingCustomEvent } from '@packages/antalmanac-types';
+import { int } from 'aws-sdk/clients/datapipeline';
```
KevinWu098:
```suggestion
```
issue: I think you intended to use the `number` type

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/ScheduleSelector.tsx
```diff
@@ -20,43 +25,45 @@ class ScheduleSelector extends PureComponent<ScheduleSelectorProps, ScheduleSele
...
+                    size="small"
+                    style={{ marginTop: 2 }}
+                    multiple
+                    value={this.state.scheduleIndices}
+                    defaultValue={this.state.scheduleIndices}
+                    onChange={this.handleChange}
+                    input={<OutlinedInput id="select-multiple-chip" label="Chip" />}
+                    renderValue={(selected) => (
+                        <Box sx={{ display: 'flex', flexWrap: 'wrap', gap: 0.5 }}>
+                            {selected.map((value: int) => {
```
KevinWu098:
```suggestion
                            {selected.map((value: number) => {
```
repeat: there is no `int` type in TS

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/ScheduleSelector.tsx
```diff
@@ -20,43 +25,45 @@ class ScheduleSelector extends PureComponent<ScheduleSelectorProps, ScheduleSele
...
+                                return <Chip key={value} label={this.props.scheduleNames[value]} />;
+                            })}
+                        </Box>
+                    )}
+                >
+                    {this.props.scheduleNames.map((name: string, index: int) => {
```
KevinWu098:
```suggestion
                    {this.props.scheduleNames.map((name: string, index: number) => {
```
repeat: there is no int type in TS

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/ScheduleSelector.tsx
```diff
@@ -20,43 +25,45 @@ class ScheduleSelector extends PureComponent<ScheduleSelectorProps, ScheduleSele
...
-            </FormGroup>
+            <FormControl style={{ marginTop: 10, maxWidth: 400 }} fullWidth variant="outlined">
+                <InputLabel id="schedule-select-label" htmlFor="select-multiple-chip">
+                    Select schedules
+                </InputLabel>
+                <Select
+                    labelId="schedule-select-label"
+                    id="schedule-select"
+                    size="small"
+                    style={{ marginTop: 2 }}
```
KevinWu098:
```suggestion
                    style={{ marginTop: 2 }}
```
nit: i think we can do without the margin top here

KevinWu098 (changes requested):
Thanks for submitting your first PR! I've added a good handful of (mostly) UI-related tweaks and code quality suggestions.

I promise that I'm not always this pedantic, but it is good to learn (and embrace) our codebase standards early! Let me know if you have any questions 🫡

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
xgraceyan: This might, and probably is, just be personal preference, but the spacing between the title--the start of the dialog--and the buttons/actions--the end--seems a bit off to me (see image below), and I thought perhaps padding out the actions so the spacing around both corners are consistent might look better. 

However, this seems to be for all MUI dialogs, and I agree it should need to be reflected across all modals if changed, so I've omitted the padding changes in my latest commit (for now, at least). It shouldn't be a huge change regardless and would look fine either way.

image

xgraceyan:
I've committed the requested changes to the best of my knowledge. I'm still familiarizing myself with Typescript, so I totally forgot I had put "int" instead of the actual number type haha. Please let me know if any other edits need to be made, and sorry for the late response--I was out for most of winter break, so I'm just getting around to this now!

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -167,42 +167,55 @@ function CustomEventDialogs(props: CustomEventDialogProps) {
                     </IconButton>
                 </Tooltip>
             )}
-            <Dialog open={open} onClose={handleClose} maxWidth={'lg'}>
+            <Dialog open={open} onClose={handleClose} maxWidth={'xs'}>
+                <DialogTitle id="form-dialog-title">
+                    {props.customEvent ? 'Edit a Custom Event' : 'Add a Custom Event'}
+                </DialogTitle>
                 <DialogContent>
```
KevinWu098:
```suggestion
                <DialogContent sx={{ display: 'flex', flexDirection: 'column', gap: 1 }}>
```

Similar to how we got rid of `marginLeft` and `marginRight` we should also avoid using margin across our components in favor of using the `gap` styling at the parent container

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -167,42 +167,55 @@ function CustomEventDialogs(props: CustomEventDialogProps) {
...
+                            required={true}
+                            value={title}
+                            onChange={handleEventNameChange}
+                            margin="dense"
+                        />
                     </FormControl>
-                    <form noValidate style={{ display: 'flex', gap: 5, marginTop: 5 }}>
+                    <FormControl
+                        fullWidth
+                        sx={{ display: 'flex', flexDirection: 'row', gap: '12px', marginTop: '12px' }}
```
KevinWu098:
```suggestion
                        sx={{ display: 'flex', flexDirection: 'row', gap: '12px' }}
```
repeat: margin is a bad pattern for maintainability

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/DaySelector.tsx
```diff
@@ -30,24 +26,25 @@ const DaySelector = ({ days = [false, false, false, false, false, false, false],
             sx={{
                 display: 'flex',
                 flexDirection: 'row',
-                justifyContent: 'center',
-                padding: 5,
+                justifyContent: 'space-between',
+                marginTop: 12,
+                marginBottom: 12,
```
KevinWu098:
```suggestion
```
repeat: margin ❌

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/ScheduleSelector.tsx
```diff
@@ -20,43 +24,44 @@ class ScheduleSelector extends PureComponent<ScheduleSelectorProps, ScheduleSele
...
-                                    color="primary"
-                                />
-                            }
-                            label={name}
-                            key={name}
-                        />
-                    );
-                })}
-            </FormGroup>
+            <FormControl style={{ marginTop: '12px', maxWidth: 400 }} fullWidth variant="outlined">
```
KevinWu098:
```suggestion
            <FormControl style={{ maxWidth: 400 }} fullWidth variant="outlined">
```

KevinWu098 (changes requested):
Nice work on the previous comments. Have a few more regarding the use of margin, but otherwise, it looks solid!

xgraceyan:
Gotcha, I've replaced all margins now -- thanks for the advice regarding gap styling.

KevinWu098 (approved):
LGTM

PR #1064, opened 2024-12-06
Write to both DynamoDB and RDS

MinhxNguyen7:
Needed to recreate because the other one had to be reset because it broke prod

Refer to #993


MinhxNguyen7:
I think (part of) the problem is because we pass the `USERDATA_TABLE_NAME` by getting it from the DDB that we create in the stack, and that isn't part of the environment variables, which is parsed by the migration script. That's why the deployment works, but the migration doesn't.

Where it's passed in.
image

We should be able to fix this by either not parsing it or hard-coding the table name as a repository/environment variable, or we can have a different env-var schema.

MinhxNguyen7:
I made different envs for RDS and DDB, which seems to have worked. I've reverted the application of the migration in staging, since multiple PRs that change the DB would break migrations.

PR #1069, opened 2024-12-07
Increase lambda timeout to 10 seconds

MinhxNguyen7:
## Summary
- We were having some timeouts on save, possibly because our DB queries were taking longer than the previous five-second limit.
- Manually increasing the limit on the current deployment fixed the issue.
- The problem might be that, with auto-save, we send multiple requests that have to written and overwritten sequentially, and the "queue time" can exceed five seconds.

## Future Follow-Up
- Have a timer so we don't auto-save every action.
- Somehow cancel the previous update to a schedule if a new one comes in. Not sure how this is possible.


ecxyzzy (approved):
LGTM, we really gotta fix the test erroring at some point...

PR #1077, opened 2024-12-14
Schedule Name Alt Text Expansion

### LINKED ISSUE #1072: Display schedule names on hover
Since the schedule dropdown has a width constraint, long schedule names are cut off. It would be nice to have the functionality to hover over a schedule name and see a small alt text pop up displaying the full schedule name, instead of having to change schedules to see its full schedule name.

CPT2412062221-282x338

Ardenjweed:
## Summary
Added an alt text pop up that displays the full names of schedules who's names were truncated.

Key features:
- Tooltip only appears for truncated schedule names
- Non-interactive tooltip that doesn't block clicking on other schedule items
- 200ms delay to prevent unwanted popups during quick mouse movements
- Maintains existing button styling and interaction behavior

AntalmanacHoverChange
## Test Plan
1. Create a schedule with a long name
2. Verify the name is truncated in the dropdown menu
3. Hover over the truncated name to confirm the tooltip appears showing the full name

## Issues

Closes #1072

### DIFF
--- a/apps/antalmanac/src/components/Calendar/toolbar/ScheduleSelect/ScheduleSelect.tsx
+++ b/apps/antalmanac/src/components/Calendar/toolbar/ScheduleSelect/ScheduleSelect.tsx
@@ -83,20 +83,20 @@ export function SelectSchedulePopover(props: { scheduleNames: string[] }) {
 
     return (
         <Box>
-            <Button
-                size="small"
-                color="inherit"
-                variant="outlined"
-                onClick={handleClick}
-                sx={{ minWidth, maxWidth, justifyContent: 'space-between' }}
-            >
-                <Tooltip title={currentScheduleName} enterDelay={200} disableInteractive placement="right">
+            <Tooltip title={currentScheduleName} enterDelay={200} disableInteractive>
+                <Button
+                    size="small"
+                    color="inherit"
+                    variant="outlined"
+                    onClick={handleClick}
+                    sx={{ minWidth, maxWidth, justifyContent: 'space-between' }}
+                >
                     <Typography whiteSpace="nowrap" textOverflow="ellipsis" overflow="hidden" textTransform="none">
                         {currentScheduleName}
                     </Typography>
-                </Tooltip>
-                <ArrowDropDownIcon />
-            </Button>
+                    <ArrowDropDownIcon />
+                </Button>
+            </Tooltip>
 
             <Popover
                 open={open}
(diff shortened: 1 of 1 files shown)


KevinWu098 (changes requested):
Thanks for working on this PR! I have some reservations about the direction that we're taking here to resolve #1072.

As implemented, `TooltipIfTruncated` certainly does work, but I think that it may be overly complicated (and potentially confusing) to other maintainers. `cloneElement` is even described by the React docs as being "uncommon and can lead to fragile code." 

I would prefer we choose a simpler path: wrap every schedule name in a tooltip, regardless of if the name is truncated

```jsx
<Tooltip title={name} enterDelay={200} disableInteractive placement="right">
```

Though certainly more "heavy-handed", I feel that the codebase would be better served by this simpler approach (primarily in readability and approachability).

Let me know if you have any questions!

Ardenjweed:
I have made the changes as requested, let me know if you need anything to be changed. 

As of right now the tooltip only shows if the user hovers over the text itself and not the whole button. Not sure if this is a change you would like to see or if how it works currently is perfect.

apps/antalmanac/src/components/Calendar/toolbar/ScheduleSelect/ScheduleSelect.tsx
```diff
@@ -90,9 +90,11 @@ export function SelectSchedulePopover(props: { scheduleNames: string[] }) {
                 onClick={handleClick}
                 sx={{ minWidth, maxWidth, justifyContent: 'space-between' }}
             >
-                <Typography whiteSpace="nowrap" textOverflow="ellipsis" overflow="hidden" textTransform="none">
-                    {currentScheduleName}
-                </Typography>
+                <Tooltip title={currentScheduleName} enterDelay={200} disableInteractive placement="right">
```
KevinWu098: issue: tooltip should apply to the entire button, not just the text

suggestion: wrap the whole button with the tooltip

KevinWu098: suggestion: Also, I think that this button (as the trigger) would look a bit better with the tooltip in the bottom (the default) position

```suggestion
                <Tooltip title={currentScheduleName} enterDelay={200} disableInteractive>
```

apps/antalmanac/src/components/Calendar/toolbar/ScheduleSelect/ScheduleSelect.tsx
```diff
@@ -116,17 +118,23 @@ export function SelectSchedulePopover(props: { scheduleNames: string[] }) {
...
                                 >
-                                    <Typography
-                                        overflow="hidden"
-                                        whiteSpace="nowrap"
-                                        textTransform="none"
-                                        textOverflow="ellipsis"
-                                    >
-                                        {name}
-                                    </Typography>
+                                    <Tooltip title={name} enterDelay={200} disableInteractive placement="right">
```
KevinWu098: repeat: wrap the entire button in the tooltip

apps/antalmanac/src/components/Calendar/toolbar/ScheduleSelect/ScheduleSelect.tsx
```diff
@@ -116,17 +118,23 @@ export function SelectSchedulePopover(props: { scheduleNames: string[] }) {
                                         justifyContent: 'flex-start',
                                         background:
                                             index === currentScheduleIndex ? theme.palette.action.selected : undefined,
+                                        '&:hover': {
+                                            backgroundColor: theme.palette.action.hover,
+                                            transition: 'background-color 0.2s',
+                                        },
```
KevinWu098: question: why are we adding this hover transition? If this is an arbitrary preference, it would be better to omit this from the PR, as AntAlmanac doesn't much use transitions

KevinWu098 (changes requested):
Thanks for the fixes! A few more comments on implementation, but otherwise it looks solid!

Ardenjweed:
I pushed new changes, let me know if these resolved the issues you had in mind!

KevinWu098 (approved):
LGTM! Thanks for working on this!

PR #1078, opened 2024-12-22
fix: leftpad section codes

KevinWu098:
## Summary
1. Courses with section codes less than 5 digits long are stored in AntAlmanac as is (`5001` -> `5001`). The API provides them as 5 digit leftpad strings. This was resulting in us not correctly loading in courses (most critically, BIO SCI).

Screenshot_2024-12-21_at_8 07 13_PM

Above: Saved AA Data
Below: API response

## Test Plan
1. Add and save courses from the 0 - 9,999 section code range
3. Loading the courses in should work as expected
4. Add and save courses from the 10,000 - 99,999 range.
5. Loading the courses should be as expected
6. Doing both together should work as expected

## Issues
Closes Bug Report


ecxyzzy (approved):
lgtm ship boat emoji etc

PR #1082, opened 2024-12-22
feat: search course info button

KevinWu098:
## Summary
1. Adds a Button to the Course Info Toolbar which routes them to the corresponding search result for that course (all lectures, discussions, etc)

chrome-capture-2024-11-22

## Test Plan
1. The button should only show on the added pane
3. Clicking the button should route to the correct search, with the correct URL parameters set (term, deptValue, courseNumber)

## Issues
From feedback:
> When making my schedule I like to go in two rounds. First, I select all the lectures I will be attending, but not the labs/discussions. Then on the second round I go back and add in the best timing of discussions/labs. In your app it is difficult to do this because there is no button that lets you immediately jump from the lecture to then seeing all of the associated discussions with that lecture, instead, I have to manually search up the name of the class.


MinhxNguyen7 (approved):
LGTM! Nice feature

PR #1083, opened 2024-12-22
feat: Course Shorcut Search From Calendar Events  

### LINKED ISSUE #1046: Adding Shortcuts for accessing courses
(This may be a feature I'm unaware of)

I've encountered situations where course conflicts require rearranging my schedule, but it’s a hassle to re-search and modify courses. It would be great to have a feature that allows quickly jumping between courses in my schedule, only if it seems feasible since it might mess with the current control flow that people are accustomed to.

image

here could be good too
image

alexespejo:
## Summary
Users can now quickly search for courses they've selected by double clicking the event in their calendar 

This is meant to enhance the user experience by reducing the redundancy of having re-search a course if a user wants to select or preview different time slots 

[Video Demo](https://vimeo.com/1041581472?share=copy#t=0)

## Test Plan
- Add different lec/dis/sem to the calendar and double clicking the event will bring up the course and all associated disc/lecs 
- Using the shortcut and searching shouldn't affect either actions (this was a problem in development)

## Issues
- This feature is locked for mobile to avoid clunky behavior 
- The search is only preformed when the user is on the Search tab. If the user is on the 'Added' or 'Map' the search won't display (this is subject to change based on how the feature feels)
 
Closes #1046

### DIFF
--- a/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
+++ b/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
@@ -9,11 +9,14 @@ import { Calendar, DateLocalizer, momentLocalizer, Views } from 'react-big-calen
 import CalendarToolbar from './CalendarToolbar';
 import CourseCalendarEvent, { CalendarEvent, CourseEvent } from './CourseCalendarEvent';
 
+import depts from '$components/RightPane/CoursePane/SearchForm/DeptSearchBar/depts';
 import locationIds from '$lib/location_ids';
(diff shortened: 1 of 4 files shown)


MinhxNguyen7:
Assuming that this PR is ready for review (since it's not a draft), can you resolve the merge conflict so that the staging workflow can run?

alexespejo:
> Assuming that this PR is ready for review (since it's not a draft), can you resolve the merge conflict so that the staging workflow can run?

Gimme a min some recent changes heavily conflict what I've been working on so I might have to restart

MinhxNguyen7:
Next time you can create a new branch and change the PR to merge from there.

PR #1086, opened 2024-12-23
feat: cache websoc

KevinWu098:
## Summary
1. Reintroduces websoc result caching, which was removed in #1031
2. Updates `eslint` at root to v9 with the [flat config](https://eslint.org/docs/latest/use/configure/migration-guide)
```js
npx @eslint/migrate-config .eslintrc.cjs
```

## Test Plan
1. Search for a course
2. Switch tabs or otherwise clear the search
3. Searching for the course again should hit the cache (no loading state), and it should also scroll to the appropriate height that you last left off on (#1001 was also regressed by #1031)
4. Searching for a different course should not hit that cache

## Issues
Closes #


MinhxNguyen7 (approved):
LGTM!

PR #1091, opened 2024-12-29
fix: add copy, rename, and delete as unsaved actions properly

### LINKED ISSUE #1090: fix: add delete and rename schedule as unsaved actions that can be restored
Currently, when a schedule is copied, it is saved locally as an unsaved action that can be restored. However, the actions for renaming and deleting schedules do not behave similarly, leaving inconsistencies in how unsaved actions are handled

jotalis:
## Summary
- Copying, renaming, and deleting schedules are now stored as unsaved actions and can be re-loaded if the user wants to 
- Fixes "an issue where duplicating a schedule that was not currently selected would incorrectly create a new schedule containing the events of the currently selected schedule, instead of duplicating the intended schedule," a bug also fixed in #1089; will fix merge conflicts after either one gets merged into main
- Slight code refactoring for consistency

## Test Plan
- Create several schedules with some events. Duplicate, rename, and delete some. Take note of the order and state of these schedules. Reload the page without saving these changes and load the unsaved changes. Ensure the loaded changes match the schedule states prior to reloading

Closes #1090

### DIFF
--- a/apps/antalmanac/src/components/dialogs/AddSchedule.tsx
+++ b/apps/antalmanac/src/components/dialogs/AddSchedule.tsx
@@ -13,7 +13,7 @@ function AddScheduleDialog({ onClose, onKeyDown, ...props }: DialogProps) {
     const isDark = useThemeStore((store) => store.isDark);
 
     const [name, setName] = useState(
-        AppStore.getNextScheduleName(AppStore.getDefaultScheduleName(), AppStore.getScheduleNames().length)
+        AppStore.getNextScheduleName(AppStore.getScheduleNames().length, AppStore.getDefaultScheduleName())
     );
 
     const handleCancel = () => {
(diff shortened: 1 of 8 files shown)


apps/antalmanac/src/components/dialogs/AddSchedule.tsx
```diff
@@ -13,7 +13,7 @@ function AddScheduleDialog({ onClose, onKeyDown, ...props }: DialogProps) {
     const isDark = useThemeStore((store) => store.isDark);
 
     const [name, setName] = useState(
-        AppStore.getNextScheduleName(AppStore.getDefaultScheduleName(), AppStore.getScheduleNames().length)
+        AppStore.getNextScheduleName(AppStore.getScheduleNames().length, AppStore.getDefaultScheduleName())
```
KevinWu098:
```suggestion
        AppStore.getNextScheduleName(() => AppStore.getScheduleNames().length, AppStore.getDefaultScheduleName())
```

issue: passing in a function call will result in it being called on rerenders, though not being used

suggestion: wrap it in an arrow function

KevinWu098 (changes requested):
LGTM, great feature improvement, however there's still a bit of bugginess as newly created schedules aren't logged in the action states, and thus actions upon newly created schedules then don't get applied

KevinWu098 (changes requested):
LGTM, just the one comment above

KevinWu098 (approved):
LGTM 🚀

PR #1092, opened 2024-12-30
feat: Class Search Shortcut

### LINKED ISSUE #1046: Adding Shortcuts for accessing courses

alexespejo:
## Summary
Users can now quickly search for courses they've selected by clicking a shortcut in in their calendar 

A new button is added to the course popup when you click on the calendar event which will auto search the course offerings. Additionally,  `cmd/ctrl` + `click` on a calendar event will perform the search.

This is meant to enhance the user experience by reducing the redundancy of having re-search a course if a user wants to select or preview different time slots 

image

https://github.com/user-attachments/assets/e0d00d72-2bad-4ae9-9b40-fc7a56ea59f3

## Test Plan
1. Add different lec/dis/sem to the calendar, click the calendar event and click the title in the popup to search for the course will bring up the course and all associated disc/lecs 
2. Do the same for 1. but use `cmd/ctrl` + `click` to search the course
3. If you're on the `Added` or `Map` pages and you try either 1. or 2. nothing should happen 
4. Using the shortcut and searching shouldn't affect either actions (this was a problem in development)

## Issues
- This feature is locked for mobile to avoid clunky behavior 
- The search is only preformed when the user is on the Search tab. If the user is on the 'Added' or 'Map' the search won't display (this is subject to change based on how the feature feels)
 
Closes #1046


apps/antalmanac/src/components/RightPane/SectionTable/CourseInfo/CourseInfoSearchButton.tsx
```diff
@@ -20,7 +21,6 @@ export function CourseInfoSearchButton({ courseDetails, term }: { courseDetails:
         RightPaneStore.updateFormValue('deptValue', deptCode);
         RightPaneStore.updateFormValue('courseNumber', courseNumber);
         RightPaneStore.updateFormValue('term', term);
-
         displaySections();
         setActiveTab(1);
     }, []);
```
KevinWu098:
```suggestion
    }, [courseNumber, deptCode, displaySections, setActiveTab, term]);
```

issue: this callback has incorrect dependency array. (I accidentally introduced this while my eslint was fritzing)

apps/antalmanac/src/stores/QuickSearchStore.ts
KevinWu098, line 1: question: what is the purpose of a Zustand store here? Or rather, why do we need to store / share the state of the quick search?

apps/antalmanac/src/components/Calendar/CalendarCourseEventWrapper.tsx
```diff
@@ -22,8 +24,12 @@ export const CalendarCourseEventWrapper = ({ children, ...props }: CalendarCours
         (e: React.MouseEvent) => {
             e.preventDefault();
             e.stopPropagation();
-
-            setSelectedEvent(e, props.event);
+            if (props.event && (e.metaKey || e.ctrlKey)) {
```
KevinWu098, line 29:
```suggestion

            if (props.event && (e.metaKey || e.ctrlKey)) {
```
nit: i like new lines 🙃

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneRoot.tsx
```diff
@@ -28,6 +29,14 @@ export function CoursePaneRoot() {
         }
     }, [displaySections, forceUpdate]);
 
+    useQuickSearchStore.subscribe((state) => {
+        const decomp = state.value;
+        RightPaneStore.updateFormValue('deptLabel', decomp.deptLabel);
+        RightPaneStore.updateFormValue('deptValue', decomp.deptValue);
+        RightPaneStore.updateFormValue('courseNumber', decomp.courseNumber);
+        handleSearch();
+    });
```
KevinWu098: issue: we shouldn't need to use a store to update the values. The event emitter, in this case, the calendar event, should be able to call `forceUpdate()` via the `CoursePaneStore`. This makes our data flow more clear, removing the store step.

note: `forceUpdate()` is pretty poorly named, as it's more akin to a refetch, but it in essence "forces" a rerender of the `CourseRenderPane` which causes it to fetch data again with the new params

suggestion: rather than using a subscriber, use logic similar to `CourseInfoSearchButton` (and add a `forceUpdate()` to it)

```jsx
    const handleClick = useCallback(() => {
        RightPaneStore.updateFormValue('deptValue', deptCode);
        RightPaneStore.updateFormValue('courseNumber', courseNumber);
        RightPaneStore.updateFormValue('term', term);
        displaySections();
        setActiveTab(1);
    }, [courseNumber, deptCode, displaySections, forceUpdate, setActiveTab, term]);
```

apps/antalmanac/src/lib/helpers.ts
```diff
@@ -18,4 +21,20 @@ export async function clickToCopy(event: MouseEvent<HTMLElement>, sectionCode: s
     openSnackbar('success', 'WebsocSection code copied to clipboard');
 }
 
+export function quickSearchForClasses(title: string, term: string) {
+    const decompCourseInfo: string[] | undefined = title.match(/^(.*)\s(\S+)$/)?.slice(1);
+    const tabNum = useTabStore.getState().activeTab;
+
+    // tabNum == 1 locks this feature to only the Search page (even on mobile)
```
KevinWu098: question: why can we only do this when we're on the search pane? I'd imagine that if a user clicks quick search, they'd expect that it would quick search, regardless of the pane they're on. 

Alternatively, if you feel strongly (and defend this design choice), the option to quick search should only be available to users when they're on the search pane.

apps/antalmanac/src/lib/helpers.ts
```diff
@@ -18,4 +21,20 @@ export async function clickToCopy(event: MouseEvent<HTMLElement>, sectionCode: s
     openSnackbar('success', 'WebsocSection code copied to clipboard');
 }
 
+export function quickSearchForClasses(title: string, term: string) {
+    const decompCourseInfo: string[] | undefined = title.match(/^(.*)\s(\S+)$/)?.slice(1);
```
KevinWu098: issue: regex is not immediately readable; any use of regex should ideally be accompanied with a clarifying comment (and, optionally, a link to a regex breakdown tool)

suggestion: add a comment

apps/antalmanac/src/lib/helpers.ts
```diff
@@ -18,4 +21,20 @@ export async function clickToCopy(event: MouseEvent<HTMLElement>, sectionCode: s
...
+    const tabNum = useTabStore.getState().activeTab;
+
+    // tabNum == 1 locks this feature to only the Search page (even on mobile)
+    if (decompCourseInfo && tabNum == 1) {
+        const deptIdx: number = depts.findIndex((item) => item.deptValue === decompCourseInfo[0]);
+        useQuickSearchStore.getState().setValue({
+            term: term,
+            deptLabel: depts[deptIdx].deptLabel,
+            deptValue: decompCourseInfo[0],
+            courseNumber: decompCourseInfo[1],
```
KevinWu098: issue: although your logic and result are not inherently wrong, we shouldn't need to re-derive the deptValue and courseNumber. 

At some point in our codebase (see below), we stripped out that information. We should probably retain that info now, given this new use case.

<details> 
  <summary>Finding the right file is left as an exercise for you ;) (answer below)</summary>
  `calendarizeHelpers.ts`
</details>

note: if you update a type as widely used in our codebase as `CourseEvent`, you'll also have to ensure that the, in this case, 23 usages do not regress. This may include updating tests and business logic.

KevinWu098 (changes requested):
Solid PR, I think users will be really happy with this improvement to the user flow around the app.

That said, I have a few questions about the approach we're taking in this PR. More specifically, I don't feel that we need to use a new store in order to implement the logic for this feature. 

Short of refactoring `RightPaneStore` to a Zustand store, I think this PR should reuse logic from `CourseInfoSearchButton` and utilize `RightPaneStore` and `CoursePaneStore` over creating another store with duplicate states

apps/antalmanac/src/stores/QuickSearchStore.ts
alexespejo: In my first PR, I implemented the feature using a store. At the time, it seemed like there were no existing stores that met the specific requirements, so creating a separate global state that components could listen to for updates made sense. I also misunderstood how the search functions worked, so I assumed sharing state between the two components (left and right panes) would be necessary. The more you know fr

apps/antalmanac/src/lib/helpers.ts
alexespejo: I'm not sure what you mean. Like I don't have to use the regex to separate the deptValue and courseNumber?

KevinWu098: Ah sorry, didn't intend my comment to be so cryptic.

To rephrase: your regex's result isn't incorrect, however the usage of regex to re-derive the Department Code and Course Number via the course title is, put less eloquently, "jank".

The type you're working with, `CourseEvent`, at some point had certain fields removed from it (dept code, course number).

What I'm suggesting is that you identify where we omitted those fields from CourseEvent, and then include them so that you can have access to the values via CourseEvent. Then, no regex and re-deriving is needed.

alexespejo: Ohh ok gotcha 👌

alexespejo:
> Solid PR, I think users will be really happy with this improvement to the user flow around the app.
> 
> That said, I have a few questions about the approach we're taking in this PR. More specifically, I don't feel that we need to use a new store in order to implement the logic for this feature.
> 
> Short of refactoring `RightPaneStore` to a Zustand store, I think this PR should reuse logic from `CourseInfoSearchButton` and utilize `RightPaneStore` and `CoursePaneStore` over creating another store with duplicate states

Apologies for the slow development speed

KevinWu098 (approved):
LGTM 🚀 

Only nit I would have is just truncating `useQuickSearchForClasses` to `useQuickSearch`, since you're shortening it anyways when you use it (`quickSearch()`)

PR #1094, opened 2024-12-30
fix: remove box messing with tooltip offset

### LINKED ISSUE #1093: fix: button tooltip offset consistency
Tooltip offset for Copy Schedule is not consistent with the other buttons.

chrome-capture-2024-12-29

jotalis:
## Summary
The tooltip offset inconsistency for `Copy Schedule` was caused by wrapping the tooltip in a Box. This was changed to an empty tag to match what was used for the other tooltips.

## Test Plan
Hover over the buttons and ensure the tooltips are all the same offset
chrome-capture-2024-12-30

Closes #1093


KevinWu098 (approved):
LGTM

PR #1095, opened 2024-12-31
Redirect to the root route on undefined route

### LINKED ISSUE #1047: Redirect to home page after invalid route
## Description
- Right now, the page just errors with a wrong route.
- That's fine, but it would be more convenient if we just redirect to the main route.

Dezert01:
## Summary

Add a redirection to the main route if the given route is undefined

## Test Plan

Any other route than /xyz should be replaced now with root route "/"

## Issues

Closes #1047 

## Future Followup

Consider enhancing this redirection logic by checking if a valid route exists higher in the navigation stack and redirecting to it instead of the root route.


Dezert01:
@MinhxNguyen7 Hi! Could you check this out and let me know if it’s what you had in mind? Open to any feedback!

MinhxNguyen7 (approved):
LGTM!

PR #1097, opened 2025-01-01
Add ISEB to building catalogue

### LINKED ISSUE #1085: fix: ISEB is not included in buildingCatalogue.ts

### Relevant Files

### Your Task

From user feedback:

vinnyho:
## Summary
- Added ISEB to building catalogue.
- Used Google Maps for coordinates.
## Test Plan
- Tested with custom event creation.
- Tested on map.
## Issues

Closes #1085


MinhxNguyen7 (approved):
Looks good, but it would be nice to have an image for it.

PR #1098, opened 2025-01-01
feat: add terms to custom events

### LINKED ISSUE #762: Download ICS: Assign CourseEvent term to CustomEvent
Right now we are not assigning specific terms to a custom event when a user creates it.

We will grab the term from CourseEvent in the schedule.
- if the terms are inconsistent (e.g. some classes are in the 2023 Fall, some are in the 2023 Spring), pick the latest one.
- if there is no CourseEvent in the schedule, use the default term.

Slo1k:
## Summary
The change addresses the issue of assigning a term to a custom event when it is created. Previously, when users created custom events, no term was assigned to them. With this update, we now ensure that each custom event is associated with a term.

The term is obtained from the associated ScheduleCourse in the schedule. If the schedule contains classes from multiple terms (e.g., Fall 2023 and Spring 2023), the latest term is selected. If no term is found in the schedule, the default term will be used.

## Test Plan
1. Test creation of a custom event with a single term in the schedule. Verify that the custom event is assigned the correct term.
2. Test creation of a custom event with multiple terms in the schedule. Ensure the latest term is assigned to the custom event.
3. Test creation of a custom event with no associated terms in the schedule. Verify that the default term is applied.
4. Ensure that edge cases like missing terms or incorrect term names in the schedule are handled correctly and fall back to the default term.
5. Verify that custom events are correctly displayed with their respective terms in calendar view - CustomEventDialog.
6. Test that after removing some scheduleCourses, the correct term is still assigned to the custom event, reflecting the updated state of the schedule.

## Issues
Closes #762

### DIFF
--- a/apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
+++ b/apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
@@ -286,20 +286,22 @@ const CourseCalendarEvent = (props: CourseCalendarEventProps) => {
             <Paper className={classes.customEventContainer} ref={paperRef}>
                 <div className={classes.title}>{title}</div>
                 {building && (
-                    <div className={classes.table}>
-                        Location:&nbsp;
-                        <Link
-                            className={classes.clickableLocation}
-                            to={`/map?location=${building ?? 0}`}
-                            onClick={focusMap}
-                        >
-                            {buildingCatalogue[+building]?.name ?? ''}
-                        </Link>
-                    </div>
+                    <tr className={classes.table}>
+                        <td className={classes.alignToTop}>Location</td>
+                        <td className={classes.rightCells}>
+                            <Link
+                                className={classes.clickableLocation}
+                                to={`/map?location=${building ?? 0}`}
+                                onClick={focusMap}
+                            >
+                                {buildingCatalogue[+building]?.name ?? ''}
+                            </Link>
+                        </td>
+                    </tr>
                 )}
                 {term && (
                     <tr className={classes.table}>
-                        <td className={classes.alignToTop}>Term:&nbsp;</td>
+                        <td className={classes.alignToTop}>Term</td>
                         <td className={classes.rightCells}>{term}</td>
                     </tr>
                 )}
--- a/apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
+++ b/apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
@@ -131,6 +131,7 @@ export interface CustomEvent extends CommonCalendarEvent {
     isCustomEvent: true;
     building: string;
     days: string[];
+    term: string;
 }
 
 export type CalendarEvent = CourseEvent | CustomEvent;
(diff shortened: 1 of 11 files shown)


Slo1k:
@MinhxNguyen7   Hi! Could you review this pull request when you have a moment? Thanks!

apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
```diff
@@ -296,6 +297,12 @@ const CourseCalendarEvent = (props: CourseCalendarEventProps) => {
                         </Link>
                     </div>
                 )}
+                {term && (
```
MinhxNguyen7, line 302: This should be styled in the same way as normal sections' pop-ups, i.e., without a colon and in two columns. 

I know that this follows the styling of the location in this popup, but that was a mistake. So, while you're at it, could you make the same change to the location row? Thanks!

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -118,6 +122,7 @@ function CustomEventDialogs(props: CustomEventDialogProps) {
             end: end,
             customEventID: props.customEvent ? props.customEvent.customEventID : Date.now(),
             building: building,
+            term: getLatestTermByShortName(termsShortNames)?.shortName,
```
MinhxNguyen7: What does this do? If it gets the latest term of all sections in the schedule, this should be assigned to an intermediate variable (something like `latestTermInSchedules`).

MinhxNguyen7 (changes requested):
Looks alright on the front-end with a minor suggestion. 

However, you need to implement the backend to save it as well. For that, you need to add the term to the DB schema and save/get it from the backend endpoints. You can find the custom event schema in `apps/backend/src/db/schema/schedule/custom_event.ts`. For saving and retrieving, edit `getGuestUserData` and `upsertGetUserData`. There's not that much to change, but you'll need to read the code to see what has already been inferred when you changed `RepeatingCustomEventSchema`. Let me know if you need help with that.

That highlights the fact that the test plan above needs to account for saving and loading user account as well.

Slo1k:
Thanks for the feedback!

I’ve made the changes on the frontend as suggested and also tried to adapt the backend by adding the term to the DB schema and updating the relevant endpoints (getGuestUserData and upsertGuestUserData). However, I’m having trouble running the backend locally to verify if the migration with the new field works correctly.

This is my first time working with tRPC, so I’m not entirely sure how to properly set up my local environment for the backend. Could you either check the changes on your side or provide a detailed explanation of how to configure the backend locally to ensure everything works as expected?

Thanks in advance!

MinhxNguyen7:
I'll check the changes on my side, and I'm working on updating the README to reflect our stack after a recent migration. 

Sorry for the inconvenience. I didn't think about this when I marked the issue as a good first task.

MinhxNguyen7:
@Slo1k You can refer to the information added in #1110 to set up locally, if you'd like. I'll try to get reviewing and testing this locally as soon as I can

Slo1k:
Thank you for your engagement! Today, I will try to test my solution with the help of the new documentation.

KevinWu098:
Closing as stale

PR #1100, opened 2025-01-07
Added ISEB and Image to Building Catalogue

### LINKED ISSUE #1085: fix: ISEB is not included in buildingCatalogue.ts

vinnyho:
## Summary
- Added ISEB to building catalogue.
- Used Google Maps for coordinates.
- Added image for its marker
## Test Plan
- Tested on map with ISEB marker

## Issues
Closes #1085

### DIFF
--- a/apps/antalmanac/src/lib/buildingCatalogue.ts
+++ b/apps/antalmanac/src/lib/buildingCatalogue.ts
@@ -1789,7 +1789,7 @@ export default {
         imageURLs: [],
     },
     83063: {
-        name: 'Interdisciplinary Science and Engineering Building  (ISEB)',
+        name: 'Interdisciplinary Science and Engineering Building (ISEB)',
         lat: 33.6431,
         lng: -117.8437,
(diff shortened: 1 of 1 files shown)


vinnyho:
@MinhxNguyen7 I apologize for the oversight of not adding the image and also creating a new pull request, this was due to me realizing I had forgotten to change branches and there was a lot of commits. I added the `imageURLs` based on the images from [Concept3D Map](https://map.concept3d.com/?id=463#!ct/61979?m/563348?s/iseb). The image link is in .png format, and although the URL includes some unusual sizing parameters, it works. Though, the other images are in .jpg format and have a different URL structure, but the ISEB link seems to function just fine. I hope this is good now! Thank you

MinhxNguyen7 (approved):
Looks good. Don't worry about it

KevinWu098:
I'm reopening the issue -- there's a more proper way to update building catalogue that slipped my mind -- for now this is good bandaid but will do a more proper write up later

PR #1068, opened 2024-12-06
feat: integrate larc

KevinWu098:
## Summary
1. Implements the [LARC endpoint from the Anteater API](https://anteaterapi.com/reference#tag/larc/GET/v2/rest/larc)

Screenshot 2024-12-06 at 2 47 52 AM

## To-Do
- [x] Refactor SearchTable to support LARC sections
- [x] Fetch LARC data
- [x] Display LARC sections alongside courses (when individually searched)
- [ ] Add courses to schedule

## Issues
Closes #432

PR #1102, opened 2025-01-09
refactor: section table body row

KevinWu098:
## Summary
1. Extracts row cells into individual components
2. Removes `withStyle`
3. Migrates to MUI v5

## Test Plan
1. Add a course, yellow highlight should be present
2. Add a course, other courses at the same time should be darkened if they conflict
3. Clicking blue links / text should still do the appropriate functionality
4. The row should look the same as it did before

## Issues
Part of #1068, chunking out the refactor first


jotalis (approved):
LGTM!

jotalis (approved):
lgtm

PR #1103, opened 2025-01-09
bug: disabled quick search from custom event

alexespejo:
## Summary
Disables the `ctrl/cmd` + `click` search feature on custom events that would give a misleading error message (see below)

image

## Issues

Closes #


KevinWu098 (approved):
LGTM

PR #1105, opened 2025-01-10
refactor: schedule management and tab store

### LINKED ISSUE #1104: Mobile map-linking opens added tab instead of map
## Description
- On mobile, clicking on the location in the section table or calendar opens the added tab.
- This is not a problem on desktop.

KevinWu098:
## Summary
This PR refactors two sections of our app: Schedule Management (the tab UI) and `TabStore` (the global tab store).

### Schedule Management
Schedule Management Tabs have been refactored to be largely simpler and require only one component for both desktop and mobile. Tab Content (i.e. Tab Panels) have been refactored into a separate component. No logic has changed here.

### Tab Store
Tab Store has also been refactored to largely be simpler. Previously, we would not render the Calendar tab on desktop, as it would be shown on the left hand side. However, this would result in inconsistent and doubled-logic to handle tab switching across both mobile and desktop.

Now, all tabs are rendered so the index of each (calendar = 0, search = 1, ...) are always consistent. The setter for `TabStore` now has two variants: `setActiveTabValue` for setting the value of `activeTab` via value (`number`) and `setActiveTab` for setting the value via a string literal union. `setActiveTab` will be used in all cases where we programmatically change the tab (e.g. `focusMap`), and `setActiveTabValue` will be used for event handlers.

Examples:

```
        setActiveTab('map');
    }, [setActiveTab]);
```

```
    const onChange = (_event: React.SyntheticEvent, value: number) => {
        setActiveTabValue(value);
    };
```

## Test Plan
0. Perform these tests on both desktop and mobile
1. Click on each tab, the correct panel should render as expected
2. Visit `/added`, `/map`, etc. The correct panel should be selected
3. Use `quickSearchForClasses`. Search panel should open.
4. Click the map link in the Section Table Body. The Map panel should open.
5. Having no signed in user should go to the search
6. Having a signed in user should go to added on desktop and calendar on mobile.

## Issues

Closes #1104


apps/antalmanac/src/components/ScheduleManagement/ScheduleManagementTabs.tsx
```diff
@@ -0,0 +1,107 @@
...
+    const isMobile = useMediaQuery(theme.breakpoints.down('sm'));
+
+    const onChange = (_event: React.SyntheticEvent, value: number) => {
+        setActiveTabValue(value);
+    };
+
+    return (
+        <Paper elevation={0} variant="outlined" square sx={{ borderRadius: '4px 4px 0 0' }}>
+            <Tabs value={activeTab} onChange={onChange} indicatorColor="primary" variant="fullWidth" centered>
+                {scheduleManagementTabs.map((tab) => {
```
MinhxNguyen7: Just use a parentheses arrow function instead of curly braces and a `return` statement. 

That being said, I think it's worthwhile to make this a component/factory function in itself. I generally think that, when you're making components in a `map`, the inside should be named instead of just a chunk of code.

apps/antalmanac/src/components/ScheduleManagement/ScheduleManagementTabsContent.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { ScheduleCalendar } from '$components/Calendar/CalendarRoot';
+import { AddedCoursePane } from '$components/RightPane/AddedCourses/AddedCoursePane';
+import { CoursePaneRoot } from '$components/RightPane/CoursePane/CoursePaneRoot';
+import { useThemeStore } from '$stores/SettingsStore';
+import { useTabStore } from '$stores/TabStore';
+
+
+const UCIMap = lazy(() => import('../Map/Map'));
+
+export function ScheduleManagementTabsContent() {
```
MinhxNguyen7: Can you add a comment (or even better yet, rename things) such that the distinction between this and `ScheduleManagementTabs`? It took me way too long to see that.

MinhxNguyen7 (changes requested):
Only a few suggestions. Everything seems to work

apps/antalmanac/src/components/ScheduleManagement/ScheduleManagementTab.tsx
```diff
@@ -0,0 +1,39 @@
...
+            icon={tab.icon}
+            iconPosition={isMobile ? 'top' : 'start'}
+            sx={{
+                ...(!isMobile
+                    ? {
+                          minHeight: 'auto',
+                          height: '44px',
+                          padding: 3,
+                          minWidth: '33%',
+                      }
```
MinhxNguyen7, line 37: I'd prefer to extract this to `mobileStyles`, but it's not a big deal

MinhxNguyen7 (changes requested):
The map link doesn't seem to work

apps/antalmanac/src/components/ScheduleManagement/ScheduleManagementTab.tsx
KevinWu098: Agree that the `sx` is rather messy here, but I'd rather keep all the styling logic together, rather than extract a chunk

MinhxNguyen7 (approved):
LGTM!

PR #1106, opened 2025-01-12
Fix mobile location tab

### LINKED ISSUE #1104: Mobile map-linking opens added tab instead of map

vinnyho:
## Summary
- Implemented mobile screen check using `useMediaQuery`
- Updated `focusMap` function to set the active tab based on screen size

## Test Plan
- Tested functionality on mobile screen
- Ensured functionality on desktop screen
## Example
Animation

## Issues
Closes #1104

### DIFF
--- a/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody/SectionTableBodyCells/LocationsCell.tsx
+++ b/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody/SectionTableBodyCells/LocationsCell.tsx
@@ -2,9 +2,9 @@ import { Box } from '@mui/material';
 import { useMediaQuery } from '@mui/material';
 import { WebsocSectionMeeting } from '@packages/antalmanac-types';
 import { Fragment, useCallback } from 'react';
-import { Link } from 'react-router-dom';
 
 import { MOBILE_BREAKPOINT } from '../../../../../globals';
+import MapLink from '../../../../buttons/MapLink';
 
 import { TableBodyCellContainer } from '$components/RightPane/SectionTable/SectionTableBody/SectionTableBodyCells/TableBodyCellContainer';
 import locationIds from '$lib/location_ids';
@@ -34,16 +34,12 @@ export const LocationsCell = ({ meetings }: LocationsCellProps) => {
                         const buildingId = locationIds[buildingName];
                         return (
                             <Fragment key={meeting.timeIsTBA + bldg}>
-                                <Link
-                                    style={{
-                                        textDecoration: 'none',
-                                    }}
-                                    to={`/map?location=${buildingId}`}
-                                    onClick={focusMap}
-                                    color={isDark ? 'dodgerblue' : 'blue'}
-                                >
-                                    {bldg}
-                                </Link>
+                                <MapLink
+                                    buildingId={buildingId}
+                                    buildingName={bldg}
+                                    focusMap={focusMap}
+                                    isDark={isDark}
+                                />
                                 <br />
                             </Fragment>
                         );
--- a/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody/SectionTableBodyCells/LocationsCell.tsx
+++ b/apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody/SectionTableBodyCells/LocationsCell.tsx
@@ -1,9 +1,7 @@
 import { Box } from '@mui/material';
-import { useMediaQuery } from '@mui/material';
 import { WebsocSectionMeeting } from '@packages/antalmanac-types';
-import { Fragment, useCallback } from 'react';
+import { Fragment } from 'react';
 
-import { MOBILE_BREAKPOINT } from '../../../../../globals';
 import MapLink from '../../../../buttons/MapLink';
 
 import { TableBodyCellContainer } from '$components/RightPane/SectionTable/SectionTableBody/SectionTableBodyCells/TableBodyCellContainer';
@@ -19,11 +17,6 @@ interface LocationsCellProps {
 export const LocationsCell = ({ meetings }: LocationsCellProps) => {
     const isDark = useThemeStore((store) => store.isDark);
     const { setActiveTab } = useTabStore();
-    const isMobileScreen = useMediaQuery(`(max-width: ${MOBILE_BREAKPOINT})`);
-
-    const focusMap = useCallback(() => {
-        setActiveTab(isMobileScreen ? 3 : 2);
-    }, [setActiveTab]);
 
     return (
         <TableBodyCellContainer>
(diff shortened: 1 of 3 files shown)


apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody/SectionTableBodyCells/LocationsCell.tsx
```diff
@@ -16,9 +19,10 @@ interface LocationsCellProps {
 export const LocationsCell = ({ meetings }: LocationsCellProps) => {
     const isDark = useThemeStore((store) => store.isDark);
     const { setActiveTab } = useTabStore();
+    const isMobileScreen = useMediaQuery(`(max-width: ${MOBILE_BREAKPOINT})`);
 
     const focusMap = useCallback(() => {
-        setActiveTab(2);
+        setActiveTab(isMobileScreen ? 3 : 2);
```
MinhxNguyen7: I believe you need to add the `isMobileScreen` to the `useCallback`'s dependency array to allow this fix to work when the screen is resized.

MinhxNguyen7 (commented):
I know this is scope creep, but seeing that his change had to be performed in two places reminded me that the link/button itself should be extracted to its own component (e.g., `MapLink`) in `components/buttons`. Would you mind doing that?

Also, it seems a little janky to set the tab index depending on a media query, but it's not obvious to be what would be a better way other than to completely refactor the tab store to use keys instead of indices, which is completely out of scope here.

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody/SectionTableBodyCells/LocationsCell.tsx
vinnyho: I think in putting the `isMobileScreen` to the `useCallback`'s dependency array causes the whole page to want to refresh for some reason, while putting it before doesn't.

vinnyho:
Please let me know if there are any changes that needs to be made, it worked well on my end with the new component on both mobile and desktop. Thank you!

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody/SectionTableBodyCells/LocationsCell.tsx
```diff
@@ -30,16 +34,12 @@ export const LocationsCell = ({ meetings }: LocationsCellProps) => {
...
-                                    to={`/map?location=${buildingId}`}
-                                    onClick={focusMap}
-                                    color={isDark ? 'dodgerblue' : 'blue'}
-                                >
-                                    {bldg}
-                                </Link>
+                                <MapLink
+                                    buildingId={buildingId}
+                                    buildingName={bldg}
+                                    focusMap={focusMap}
```
MinhxNguyen7: The point to the MapLink is to move the map-focusing business logic into there so that we don't have to keep re-writing the `focusMap` component.

apps/antalmanac/src/components/buttons/MapLink.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+
+interface MapLinkProps {
+    buildingId: number;
+    buildingName: string;
```
MinhxNguyen7: This is a misnomer, since it has both the building and the room number

MinhxNguyen7 (changes requested):
Just a few little things

vinnyho:
I don't get why but I'm getting `AntAlmanac\apps\antalmanac\src\components\Calendar\CourseCalendarEvent.tsx`
 `error  'Link' is defined but never used. Allowed unused vars must match /^_/u  @typescript-eslint/no-unused-vars` even if `Link` was never imported. Could you check it out for me based on the changes made? 

Other than that, I moved the focusMap into the MapLink component, changed buildingName to room, and added the MapLink component to custom events. 

I will make further necessary changes to fix that error without needed to not verify it. Thank you again

KevinWu098 (changes requested):
This PR's primary consideration, the buggy mobile tab switching, was implemented in #1105, which does in fact refactor our tabs into absolute string keys and removes differences between desktop and mobile tab handling. 

Still, I like the work being done here on `MapLink` and would recommend splitting out that work into a separate PR that we can get merged separate to the tabs

CC @MinhxNguyen7

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody/SectionTableBodyCells/LocationsCell.tsx
```diff
@@ -30,16 +27,12 @@ export const LocationsCell = ({ meetings }: LocationsCellProps) => {
...
-                                    onClick={focusMap}
-                                    color={isDark ? 'dodgerblue' : 'blue'}
-                                >
-                                    {bldg}
-                                </Link>
+                                <MapLink
+                                    buildingId={buildingId}
+                                    room={bldg}
+                                    setActiveTab={setActiveTab} 
+                                    isDark={isDark}
```
KevinWu098, line 34:
```suggestion
```
issue: `MapLink` should directly call the stores, rather than having them passed in as props

apps/antalmanac/src/components/buttons/MapLink.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { MOBILE_BREAKPOINT } from '../../globals';
+
+interface MapLinkProps {
+    buildingId: number;
+    room: string;
+    isDark: boolean;
+    setActiveTab: (tab: number) => void;
+}
+
+const MapLink: React.FC<MapLinkProps> = ({ buildingId, room, isDark, setActiveTab }) => {
```
KevinWu098, line 13:
```suggestion
const MapLink = ({ buildingId, room, isDark, setActiveTab }: MapLinkProps) => {
```
chore: consistency in how we type components

apps/antalmanac/src/components/buttons/MapLink.tsx
```diff
@@ -0,0 +1,35 @@
...
+                textDecoration: 'none',
+                color: isDark ? 'dodgerblue' : 'blue',
+            }}
+        >
+            {room}
+        </Link>
+    );
+};
+
+export default MapLink;
```
KevinWu098, line 35:
```suggestion
```
chore (repeat): prefer named exports

apps/antalmanac/src/components/buttons/MapLink.tsx
KevinWu098, line 13:
```suggestion
export const MapLink: React.FC<MapLinkProps> = ({ buildingId, room, isDark, setActiveTab }) => {
```
chore: prefer named exports

KevinWu098 (changes requested):
also some additional feedback on `MapLink`

vinnyho:
Got it, thanks Kevin for the suggestions I'll get to it

KevinWu098:
> Got it, thanks Kevin for the suggestions I'll get to it

Just to clarify, the ask is primarily that the PR shift its attention to refactoring out the repeated `focusMap` component. You can either create a new PR, or modify this PR, removing the now-defunct mobile tab fixes and adding the suggested fixes. thanks!

PR #1108, opened 2025-01-13
feat: consolidate "help" elements into help menu

### LINKED ISSUE #1053: feat: consolidate "help" elements into help menu

chrome-capture-2024-10-22

vicksey:
## Summary
Aggregated the "help" elements, Help Box, Feedback Form, and Tutorial, into one help menu button. Once hovered the three elements are shown and can be clicked to be used.

## Test Plan
- [ ] Hover over the Help Menu to ensure all necessary components appear
- [ ] Click through each element to test that they work like they normally do elsewhere
- [ ] Restart Tutorial button works
- [ ] Feedback Form opens in new tab
- [ ] Help menu appears
- [ ] See if size and design is good

## Issues

Closes #1053


.turbo/daemon/759b514fb2013120-turbo.log.2024-12-20
```diff
@@ -0,0 +1,2 @@
+2024-12-20T19:43:09.183783Z  WARN daemon_server: turborepo_lib::commands::daemon: daemon already running
```
adcockdalton, line 1: next time un-stage the log file:

`git restore --staged <pathspec>`

or only add the files you worked on

adcockdalton (changes requested):
I'm hoping to get a better idea of where we're at with this PR as far as closing #1053 . I see we're missing the patch notes still?

More importantly, it seems like there are still some issues with functionality that need to be worked through:

- [ ] I am noticing that the tutorial does not pop up right away by default; this should be corrected
- [ ] There is inconsistent behavior with how to open and close the menu. Is it supposed to open and close on click, or mouse-over? It seems to inconsistently open on mouse over (i.e. if you mouse-off on the icons it closes, but not if you mouse-off from the bulb).
- [ ] (If it's meant to be mouse-over, then the button should not ripple on click--it's misleading)


We can get more into the code if you are simply struggling with implementation--but I'd like to hear more about which of these design choices is intentional, and what your suggestions pertaining to my above design questions might be.

adcockdalton:
Also... you're still on a fork ;)

I was able to pull from your fork and look at it locally, so for this PR I can work with you from here, but for future PRs it's preferred for members of the project team to make branches from the primary repo

vicksey:
@adcockdalton Sorry I forgot to make the branch on the main repo. I'll open a new PR (since I can't figure out how to change the source of a PR) and I have it now so you can either hover or/and click the main help menu button to access the subcomponents. 
Also, I'm able to see the tutorial popup right away when clicked on my end so I'm not sure how to fix that.

adcockdalton:
I can show you what's going on in the meeting

Issue #1114, opened 2025-01-13
Optimize schedule save


MinhxNguyen7:
## Description
- Large schedules can take a long time to save, exceeding the old lambda limit of 5 seconds, and sometimes exceeding the current limit of 10 seconds.
- Saves take such a long time because of operations on the database, which are not very efficient right now.
- This is probably because we delete all of the user's data and put it back in every save.

PR #1115, opened 2025-01-13
Increase backend lambda timeout to 20 seconds

MinhxNguyen7:
## Summary
- Some large schedules are taking more than 10 seconds to save, which causes the lambda to time out.
- This is only a temporary fix. We need to optimize schedule saves.

## Future Followup
- #1114


ecxyzzy (approved):
lgtm

PR #1118, opened 2025-01-15
 feat: consolidate "help" elements into help menu

### LINKED ISSUE #1053: feat: consolidate "help" elements into help menu

vicksey:
## Summary
Aggregated the "help" elements, Help Box, Feedback Form, and Tutorial, into one help menu button. Once hovered the three elements are shown and can be clicked to be used.

## Test Plan
- [ ] Hover over the Help Menu to ensure all necessary components appear and can be accessed
- [ ] Click on the Help Menu to ensure children components appear/disappear
- [ ] Click through each element to test that they work like they normally do elsewhere
- [ ] Restart Tutorial button works
- [ ] Feedback Form opens in new tab
- [ ] Help menu appears
- [ ] See if size and design is good

## Issues

Closes #1053


adcockdalton (changes requested):
Good work so far, but there are still a couple of bugs I'm noticing:

1. Help box does not re-open every time icon is clicked. Steps to repro:
- [x] Click on help box icon
- [x] X it out
- [x] Click on help box icon again. It will not show, since clicking it just attempts to close the already-closed help box.

2. Mouse-off delay is slightly too long. I like the feature (great suggestion by @KevinWu098 ) but a collapse timeout of 1 second is quite long. Try out some smaller times and see how it feels. Maybe half a second? 

3. Mouse-off delay only works from the blue circle. If there is no way to implement this natively with mui then we can settle with a short mouse-off delay and not address this, but it would be apt for the delay to also work from the icon buttons (right now it doesn't).

4. The tutorial dies if you mouse over the help menu while the tutorial is running.
- [x] Click restart tutorial
- [x] Mouse-off the menu
- [x] Mouse back onto the menu once it closes

If you need help with implementation or have questions about code, let us know in the Discord ASAP

LGTM: light mode, help box design

apps/antalmanac/src/components/RightPane/RightPaneStore.ts
```diff
@@ -29,6 +29,7 @@ class RightPaneStore extends EventEmitter {
     private formData: Record<string, string>;
     private doDisplaySearch: boolean;
     private openSpotAlertPopoverActive: boolean;
+    private isHelpBoxVisible: boolean;
```
KevinWu098, line 32: Ideally, we should try to not add to the large state objects that our `RightPaneStore` and our `AppStore` have become. I think, in this case, it makes sense for us to create a new Zustand store that manages the states for all of our help/user-tip elements

apps/antalmanac/src/components/Tutorial.tsx
KevinWu098, line 1: Tutorial shouldn't be available on mobile (it isn't designed for mobile (yet))

KevinWu098 (changes requested):
Just throwing up some comments to work on

apps/antalmanac/src/components/Tutorial.tsx
vicksey: Ok thanks I didn't know I took it out for mobile

vicksey:
@adcockdalton I've fixed everything and cleaned up my code it should be ready for a final review now. Let me know if anything comes up!

MinhxNguyen7 (approved):
Using event emitters and listeners in `RightPaneStore` is a broken window, but I'll let it slide because we'll be refactoring this.

MinhxNguyen7:
@KevinWu098 @adcockdalton Can you check if your comments have been addressed? This has been sitting for a while

KevinWu098:
> @KevinWu098 @adcockdalton Can you check if your comments have been addressed? This has been sitting for a while

mm, I had a comment about refactoring logic into a Zustand store specifically for the help elements, but that can be in a follow up -- don't think I have anything blocking now

adcockdalton:
> @KevinWu098 @adcockdalton Can you check if your comments have been addressed? This has been sitting for a while

@vicksey 
I only have nitpicks at this point. The changes look great.

1. It doesn't really make sense from a design standpoint to have the help button ripple on click if clicking it has no effect. If this is removable that would be advisable.
2. The best behavior/appearance of the help button *during* the tutorial is sort of a gray area. Right now, it just sits there and does not open on hover. I'll leave it up to you what behavior you think it should have (I'm good to approve as-is, but I want to make sure it's intentional).

Nice PR!

vicksey:
@adcockdalton I've taken away the ripple effect for desktop. I don't think the button should do anything when in the tutorial so I've left it as is (aka it doesn't do anything when hovered besides it still darkens and says help menu). 

I think kevin's comments are outside of this PR and I'll probably be re-refactoring everything once auth is done and stuff.

Besides that I think it's good to merge after you do a final review!

KevinWu098:
Superseded by #1175

PR #1119, opened 2025-01-15
fix: Reformat catalogue builder (+ add ISEB)

### LINKED ISSUE #1085: fix: ISEB is not included in buildingCatalogue.ts

xgraceyan:
## Summary
- Reformatted `catalogue_builder.js` into `catalogueBuilder.ts`
- catalogueBuilder builds `locations.ts` alongside `buildingCatalogue.ts`
- Replaced `locations.ts` and `buildingCatalogue.ts` with its recently updated version
- Moved `catalogueBuilder.ts, buildingCatalogue.ts, locations.ts` into `apps/antalmanac/src/lib/locations`

### Important note/potential issue
Currently, the data in `buildingCatalogue.ts` has its building ID keys as strings instead of numbers. This is because json requires all keys be strings, and `JSON.stringify` is used in order to write the scraped data to buildingCatalogue.

I can't find a workaround that writes to the file without using `JSON.stringify` as of right now, so I've just left the IDs as strings for now. To my knowledge, you can still use numbers to index the data, and locations are still able to be fetched properly from other scripts. However, this is still pretty important to note down, and might need another solution soon.

## Test Plan
- Run `locations.test.ts` and make sure all tests pass (all tests passed when I ran it on my end)
- Ensure locations are able to be looked up on the map
- Ensure location name, image, directions, and info are displayed when searched on map
- Ensure any "missing" locations are present -- ie, ISEB and any others
- Ensure locations are able to be fetched when clicked from a class lookup
- Ensure locations are able to be saved properly when adding a custom event
- Ensure location abbreviations are displayed on the calendar (and wherever applicable)

## Issues

Closes #1085


apps/antalmanac/src/lib/locations/catalogueBuilder.ts
```diff
@@ -1,34 +1,48 @@
-const axios = require('axios').default;
-const fs = require('fs');
+import fs from 'fs';
+
```
KevinWu098, line 2:
```suggestion
```
chore: random line break

apps/antalmanac/src/lib/locations/catalogueBuilder.ts
```diff
@@ -1,34 +1,48 @@
...
+import axios from 'axios';
 // https://map.uci.edu/?id=463#!s/?ct/8424,8309,8311,8392,8405,44392,44393,44394,44395,44396,44397,44398,44400,44401,44402,44538,44537,44399,8396,11907,8400,10486,11906,11889,8310,8312,8393,8394,8397,8398,8399,8404,8407,8408,11891,11892,11899,11900,11902,21318,8406,11908,11935
```
KevinWu098: chore: maybe colocate this with our `CATEGORIES` const to explain better where we source our category ids from. adding a brief comment to further illustrate this would be good too

apps/antalmanac/src/lib/locations/catalogueBuilder.ts
```diff
@@ -39,19 +53,43 @@ async function fetchLocations() {
...
+                ? location.name.substring(location.name.indexOf('(') + 1, location.name.indexOf(')'))
+                : location.name;
+            locationIds[locationName] = location.id;
+
+            await sleep(250);
         }
     }
 }
 
 (async () => {
```
KevinWu098: issue: looks like this code works, but it's a tad hard on the eyes to read.

suggestion: rewrite this function as a `main()` and call that at the bottom of the file. Further, break up the string literals into separate variables

```ts
const foo = `
    export interface Building {
    imageURLs: string[];
    lat: number;
    lng: number;
    name: string;
    }\n\n
`;
```

apps/antalmanac/package.json
```diff
@@ -39,6 +39,7 @@
         "@tanstack/react-query": "^4.24.4",
         "@trpc/client": "^10.30.0",
         "@trpc/server": "^10.30.0",
+        "axios": "^1.7.9",
```
KevinWu098:
```suggestion
```
issue: would be nice to not add an additional dependency; we can do this with just fetch

apps/antalmanac/src/lib/locations/catalogueBuilder.ts
```diff
@@ -1,34 +1,48 @@
-const axios = require('axios').default;
-const fs = require('fs');
+import fs from 'fs';
+
+import axios from 'axios';
```
KevinWu098: repeat: lets not use a new dependency; `fetch` is fine here

apps/antalmanac/src/lib/locations/catalogueBuilder.ts
```diff
@@ -1,34 +1,48 @@
...
 
-function sleep(ms) {
+function sleep(ms: number): Promise<number> {
     return new Promise((resolve) => setTimeout(resolve, ms));
 }
 
 async function fetchLocations() {
     // Hits locations API to get a list of every single location
-    let locations = await axios.get(LOCATIONS_LIST_API);
+    let locations: any = await axios.get<BuildingLocation>(LOCATIONS_LIST_API);
```
KevinWu098: issue: using `any` explicitly and naming both the response and the resulting data the same variable can create a surface for potential errors

suggestion: aside from switching out Axios for Fetch, I'd recommend naming the variables more clearly and not using `any`. When you must make an assertion (which you'll have to), I'd recommend asserting the type of the decoded data.

Would also be good to write some validator for the data, but I think we can skip on that for now.

KevinWu098 (changes requested):
Great improvement to our system! Just some minor suggestions, but overall very thoroughly put together!

KevinWu098:
I'm trolling, didn't see the re-request, will review asap 🫡

KevinWu098:
In the meantime, would you mind resolving the merge conflict (should be very quick)

apps/antalmanac/src/lib/locations/catalogueBuilder.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    // Go through all the locations
+    for (const location of locations) {
+        // Check if location was already recorded
+        const locationExists: boolean = Object.values(locationsCatalogue).some(
+            (existingLocation) => existingLocation.name === location.name
+        );
+
+        // If one of the locations belongs to one of the categories above, and is not a duplicate
+        if (CATEGORIES.has(location.catId) && !locationExists) {
```
KevinWu098: nit (non-blocking): An (arguably) better way to write this function would be with a few early returns/guard clauses

```ts
if (!CATEGORIES.has()) {
    return;
}

if (!locationExists) {
    return;
}

// do logic after guard clauses
...

KevinWu098 (approved):
Very good work! LGTM 🚀 

I think that, like a few of our other "generated" data lists (e.g. term data, fuzzy search), locations would also be a good one to generate at build time!

pnpm-lock.yaml
KevinWu098, line 1: Hmm... this lockfile looks wrong, actually... 

Why are we removing so many dependencies? Can you rerun `pnpm install` at the root and commit the resulting file? If it's the same, just lmk

KevinWu098 (changes requested):
Lockfile looks funny ⬇️

pnpm-lock.yaml
xgraceyan: Yep just committed the resulting lockfile but there still seems to be some merge conflicts with main(?)

In the meantime, it looks like some other person was working on the same file LocationsCell.tsx and got merged first. I tried to pull the changed files necessary to resolve the merge conflict from main, but it didn't quite work out as intended so I just reverted it back. How should I resolve this conflict?

KevinWu098: Er... I'm on mobile so I don't have the best context of what's going on... but the merge resolution on LocationCell should be very straightforward. It should just be fixing the imports.

For the lockfile, running pnpm install and committing the result should be all that you should need to do.

Can you try that out (or if you already have done so), then lmk the result?

xgraceyan: Oh, you're right about LocationCell. I have no idea what I did earlier then lol
As for the lockfile, I ran pnpm install again already and committed the result (commit 9bff something?) -- no idea why there's still conflicts with main

KevinWu098:
Decided to go in and resolve the conflict -- in the future, deleting the lockfile, marking the merge as complete, then re-installing usually does the trick!

KevinWu098 (approved):

jotalis (approved):
lgtm

PR #1127, opened 2025-01-17
feat: Implement Google Auth

alexespejo:
## Summary
- Updates to saving schedules. 
- To save a schedule users can either sign in with their **Gmail** account or use a **username** (similar to how we save/load schedules in the original AntAlmanac).  

### Sign in
Users can create schedules  without signing in but will be prompted to sign in when they attempt to **save** 
image

### Guest Account
image

### Google Account
image

### Signing in with a existing schedules 
image

_Schedules created before signing in area marked (RESTORED)_
image

### Importing Old AA schedules (schedule names) into accounts 
 _When current users create a google account they can import their schedules from their prior accounts_
- This feature is only enabled if the user is signed in with either ***guest*** or ***Google***
- Cannot import schedules tied to Google accounts (you technically can but def won't happen)
- Imported schedules are prefixed with (Imported)
image
image

## 
[Workflow Diagram](https://www.figma.com/board/tcYkHlSRcMlsBlS6cmiztz/AntAlmanac-User-Auth-Workflow?node-id=0-1&t=6MbDWFKq4mXYRR0t-1)

## Test Plan
- [ ] Make sure users with old schedules can still access them
- [ ] Ensure Google auth creates and persists sessions with schedules 
- [ ] Ensure Guest auth creates and persists sessions with schedules 
- [ ] All Guest accounts have to have a unique name (no duplicate Guests can be created) 
- [ ] Import schedule should not be affected 
- [ ] Create schedule before creating an acocunt (schedule should persist into the newly created account)
- [ ] Sessions should expire after 7 days, expired sessions no errors should be thrown and schedule data shouldn't appear after expire 
- [ ] Data schedule data should not appear after sign out 
- [ ] Local storage for storing pre-login schedules might break shit
- [ ] Schedules created before auth update should be loadable with Guest Username

## Migration Plan
- [x] Dump DB into S3 (Eddy).
- [ ] Merge (executes migration).
- [ ] Smoke test to verify that nothing is catastrophically broken.
- [ ] Tentatively celebrate.

## Issues

Closes #

## Tasks 

- [X] Setup Google auth provider
- [X] Setup tRCP endpoint for handling Google
- [X] Wireup Sessions and auth  
- [X] Build out sign in sign out UI and logic
- [X] Connect Schedules
- [x] Setup Guest login
- [x] Remove the ID from save alerts 
- [X] Pain
- [x] Add `email` column to `user` 
- [x] Guests are all unique (I forgot about this)
- [x] Cache existing schedule to local storage 

## Future Followup
- [x] Configure secrets for prod and dev
- [x] Clean up 🍝 code (subjective)
- [ ] Programmatically connect redirect uri


vicksey:
Sorry, I was checking out the Google Auth and accidentally marked this as ready for review. Can the author switch it back?

alexespejo:
> Sorry, I was checking out the Google Auth and accidentally marked this as ready for review. Can the author switch it back?

Nah I can't

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -85,24 +86,16 @@ function isEmptySchedule(schedules: ShortCourseSchedule[]) {
...
 
-export const saveSchedule = async (userID: string, rememberMe: boolean) => {
-    logAnalytics({
-        category: analyticsEnum.nav.title,
-        action: analyticsEnum.nav.actions.SAVE_SCHEDULE,
-        label: userID,
-        value: rememberMe ? 1 : 0,
-    });
+export const saveSchedule = async (userID: string) => {
+    // logAnalytics({
```
alexespejo: What are these analytics for?

alexespejo:
@MinhxNguyen7 I'm still going through and bug testing everything, lmk if there are any naming conventions or patterns I should change. I wasn't very familiar with a lot of the backend stuff so a lot of my code was learn as you go lol.

MinhxNguyen7:
I'll try to look at this *at some point*

apps/antalmanac/src/actions/AppStoreActions.ts
MinhxNguyen7: It's Google Analytics. Don't delete it.

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -85,24 +86,16 @@ function isEmptySchedule(schedules: ShortCourseSchedule[]) {
...
+    //     category: analyticsEnum.nav.title,
+    //     action: analyticsEnum.nav.actions.SAVE_SCHEDULE,
+    //     label: userID,
+    //     value: rememberMe ? 1 : 0,
+    // });
```
MinhxNguyen7: Don't remove analytics

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -170,49 +159,59 @@ export async function autoSaveSchedule(userID: string) {
...
+            if (scheduleCache) {
+                const cacheSchedule = scheduleCache.map((schedule: ShortCourseSchedule) => {
+                    return {
+                        ...schedule,
+                        scheduleName: '(RESTORED)-' + schedule.scheduleName,
+                    };
+                });
+                scheduleSaveState.schedules.push(...cacheSchedule);
             }
+        }
```
MinhxNguyen7: It's kind of confusing what this does. I think it should be broken into its own function.

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -170,49 +159,59 @@ export async function autoSaveSchedule(userID: string) {
...
+        //     value: account.AccountType === 'GUEST' ? 1 : 0,
+        // });
+
+        const res: User = await trpc.users.getUserData.query({ userId: userId });
+        const scheduleSaveState = res && 'userData' in res ? res.userData : res;
+
+        if (isEmptySchedule(scheduleSaveState.schedules)) return;
+        if (loadCache) {
+            if (scheduleCache) {
+                const cacheSchedule = scheduleCache.map((schedule: ShortCourseSchedule) => {
```
MinhxNguyen7:
```suggestion
                const shortCourseSchedules = scheduleCache.map((schedule: ShortCourseSchedule) => {
```

apps/antalmanac/src/lib/localStorage.ts
```diff
@@ -14,11 +14,24 @@ enum LocalStorageKeys {
...
     helpBoxDismissalTime = 'helpBoxDismissalTime',
     columnToggles = 'columnToggles',
+    sessionId = 'sessionId',
+    scheduleCache = 'scheduleCache',
 }
 
 const LSK = LocalStorageKeys;
 
-// Helper functions for userId
+export function setLocalStorageScheduleCache(value: string) {
```
MinhxNguyen7: It's not a schedule cache, per se. It's a user data cache

apps/antalmanac/src/lib/download.ts
```diff
@@ -256,13 +256,14 @@ export function getRRule(bydays: string[], quarter: string) {
 
 export function getEventsFromCourses(
     events = AppStore.getEventsWithFinalsInCalendar(),
-    term = getDefaultTerm(events).shortName
+    _term = getDefaultTerm(events).shortName
```
MinhxNguyen7, line 259: What does this have to do with this PR?

apps/antalmanac/src/components/dialogs/SignInDialog.tsx
```diff
@@ -0,0 +1,165 @@
...
+    const handleUserNameOptionOpen = () => {
+        setOpenUserNameOption(true);
+    };
+
+    const handleUserNameOptionClose = () => {
+        setOpenUserNameOption(false);
+    };
+
+    return (
+        <AuthDialog
```
MinhxNguyen7: The `AuthDialog` should be folded into this component, and the two options for the ternary expression should be broken out into their components (if that's not too much work).

apps/antalmanac/src/components/dialogs/SignInDialog.tsx
```diff
@@ -0,0 +1,165 @@
...
+            open={open}
+            onClose={handleClose}
+            title={!openUserNameOption ? 'Sign in to save' : 'Username Login'}
```
MinhxNguyen7: I think it should be "Sign in" and "Guest Username"/"Guest Account"

apps/antalmanac/src/lib/localStorage.ts
```diff
@@ -31,6 +44,17 @@ export function removeLocalStorageUserId() {
     window.localStorage.removeItem(LSK.userId);
 }
 
+export function setLocalStorageSessionId(value: string) {
```
MinhxNguyen7, line 48: Generally speaking, cookies should be set by the server, and client code should not access it. Using `localStorage` is technically less secure because it is vulnerable to XSS and token hijack. That being said, it's not a big deal, and we can address this in a future PR for the sake of shipping.

apps/antalmanac/src/stores/SessionStore.ts
```diff
@@ -0,0 +1,39 @@
...
+    clearSession: () => Promise<void>;
+}
+
+export const useSessionStore = create<SessionState>((set) => {
+    const localSessionId = getLocalStorageSessionId();
+    return {
+        session: localSessionId,
+        validSession: false,
+        setSession: async (session) => {
+            if (session) {
```
MinhxNguyen7, line 19: If the passed-in session is nullish, shouldn't that just invalidate the session as well?

apps/backend/src/lib/rds.ts
```diff
@@ -22,6 +31,106 @@ export class RDS {
...
+    /**
+     * Retrieves an account with the specified user ID and account type 'GOOGLE'.
+     *
+     * @param db - The database or transaction object.
+     * @param userId - The ID of the user whose account is to be retrieved.
+     * @returns A promise that resolves to the account object if found, otherwise null.
+     */
+    static async getAccountByProviderId(
+        db: DatabaseOrTransaction,
+        accountType: 'GOOGLE' | 'GUEST',
```
MinhxNguyen7: This should be imported from `account`. 

Also, the docstring doesn't match what this function does, and this is a generalization of `guestUserIdWithNameOrNull`, so that function should be refactored to call this one.

apps/backend/src/lib/rds.ts
```diff
@@ -22,6 +31,106 @@ export class RDS {
...
+            tx
+                .select()
+                .from(accounts)
+                .where(and(eq(accounts.accountType, accountType), eq(accounts.providerAccountId, providerId)))
+                .limit(1)
+                .then((res) => res[0] ?? null)
+        );
+    }
+
+    static async getAccountUserByToken(db: DatabaseOrTransaction, refreshToken: string) {
```
MinhxNguyen7: If this gets the account and the user, it should be `getUserAndAccountByToken`.

apps/backend/src/lib/rds.ts
```diff
@@ -22,6 +31,106 @@ export class RDS {
...
+    ) {
+        const existingAccount = await this.getAccountByProviderId(db, accountType, providerId);
+        if (!existingAccount) {
+            const userId = crypto.randomUUID();
+            await db.insert(users).values({
+                id: userId,
+                avatar: avatar ?? '',
+                name: name,
+                email: email ?? '',
+            });
```
MinhxNguyen7: You don't need to generate a user ID here. There's already a default for that column in `users`.


```suggestion
            const { userId } = await db.insert(users).values({
                id: userId,
                avatar: avatar ?? '',
                name: name,
                email: email ?? '',
            }).returning({userId: users.id});
```

apps/backend/src/lib/rds.ts
```diff
@@ -103,28 +210,20 @@ export class RDS {
...
      */
-    static async upsertGuestUserData(db: DatabaseOrTransaction, userData: User): Promise<string> {
+    static async upsertUserData(db: DatabaseOrTransaction, userData: User): Promise<string> {
         return db.transaction(async (tx) => {
-            const userId = await this.createGuestUserOptional(tx, userData.id);
+            let userId = userData.id;
+            let user = await this.getUserById(db, userId);
+            if (!user) {
+                userId = await this.createGuestUserOptional(tx, userId);
+            }
```
MinhxNguyen7: What are these changes for?

apps/backend/src/lib/rds.ts
```diff
@@ -22,6 +31,106 @@ export class RDS {
             .then((xs) => xs[0]?.id ?? null);
     }
 
+    /**
```
MinhxNguyen7: This class is getting very bloated and should be broken up, but you can just leave it as is and clean up later.

apps/backend/src/routers/users.ts
```diff
@@ -30,33 +37,104 @@ const usersRouter = router({
...
+        .input(z.object({ sessionToken: z.string() }))
+        .query(async ({ input }) => {
+            return await RDS.getAccountUserByToken(db, input.sessionToken);
+        }),
+    getAccountByUid: procedure.input(z.object({ userid: z.string() })).query(async ({ input }) => {
+        return await RDS.getAccountById(db, input.userid);
+    }),
+    getUserByUid: procedure.input(z.object({ userid: z.string() })).query(async ({ input }) => {
+        return await RDS.getUserById(db, input.userid);
+    }),
```
MinhxNguyen7: Why are these endpoints here? If they're for testing while development, they should be moved to a different, clearly-marked router.

apps/backend/src/routers/index.ts
```diff
@@ -5,13 +5,15 @@ import courseRouter from './course';
...
-import searchRouter from "./search";
+import searchRouter from './search';
+import sessionRouter from './session';
 
 const appRouter = router({
     course: courseRouter,
     enrollHist: enrollHistRouter,
     grades: gradesRouter,
     search: searchRouter,
+    session: sessionRouter,
```
MinhxNguyen7: Instead of having a `sessionRouter` and a `usersRouter`, there should be an `authRouther` and a `userData` router.

apps/backend/.env.example
```diff
@@ -9,4 +9,10 @@ MAPBOX_ACCESS_TOKEN=pk.abc123
...
 # Connection string for the postgres database
-DB_URL=postgres://USERNAME:PASSWORD@HOST:PORT/antalmanac-dev
\ No newline at end of file
+DB_URL=postgres://USERNAME:PASSWORD@HOST:PORT/antalmanac-dev
+
+GOOGLE_CLIENT_ID=''
+
+GOOGLE_CLIENT_SECRET=''
+
+REDIRECT_URI=http://localhost:5173
```
MinhxNguyen7:
```suggestion
GOOGLE_REDIRECT_URI=http://localhost:5173
```

MinhxNguyen7 (changes requested):
I've only looked at the code, not the functionality

apps/antalmanac/src/lib/download.ts
alexespejo: After pulling from main there was some linting error in that file

apps/antalmanac/src/components/dialogs/SignInDialog.tsx
alexespejo: It's poorly named but I use it for another component so I'm using it as a generic

apps/antalmanac/src/stores/SessionStore.ts
alexespejo: I'm not sure what you mean. If you're referring to `validSession,` it's there because an expired session it might still be in local storage

apps/backend/src/routers/userData.ts
```diff
@@ -0,0 +1,149 @@
...
+import { mangleDupliateScheduleNames } from 'src/lib/formatting';
+import { RDS } from 'src/lib/rds';
+import { TRPCError } from '@trpc/server';
+import { procedure, router } from '../trpc';
+import { z } from 'zod';
+
+const GOOGLE_CLIENT_ID = process.env.GOOGLE_CLIENT_ID!;
+const GOOGLE_CLIENT_SECRET = process.env.GOOGLE_CLIENT_SECRET!;
+const REDIRECT_URI = 'http://localhost:5173';
+
```
MinhxNguyen7: You need to grab these from `env.ts`. 

Specifically, you need `const { GOOGLE_CLIENT_ID, GOOGLE_CLIENT_SECRET, REDIRECT_URI } = backendEnvSchema.parse(process.env)`. This both checks that all of the variables are avialable and get rid of the type warning.

apps/backend/src/routers/userData.ts
```diff
@@ -0,0 +1,149 @@
...
+     * Logs in or signs up a user and creates user's session
+     */
+    handleGoogleCallback: procedure
+        .input(z.object({ code: z.string(), token: z.string() }))
+        .query(async ({ input }) => {
+            const { tokens } = await oauth2Client.getToken({ code: input.code });
+            oauth2Client.setCredentials(tokens);
+
+            const ticket = await oauth2Client.verifyIdToken({
+                idToken: tokens.id_token!,
```
MinhxNguyen7: Needs null check

apps/backend/src/routers/userData.ts
```diff
@@ -0,0 +1,149 @@
...
+                audience: GOOGLE_CLIENT_ID,
+            });
+
+            const payload = ticket.getPayload()!;
```
MinhxNguyen7: Needs null check

apps/backend/src/routers/userData.ts
```diff
@@ -0,0 +1,149 @@
...
+                payload.name ?? '',
+                'GOOGLE',
+                payload.email ?? '',
+                payload.picture ?? ''
+            );
+
+            const userId: string = account.userId;
+
+            if (userId.length > 0) {
+                let session = await RDS.upsertSession(db, userId, input.token);
```
MinhxNguyen7: Use `const`. Doesn't this show up as an error for you?

apps/backend/src/routers/userData.ts
```diff
@@ -0,0 +1,149 @@
...
+                return session?.refreshToken;
+            }
+            return null;
+        }),
+    /**
+     * Logs in or signs up existing user
+     */
+    handleGuestLogin: procedure.input(z.object({ name: z.string() })).query(async ({ input }) => {
```
MinhxNguyen7, line 139: Since this has an `upsert` operation, it should be a `mutation`, not `query`

apps/backend/src/routers/userData.ts
```diff
@@ -0,0 +1,149 @@
...
+            scope: ['profile', 'email'],
+        });
+        return url;
+    }),
+    /**
+     * Logs in or signs up a user and creates user's session
+     */
+    handleGoogleCallback: procedure
+        .input(z.object({ code: z.string(), token: z.string() }))
+        .query(async ({ input }) => {
```
MinhxNguyen7: Use `mutation`

apps/antalmanac/src/components/dialogs/SignInDialog.tsx
alexespejo: I changed it to `InputDialog`

apps/antalmanac/src/lib/TutorialHelpers.tsx
```diff
@@ -186,9 +186,8 @@ export function namedStepsFactory(goToStep: (step: number) => void): Record<Tour
             selector: '#load-save-container',
             content: (
                 <>
-                    <b>Save</b> your schedule when you&apos;re done. <br />
-                    <b>Load</b> your schedule when you need it again. <br />
-                    <hr />
+                    <b>Sign in</b> with your <b>Google</b> account or a usersname to save your schedule when you&apos;re
```
MinhxNguyen7:
```suggestion
                    <b>Sign in</b> to save your schedule when you&apos;re
```

apps/antalmanac/src/stores/SessionStore.ts
MinhxNguyen7: I meant that this function should be called `updateSession` or something because it optionally sets the session token. The function doesn't actually *set* if the passed-in value is `null`.

apps/antalmanac/src/stores/SessionStore.ts
```diff
@@ -0,0 +1,39 @@
...
+                const validSession: boolean = await trpc.auth.validateSession.query({ token: session });
```
MinhxNguyen7:
```suggestion
                const sessionIsValid: boolean = await trpc.auth.validateSession.query({ token: session });
```

apps/antalmanac/src/stores/SessionStore.ts
```diff
@@ -0,0 +1,39 @@
+import { create } from 'zustand';
+
+import trpc from '$lib/api/trpc';
+import { getLocalStorageSessionId, removeLocalStorageSessionId, setLocalStorageSessionId } from '$lib/localStorage';
+
+interface SessionState {
+    session: string | null;
+    validSession: boolean;
```
MinhxNguyen7:
```suggestion
    sessionIsValid: boolean;
```

apps/backend/src/lib/rds.ts
```diff
@@ -337,4 +398,75 @@ export class RDS {
...
+     * @param db - The database or transaction object to use for the operation.
+     * @param userID - The ID of the user for whom the session is being created.
+     * @returns A promise that resolves to the created session object or null if the creation failed.
+     */
+    static async createSession(db: DatabaseOrTransaction, userID: string): Promise<Session | null> {
+        return db.transaction((tx) =>
+            tx
+                .insert(sessions)
+                .values({
+                    id: crypto.randomUUID(),
```
MinhxNguyen7: You can specify a default value generator in the table definition instead of generating it here.

apps/backend/src/lib/rds.ts
```diff
@@ -103,31 +167,20 @@ export class RDS {
...
      *
      * @param db The Drizzle client or transaction object
      * @param userData The object of data containing the user's schedules and courses
      * @returns The user's ID
      */
-    static async upsertGuestUserData(db: DatabaseOrTransaction, userData: User): Promise<string> {
+    static async upsertUserData(db: DatabaseOrTransaction, userData: User): Promise<string> {
         return db.transaction(async (tx) => {
-            const userId = await this.createGuestUserOptional(tx, userData.id);
+            let userId = userData.id;
```
MinhxNguyen7:
```suggestion
            const userId = userData.id;
```

apps/backend/src/lib/rds.ts
```diff
@@ -103,31 +167,20 @@ export class RDS {
...
 
-            if (userId === undefined) {
-                throw new Error(`Failed to create guest user for ${userData.id}`);
+            let user = await this.getUserById(db, userId);
```
MinhxNguyen7: Don't you get errors in the editor for this?
```suggestion
            const user = await this.getUserById(db, userId);
```

apps/backend/src/routers/auth.ts
```diff
@@ -0,0 +1,49 @@
...
+import { z } from 'zod';
+import { db } from 'src/db';
+import { RDS } from 'src/lib/rds';
+
+const authRouter = router({
+    handleGuestSession: procedure.input(z.object({ name: z.string() })).query(async ({ input }) => {
+        const account = await RDS.registerUserAccount(db, input.name, input.name, 'GUEST');
+
+        if (account.userId.length > 0) {
+            let session = await RDS.upsertSession(db, account.userId);
```
MinhxNguyen7:
```suggestion
            const session = await RDS.upsertSession(db, account.userId);
```

apps/backend/src/routers/auth.ts
```diff
@@ -0,0 +1,49 @@
+import { procedure, router } from '../trpc';
+import { z } from 'zod';
```
MinhxNguyen7:
```suggestion
import { z } from 'zod';

import { procedure, router } from '../trpc';


```

apps/backend/src/routers/auth.ts
```diff
@@ -0,0 +1,49 @@
...
+        const account = await RDS.registerUserAccount(db, input.name, input.name, 'GUEST');
+
+        if (account.userId.length > 0) {
+            let session = await RDS.upsertSession(db, account.userId);
+            return session?.refreshToken;
+        }
+        return null;
+    }),
+    /**
+     * Returns the current session, returns true if the session exists exist and hasn't expired
```
MinhxNguyen7: So which one does it return?

apps/backend/src/routers/auth.ts
```diff
@@ -0,0 +1,49 @@
+import { procedure, router } from '../trpc';
+import { z } from 'zod';
+import { db } from 'src/db';
+import { RDS } from 'src/lib/rds';
+
+const authRouter = router({
+    handleGuestSession: procedure.input(z.object({ name: z.string() })).query(async ({ input }) => {
```
MinhxNguyen7, line 12: Docstring

apps/backend/src/routers/auth.ts
```diff
@@ -0,0 +1,49 @@
...
+    validateSession: procedure.input(z.object({ token: z.string() })).query(async ({ input }) => {
+        if (input.token === '') return false;
+        const session = await RDS.getCurrentSession(db, input.token);
+
+        return session !== null && session.expires > new Date();
+    }),
+    /**
+    Removes a session from the database
+     */
+    removeSession: procedure.input(z.object({ token: z.string() })).mutation(async ({ input }) => {
```
MinhxNguyen7: Since this is the opposite of `validateSession`, from my understanding, it should be called `invalidateSession`

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -170,49 +157,68 @@ export async function autoSaveSchedule(userID: string) {
     }
 }
 
-export const loadSchedule = async (userId: string, rememberMe: boolean) => {
-    logAnalytics({
-        category: analyticsEnum.nav.title,
-        action: analyticsEnum.nav.actions.LOAD_SCHEDULE,
-        label: userId,
-        value: rememberMe ? 1 : 0,
+const mergeSchedules = (schedules: ShortCourseSchedule[], incomingSchedule: ShortCourseSchedule[]) => {
```
MinhxNguyen7: What does this do/how is this used? Can you put it in a docstring?

MinhxNguyen7 (changes requested):
Just looked at the code. Mostly small code quality comments.

Also, if anything fails/errors, you *must* display a message in the snackbar. Right now, if login fails, it doesn't say anything.

.github/workflows/deploy_production.yml
```diff
@@ -123,6 +123,10 @@ jobs:
 
             - name: Deploy backend production CloudFormation stack
               run: pnpm --filter "antalmanac-cdk" backend-production deploy
+              env:
+                  GOOGLE_CLIENT_ID: ${{ secrets.GOOGLE_CLIENT_ID }}
+                  GOOGLE_CLIENT_SECRET: ${{ secrets.GOOGLE_CLIENT_SECRET }}
+                  GOOGLE_REDIRECT_URI: https://api.antalmanac.com/
```
MinhxNguyen7: Needs correcting to match endpoint

.github/workflows/deploy_staging.yml
```diff
@@ -176,6 +176,10 @@ jobs:
 
             - name: Deploy backend staging CloudFormation stack
               run: pnpm --filter "antalmanac-cdk" backend-staging deploy
+              env:
+                  GOOGLE_CLIENT_ID: ${{ secrets.GOOGLE_CLIENT_ID }}
+                  GOOGLE_CLIENT_SECRET: ${{ secrets.GOOGLE_CLIENT_SECRET }}
+                  GOOGLE_REDIRECT_URI: ${{ needs.get_staging_configuration.outputs.backend_environment_url }}
```
MinhxNguyen7: Needs correcting to match endpoint

MinhxNguyen7: `${{ needs.get_staging_configuration.outputs.backend_environment_url }}` resolves to the backend URL of this staging instance. If this should be the frontend URL, change it to `${{ needs.get_staging_configuration.outputs.frontend_environment_url }}`. You can add text after it to specify the endpoint. (Note that spaces are significant.)

MinhxNguyen7 (changes requested):
I've fixed the env vars problem now. I don't know what I was on the other day. I can tell you what I did next meeting (or you can look at the diff). 

I'm getting `redirect_uri_mismatch`, now, because I don't know what `GOOGLE_REDIRECT_URI` should be. I've commented where you can change it.

IsaacNguyen:
I think everything looks and flows great. The only minor critique I have is that I think there should be better indication of when a user signs out. Maybe just a little popup at the bottom notifying them that they have been signed out of whatever account they were on. Otherwise, I think it looks great!

apps/antalmanac/src/components/Header/Login.tsx
```diff
@@ -0,0 +1,98 @@
...
+    const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null);
+    const [user, setUser] = useState<null | User>(null);
+
+    const { clearSession } = useSessionStore();
+    const navigate = useNavigate();
+
+    const handleLogout = async () => {
+        clearSession();
+        navigate('/');
+    };
```
IsaacNguyen, line 26: I think everything looks and flows great. The only minor critique I have is that I think there should be better indication of when a user signs out. Maybe just a little popup at the top notifying them that they have _successfully_ been signed out of whatever account they were on. Otherwise, good looks Alex!!

alexespejo: yeah i did that cuz instead of reloading the data i just refresh the  page lol, i'll patch post release

MinhxNguyen7:
Did you remove the loading icon from the save button? That is pretty important.

Also, we need to make sure that the user cannot attempt to save *while* the previous save is being processed.

apps/antalmanac/src/components/Header/Header.tsx
```diff
@@ -1,13 +1,41 @@
...
+                navigate('/');
+            }
+        } catch (error) {
+            console.error('Error during authentication', error);
+        }
+    };
+
+    useEffect(() => {
+        handleCallback();
+    }, [searchParams]);
```
MinhxNguyen7: This is kind of confusing. If you're watching for the change in `searchParams`, you should read it in this scope and pass it into the handler (and change it to `handleSearchParamsChange`)

apps/antalmanac/src/components/Header/Import.tsx
```diff
@@ -49,75 +53,99 @@ function Import() {
         setOpen(false);
     }, []);
 
+    const ZOT_COURSE_IMPORT = 'zotcourse';
+    const STUDY_LIST_IMPORT = 'studylist';
+    const AA_USERNAME_IMPORT = 'username';
```
MinhxNguyen7: This should be an `Enum` defined at the top level.

apps/antalmanac/src/components/Header/Login.tsx
```diff
@@ -0,0 +1,122 @@
...
+            const userId = await trpc.auth.getSessionUserId.query({ token: session ?? '' });
+            if (userId) {
+                setUser(await trpc.userData.getUserByUid.query({ userId: userId }));
+            }
+        }
+    };
+
+    useEffect(() => {
+        setSession(session);
+        handleUser();
```
MinhxNguyen7: What does this mean? Do you mean something like `handleAuthChange`?

apps/antalmanac/src/components/Header/Login.tsx
```diff
@@ -0,0 +1,122 @@
...
+        setTimeout(() => {
+            if (reLogin && !validSession && currentSession.current) {
+                setOpenSignIn(true);
+                setRelogin(false);
+            }
+        }, 1000);
+    }, [session, validSession, user, openSignIn]);
```
MinhxNguyen7: I think there's a misunderstanding of the use of `useEffect` (and `useCallback`), here.

Within a hook, you must specify its *immediate* dependencies. Here, the immediate dependency is just `handleUser`. You can, of course, specify more dependencies for the hook in the way that you're doing here, but you might be overdoing it.

For `handleUser` to change reactively, you do this:

```typescript
const handleUser = useCallback(async () => {
    if (validSession) {
        const userId = await trpc.auth.getSessionUserId.query({ token: session ?? '' });
        if (userId) {
            setUser(await trpc.userData.getUserByUid.query({ userId: userId }));
        }
    }
}, [session, validSession])
```

Let me know if this doesn't make sense

MinhxNguyen7 (changes requested):
Only a few minor things about useEffect, in addition to the above [comment](https://github.com/icssc/AntAlmanac/pull/1127#issuecomment-2708677237). We're almost there 🤏

Fyi, I did not review `rds.ts` again. No time. Sorry for taking so long as well.

apps/antalmanac/src/components/Header/Import.tsx
alexespejo: Is there a specific place we put constants? For now I just copied the pattern for `EXCLUDE_RESTRICTION_CODES_OPTIONS`

MinhxNguyen7 (approved):
```
                __/___            
          _____/______|           
  _______/_____\_______\_____     
  \              < < <       |    
```

PR #1135, opened 2025-01-29
feat: improvements to save state

### LINKED ISSUE #1133: Undo with Autosave can wipe schedules unexpectedly
## Context
AntAlmanac implements a basic "history" feature which stores "states" of schedules. We can revert to a prior "state" via undo (e.g. ctrl+z). 
Relevant files: `apps/antalmanac/src/stores/Schedules.ts`

We also implemented auto save via #891 which automatically saves the latest "state" to the saved user in localStorage.
Relevant files: `apps/antalmanac/src/actions/ActionTypesStore.ts`

## Problem
Loading in with no schedule/no username starts our state as empty: `[]`. The auto-loading in of a schedule (if a username is saved in localStorage) also counts as a "state" `[{some object}]`.

If a user's first action is to press `ctrl+z` after their schedule is loaded in, they will revert to the prior snapshot (empty state `[]`). This becomes **extremely problematic** because auto save will now save the empty state (without warning or confirmation) to the username stored in localStorage.

KevinWu098:
## Summary
1. Prevents "undo" if it would return the user to an empty state. This prevents autosave from wiping the user's schedule since we can never return to an empty state
2. Clears "states" when loading in a new username. This prevents undoing to another user's schedule and having that autosaved.

## Test Plan
1. Load a schedule. Try and undo (`ctrl + z`). Nothing should happen.
2. Undoing should still work, but not on the very first action you take
3. Load a schedule. Load a different schedule. Undo should do nothing.

## Issues

Closes #1133

### DIFF
--- a/apps/antalmanac/src/stores/AppStore.ts
+++ b/apps/antalmanac/src/stores/AppStore.ts
@@ -336,6 +336,7 @@ class AppStore extends EventEmitter {
         this.unsavedChanges = false;
 
         await actionTypesStore.loadScheduleFromLocalSave();
+        this.schedule.clearPreviousStates();
 
         this.emit('addedCoursesChange');
         this.emit('customEventsChange');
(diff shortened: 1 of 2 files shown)


MinhxNguyen7 (approved):
LGTM!

PR #1139, opened 2025-02-02
AANTS Backend

IsaacNguyen:
## Summary
- Setup AANTS folder with dependencies 
- Wrote script that implements email notification logic
- Actual PR is #1164 

## Brief Overview of the Logic
- Makes call to DB to obtain _unique_ subscription section codes and their terms
- Groups each section code by term (e.g. group 2025-WINTER and 2025-FALL classes together) into a hash map
- Creates batches of section codes for each term (in case of AAPI batch limit)
- Makes call to AAPI endpoint to receive data about updated classes
- Parses the output
- If a course's enrollment status _has changed_ since the last poll, it will prompt a notification
  - If it has not changed, the section is ignored
- To make a notification, calls DB for users associated wtih any subscriptions that are signed up for current update (e.g. class is full or open)
  - In parallel, also calls AAPI for course information (used later for email)
- Then will make a call to SES to send email notifications in batches
- Finally update enrollment status of subscriptions

## To-do
- [x] Wait for email column in DB to be setup when Google OAuth is implemented
- [x] Implement and test SES
- [x] Get SES production
- [x] Integrate functionality for restriction codes (when AAPI sets up endpoint).
- [ ] Set up deployment.

## Test Plan
- [x] Ensure only sends emails when enrollment status/restriction codes HAVE changed
- [x] Ensure email format is correct
- [ ] Block out notifications for non-signed in users
- [ ] Notifications are global and accessible between schedules

## Issues
- Still not fully implemented (uses a dummy output because AAPI endpoint hasn't been setup yet)
- There could be some dependency issues but the script itself seems to work just fine

Closes #


KevinWu098:
For AANTS frontend, we'll need an endpoint which will `upsert` a row in the notifications table when accessed. Should be pretty straightforward, but the endpoint should accept all fields which can be edited (the various notification statuses), as well as being able to remove a notification.

All actions on notifications are global (i.e. removing notification status is shared between schedules for the same sections)

PR #1142, opened 2025-02-05
feat: Larc API

KevinWu098:
## Summary
1. Implements the backend endpoint, types for the LARC
2. Minor refactoring of `SectionTable`

## Test Plan
1. Since the backend isn't being called by the frontend, that doesn't necessarily need to be tested here, and can be mostly handled when the frontend is implemented
2. Does the section table still look right?
    - Light/Dark Mode
    - Loading Message
    - Error Message

## Issues
Closes #


.eslintignore
KevinWu098, line 1: Don't worry about this, it was extraneous

KevinWu098:
bump @adcockdalton

KevinWu098:
Closing as stale

PR #1151, opened 2025-02-07
Fix first undo after load

### LINKED ISSUE #1150: Can't undo the first action after loading
## Description
- After loading a schedule, the first action you take can't be undone.
    - To reproduce, load a schedule, delete one section, then try to undo.
- This is likely an off-by-one error in #1135.
    - I guess I should've caught that.

MinhxNguyen7:
## Summary
- Previously, we checked whether the length of the save states is `<= 1` to disallow, which means that the first state after loading can't be undone.
- That check has been removed, so the state is revertible as long as there is a `previousState`.

## Test Plan
- [ ] Undo works properly before loading a schedule by adding one section and performing undo.
- [ ] Undo does not clear a schedule when it is loaded in.
- [ ] Undo works on the first action (e.g., removing a section) after loading a schedule.
- [ ] Multiple undo's can be performed.

## Issues

Closes #1150

### DIFF
--- a/apps/antalmanac/src/stores/Schedules.ts
+++ b/apps/antalmanac/src/stores/Schedules.ts
@@ -481,7 +481,7 @@ export class Schedules {
      */
     revertState() {
         // prevent the user from undoing to an empty state
-        if (this.previousStates.length <= 1) {
+        if (this.previousStates.length === 0) {
             return;
         }
 
--- a/apps/antalmanac/src/stores/Schedules.ts
+++ b/apps/antalmanac/src/stores/Schedules.ts
@@ -480,11 +480,6 @@ export class Schedules {
      * All actions that call `addUndoState()` can be reverted.
      */
     revertState() {
-        // prevent the user from undoing to an empty state
-        if (this.previousStates.length === 0) {
-            return;
-        }
-
         const state = this.previousStates.pop();
         if (state !== undefined) {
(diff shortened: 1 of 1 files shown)


apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -480,11 +480,6 @@ export class Schedules {
      * All actions that call `addUndoState()` can be reverted.
      */
     revertState() {
-        // prevent the user from undoing to an empty state
-        if (this.previousStates.length <= 1) {
```
KevinWu098, line 484: Wouldn't removing this check would reintroduce the bug which deletes user's schedules without warning Preventing undo if it would return to an "empty state" is critical for preventing autosave from nuking the schedule

KevinWu098 (changes requested):
I think the solution here may be to treat loading in a schedule as a state and keeping the check.

```
     async loadSchedule(savedSchedule: ScheduleSaveState) {
         await actionTypesStore.loadScheduleFromLocalSave();
         this.schedule.clearPreviousStates();
+        this.schedule.addUndoState();
```

However, I'm not sure if that would produce problems with how autosave is implemented.

KevinWu098 (approved):
After testing, LGTM -- thanks for catching this bug and fixing it 🫡

PR #1155, opened 2025-02-08
feat/refactor: Spring 25, select most recent term

KevinWu098:
## Summary
1. Term data now selects `defaultTerm` based on the term which has the most recent and available SOC. If none are present, it defaults to index 0 (though it will always be present).
2. Adds Spring 2025 through Fall 2025

## Test Plan
1. Loading in should select the correct default term (Spring 2025, once it becomes 2/8) -- best way to test would be to modify `getDefaultTerm` and adjust what day it is

```ts
function getDefaultTerm(events: (CustomEvent | CourseEvent)[] = []): Term {
...
        return new Date() >= new Date(year, month, day); // edit this line
    });
...
}
```

## Issues

Closes #

### DIFF
--- a/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
+++ b/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
@@ -16,6 +16,12 @@ import AppStore from '$stores/AppStore';
 import { useHoveredStore } from '$stores/HoveredStore';
 import { useTimeFormatStore } from '$stores/SettingsStore';
 
+moment.updateLocale('es-us', {
+    week: {
+        dow: 6,
+    },
+});
+
 const CALENDAR_LOCALIZER: DateLocalizer = momentLocalizer(moment);
 const CALENDAR_VIEWS: ViewsProps<CalendarEvent, object> = [Views.WEEK, Views.WORK_WEEK];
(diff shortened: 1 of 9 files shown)


apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
```diff
@@ -5,19 +5,23 @@ import { Box } from '@material-ui/core';
...
 import { CalendarEventPopover } from '$components/Calendar/CalendarEventPopover';
+import { CalendarEvent, CourseEvent } from '$components/Calendar/CourseCalendarEvent';
 import { CalendarToolbar } from '$components/Calendar/Toolbar/CalendarToolbar';
-import { getDefaultFinalsStartDate, getFinalsStartDateForTerm } from '$lib/termData';
+import { getTerm } from '$lib/termData';
 import AppStore from '$stores/AppStore';
 import { useHoveredStore } from '$stores/HoveredStore';
 import { useTimeFormatStore } from '$stores/SettingsStore';
 
+moment.updateLocale('es-us', {
```
KevinWu098, line 19: We should always start our week on Saturday -- on `views.WORK_WEEK`, this has no effect, but it most certainly is important for displaying finals dates correctly (since finals start on Saturday)

PR #1160, opened 2025-02-12
feat: One row calendar top bar on small screens

### LINKED ISSUE #1026: Calendar top bar becomes two rows for smaller screens
## Description
- Calendar top bar becomes two rows for smaller screens
Image

IsaacNguyen:
## Summary
Changed the "FINALS" button to become a smaller icon on smaller screens, preventing it from becoming two rows.

BEFORE:
image

AFTER:
- When the icon is not pressed
image

- When the icon IS pressed
image

## Test Plan
- [ ] Ensure works properly on all kinds of small devices

## Issues
- I make it become an icon at around a 400px width, but it still becomes two rows for very very small screens. But I doubt there are even screens this tiny. If there are, I can change it further.
   - image
- Could change the icon but I felt like this one was suiting.

Closes #1026

### DIFF
--- a/apps/antalmanac/src/components/Calendar/Toolbar/CalendarToolbar.tsx
+++ b/apps/antalmanac/src/components/Calendar/Toolbar/CalendarToolbar.tsx
@@ -1,5 +1,9 @@
-import { Undo as UndoIcon, Description as DescriptionIcon } from '@mui/icons-material';
-import { useMediaQuery, Box, Button, IconButton, Paper, Tooltip } from '@mui/material';
+import {
+    Undo as UndoIcon,
+    Description as DescriptionIcon,
+    DescriptionOutlined as DescriptionOutlinedIcon,
+} from '@mui/icons-material';
+import { useTheme, useMediaQuery, Box, Button, IconButton, Paper, Tooltip } from '@mui/material';
 import { useState, useCallback, useEffect, memo } from 'react';
 
 import { undoDelete } from '$actions/AppStoreActions';
@@ -30,11 +34,12 @@ export interface CalendarPaneToolbarProps {
  * The root toolbar will pass down the schedule names to its children.
  */
 export const CalendarToolbar = memo((props: CalendarPaneToolbarProps) => {
+    const theme = useTheme();
     const { showFinalsSchedule, toggleDisplayFinalsSchedule } = props;
     const [scheduleNames, setScheduleNames] = useState(AppStore.getScheduleNames());
     const [skeletonMode, setSkeletonMode] = useState(AppStore.getSkeletonMode());
     const [skeletonScheduleNames, setSkeletonScheduleNames] = useState(AppStore.getSkeletonScheduleNames());
-    const isSmallScreen = useMediaQuery('(max-width: 400px)');
+    const isSmallScreen = useMediaQuery(theme.breakpoints.down('xxs'));
 
     const handleToggleFinals = useCallback(() => {
         logAnalytics({
@@ -92,7 +97,7 @@ export const CalendarToolbar = memo((props: CalendarPaneToolbarProps) => {
                             id={showFinalsSchedule ? 'finals-button-pressed' : 'finals-button'}
                             disabled={skeletonMode}
                         >
-                            <DescriptionIcon />
+                            {showFinalsSchedule ? <DescriptionIcon /> : <DescriptionOutlinedIcon />}
                         </IconButton>
                     ) : (
                         <Button
@@ -108,7 +113,6 @@ export const CalendarToolbar = memo((props: CalendarPaneToolbarProps) => {
                     )}
                 </Tooltip>
             </Box>
-
             {!isSmallScreen && <Box flexGrow={1} />}
 
             <Box display="flex" flexWrap="wrap" alignItems="center" gap={0.5}>
--- a/apps/antalmanac/src/components/Calendar/Toolbar/CalendarToolbar.tsx
+++ b/apps/antalmanac/src/components/Calendar/Toolbar/CalendarToolbar.tsx
@@ -113,7 +113,7 @@ export const CalendarToolbar = memo((props: CalendarPaneToolbarProps) => {
                     )}
                 </Tooltip>
             </Box>
-            {!isSmallScreen && <Box flexGrow={1} />}
+            <Box flexGrow={1} />
 
             <Box display="flex" flexWrap="wrap" alignItems="center" gap={0.5}>
                 <ScreenshotButton />
--- a/apps/antalmanac/src/components/Calendar/Toolbar/CalendarToolbar.tsx
+++ b/apps/antalmanac/src/components/Calendar/Toolbar/CalendarToolbar.tsx
@@ -99,7 +99,7 @@ export const CalendarToolbar = memo((props: CalendarPaneToolbarProps) => {
                             size="small"
                             sx={{
                                 border: '1px solid',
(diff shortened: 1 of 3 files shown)


apps/antalmanac/src/components/Calendar/Toolbar/CalendarToolbar.tsx
```diff
@@ -34,6 +34,7 @@ export const CalendarToolbar = memo((props: CalendarPaneToolbarProps) => {
     const [scheduleNames, setScheduleNames] = useState(AppStore.getScheduleNames());
     const [skeletonMode, setSkeletonMode] = useState(AppStore.getSkeletonMode());
     const [skeletonScheduleNames, setSkeletonScheduleNames] = useState(AppStore.getSkeletonScheduleNames());
+    const isSmallScreen = useMediaQuery('(max-width: 400px)');
```
KevinWu098: issue: rather than specifying an arbitrary ("magic") number, please reference our theme's breakpoints (there are some examples in the codebase: search for `isMobile`)

apps/antalmanac/src/components/Calendar/Toolbar/CalendarToolbar.tsx
```diff
@@ -84,20 +85,31 @@ export const CalendarToolbar = memo((props: CalendarPaneToolbarProps) => {
...
-                        onClick={handleToggleFinals}
-                        size="small"
-                        id={showFinalsSchedule ? 'finals-button-pressed' : 'finals-button'}
-                        disabled={skeletonMode}
-                    >
-                        Finals
-                    </Button>
+                    {isSmallScreen ? (
+                        <IconButton
+                            color={showFinalsSchedule ? 'primary' : 'inherit'}
```
KevinWu098, line 95: issue: it's not clear that the finals button is a toggle. Choose another variant (`outlined`?) which more clearly show that this is a button different from the rest of the toolbar icon buttons

KevinWu098: Alternatively, just continue using `Button` as the parent container

apps/antalmanac/src/components/Calendar/Toolbar/CalendarToolbar.tsx
KevinWu098: I'd recommend perhaps adding `xxs` at 24rem (384px). Make sure to add it to both Theme and ThemeV5!

apps/antalmanac/src/components/Calendar/Toolbar/CalendarToolbar.tsx
```diff
@@ -84,20 +85,31 @@ export const CalendarToolbar = memo((props: CalendarPaneToolbarProps) => {
...
+                            disabled={skeletonMode}
+                        >
+                            Finals
+                        </Button>
+                    )}
                 </Tooltip>
             </Box>
 
-            <Box flexGrow={1} />
+            {!isSmallScreen && <Box flexGrow={1} />}
```
KevinWu098: question: why are we adding this conditional?

KevinWu098 (changes requested):
Thanks for working on this PR Isaac! Very happy that we're address this area of the UI. 

I've left some comments regarding some code consistency and visual clarity 🫡

apps/antalmanac/src/components/Calendar/Toolbar/CalendarToolbar.tsx
IsaacNguyen: Gotcha! I added `xxs` at 400px though because at 384px, the overflow was still present.

apps/antalmanac/src/components/Calendar/Toolbar/CalendarToolbar.tsx
IsaacNguyen: For some reason, on the smaller screens, this box would push the icons to the second row. Do you suggest any other way to fix this?

apps/antalmanac/src/components/Calendar/Toolbar/CalendarToolbar.tsx
KevinWu098: Interesting, it wouldn’t be bad to do it at a higher breakpoint then! Nothing wrong with that — try it out, choose what you think works best visually and code wise

apps/antalmanac/src/components/Calendar/Toolbar/CalendarToolbar.tsx
KevinWu098: Uhh, that should just be the flex wrapping… just sounds like overflow and we should make the finals button smaller, earlier

IsaacNguyen: Wait, are you sure? For a Samsung Galaxy S8+ (360px wide, which beneath the new 400px `xxs` breakpoint), this is what it looks like with the box _not_ rendered.
image

When it _is_ being rendered, it looks like this.
image

Sorry, I could be misinterpreting what you mean, but making the finals button smaller earlier wouldn't fix this right?

KevinWu098: well, at some point, the elements _will_ wrap on a smaller screen. no way to really get around that. To be clear, the unofficial requirement is that at 390px, thing do not wrap. Why 390px? Because that's my iPhone's pixel width (12 Pro)

IsaacNguyen: Oh okay that makes more sense. I was basing it off the Samsung Galaxy S8+ because that was the screen size displayed in the original issue (#1026). I will just revert this change then.

apps/antalmanac/src/components/Calendar/Toolbar/CalendarToolbar.tsx
```diff
@@ -84,19 +90,29 @@ export const CalendarToolbar = memo((props: CalendarPaneToolbarProps) => {
...
-                        variant={showFinalsSchedule ? 'contained' : 'outlined'}
-                        onClick={handleToggleFinals}
-                        size="small"
-                        id={showFinalsSchedule ? 'finals-button-pressed' : 'finals-button'}
-                        disabled={skeletonMode}
-                    >
-                        Finals
-                    </Button>
+                    {isSmallScreen ? (
+                        <IconButton
```
KevinWu098, line 94: issue: the button still does not look like a toggle. try out an outline, perhaps

apps/antalmanac/src/providers/Theme.tsx
```diff
@@ -36,6 +36,12 @@ interface Props {
     children?: React.ReactNode;
 }
 
+declare module '@material-ui/core/styles/createBreakpoints' {
```
KevinWu098, line 39: question: why are we adding the breakpoint like this? Is this the officially recommended way from the MUI docs?

apps/antalmanac/src/providers/Themev5.tsx
```diff
@@ -34,6 +34,12 @@ interface Props {
     children?: React.ReactNode;
 }
 
+declare module '@mui/material/styles' {
```
KevinWu098, line 37: repeat: is this the recommended way to add a new breakpoint?

KevinWu098 (changes requested):

MinhxNguyen7:
Border not visible in light mode

KevinWu098 (approved):
LGTM

PR #1162, opened 2025-02-19
feat: aants frontend

KevinWu098:
## Summary
1. Implements adding notifications per section via AddedPane
2. Implements managing global notifications 
3. Specifies tRPC endpoints for reading and writing

### Notification for a section
chrome-capture-2025-1-23

### Global notification table
chrome-capture-2025-1-23 (1)

## Test Plan
1. Clicking a checkbox should update the value of the notification. That value should persist across an AA session (it will persist beyond a session once it's hooked up to our DB)

## Issues
Closes #

### Deleting Notifications
Notifications exist somewhat outside of our existing AntAlmanac model of schedules and saving. All writes are (more or less) immediate. Notifications do not belong to any given schedule, and can be edited across multiple schedules which share the same section.

To that end, deleting a section **currently does not constitute deleting the associated notification**. In the upcoming PR linking this and #1164, a flow to delete a notification (probably via the global notification table) will need to be implemented.

PR #1164, opened 2025-02-19
Rebased AANTS PR

IsaacNguyen:
## Summary
- Rebased AANTS on top of #1127
- Most current branch on #1241 

## Issues

Associated with #1139


KevinWu098:
Small change in scope: I'll be handling writing the TRPC endpoints for the `get` and `set` of AANTS on my PR #1162, specifically writing to the Subscription table. 

This shouldn't affect (or block) anything you're doing as far as dispatching notifications, but let me know if that's not the case

IsaacNguyen:
> Small change in scope: I'll be handling writing the TRPC endpoints for the `get` and `set` of AANTS on my PR #1162, specifically writing to the Subscription table.
> 
> This shouldn't affect (or block) anything you're doing as far as dispatching notifications, but let me know if that's not the case

Yeah I think it should be fine!

MinhxNguyen7:
A few to-dos:
- Refactor the main script to be multiple functions. One tip is to return promises and await them instead of nesting them.
- Push the deployment code that we worked on.
- Resolve the merge conflicts.
- See if it works. Note that if you don't expect it to work because of the current state of the DB, at least console log some things, then we can see that it's running in CloudWatch logs.

MinhxNguyen7:
Superceded by #1241

PR #1166, opened 2025-02-23
fix: revert eslint config

KevinWu098:
## Summary
1. We updated to a newer verson of the eslint config (v9) recently, but that migration did not work very well — this PR reverts us to our prior settings

## Issues

Closes #


MinhxNguyen7 (approved):
Sure

PR #1169, opened 2025-02-25
Bump node version to 22

MinhxNguyen7:
## Summary
- What it says on the tin.
- Also removes the node version check in the `cdk` sub-repo.

## Test Plan
- See if prod breaks (because actions need to be merged to actually work).


KevinWu098 (approved):
lgtm

PR #1175, opened 2025-02-28
feat: help menu

### LINKED ISSUE #1053: feat: consolidate "help" elements into help menu

KevinWu098:
## Summary
This PR supersedes #1118 and implements a Help Menu for AntAlmanac. The Help Menu consists of user facing elements which are not specific to any given page:
1. Help Box
2. Feedback Form
3. Tutorial
4. Patch Notes

Screenshot 2025-02-28 at 8 15 09 PM

chrome-capture-2025-1-28

## Test Plan
1. The Help Menu is triggered `onClick` and not `onMouseEnter`. The menu should open and close appropriately when it, its backdrop, or any action is clicked.
2. Each action should work appropriately
    a. Help Box should navigate correctly and open correctly
    b. Feedback Form is a new tab
    c. Tutorial and Patch Notes both open as expected if their conditions are valid. Note: this PR makes it so patch notes do not show IF the tutorial is also showing. The expected behavior on a completely new account (no saved username) would be for the tutorial to show, but not patch notes. But on a subsequent reload, patch notes would show.
3. On every tab (both mobile and desktop), the help menu should be positioned appropriately without blocking (too much) content.
4. Light and Dark mode
5. All actions should be responsive to both mobile and desktop (e.g. We disable tutorial on mobile) 

## Issues
Closes #1053

## Future Followup
Add Install PWA to Help Menu


apps/antalmanac/tsconfig.json
```diff
@@ -21,6 +21,7 @@
             "$api/*": ["src/api/*"],
             "$components/*": ["src/components/*"],
             "$lib/*": ["src/lib/*"],
+            "$hooks/*": ["src/hooks/*"],
```
KevinWu098, line 24: Fixes my import adjustment from #1171

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/FuzzySearch.tsx
```diff
@@ -187,11 +187,9 @@ class FuzzySearch extends PureComponent<FuzzySearchProps, FuzzySearchState> {
                 renderInput={(params) => (
                     <TextField
                         {...params}
-                        inputRef={(input: HTMLInputElement | null) => {
-                            if (input && !isMobile()) {
-                                input.focus();
-                            }
-                        }}
+                        // eslint-disable-next-line jsx-a11y/no-autofocus
+                        autoFocus={!isMobile()}
```
KevinWu098, line 191: Instead of constantly focusing the search (which is poor behavior AFAIK), we now only autofocus once.

MinhxNguyen7 (approved):
Sure

MinhxNguyen7 (changes requested):
I just realized, you're hiding the patch notes by default, right? We need it to be visible as much as is reasonable. Can you make it only show up if the tour was previously dismissed?

KevinWu098:
> I just realized, you're hiding the patch notes by default, right? We need it to be visible as much as is reasonable. Can you make it only show up if the tour was previously dismissed?

You might be catching something in my logic I didn't intend, but I believe the logic is that:

If tutorial would show, do not show patch notes (but also do not adjust the last seen date). 

Then, when the tutorial would not show (e.g. on the next visit), then the patch notes would show.

MinhxNguyen7 (approved):
I didn't check the logic. I just tested it, but I think I did it wrong. Good to see we were on the same page

PR #1189, opened 2025-03-06
fix: enrollment history graph off-by-one

### LINKED ISSUE #1180: bug: Past Enrollment has 0 in month, day fields
image

The past enrollment component renders December as 0, and some dates (the first?) as 0. Likely off-by-one error...

Original Feedback Form: https://discord.com/channels/772739905981644850/1345662911577849917/1345662911577849917

laggycomputer:
## Summary

Fix off-by-one bugs when parsing AAPI enrollment history response.
Months as returned by `Date` are zero-based, and dates (days of the month) are one-based. Simple fix.

## Test Plan
Try some more cases.

Here's the screenshot from the issue: 
image

## Issues

Closes #1180.

### DIFF
--- a/apps/antalmanac/src/lib/enrollmentHistory.ts
+++ b/apps/antalmanac/src/lib/enrollmentHistory.ts
@@ -91,7 +91,7 @@ export class DepartmentEnrollmentHistory {
             for (const [i, dateString] of enrollmentHistory.dates.entries()) {
                 const [day = '', month = '', year = ''] = dateString.split('-');
                 const date = new Date(Number(day), Number(month), Number(year));
-                const formattedDate = `${date.getMonth()}/${date.getDate() - 1}/${date.getFullYear()}`;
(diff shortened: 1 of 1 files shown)


MinhxNguyen7:
Are you sure that it's an off-by-one error and not a "0 is 12 in mod space"? I don't think you're wrong; I just want to make sure

laggycomputer:
> Are you sure that it's an off-by-one error and not a "0 is 12 in mod space"? I don't think you're wrong; I just want to make sure

Not sure what you're referencing; I can trace the chain of custody of that data all the way to the underlying AAPI call and there's never any modulo. The date string returned by AAPI is split on `-` and parsed as a `Date` and the UI extracts the month/date/year from that `Date`.

The fix I apply here is also to blatantly incorrect code (see `Date` docs).

MinhxNguyen7 (approved):
Alright

PR #1191, opened 2025-03-08
feat: reduce cache lifetime

KevinWu098:
## Summary
1. Drops AA's client side Websoc cache to 5 minutes, more closely matching AAPI's ~5 minute scraping cadence

## Test Plan
1. Make a query
2. Make the same query; this should be instant (cached)
3. Wait at least 5 minutes but no more than 15 minutes
4. Query should not be cached

## Issues
Closes #


alexespejo:
lgtm!

alexespejo (commented):
seems legit

alexespejo (approved):

PR #1192, opened 2025-03-08
Restyle PWA Banner

### LINKED ISSUE #1182: Restyle PWA Banner
The PWA banner is very visible to users, we should leverage more default/standard MUI components and rely less on custom styles. Ideal goal would be a more elegant, and not too loud restyle (but still very visually apparent for users)

Image

hashedonemilliontimes:
Restyle PWA Banner #1182: Use Standard MUI Components

## Summary
Restyled the PWA banner to be more elegant and visually appealing by leveraging standard MUI components instead of custom styles. Simplified the design to maintain visibility without overwhelming users.

## Test Plan
- Verified the new PWA banner styling on desktop and mobile views.
- Tested interactions like close button and responsiveness.
- Ensured no layout shifts or console errors.

## Issues
Closes #1182

Closes #

### DIFF
--- a/apps/antalmanac/src/components/InstallPWABanner.tsx
+++ b/apps/antalmanac/src/components/InstallPWABanner.tsx
@@ -7,7 +7,6 @@ import { shallow } from 'zustand/shallow';
 
 import { getLocalStoragePWADismissalTime, setLocalStoragePWADismissalTime } from '$lib/localStorage';
 import { BeforeInstallPromptEvent, usePWAStore } from '$stores/PWAStore';
-import { useThemeStore } from '$stores/SettingsStore';
 
 function InstallPWABanner() {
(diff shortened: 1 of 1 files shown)


xgraceyan:
Unfortunately the banner restyle is no longer needed; after further discussion about how to most effectively implement the PWA prompt, we're going to temporarily disable the banner for now while we figure out another solution. 

Thanks for your contribution regardless -- we'll update this if needed!

KevinWu098:
Just to add some additional context — we discovered some compatibility issues with the popup on certain browsers which made restarted discussion about how we could best present the PWA option to users, which we ultimately determined to (probably) not be the popup. Still, we appreciate your time and contribution!

hashedonemilliontimes:
Thank you for the info! I will keep on the lookout for another good first issue.

PR #1194, opened 2025-03-09
feat: Auto Display Import When Using Google Auth

alexespejo:
## Summary
This is meant to only happen on the first login, and if the user doesn't already have a schedule (new user)

## Issues

Closes #


alexespejo:
@MinhxNguyen7 When you sign in with Google it's not obvious that you can import your schedules from "old" AntAlmanac, so this auto prompts the user to import with their old schedule id

MinhxNguyen7 (approved):
Sure

PR #1200, opened 2025-03-10
feat: outage page

KevinWu098:
## Summary
1. Adds an outage page which can be manually triggered by flipping a boolean in `App.tsx`

Screenshot 2025-03-09 at 10 44 36 PM

## Test Plan
1. Navigating to `/outage` should, by default, do nothing.
2. If you manually change the boolean flag locally, then all routes should display the Outage page

## Issues
Closes #


MinhxNguyen7 (approved):
LGTM!

PR #1209, opened 2025-03-16
refactor: fallback schedules

### LINKED ISSUE #1208: Refactor Skeleton Mode
We should refactor skeleton mode out of `AppStore` and into its own dedicated Zustand store. Probably name it something else too...

KevinWu098:
## Summary
Refactors away the SkeletonSchedule from our (soon-to-be legacy) Schedules store to a Zustand store. Removes event emitters, updates logic for reading and writing skeleton schedules.

This PR largely removes the old event emitter + event listener + useEffect pattern we had in many of our components in favor of Zustand's useStore hook. In some cases, this required additional logic changes (e.g. ScheduleSelect), but the PR is _mostly_ refactoring.

## Test Plan
Skeleton Schedules trigger when a user's schedule cannot be fully loaded because the AnteaterAPI is down. This results in us only having access to select fields (what AA stores on our DB), which are primarily the schedule names, custom events, schedule notes, and the course ID of the added courses.

For testing, you'll need to block requests to the AnteaterAPI via [Chrome Dev Tools](https://developer.chrome.com/docs/devtools/network-request-blocking), then attempt to load a schedule **with courses added**. Fallback does not trigger if there are no courses, as it would make no requests to the API. Blocking the keyword `websoc` should work here.

Fallback should trigger immediately when the API request fails, displaying schedules correctly and blocking features which could require API data (e.g. save, load, editing).

## Issues
Closes #1208


apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -325,7 +322,7 @@ export class Schedules {
     /**
      * Get a reference ito the custom event that matches the ID.
      */
-    getExistingCustomEvent(customEventId: number) {
+    getExistingCustomEvent(customEventId: string | number) {
```
KevinWu098, line 325: note: the type updates in these files aren't a part of the broader PR, but they fix type errors that have been lingering since the RDS changes (?), ~a few months maybe

apps/antalmanac/src/stores/AppStore.ts
```diff
@@ -347,11 +333,7 @@ class AppStore extends EventEmitter {
     }
 
     private async loadScheduleFromSaveState(savedSchedule: ScheduleSaveState) {
-        try {
-            await this.schedule.fromScheduleSaveState(savedSchedule);
-        } catch {
-            return false;
-        }
```
KevinWu098, line 377: note: this was some pretty gnarly logic — now we allow the error to propagate upwards and be handled by the caller. See `AppStoreActions`

apps/antalmanac/src/routes/Home.tsx
```diff
@@ -92,7 +92,7 @@ export default function Home() {
     return (
         <MuiPickersUtilsProvider utils={DateFnsUtils}>
             <PatchNotes />
-            <InstallPWABanner />
+            {/* <InstallPWABanner /> */}
```
KevinWu098: `InstallPWABanner` was already disabled via #1213, this just further ensures none of its logic is applied unnecessarily

MinhxNguyen7 (changes requested):
This doesn't seem to work. I tried to load the `test` schedule (which has classes), and nothing showed up in the added pane.

MinhxNguyen7:
@KevinWu098 Reminder

apps/antalmanac/src/lib/api/trpc.ts
```diff
@@ -13,8 +13,22 @@ function getEndpoint() {
     return import.meta.env.MODE === 'development' ? `https://dev.api.antalmanac.com` : `https://api.antalmanac.com`;
 }
 
+const url = getEndpoint() + '/trpc';
```
MinhxNguyen7: This should just be folded into `getEndpoint` and renamed to `getTrpcEndpoint`.  The only other place that uses that function appends `/trpc` onto it anyway.

apps/antalmanac/src/lib/api/trpc.ts
```diff
@@ -13,8 +13,22 @@ function getEndpoint() {
...
+                return op.context.skipBatch === true;
+            },
+            true: httpLink({
+                url,
+            }),
+            // when condition is false, use batching
+            false: httpBatchLink({
+                url,
+            }),
+        }),
```
MinhxNguyen7, line 31: What does this do? Can you add a comment?

apps/antalmanac/src/routes/Home.tsx
MinhxNguyen7: Just throw it on another PR. It's one line. I'll review it in a snap.

apps/antalmanac/src/stores/AppStore.ts
```diff
@@ -257,7 +243,7 @@ class AppStore extends EventEmitter {
         this.emit('customEventsChange');
     }
 
-    deleteCustomEvent(customEventId: number, scheduleIndices: number[]) {
+    deleteCustomEvent(customEventId: number | string, scheduleIndices: number[]) {
```
MinhxNguyen7, line 247: Why is this suddenly accepting a string? What's the type for a custom event on the backend? It should just be consistent instead of allowing anything

apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -426,8 +422,8 @@ export class Schedules {
     /**
      * Checks if a schedule contains the custom event ID
      */
-    doesCustomEventExistInSchedule(customEventId: number, scheduleIndex: number) {
-        for (const customEvent of this.schedules.at(scheduleIndex)?.customEvents ?? []) {
+    doesCustomEventExistInSchedule(customEventId: string | number, scheduleIndex: number) {
```
MinhxNguyen7, line 425: Same thing about the types

MinhxNguyen7 (changes requested):
This is what I see when I try to load the user `test`, even without blocking requests. I think you broke schedule loading in general,
image

Have you tested this?

apps/antalmanac/src/stores/AppStore.ts
KevinWu098: This fixes some type errors we've had for a while [post-RDS](https://github.com/icssc/AntAlmanac/blob/6301c47214819f5944396965fba38faa4c5fcae1/packages/types/src/customevent.ts#L8)

KevinWu098:
> This is what I see when I try to load the user `test`, even without blocking requests. I think you broke schedule loading in general, image
> 
> Have you tested this?

Loading works locally — isn't staging broken?

apps/antalmanac/src/lib/api/trpc.ts
KevinWu098: Personal preference mostly, but I'd rather not write `/trpc` across four branches — I've consolidated usage to all draw from `const trpcUrl`

MinhxNguyen7:
> isn't staging broken

Oops. I forgot about that

KevinWu098:
i give up

PR #1213, opened 2025-03-19
Temporarily disable PWA banner

### LINKED ISSUE #1182: Restyle PWA Banner

xgraceyan:
## Summary
Temporarily disables PWA banner from rendering until more effective implementation has been figured out

## Test Plan
- Ensure PWA banner doesn't appear (ideally in a new browser)

## Issues

Closes #1182


KevinWu098 (approved):
I'd maybe suggest changing the logic in the store or somewhere else in the implementation (i.e. so we don't even have to consider rendering the component), but this is solid for our purposes

Issue #1158, opened 2025-02-10
Clean up RDS post auth


alexespejo:
This class is getting very bloated and should be broken up, but you can just leave it as is and clean up later.

_Originally posted by @MinhxNguyen7 in https://github.com/icssc/AntAlmanac/pull/1127#discussion_r1948010608_

KevinWu098:
PR will need to consider whether public methods should be called w/ a transaction or not: https://github.com/icssc/AntAlmanac/pull/1214/files?diff=split#r2004589736

PR #1214, opened 2025-03-20
feat: decrease save times

### LINKED ISSUE #1114: Optimize schedule save

KevinWu098:
## Summary
The migration from Dynamo to RDS (#993, #1064) resulted in significant performance degradation for saving schedules. Users with (reasonably) large schedules across multiple quarters began experiencing save times of >15 seconds, which timed out our lambdas and required a hotfix patch (#1115).

This PR removes redundant database operations and nested transactions which significantly increased time required to save a schedule.

On a user with 17 schedules, 11 of which have 16 elements (courses+custom events) (and the rest have a regular amount ~8), this PR brings save times down to ~5.75 seconds. On live, this operation takes ~16 seconds. That's an improvement of 64% 🤯. 

On even more extreme testing with 15 large schedules that each have 26 elements, plus 6 normal schedules, save times reach ~22 seconds on live. This PR achieves ~10 seconds, at a ~55% improvement.

## Test Plan
On staging, create and populate a schedule with a bunch of courses. Duplicate that schedule a few times (12 or so should be reasonable).

1. Save the user and look at how long it took. This can be checked via the Network time in Chrome Dev Tools, look for a call to `users.saveUserData` and look at its timing tab.
2. Make edits to a schedule, save, and reload. Changes should be persisted. There also shouldn't be any "dangling" courses or schedules in the database. Checking this would require access to the DB, so please contact myself or @MinhxNguyen7 or @adcockdalton for access.
3. Try your best to stress test this! Change schedule names. Save it to another username. Move the order around.

> [!NOTE]
> Attempting to test this feature on a local DB will likely not be very effective. The most accurate results will be on the staging instance.

## Issues
Closes #1114

## Future Followup
The `RDS` class is a mess. Methods are misnamed (e.g. `upsert` when there's only an `insert`), the class is seemingly not necessary beyond being functioning as an index for methods, etc. 

See: #1158

### DIFF
--- a/apps/antalmanac/src/components/Calendar/Toolbar/ScheduleSelect/schedule-select-buttons/DeleteScheduleButton.tsx
+++ b/apps/antalmanac/src/components/Calendar/Toolbar/ScheduleSelect/schedule-select-buttons/DeleteScheduleButton.tsx
@@ -1,5 +1,5 @@
 import { Clear as ClearIcon } from '@mui/icons-material';
-import { Box, IconButton, Tooltip } from '@mui/material';
+import { IconButton, Tooltip } from '@mui/material';
 import { useCallback, useState } from 'react';
 
 import DeleteScheduleDialog from '$components/dialogs/DeleteSchedule';
@@ -22,8 +22,8 @@ export function DeleteScheduleButton({ index, disabled }: DeleteScheduleButtonPr
     }, []);
 
     return (
-        <Box>
-            <Tooltip title="Delete Schedule">
+        <>
+            <Tooltip title="Delete Schedule" disableInteractive>
                 <span>
                     <IconButton
                         onClick={handleOpen}
@@ -35,6 +35,6 @@ export function DeleteScheduleButton({ index, disabled }: DeleteScheduleButtonPr
                 </span>
             </Tooltip>
             <DeleteScheduleDialog fullWidth open={open} index={index} onClose={handleClose} />
-        </Box>
+        </>
     );
 }
--- a/apps/backend/src/lib/rds.ts
+++ b/apps/backend/src/lib/rds.ts
@@ -23,8 +23,8 @@ export class RDS {
     /**
      * If a guest user with the specified name exists, return their ID, otherwise return null.
      */
-    static async guestUserIdWithNameOrNull(db: DatabaseOrTransaction, name: string): Promise<string | null> {
-        return db
+    private static async guestUserIdWithNameOrNull(tx: Transaction, name: string): Promise<string | null> {
+        return tx
             .select({ id: accounts.userId })
             .from(accounts)
             .where(and(eq(accounts.accountType, 'GUEST'), eq(accounts.providerAccountId, name)))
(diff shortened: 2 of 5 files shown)


MinhxNguyen7:
Very cool.

I was under the impression that nested transactions are automatically resolved by the ORM, but I guess that's the pitfall of using ORMs; you don't actually know what's going on.

apps/antalmanac/src/components/Calendar/Toolbar/ScheduleSelect/schedule-select-buttons/DeleteScheduleButton.tsx
KevinWu098, line 1: While testing, I encountered significant annoyance with tooltips that didn't have `disableInteractive`. This PR includes a few quick fixes to tooltips used in components relating to the schedule and adding courses

apps/backend/src/lib/rds.ts
```diff
@@ -30,39 +37,40 @@ export class RDS {
      * @returns The new/existing user's ID
      */
     static async createGuestUserOptional(db: DatabaseOrTransaction, name: string) {
-        return db.transaction(async (tx) => {
```
KevinWu098, line 33: This operation is always called within an existing transaction. Like other operations in this PR which had their nested transactions removed, care will need to be taken that public methods are called with transactions.

Something to note for #1158, I think

apps/backend/src/lib/rds.ts
```diff
@@ -149,15 +154,15 @@ export class RDS {
 
     /** Deletes and recreates all of the user's schedules and contents */
     private static async upsertSchedulesAndContents(
-        db: DatabaseOrTransaction, userId: string, scheduleArray: ShortCourseSchedule[]
+        db: DatabaseOrTransaction,
+        userId: string,
+        scheduleArray: ShortCourseSchedule[]
     ): Promise<string[]> {
         // Drop all schedules, which will cascade to courses and custom events
         await db.delete(schedules).where(eq(schedules.userId, userId));
```
KevinWu098: This cascade is why we can remove the operations deleting all custom events and courses from existing schedules. We were needlessly duplicating operations here.

apps/antalmanac/src/components/Calendar/Toolbar/ScheduleSelect/schedule-select-buttons/DeleteScheduleButton.tsx
MinhxNguyen7: Hm ok

MinhxNguyen7 (changes requested):
Looks good, generally. 

To communicate the necessity of using a transaction, you can change the `DatabaseOrTransaction` type to just `Transaction` (and chang the argument name accordingly). It won't stop you drom passing in a `Database`, but it's at least good indication.

MinhxNguyen7:
Also, I'm taking your word that this runs.

apps/backend/src/lib/rds.ts
MinhxNguyen7: Feel free to add this to that issue

apps/antalmanac/src/components/Calendar/Toolbar/ScheduleSelect/schedule-select-buttons/DeleteScheduleButton.tsx
KevinWu098: For context, because the icons are clustered so close to one another, attempting to hover the edit button a row below would result in your mouse hovering the tooltip from the above row, resulting in it blocking the intended action — you can see this on live pretty easily

chrome-capture-2025-3-23

More context from MUI docs: https://mui.com/material-ui/react-tooltip/#interactive

KevinWu098:
Going to advise that this PR merges in no earlier than 4/1 so the whole team will be back at UCI and can more easily receive + triage bug reports

<sub>crashing antalmanac on april fools? what a joke 🥁</sub>

MinhxNguyen7 (approved):
It's probably fine. That being said, unless you want to be on call to address problems, it should be merged when we're actually back

PR #1215, opened 2025-03-20
style: override rbc-today

### LINKED ISSUE #1210: React Big Calendar on current day is not styled appropriately
Image

On Finals calendars, we actually set the date correctly, so when it becomes the current day on the calendar, other styles are applied by default. We should either implement a proper, AA-styled current day or completely override it

KevinWu098:
## Summary
React Big Calendar has a set of default styles. One of them, `rbc-today` applies to the current day and is only visible when the Finals view is toggled on. This is because we normally set the date to an arbitrary date, but set it accurately for finals.

This PR overrides the style by using the built-in `dayPropsGetter` to override the styling on dark mode.

### Before
Screenshot 2025-03-19 at 7 20 39 PM

### After
Screenshot 2025-03-19 at 7 20 44 PM

## Test Plan
Load a schedule with winter quarter finals.

On light mode, the current day should be blue. On dark mode, the current day should be a lighter gray.

## Issues
Closes #1210

### DIFF
--- a/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
+++ b/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
@@ -1,10 +1,11 @@
 import 'react-big-calendar/lib/css/react-big-calendar.css';
 import './calendar.css';
 
-import { Box } from '@material-ui/core';
+import { Box, useTheme } from '@mui/material';
 import moment from 'moment';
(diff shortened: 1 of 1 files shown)


jotalis (approved):
lrstm !

KevinWu098:
> lrstm ! 

What does lrstm mean?

MinhxNguyen7:
Reasonably sound?

PR #1216, opened 2025-03-20
fix: 🐛 finals throws error

alexespejo:
## Summary
Allows to search course from final. The problem was that final calendar events didn't pass the same props to the popover so it would pass in undefined values to the search. This change checks if the values are undefined and if they are then it will use the sectionCode to find the original course info in AppStore 
## Issues
Quick search on a final blows up 

Closes #


apps/antalmanac/src/hooks/useQuickSearch.tsx
```diff
@@ -11,13 +13,22 @@ export function useQuickSearch() {
...
+        (deptValue: string, courseNumber: string, termValue: string, sectionCode = '') => {
             const queryParams = {
                 term: termValue,
                 deptValue: deptValue,
                 courseNumber: courseNumber,
             };
 
+            if ((!termValue || !deptValue || !courseNumber) && sectionCode.length > 0) {
+                const course: ScheduleCourse = (AppStore.getAddedCourses() as unknown as ScheduleCourse[]).filter(
+                    (course) => 'sectionCode' in course && course.sectionCode === sectionCode
```
alexespejo: linter didn't like setionCode

MinhxNguyen7 (commented):
The page doesn't blow up anymore, but can you make it so that the search either actually works, or at least don't make the course name a button *suggesting* that it would work?

alexespejo:
> The page doesn't blow up anymore, but can you make it so that the search either actually works, or at least don't make the course name a button _suggesting_ that it would work?

Sorry, bad description this PR allows the search to work

MinhxNguyen7:
If that's the case, it's not working with next quarter's ICS 6B

alexespejo:
I think there's a discrepancy in how each schedule is stored it seems like old records use `CourseEvent` while new ones use `ScheduleCourse`? Note how `26` doesn't have the `sectionCode` attribute. This is ready btw

image

apps/antalmanac/src/hooks/useQuickSearch.tsx
```diff
@@ -11,13 +13,24 @@ export function useQuickSearch() {
...
-        (deptValue: string, courseNumber: string, termValue: string) => {
+        (deptValue: string, courseNumber: string, termValue: string, sectionCode = '') => {
             const queryParams = {
                 term: termValue,
                 deptValue: deptValue,
                 courseNumber: courseNumber,
             };
 
+            if ((!termValue || !deptValue || !courseNumber) && sectionCode.length > 0) {
+                const course: ScheduleCourse = (AppStore.getAddedCourses() as unknown as ScheduleCourse[]).filter(
```
MinhxNguyen7: Why is this cast necessary. I don' see any errors here.
```suggestion
                const course: ScheduleCourse = AppStore.getAddedCourses().filter(
```

MinhxNguyen7 (dismissed):
Just one small thing.

apps/antalmanac/src/hooks/useQuickSearch.tsx
KevinWu098: issue: If you're looking for the first matching element, prefer `find` instead of filtering then accessing the first index!

KevinWu098 (changes requested):

apps/antalmanac/src/hooks/useQuickSearch.tsx
alexespejo: I was following copilots orders to appease the linter

KevinWu098 (changes requested):
It's been a minute on this PR, to say the least — err... upon second inspection, I don't understand why we're doing essentially a hack to find the relevant data...

I won't tell you outright what the fix is, but it's a two line fix and it involves none of the files currently in the diff... as a hint, I'd suggest inspecting what differs between the a course event and a finals course event... where do those values come from?

<details>
  <summary>Answer:</summary>

`calendarizeFinals` in `calendarizeHelpers` is not returning the right type...
  
</details>

KevinWu098 (approved):
LGTM

alexespejo:
> It's been a minute on this PR, to say the least — err... upon second inspection, I don't understand why we're doing essentially a hack to find the relevant data...
> 
> I won't tell you outright what the fix is, but it's a two line fix and it involves none of the files currently in the diff... as a hint, I'd suggest inspecting what differs between the a course event and a finals course event... where do those values come from?
> 
> Answer:

lol I wouldnt have thought of that

KevinWu098 (changes requested):
actually uh let's fix that test

KevinWu098:
> > It's been a minute on this PR, to say the least — err... upon second inspection, I don't understand why we're doing essentially a hack to find the relevant data...
> > I won't tell you outright what the fix is, but it's a two line fix and it involves none of the files currently in the diff... as a hint, I'd suggest inspecting what differs between the a course event and a finals course event... where do those values come from?
> > Answer:
> 
> lol I wouldnt have thought of that

Yeah... we really should be stricter on our builds about type errors and linting errors... 🤔  — food for thought

KevinWu098 (approved):

PR #1220, opened 2025-04-01
Day Filtering in Advanced Search

### LINKED ISSUE #1190: Advanced Search should also support filtering by Day
The AAPI Websoc endpoint supports filtering by day: https://anteaterapi.com/reference#tag/websoc/GET/v2/rest/websoc

Image

IsaacNguyen:
## Summary
Added filtering by days for classes in advanced search
Screen Recording 2025-04-01 at 12 06 59 AM

## Test Plan

- [x] Make sure it works for one day selected
- [x] Make sure it works when multiple days are selected (TuTh or MWF or MTuF)
- [x] Make sure it works when no days are selected

## Issues

Closes #1190

### DIFF
--- a/apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearchTextFields.tsx
+++ b/apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearchTextFields.tsx
@@ -2,7 +2,11 @@ import { TextField, Box, FormControl, InputLabel, Select, Switch, FormControlLab
 import { MenuItem } from '@mui/material';
 import { useState, useEffect, useCallback } from 'react';
 
-import { EXCLUDE_RESTRICTION_CODES_OPTIONS } from '$components/RightPane/CoursePane/SearchForm/AdvancedSearch/constants';
+import {
+    EXCLUDE_RESTRICTION_CODES_OPTIONS,
+    DAYS_OPTIONS,
+    DAY_ORDER,
+} from '$components/RightPane/CoursePane/SearchForm/AdvancedSearch/constants';
 import RightPaneStore from '$components/RightPane/RightPaneStore';
 
 export function AdvancedSearchTextFields() {
@@ -17,6 +21,11 @@ export function AdvancedSearchTextFields() {
     const [excludeRestrictionCodes, setExcludeRestrictionCodes] = useState(
         RightPaneStore.getFormData().excludeRestrictionCodes
     );
+    const [days, setDays] = useState(() => {
+        const daysString = RightPaneStore.getFormData().days;
+        const daysArray = daysString ? daysString.split(/(?=[A-Z])/) : [];
+        return daysArray.sort((a, b) => DAY_ORDER.indexOf(a) - DAY_ORDER.indexOf(b));
+    });
 
     const resetField = useCallback(() => {
         const formData = RightPaneStore.getFormData();
@@ -29,6 +38,7 @@ export function AdvancedSearchTextFields() {
         setRoom(formData.room);
         setDivision(formData.division);
         setExcludeRestrictionCodes(formData.excludeRestrictionCodes);
+        setDays([]);
     }, []);
 
     useEffect(() => {
@@ -97,6 +107,16 @@ export function AdvancedSearchTextFields() {
                     case 'excludeRestrictionCodes':
                         setExcludeRestrictionCodes(stringValue);
                         break;
+                    case 'days': {
+                        const daysArray = Array.isArray(value) ? value : [value];
+                        const orderedDaysArray = daysArray.sort((a, b) => {
+                            return DAY_ORDER.indexOf(a) - DAY_ORDER.indexOf(b);
+                        });
+
+                        setDays(orderedDaysArray);
+                        RightPaneStore.updateFormValue('days', orderedDaysArray.join(''));
+                        break;
+                    }
                     default:
                         break;
                 }
@@ -269,6 +289,37 @@ export function AdvancedSearchTextFields() {
                     renderValue={(selected) => (selected as string[]).join(', ')}
                 >
                     {EXCLUDE_RESTRICTION_CODES_OPTIONS.map((option) => (
+                        <MenuItem
+                            key={option.value}
+                            value={option.value}
+                            style={{
+                                maxWidth: 240,
+                            }}
+                        >
+                            <span
+                                style={{
+                                    overflow: 'hidden',
+                                    textOverflow: 'ellipsis',
+                                    whiteSpace: 'nowrap',
+                                }}
+                            >
+                                {option.label}
+                            </span>
+                        </MenuItem>
+                    ))}
+                </Select>
+            </FormControl>
+
+            <FormControl style={{ minWidth: 150 }}>
+                <InputLabel id="days-label">Days</InputLabel>
+                <Select
+                    multiple
+                    labelId="days-label"
+                    value={days}
+                    onChange={handleChange('days')}
+                    renderValue={(selected) => (selected as string[]).join(', ')}
+                >
+                    {DAYS_OPTIONS.map((option) => (
                         <MenuItem
                             key={option.value}
                             value={option.value}
--- a/apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearchTextFields.tsx
+++ b/apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearchTextFields.tsx
@@ -309,8 +309,8 @@ export function AdvancedSearchTextFields() {
                     renderValue={(selected) =>
                         (selected as string[])
                             .sort((a, b) => {
-                                const orderA = DAYS_OPTIONS.find((day) => day.value === a)?.order ?? Infinity;
-                                const orderB = DAYS_OPTIONS.find((day) => day.value === b)?.order ?? Infinity;
+                                const orderA = DAYS_OPTIONS.findIndex((day) => day.value === a);
+                                const orderB = DAYS_OPTIONS.findIndex((day) => day.value === b);
                                 return orderA - orderB;
                             })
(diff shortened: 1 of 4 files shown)


apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/constants.ts
```diff
@@ -18,3 +18,13 @@ export const EXCLUDE_RESTRICTION_CODES_OPTIONS = [
...
     { value: 'X', label: 'X: Separate authorization codes required to add, drop, or change enrollment' },
 ];
+
+export const DAYS_OPTIONS = [
+    { value: 'M', label: 'M: Monday' },
+    { value: 'Tu', label: 'Tu: Tuesday' },
+    { value: 'W', label: 'W: Wednesday' },
+    { value: 'Th', label: 'Th: Thursday' },
+    { value: 'F', label: 'F: Friday' },
+];
```
KevinWu098: issue: courses can be on the weekend too :)

```
{
  "ok": false,
  "message": "...Valid days of the week are M, Mo, Mon, Tu, Tue, Tues, W, We, Wed, Th, Thu, Thur, F, Fr, Fri, S, Sa, Sat, Su, Sun."
}
```

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/constants.ts
```diff
@@ -18,3 +18,13 @@ export const EXCLUDE_RESTRICTION_CODES_OPTIONS = [
...
+
+export const DAY_ORDER = ['M', 'Tu', 'W', 'Th', 'F', 'Sa', 'Su'];
```
KevinWu098: nit: could this be combined w/ `DAYS_OPTIONS`?

ex: `{ value: 'M', label: 'M: Monday', order: 0}`

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearchTextFields.tsx
```diff
@@ -29,6 +38,7 @@ export function AdvancedSearchTextFields() {
         setRoom(formData.room);
         setDivision(formData.division);
         setExcludeRestrictionCodes(formData.excludeRestrictionCodes);
+        setDays([]);
```
KevinWu098: question: why are we setting an empty array rather than matching how the above lines are doing it? It's been a while since I worked in this file...

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearchTextFields.tsx
```diff
@@ -17,6 +21,11 @@ export function AdvancedSearchTextFields() {
     const [excludeRestrictionCodes, setExcludeRestrictionCodes] = useState(
         RightPaneStore.getFormData().excludeRestrictionCodes
     );
+    const [days, setDays] = useState(() => {
+        const daysString = RightPaneStore.getFormData().days;
+        const daysArray = daysString ? daysString.split(/(?=[A-Z])/) : [];
+        return daysArray.sort((a, b) => DAY_ORDER.indexOf(a) - DAY_ORDER.indexOf(b));
+    });
```
KevinWu098: issue: this seems inconsistent with how we've implemented other multi-select fields (e.g. restriction codes). Is there a reason we're not storing it as a string?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearchTextFields.tsx
```diff
@@ -97,6 +107,16 @@ export function AdvancedSearchTextFields() {
...
+                    case 'days': {
+                        const daysArray = Array.isArray(value) ? value : [value];
+                        const orderedDaysArray = daysArray.sort((a, b) => {
+                            return DAY_ORDER.indexOf(a) - DAY_ORDER.indexOf(b);
+                        });
+
+                        setDays(orderedDaysArray);
+                        RightPaneStore.updateFormValue('days', orderedDaysArray.join(''));
+                        break;
+                    }
```
KevinWu098: repeat (issue): this seems inconsistent with our other fields

KevinWu098 (changes requested):
Thanks for working on this feature — tested it on the deployment and it works great. Have added some review comments below 🫡

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearchTextFields.tsx
IsaacNguyen: My original reason for wanting to store it as an array was because I was having trouble implementing it similar to how we store restriction codes to exclude (it would store `Th` as `T,h` and `Tu` as `T,u`). Looking back at it now, I could have implemented it differently, but I think storing the days as an array allows me to maintain relative ordering of days (TuMTh becomes MTuTh) in the textbox. So I think this would be the correct way to store the days unless you have a more efficient way to maintain the relative ordering.

KevinWu098: If you stored it as a string `MTuTh`, couldn't you use this snippet you also wrote: `formData.days.split(/(?=[A-Z])/)`

KevinWu098: Also, does relative ordering matter? If you think it's important for the select, then when rendering values, you can sort by order there

IsaacNguyen: Yeah you're right. For some reason I thought AAPI required that we input the days in order for it work properly, but testing it now I realized that is not the case. I will try to fix all of your concerns by tonight.

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearchTextFields.tsx
```diff
@@ -315,9 +304,17 @@ export function AdvancedSearchTextFields() {
...
-                    renderValue={(selected) => (selected as string[]).join(', ')}
+                    renderValue={(selected) =>
+                        (selected as string[])
+                            .sort((a, b) => {
+                                const orderA = DAYS_OPTIONS.find((day) => day.value === a)?.order ?? Infinity;
+                                const orderB = DAYS_OPTIONS.find((day) => day.value === b)?.order ?? Infinity;
+                                return orderA - orderB;
+                            })
+                            .join(', ')
+                    }
```
KevinWu098: Sorry for going back and forth w/ you on this, but I have to say that the nullish coalesce + Infinity combo is a little suspect...

Having `DAY_ORDER` separate like you did previously and sorting using that might be better here 😅

KevinWu098 (changes requested):

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearchTextFields.tsx
IsaacNguyen: I’m not really sure what you mean by suspect, but I really couldn’t figure out a better way to go about bypassing the non-null type checking. I’ll switch to using `DAY_ORDER` but if we were gonna do it like this, do you have a suggestion on a better way to go about it?

KevinWu098: By suspect, I mean that the code feels sus /hj

More specifically, it’s just doesn’t feel super great having to deal with potentially undefined values here and then having to use Infinity to patch that up

KevinWu098: If you wanted to keep it like this, it could actually be cleaner / less code to just use findIndex, so you're guaranteed a `number` return. No need for `DAY_ORDER` 

```
                            .sort((a, b) => {
                                const orderA = DAYS_OPTIONS.findIndex((day) => day.value === a);
                                const orderB = DAYS_OPTIONS.findIndex((day) => day.value === b);
                                return orderA - orderB;
                            })
```

KevinWu098:
@IsaacNguyen Is this PR ready for review?

IsaacNguyen:
> @IsaacNguyen Is this PR ready for review?

Yes

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/constants.ts
```diff
@@ -18,3 +18,13 @@ export const EXCLUDE_RESTRICTION_CODES_OPTIONS = [
...
+
+export const DAYS_OPTIONS = [
+    { value: 'Su', label: 'Su: Sunday', order: 0 },
+    { value: 'M', label: 'M: Monday', order: 1 },
+    { value: 'Tu', label: 'Tu: Tuesday', order: 2 },
+    { value: 'W', label: 'W: Wednesday', order: 3 },
+    { value: 'Th', label: 'Th: Thursday', order: 4 },
+    { value: 'F', label: 'F: Friday', order: 5 },
+    { value: 'Sa', label: 'Sa: Saturday', order: 6 },
+];
```
KevinWu098: issue: I don't believe we're using the `order` field here, so I think we can remove this

KevinWu098 (changes requested):
One last change, then LGTM!

KevinWu098 (approved):
LGTM!

PR #1221, opened 2025-04-01
feat: pregenerate terms

### LINKED ISSUE #1123: TermData should be generated `preStart`
Similarly to how we generate the fuzzy search data, one convenient way to keep our term data updated would be to have it fetched from AAPI at `preStart`.

Fetch calendar and term data from the AAPI at `preStart`, write it to a generated file, then consume it via a new TRPC route on the frontend.

jotalis:
## Summary
- Introduced a script to fetch calendar term data from the Anteater API on prebuild / start
  -  Displays terms from the last 10 years and only those with WebSoc available
- Updated termData structure to use Date objects instead of arrays for start and finals dates.
- Fixed jankiness and bugs concerning finals start date calculations, particularly with schedules that have finals on weekends.
- Nuked... old code

## Test Plan
- Ensure terms are generated and displayed properly in term drop down
- Ensure courses can be successfully searched using various different terms 
- Ensure finals schedule is displayed properly with and without a weekend final (e.g., Math 2B)
- Ensure download as .ics works as expected
  - Event dates are correct
  - Both class and final events show

## Issues
- Removes manual override of `2022 Summer10wk` and `2022 Summer1` dates to the 20th, originally there because Juneteenth is observed 6/20/22.  
  
Closes #1123


apps/antalmanac/src/lib/termData.ts
```diff
@@ -119,43 +48,32 @@ function getDefaultTerm(events: (CustomEvent | CourseEvent)[] = []): Term {
     return term;
 }
 
-// Returns the default finals start as array
-function getDefaultFinalsStart() {
-    // FIXME: Un-offset once Spring starts, or figure out a proper fix
-    return termData[defaultTerm + 1].finalsStartDate;
+function getDefaultFinalsStartDate() {
+    const defaultFinalsDate = termData[defaultTerm]?.finalsStartDate;
```
KevinWu098: issue: I see that you’ve handled this value being undefined, but prefer using .at() when operating on an array

jotalis: why do we prefer using .at()? we'd still need the optional chaining operator for the cases where `finalsStartDate` is undefined; seems redundant, especially if we never deal with negative indices

KevinWu098: `.at()`'s value isn't just in that it handles negative indices, but that it explicitly types the result as potentially undefined, which accessing via `[]` does not

KevinWu098: i.e. you could remove the optional chain from `termData[defaultTerm]` and not get a type error, even though that object might not exist

KevinWu098: (it's mostly a nit, but it is something I do like to note (as an issue) when working w/ arrays)

jotalis: removing the optional chain from `termData[defaultTerm]` actually does throw a type error. 

sure, using `.at()` makes it explicit that an object can be undefined, but that doesn't guarantee safety when trying to access `finalsStartDate` with regular dot notation when the object doesn't exist

then in this case, we should use the optional chain, which makes using `.at()` redundant

apps/antalmanac/scripts/fetch-term-data.ts
```diff
@@ -0,0 +1,96 @@
+import { mkdir, writeFile } from 'node:fs/promises';
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+import { CalendarTerm } from '@packages/antalmanac-types';
+
```
KevinWu098, line 6: issue: this type isn't actually exported from `@packages/antalmanac-types`. I belive it's missing from `types/src/index/ts`

apps/antalmanac/scripts/fetch-term-data.ts
```diff
@@ -0,0 +1,96 @@
...
+    Summer2: 'Summer Session 2',
+    Fall: 'Fall Quarter',
+    Winter: 'Winter Quarter',
+    Spring: 'Spring Quarter',
+};
+
+function sanitizeTermName(year: string, quarter: string): `${string} ${string}` {
+    return `${year} ${QUARTER_MAP[quarter] ?? quarter}`;
+}
+
```
KevinWu098: suggestion: `QUARTER_MAP` is better defined as `const` rather than as `Record<string, string>`.

```ts
const QUARTER_MAP = {
    Summer1: 'Summer Session 1',
    Summer10wk: '10-wk Summer',
    Summer2: 'Summer Session 2',
    Fall: 'Fall Quarter',
    Winter: 'Winter Quarter',
    Spring: 'Spring Quarter',
} as const;
```

This narrows our map's type, and means we can also do the following simplification (and narrowing) on `sanitizeTermName`:

```ts
function sanitizeTermName(year: string, quarter: keyof typeof QUARTER_MAP): `${string} ${string}` {
    return `${year} ${QUARTER_MAP[quarter]}`;
}
```

apps/antalmanac/scripts/fetch-term-data.ts
```diff
@@ -0,0 +1,96 @@
...
+async function main() {
+    console.log('Fetching all calendar terms from Anteater API...');
+    const calendarTerms = await fetchCalendarTerms();
+    console.log(`Fetched ${calendarTerms?.length} calendar terms.`);
+
+    calendarTerms.sort((a, b) => {
+        const dateA = new Date(a.instructionStart ?? a.finalsStart ?? 0).getTime();
+        const dateB = new Date(b.instructionStart ?? b.finalsStart ?? 0).getTime();
```
KevinWu098: question: Is `instructionStart` not always provided? Just curious about all of the nullish coalescing going on here... if I have a quarter whose start date _is_ before another, but for some reason is not present, is that first quarter's `finalsStart` guaranteed to always be before a subsequent quarter's `instructionStart`...? Is that true for our summer terms?

apps/antalmanac/scripts/fetch-term-data.ts
```diff
@@ -0,0 +1,96 @@
...
+        return dateB - dateA;
+    });
+
+    // Save only the last 10 years of terms
+    const currentYear = new Date().getFullYear();
```
KevinWu098: thought (non-blocking): We've discussed this prior, but I see no reason to limit AA's search to the last 10 years... it would already be 60 terms w/ 10 years of data... not sure that there is a negative impact to user experience by including the other ~5 years of data we have access to

apps/antalmanac/scripts/fetch-term-data.ts
```diff
@@ -0,0 +1,96 @@
...
+    return `new Date(${year}, ${month - 1}, ${day})`;
+}
+
+function serializeTerm(term: CalendarTerm): string {
+    const { year, quarter, instructionStart, finalsStart, socAvailable } = term;
+    const shortName = `${year} ${quarter}`;
+    const longName = sanitizeTermName(year, quarter);
+    const isSummerTerm = quarter.toLowerCase().includes('summer');
+
+    return `    {
```
KevinWu098, line 62: thought: I don't have a better solution, so this is non-blocking, but I really am curious if there are better ways or patterns to handle generated data...

apps/antalmanac/src/lib/download.ts
```diff
@@ -12,9 +12,7 @@ import { getDefaultTerm, termData } from '$lib/termData';
 import AppStore from '$stores/AppStore';
 
 export const quarterStartDates = Object.fromEntries(
-    termData
-        .filter((term) => term.startDate !== undefined)
-        .map((term) => [term.shortName, term.startDate as [number, number, number]])
+    termData.filter((term) => term.startDate !== undefined).map((term) => [term.shortName, term.startDate as Date])
```
KevinWu098: thought: it's a shame TS can't infer that `startDate` exists here without some nasty guards and whatever

apps/antalmanac/src/lib/termData.ts
KevinWu098: What type error do you see, when removing `?`. To be clear, I'm not saying that `termData.at(defaultTerm).finalsStartDate` would be valid _without_ the optional chain, but I _am_ saying that `termData.at(defaultTerm)` is preferable to `termData[defaultTerm]`

KevinWu098 (changes requested):
Overall a fantastic PR. I've got (mostly) nits and thoughts about some code quality items, but generally looks good to me :)

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
```diff
@@ -19,6 +19,17 @@ import AppStore from '$stores/AppStore';
...
+/*
+ * Always start week on Saturday for finals potentially on weekends.
+ * CALENDAR_VIEWS will set the correct day range
+ */
+
+moment.updateLocale('es-us', {
+    week: {
+        dow: 6,
+    },
+});
```
KevinWu098, line 31: 🔥🔥🔥

jotalis:
I realize now that we have so many unnecessary checks for whether `Term` objects and their properties (`startDate`, `finalsStartDate`, `socAvailable`) exist or not. This was enforced in the legacy code because we manually created the `Term` objects and often didn't populate all the necessary Date objects.

Now that we're fetching Term data directly from [AAPI](https://anteaterapi.com/reference#tag/calendar/GET/v2/rest/calendar), which marks these dates are required in a successful response (see 200 Response => Data), we can assume that these properties will always exist on a `Term` object and thus simplify logic in many places. For extra validation, I've let the script throw an error if we ever do not receive a fully complete response from AAPI.

apps/antalmanac/src/lib/termData.ts
jotalis: thread now moot; see below

apps/antalmanac/src/lib/download.ts
```diff
@@ -11,9 +11,7 @@ import buildingCatalogue from '$lib/locations/buildingCatalogue';
 import { getDefaultTerm, termData } from '$lib/termData';
 import AppStore from '$stores/AppStore';
 
-export const quarterStartDates = Object.fromEntries(
-    termData.filter((term) => term.startDate !== undefined).map((term) => [term.shortName, term.startDate as Date])
-);
+export const quarterStartDates = Object.fromEntries(termData.map((term) => [term.shortName, term.startDate]));
```
KevinWu098, line 14: 🔥

KevinWu098 (approved):
LGTM, really excited about us being able to pregnant terms now :)

PR #1222, opened 2025-04-01
feat: remove dept label

KevinWu098:
## Summary
Refactors and simplifies our logic for `deptLabel` and `deptValue` by completing removing `deptLabel` (it doesn't exist in our codebase anymore!)

For context, `deptLabel` is the human-friendly text (e.g. Computer Science) versus `deptValue` (e.g. COMPSCI). In a lot of our codebase, these two values are extremely coupled and are basically not used without each other. This increased complexity and introduced a long-running UI bug:

1. In the video below, I visit https://antalmanac.com/?deptValue=COMPSCI. 
2. It sends me to the `added` page because I'm signed in
    - this should not happen — if params are present, we should be given the `search` page (implemented in this PR)
4. When I visit the `search` page, the results are already populated because `deptValue` is a "required" param
5. When I click the back arrow to return to the search form, you can see that the Department Search Bar appears to be `ALL`
    - this should not happen — this occurs because although `deptValue` is correctly set to `COMPSCI` (more on that below), `deptLabel` was not being set
6. Clicking search again searches for COMPSCI, rather than "ALL"
7. Clicking back and opening the search bar shows that `COMPSCI` is _actually_ the selected value

https://github.com/user-attachments/assets/bc0c2d76-8861-4bea-8301-ee433ea677e1

To fix this problem, we now only use `deptValue` in our codebase for handling departments and only use department labels when rendering the department search bar.

## Test Plan

1. Loading in a URL without params should function as expected (e.g. `added` tab, if signed in)
2. Loading in a URL with params should direct to the search page, with values correctly filled
8. Selecting a department should correctly mark it as a recently searched
9. Using the dept search bar should still work as before

## Issues
Closes #


apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/DeptSearchBar/MobileDeptSelector.tsx
KevinWu098, line 1: unused file

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/DeptSearchBar/constants.ts
KevinWu098, line 1: This file takes the deleted `depts.ts` and set the key to `deptValue` and value to `deptLabel`

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearch.tsx
```diff
@@ -69,7 +69,7 @@ class AdvancedSearch extends PureComponent<AdvancedSearchProps, AdvancedSearchSt
 
         const formData = RightPaneStore.getFormData();
         for (const key of Object.keys(formData)) {
-            if (key === 'deptLabel' || key === 'deptValue') {
+            if (key === 'deptValue') {
                 continue;
```
jotalis: question: why are we skipping if the key equals `deptValue`?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearch.tsx
```diff
@@ -43,7 +43,7 @@ class AdvancedSearch extends PureComponent<AdvancedSearchProps, AdvancedSearchSt
         const formData = RightPaneStore.getFormData();
         const defaultFormData = RightPaneStore.getDefaultFormData();
         for (const [key, value] of Object.entries(formData)) {
-            if (key === 'deptLabel' || key === 'deptValue') {
+            if (key === 'deptValue') {
                 continue;
```
jotalis: question: why are we skipping if the key equals deptValue?

apps/antalmanac/src/stores/CoursePaneStore.ts
```diff
@@ -42,7 +41,7 @@ function paramsAreInURL() {
     return searchParams.some((param) => search.get(param) !== null);
 }
 
-function requiredParamsAreInURL() {
+export function requiredParamsAreInURL() {
```
jotalis: nit: if this isn't imported anywhere and only used a toggle in `useCoursePaneStore`, do we need to export it?

apps/antalmanac/src/components/ScheduleManagement/ScheduleManagement.tsx
```diff
@@ -51,7 +51,9 @@ export function ScheduleManagement() {
 
         const userId = getLocalStorageUserId();
 
-        if (userId === null) {
+        // Change the tab to the "added classes" tab if the user was previously logged in
+        // or if a search parameter is in the URL
+        if (userId === null || paramsAreInURL()) {
             setActiveTab('search');
         } else if (isMobile) {
             setActiveTab('calendar');
```
jotalis: issue (non-blocking): comment is inaccurate and confusing

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/DeptSearchBar/DeptSearchBar.tsx
```diff
@@ -1,175 +1,101 @@
...
-
-import RightPaneStore from '../../../RightPaneStore';
-
-import depts from './depts';
+import { Autocomplete, Box, TextField } from '@mui/material';
+import { useCallback, useEffect, useState } from 'react';
 
+import { DEPARTMENT_MAP } from '$components/RightPane/CoursePane/SearchForm/DeptSearchBar/constants';
+import RightPaneStore from '$components/RightPane/RightPaneStore';
 import { getLocalStorageFavorites, setLocalStorageFavorites } from '$lib/localStorage';
```
jotalis: nit: can we refactor everywhere 'favorites' is mentioned to something more clear? e.g., `getLocalStorageRecentSearches` or `getLocalStorageSearchHistory`

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/DeptSearchBar/DeptSearchBar.tsx
```diff
@@ -1,175 +1,101 @@
...
+            return data;
+        }
 
-interface DeptSearchBarState {
-    value: Department;
-    favorites: Department[];
-}
+        if (data.every((x) => typeof x === 'object' && x !== null && 'deptValue' in x)) {
+            return data.map((x) => x.deptValue);
+        }
```
jotalis: nice 😎

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/DeptSearchBar/DeptSearchBar.tsx
```diff
@@ -1,175 +1,101 @@
...
-                    groupBy={(dept) => (dept.isFavorite ? 'Recent Departments' : 'Departments')}
-                    renderInput={(params) => (
-                        <TextField {...params} label="Department" type="search" InputLabelProps={{ shrink: true }} />
-                    )}
-                />
-            </div>
-        );
-    }
-}
+            if (newValue == null || newValue === 'ALL') return;
```
jotalis: nit: redundant to check for null
```suggestion
            if (newValue === 'ALL') return;
```

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/DeptSearchBar/DeptSearchBar.tsx
```diff
@@ -1,175 +1,101 @@
...
 
-    componentDidMount() {
-        RightPaneStore.on('formReset', this.resetField);
-    }
+            urlParam.delete('deptValue');
 
-    componentWillUnmount() {
-        RightPaneStore.removeListener('formReset', this.resetField);
-    }
+            if (newValue && newValue != 'ALL') {
```
jotalis: nit: redundant to check for truthyness
```suggestion
            if (newValue != 'ALL') {
```

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/DeptSearchBar/DeptSearchBar.tsx
```diff
@@ -1,175 +1,101 @@
...
+            favorites.includes(newValue)
+                ? setFavorites((prev) =>
+                      prev.sort((a, b) => {
+                          return a === newValue ? -1 : b === newValue ? 1 : 0;
+                      })
+                  )
+                : setFavorites((prev) => [newValue, ...prev].slice(0, 5));
+
+            setLocalStorageFavorites(JSON.stringify(favorites));
+        },
```
jotalis: issue: there's a race condition with the `favorites` state here

`setLocalStorageFavorites` is often called with the current favorites value before the asynchronous `setFavorites` state update completes, causing localStorage to save stale data instead of the newly updated favorites list.

consider moving `setLocalStorageFavorites(JSON.stringify(favorites));` outside and call it in a useEffect below e.g.,
``` tsx
    useEffect(() => {
        setLocalStorageFavorites(JSON.stringify(favorites));
    }, [favorites]);
```

jotalis (changes requested):
Overall solid PR and nice reimplementation of DeptSearchBar with a few nits and small comments

jotalis:
note: trying `deptValue=All` in the url leads to a TRPC error and infinite search screen, rather than erroring like `deptValue = {random_invalid_string}` does. 

not sure if this is expected behavior (it happens on prod too), but it would be better to catch it and throw a warning / cancel the search

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearch.tsx
KevinWu098: Ah, that's because the presence of deptValue doesn't imply that advanced settings have been set (which is what is checked/handled in lines 50-53)

KevinWu098: wait no, I'm trolling

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearch.tsx
KevinWu098: Good question, I imagine it was a legacy thing — I can't see anything breaking as a result of removing this on my local branch

KevinWu098: (removed)

apps/antalmanac/src/components/ScheduleManagement/ScheduleManagement.tsx
KevinWu098: fixed comment

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/DeptSearchBar/DeptSearchBar.tsx
KevinWu098: I've chosen to not rename the localStorage functions, but I have updated the component to use `recentSearches` over favorites. Main reason being that I'd like to keep our helper functions as close to the actual key they're modifying as possible.

Thoughts?

jotalis: you should just refactor the localStorage as well, since those functions and `favorites` are only being used for this purpose. for clarity, it would make sense to refactor the localStorage names to `recentDepartmentSearches` as opposed to my previous suggestions, since localStorage covers a broader scope

KevinWu098:
> note: trying `deptValue=All` in the url leads to a TRPC error and infinite search screen, rather than erroring like `deptValue = {random_invalid_string}` does.
> 
> not sure if this is expected behavior (it happens on prod too), but it would be better to catch it and throw a warning / cancel the search

Hmm, this is super bizarre — looking into it

KevinWu098:
Fix was found @jotalis — okay so, technically, "all" is a valid input to the API now. The request is so large, however, that it times out our lambda function so TRPC (which batches calls) will also fail our user loading (among other things).

The reason this was possible was because we were checking for the existence of a required parameter (e.g. `deptValue`) but not that the parameter was valid `RightPaneStore.formDataIsValid()`.

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/DepartmentSearchBar/DepartmentSearchBar.tsx
```diff
@@ -0,0 +1,103 @@
...
+    }, [recentSearches]);
+
+    return (
+        <Box sx={{ flexGrow: 1, width: '50%' }}>
+            <Autocomplete
+                value={value}
+                options={Array.from(new Set<string>([...recentSearches, ...options]))}
+                autoHighlight={true}
+                openOnFocus={true}
+                getOptionLabel={(option) => DEPARTMENT_MAP[option.toUpperCase() as keyof typeof DEPARTMENT_MAP]}
```
jotalis, line 89: question(non-blocking): why is `toUpperCase()` needed if we're populating autocomplete with DEPARTMENT_MAP options? even if they enter invalid text, it either forces the first available function or shows `noOptionsText`

apps/antalmanac/src/components/RightPane/RightPaneStore.ts
```diff
@@ -101,7 +97,9 @@ class RightPaneStore extends EventEmitter {
 
     formDataIsValid = () => {
         const { ge, deptValue, sectionCode, instructor } = this.formData;
-        return ge !== 'ANY' || deptValue !== 'ALL' || sectionCode !== '' || instructor !== '';
+        return (
+            ge.toUpperCase() !== 'ANY' || deptValue.toUpperCase() !== 'ALL' || sectionCode !== '' || instructor !== ''
+        );
```
jotalis, line 102: issue: you should perform the uppercase validation when we parse the URL

apps/antalmanac/src/components/ScheduleManagement/ScheduleManagement.tsx
```diff
@@ -51,7 +51,9 @@ export function ScheduleManagement() {
 
         const userId = getLocalStorageUserId();
 
-        if (userId === null) {
+        // Change the tab to the search tab if the user is not logged in
+        // or if a search parameter is in the URL
+        if (userId === null || paramsAreInURL()) {
```
jotalis: nit: make this comment read better or delete it.

jotalis (approved):

jotalis (approved):
l g t m  🐮

PR #1223, opened 2025-04-01
chore(cd): unify frontend prod and staging deployment

### LINKED ISSUE #1149: Unify staging and deploy frontend
## Description
- Right now, we deploy to prod with GitHub pages but deploy to staging with Cloudfront.
- There's little reason to maintain two versions of the stack like this, which incurs additional development costs.
    - E.g., prod deployments stopped working because the action's version was out of date.

jotalis:
## Summary
- Unifies frontend prod and staging deployment with CloudFront

## Issues

Closes #1149

### DIFF
--- a/apps/cdk/src/app/frontend-production.ts
+++ b/apps/cdk/src/app/frontend-production.ts
@@ -1,0 +1,16 @@
+import { App } from 'aws-cdk-lib';
+
+import { waitForStackIdle } from '../lib/wait-for-stack-idle';
+import { FrontendStack } from '../stacks/frontend';
+
+async function main() {
+    const stackName = 'antalmanac-frontend-production';
+
+    await waitForStackIdle(stackName);
+
+    const app = new App({ autoSynth: true });
+
+    new FrontendStack(app, stackName, { env: { region: 'us-east-1' } });
+}
+
+main();
--- a/apps/cdk/src/app/frontend-production.ts
+++ b/apps/cdk/src/app/frontend-production.ts
@@ -10,7 +10,7 @@ async function main() {
 
     const app = new App({ autoSynth: true });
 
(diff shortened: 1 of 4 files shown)


adcockdalton:
Is this PR still something you want to deploy?

apps/cdk/src/app/frontend-production.ts
```diff
@@ -0,0 +1,16 @@
...
+import { FrontendStack } from '../stacks/frontend';
+
+async function main() {
+    const stackName = 'antalmanac-frontend-production';
+
+    await waitForStackIdle(stackName);
+
+    const app = new App({ autoSynth: true });
+
+    new FrontendStack(app, stackName, { env: { region: 'us-east-1' } });
```
MinhxNguyen7: This should be US West 1, I think. Generally, staging is US East, and production is US West because the former has more features, but the latter is closer to users.

MinhxNguyen7 (changes requested):
Just one thing, and this should be ready to merge.

apps/cdk/src/app/frontend-production.ts
```diff
@@ -0,0 +1,16 @@
...
+import { FrontendStack } from '../stacks/frontend';
+
+async function main() {
+    const stackName = 'antalmanac-frontend-production';
+
+    await waitForStackIdle(stackName);
+
+    const app = new App({ autoSynth: true });
+
+    new FrontendStack(app, stackName, { env: { region: 'us-west-1' } });
```
MinhxNguyen7, line 13: This stack was written for staging. You need to change it to support prod deployment. Also, check with @alexanderl19 about how this would affect DNS resolution, since it's been pointing at GithubPages.

MinhxNguyen7 (changes requested):

KevinWu098:
Closing as superceded by #1237

PR #1247, opened 2025-05-17
fix: catch invalid username

### LINKED ISSUE #1246: Unhelpful Schedule Not Found Message
## Description
- Now, when a legacy username is not found, an internal server error is returned, and the frontend says "Failed to load schedules..."
- It needs to say specifically that the username is not found so that people who mistake their username know instead of thinking that their data is just not loading.
- Among others, this indexing assuming that results exist is a culprit.
Image

alexespejo:
## Summary
When fetching the account associated with the username returns null. Specify the error has to do with the username not existing.

image

## Test Plan
- Type in a username that doesn't exist -> `Couldn't find [...]" 
-
## Issues
Hold up haven't tested. I'm at 7 grams rn

Closes #1246

### DIFF
--- a/apps/antalmanac/src/actions/AppStoreActions.ts
+++ b/apps/antalmanac/src/actions/AppStoreActions.ts
@@ -280,17 +280,10 @@ export const loadSchedule = async (providerId: string, rememberMe: boolean, acco
                     providerId,
                 });
 
-                if (account == null) {
-                    openSnackbar('error', `Couldn't find account for username "${providerId}".`);
-                    return;
-                }
-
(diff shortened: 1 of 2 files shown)


MinhxNguyen7 (changes requested):
This doesn't work. As I said, the server is attempting to index an empty list of results and returning 500. You need to handle this server-side, return a `404`, then handle this on the frontend.

Also, if the code isn't ready for review, mark it as draft.

MinhxNguyen7 (approved):
LGTM!

PR #1249, opened 2025-05-18
Chore: Nicer looking mobile loading screen

alexespejo:
## Summary
On mobile the loading screen just looks a little weird 

Old:
image

Updated:
image

## Issues

Closes #


apps/antalmanac/src/components/LoadingScreen.tsx
```diff
@@ -11,6 +11,8 @@ const FUN_FACTS = [
     'Did you know you can search for classes by pressing "CTRL/CMD" + clicking on your schedule item!',
     'I 💖 Kevin Wu',
     'Need a 4 year plan? Checkout PeterPortal!',
+    'Congratulations class of 2025! 🎓',
+    'Thank you Minh Nguyen and Eddy Chen for your hard work and contributions to Antalmanac!',
```
KevinWu098: As much as we might enjoy these, let's not add people-based and time-based fun facts. mm, 2025 I might let slide, but then you better set yourself a reminder to remove it in a month's time...

apps/antalmanac/src/components/LoadingScreen.tsx
```diff
@@ -38,12 +40,26 @@ export function LoadingScreen(props: LoadingScreenProps) {
...
+                        display="flex"
+                        flexDirection="column"
+                        alignItems="center"
+                    >
+                        <LinearProgress sx={{ width: { default: '100%', md: '50%' } }} />
+                        <Box
+                            fontStyle="italic"
+                            color="white"
+                            fontSize="h6.fontSize"
+                            sx={{ textAlign: 'center', height: '50%' }}
```
KevinWu098: question: why does text need height?

apps/antalmanac/src/providers/Theme.tsx
KevinWu098, line 1: question: why are we editing the styles? What does a default that's at 0 do?

KevinWu098 (changes requested):

apps/antalmanac/src/components/LoadingScreen.tsx
KevinWu098: (this includes removing the one referencing myself)

alexespejo: 😔

apps/antalmanac/src/providers/Theme.tsx
alexespejo: When we use break points like `width: {{xxs: '100%'}}` any screen smaller than `xxs` won't be given styling. So default is to create a lower bound for smaller phones (`mobile` would be more descriptive)

apps/antalmanac/src/components/LoadingScreen.tsx
alexespejo: mb leftovers

KevinWu098 (approved):
LGTM

PR #1250, opened 2025-05-19
Advanced Search

### LINKED ISSUE #1186: Fix Advanced Search
## Description
- Some things in advanced search are broken.
- We're receiving many bug reports about advanced search being finnicky, but nothing's been reproducible.
- Someone needs to do an investigation to see what the problem is/might be.

vicksey:
## Summary
Temporary solution for advanced search placement. This is to help bring more users to use our manual search functionality.

## Issues

Closes #1186


MinhxNguyen7:
image
This but do quick search vs. manual search, where the latter has a dropdown for advanced options.
Remember to make sure that, if the user hides the advanced search options, that it's not passed into the search logic.

PR #1252, opened 2025-05-19
fix: Load schedule error propogation

alexespejo:
## Summary
Display an error to the user when a schedule fails to load

This is a hypothosized fix for empty schedules being loaded. Currently, when `fromScheduleSaveState` throws an error nothing happens and `loadSchedule` returns `true` 

## Issues

Closes #

### DIFF
--- a/apps/antalmanac/src/stores/AppStore.ts
+++ b/apps/antalmanac/src/stores/AppStore.ts
@@ -355,7 +355,8 @@ class AppStore extends EventEmitter {
     }
 
     async loadSchedule(savedSchedule: ScheduleSaveState) {
-        if (!(await this.loadScheduleFromSaveState(savedSchedule))) {
+        const loadSuccess = await this.loadScheduleFromSaveState(savedSchedule);
+        if (!loadSuccess) {
             return false;
         }
         this.unsavedChanges = false;
--- a/apps/antalmanac/src/stores/AppStore.ts
+++ b/apps/antalmanac/src/stores/AppStore.ts
@@ -348,13 +348,16 @@ class AppStore extends EventEmitter {
     private async loadScheduleFromSaveState(savedSchedule: ScheduleSaveState) {
         try {
             await this.schedule.fromScheduleSaveState(savedSchedule);
+            return true;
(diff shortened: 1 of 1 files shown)


alexespejo:
Is this an issue with auth? When I initially reported the error I was getting them with legacy schedules

KevinWu098:
This is a bit confusing @alexespejo. Isn't the problem that the schedules are being decoded improperly? Catching it shouldn't resolve the base problem of us not reading it correctly, no?

alexespejo:
> This is a bit confusing @alexespejo. Isn't the problem that the schedules are being decoded improperly? Catching it shouldn't resolve the base problem of us not reading it correctly, no?

"Fix" is the wrong word.. Since I couldn't repro it I figured the second best option is to inform the user something went wrong and they'll instinctively refresh the webpage rather than thinking their data got popped.

alexespejo:
I think I'm able to reproduce the error by messing with unsaved changes in prod but not locally. All I'm doing is making changes -> refresh the page with out saving -> cancel loading unsaved changes -> error from websoc

apps/antalmanac/src/stores/AppStore.ts
```diff
@@ -348,13 +348,16 @@ class AppStore extends EventEmitter {
...
             await this.schedule.fromScheduleSaveState(savedSchedule);
+            return true;
         } catch {
             return false;
         }
     }
 
     async loadSchedule(savedSchedule: ScheduleSaveState) {
-        await this.loadScheduleFromSaveState(savedSchedule);
+        if (!(await this.loadScheduleFromSaveState(savedSchedule))) {
```
KevinWu098: Maybe extract the result out to a value like result, then say if !result?

KevinWu098 (changes requested):
Update the title to clarify this is more of a load schedule error propagation fix, rather than the actual fix.

Also my one nit

KevinWu098:
Basically, I want the user to see the error, and we'll continue diagnosing the problem in another PR

KevinWu098 (approved):

PR #1253, opened 2025-05-19
feat: section code search v2

### LINKED ISSUE #1051: Course code in fuzzy search
## Description
- Functionality to search for a course code in fuzzy search.
- The easiest way to do that would be to see if the query is a five-digit number and say "search for section xxxxx"
- The better UX would be to show the section itself in the search suggestions.
    - This would necessitate changing the fuzzy search drop-down and re-implementing fuzzy search on a per-term basis (which would probably be a nicer experience) because section codes are resused.

jotalis:
## Summary
- Search for course section by its sectionCode
- Fetch and store 400,000 course codes by term using Anteater API's GraphQL endpoint
- Autofill suggestions if 4 digits are inputted
- Generates terms (quarters + years) on build
- Shards course code data into jsons by term
- Zips code for lambda for non local deployments

## Test Plan
- Run the backend locally
- Search for courses by their course code using 4-5 digits (including left padded ones e.g., 05001)
- Ensure correct course is displayed
- Ensure staging deployment actually works with new lambda bundle

Closes #1051


apps/backend/package.json
```diff
@@ -5,9 +5,12 @@
...
         "get-search-data": "tsx scripts/get-search-data.ts",
-        "prebuild": "pnpm get-search-data",
-        "build": "node scripts/build.mjs",
-        "prestart": "pnpm get-search-data",
+        "fetch-term-data": "tsx scripts/fetch-term-data.ts",
+        "copy-terms-to-dist": "tsx scripts/copy-terms-to-dist.ts",
+        "zip-dist": "tsx scripts/zip-dist.ts",
+        "prebuild": "pnpm fetch-term-data && pnpm get-search-data && pnpm copy-terms-to-dist",
+        "build": "node scripts/build.mjs && pnpm zip-dist",
+        "prestart": "pnpm fetch-term-data && pnpm get-search-data",
```
KevinWu098: question/suggestion: is it possible to merge all of these scripts into one prestart script? Something like `node scripts/prestart.mjs`. Similarly, can `zip-dist` be consolidated into `scripts/build`?

apps/backend/scripts/fetch-term-data.ts
KevinWu098, line 1: question: If I'm not mistaken... this is the same script as the one in `apps/antalmanac`? Have you given thought to how we can these in sync? I'm not particularly against duplicating code... but this is duping ~100 lines of the same script into two directories

apps/backend/scripts/get-search-data.ts
```diff
@@ -1,62 +1,110 @@
...
+                }
+            }
+        }
+    }`;
+    await mkdir(join(__dirname, '../src/generated/'), { recursive: true });
+    await mkdir(join(__dirname, '../src/generated/terms/'), { recursive: true });
+    await writeFile(
+        join(__dirname, '../src/generated/searchData.ts'),
+        `
+    import type { CourseSearchResult, DepartmentSearchResult, SectionSearchResult } from "@packages/antalmanac-types";
```
KevinWu098: thought: I think we had this discussion long long ago, but isn't it lowkey kinda weird that we write TS files...? I swear I was on the side of the TS files but now I can't think of a justification? This isn't blocking to this PR, but uhh do you know why we're writing TS and not some JSON we parse later?

apps/backend/scripts/get-search-data.ts
```diff
@@ -1,62 +1,110 @@
...
+    `
+    );
+    let count = 0;
+    for (const term of termData) {
+        const [year, quarter] = term.shortName.split(' ');
+        const parsedTerm = `${quarter}_${year}`;
+        const query = QUERY_TEMPLATE.replace('$$YEAR$$', year).replace('$$QUARTER$$', quarter);
+        const res = await queryGraphQL<SectionCodesGraphQLResponse>(query);
+        if (!res) {
+            throw new Error('Error fetching section codes.');
```
KevinWu098: suggestion: maybe also pass in which term it failed on?

apps/backend/scripts/get-search-data.ts
```diff
@@ -1,62 +1,110 @@
...
-    await mkdir(join(__dirname, "../src/generated/"), { recursive: true });
-    await writeFile(join(__dirname, "../src/generated/searchData.ts"), `
-    import type { CourseSearchResult, DepartmentSearchResult } from "@packages/antalmanac-types";
-    export const departments: Array<DepartmentSearchResult & { id: string }> = ${JSON.stringify(Array.from(deptMap.values()))};
-    export const courses: Array<CourseSearchResult & { id: string }> = ${JSON.stringify(Array.from(courseMap.values()))};
-    `)
-    console.log("Cache generated.");
+
+    const QUERY_TEMPLATE = `{
+        websoc(query: {year: "$$YEAR$$", quarter: $$QUARTER$$}) {
```
KevinWu098, line 57: question: is `$$FOO$$` standard syntax?

apps/backend/src/lib/helpers.ts
```diff
@@ -0,0 +1,20 @@
...
+        headers: {
+            'Content-Type': 'application/json',
+            Accept: 'application/json',
+        },
+        body: query,
+    });
+
+    const json = await res.json();
+
+    if (!res.ok || json.data === null) return null;
```
KevinWu098: nit: brace your conditionals
```suggestion
    if (!res.ok || json.data === null) {
    	return null;
    }
```

apps/backend/src/lib/helpers.ts
```diff
@@ -0,0 +1,20 @@
...
+
+    return json as Promise<PromiseReturnType>;
```
KevinWu098, line 21: suggestion: is it possible to use something like Zod here to validate the type so we don't have to assert? This isn't blocking, but just curious if you've considered using Zod / would want to

apps/backend/src/lib/term-section-codes.ts
```diff
@@ -0,0 +1,69 @@
+import { SectionSearchResult } from '@packages/antalmanac-types';
+import { terms } from '../generated/termData';
```
KevinWu098: nit: can we do absolute import?

apps/backend/src/lib/term-section-codes.ts
```diff
@@ -0,0 +1,69 @@
...
+
+interface SectionData {
+    sectionCode: string;
+    sectionType: string;
+    sectionNum: string;
+}
```
KevinWu098: question: do these types not exist via the AAPI type package? why are we defining our own?

apps/backend/src/routers/search.ts
```diff
@@ -31,24 +39,64 @@ const toMutable = <T>(arr: readonly T[]): T[] => arr as T[];
...
 const searchRouter = router({
     doSearch: procedure
-        .input(z.object({ query: z.string() }))
+        .input(z.object({ query: z.string(), term: z.string() }))
         .query(async ({ input }): Promise<Record<string, SearchResult>> => {
             const { query } = input;
+            const [year, quarter] = input.term.split(' ');
+            const parsedTerm = `${quarter}_${year}`;
+
+            let termData: Record<string, SectionSearchResult>;
```
KevinWu098: suggestion: I understand `termData` as used in `term-section-codes` is an array of terms which is returned by the AAPI. Any thoughts on renaming the usage here so we don't shadow the name?

apps/backend/scripts/get-search-data.ts
```diff
@@ -1,62 +1,110 @@
...
+    export const departments: Array<DepartmentSearchResult & { id: string }> = ${JSON.stringify(
+        Array.from(deptMap.values())
+    )};
+    export const courses: Array<CourseSearchResult & { id: string }> = ${JSON.stringify(
+        Array.from(courseMap.values())
+    )};
+    `
+    );
+    let count = 0;
+    for (const term of termData) {
```
KevinWu098: question: running locally, it's about 60 seconds to fetch all the data and build the cache. Is it possible to parallelize the queries? Not blocking, but wondering if this is something you'd like to tackle here, if possible.

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/FuzzySearch.tsx
```diff
@@ -76,6 +77,10 @@ class FuzzySearch extends PureComponent<FuzzySearchProps, FuzzySearchState> {
                 RightPaneStore.updateFormValue('courseNumber', ident[0].split(' ').slice(-1)[0]);
                 break;
             }
+            case emojiMap.SECTION: {
+                RightPaneStore.updateFormValue('sectionCode', ident[0].split(' ').slice(0)[0]);
```
KevinWu098: issue: I was curious about this line so I went and logged out `ident` for my search of `30000`: 
```
[
    "30000 lec a",
    " med hum 1"
]
```

The first index -> `30000 lec a`
Split by space -> [30000, lec, a]
Slice on `0` (which is a no-op) -> [30000, lec, a]
First index -> `30000`.

TL;DR seems like `.slice(0)` is redundant

apps/backend/src/routers/search.ts
```diff
@@ -31,24 +39,64 @@ const toMutable = <T>(arr: readonly T[]): T[] => arr as T[];
...
-                keys: ['id', 'name', 'alias', 'metadata.department', 'metadata.number'],
-                limit: 10 - matchedDepts.length
-            })
-            return Object.fromEntries(
-                [...matchedDepts.map(x => [x.obj.id, x.obj]),
-                ...matchedCourses.map(x => [x.obj.id, x.obj]),]
-            );
+
+            const matchedDepts =
+                matchedSections.length === 10
```
KevinWu098: suggestion: I think why we limit to 10 results predates me, but take for instance a course like `EURO ST 10: Topics in Historical Foundations (1500-1800)`. If someone (for some wild reason) searched 1500, they'd only see 1500X section results.

Recognizing that, yes, this is a very rare edge case, I'm thinking it probably can't hurt to show the top 12 results rather than 10?

KevinWu098: suggestion: oh also, while we're here, maybe extract `10` into a const `MAX_AUTOCOMPLETE_RESULTS` or smth so we don't end up with discrepancies from human error

KevinWu098 (changes requested):
Great PR. Tested locally and on the deployment. A joy to review (mostly questions and thoughts about certain aspects, virtually no issues). Most certainly will be a great surprise and improvement for users searching. 

🔥

apps/backend/scripts/fetch-term-data.ts
jotalis: the backend and frontend are deployed separately, so they can't really be in sync unless you make the backend the one 'source of truth'. e.g., in some earlier discussions for the pregenerate terms PR, I considered having this script only run in the backend as one source of truth and setting up a tRPC route for the frontend to hit to fetch the terms, but this meant unnecessary calls to the backend and potential latency when rendering terms on the frontend.

since they are deployed separately and they both need termData, i think it's okay that the scripts are nearly identical...

apps/backend/package.json
jotalis: i think it is valid to keep the scripts separated, but i can some batch together as a new script alias in package.json

e.g.,
`fetch-data": "tsx scripts/fetch-term-data.ts && tsx scripts/get-search-data.ts`
then
`
        "prebuild": "pnpm fetch-data && pnpm copy-terms-to-dist",
        "prestart": "pnpm fetch-data"
`

KevinWu098: Sounds reasonable!

apps/backend/scripts/fetch-term-data.ts
KevinWu098: Would you mind adding a comment block to both of them for posterity? Something like:

NB: blah blah the same script exists in both `apps/antalmanac` and `apps/backend` for the purpose of fetching term data... blah blah if you're making changes in one, you most likely should make the same change to backend

apps/backend/scripts/get-search-data.ts
jotalis: this was from Eddy's original implementation, but I believe just for that extra type checking and validation that the searchData fetched from AAPI is still in the format we think it is

KevinWu098: hrm... i feel like we should really be doing this validation via zod but I'd rather not have this PR tackle that since that feels like more of a dedicated Zod-ification PR for all our scripts

apps/backend/scripts/get-search-data.ts
jotalis: if you're asking if its a common placeholder syntax, then yeah

apps/backend/src/lib/term-section-codes.ts
jotalis: i'm requesting and storing only specific information using graphQL; i believe their type encapsulates more

apps/backend/src/routers/search.ts
jotalis: yeah makes sense

apps/backend/src/lib/term-section-codes.ts
KevinWu098: err, I think you could do something like `Pick<AAPIType, 'sectionCode' | 'sectionType' | 'sectionNum'`?

KevinWu098: https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys

apps/backend/src/routers/search.ts
jotalis: yeah this is a good point

apps/backend/src/lib/helpers.ts
jotalis: now that i use typing directly from AAPI, it should be fine? since this is a general function, to use zod i'd probably need to pass in manually pre-defined zod schemas for every endpoint i choose to hit, which is a bit annoying

apps/backend/scripts/get-search-data.ts
jotalis: i don't seem to be rate limited when i do parallelize so 🚀

KevinWu098 (approved):
LGTM! Great PR!

KevinWu098 (approved):

PR #1254, opened 2025-05-19
ci: log websoc data in prod

KevinWu098:
## Issues

Closes #


MinhxNguyen7 (approved):
LGTM

MinhxNguyen7:
I'm pretty sure that the test failing is an issue with turbo caching, which we have observed before. I don't know how to clear it, but this PR is so simple, I'm going to let that slide.

KevinWu098:
@MinhxNguyen7 Let's revert this one too?

PR #1255, opened 2025-05-19
Redeploy every 15 minutes

MinhxNguyen7:
## Summary
- Deploy every 15 minutes to alleviate an unknown issue that is fixed by redeployment.
- This is a temporary fix so the page stays up during enrollment.


MinhxNguyen7:
We'll merge this if the page breaks again

Issue #1260, opened 2025-05-19
Missing API keys should throw errors


KevinWu098:
Something in how we're parsing API keys with Zod AND/OR something in our deploy pipeline is not correctly throwing errors when our API keys are missing.

CC @MinhxNguyen7 for any additional context

MinhxNguyen7:
Closed by #1261

PR #1261, opened 2025-05-19
Add ANTEATER_API_KEY to backend vars parsing

MinhxNguyen7:
## Summary
- ANTEATER_API_KEY was omitted from the Zod environment variables parsing.
    - I think the problem was that AAPI key change happened before the start of auth, and its inclusion was lost during merging.
    - Perhaps I should've caught that during review, but I glossed over it because I was the one that wrote it.
- This adds it back, along with reformatting the Zod declaration to be clearer.

## Test Plan
- See that the environment variable shows up in the staging lambda.
- Deploy to prod and see the same in prod.

Closes (partially) #1260

## Future Follow-Up
- Right now, the backend parses environment variables when it starts, but then the variables are still pulled from `process.env`. That bypasses type-checking, which partly explains why we didn't see this omission earlier.

### DIFF
--- a/apps/backend/src/env.ts
+++ b/apps/backend/src/env.ts
@@ -24,10 +24,16 @@ export const mapboxEnvSchema = z.object({
 });
 
 /**
+ * Environment variables required by the backend to connect to the Anteater API.
+ */
+export const aapiEnvSchema = z.object({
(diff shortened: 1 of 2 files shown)


KevinWu098 (approved):
Tested locally, works

PR #1265, opened 2025-05-20
fix: spontaneous user logout on refresh

### LINKED ISSUE #1262: Reloading in quick succession signs out the user
If you reload while AA is loading in your user, you'll become signed out. This shouldn't happen.

alexespejo:
## Summary
It seems like when you spam Refresh the `loadScheduleWithSessionToken` catches an error that removes the session token from local storage

## Reproduce on prod 
1. Login with Google account with some data
2. Quickly spam refresh on the webpage before it can load 
3. After several refreshes the page your account should be logged out and you'll have to log back in 
4. (Additional note): Inspect Page -> in `Console` type: `localStorage` and the `sessionID` should be missing 

## Test Plan
1. Follow the same steps to reproduce the error but your account should remained logged in

## Issues

Closes #1262

### DIFF
--- a/apps/antalmanac/src/actions/AppStoreActions.ts
+++ b/apps/antalmanac/src/actions/AppStoreActions.ts
@@ -13,7 +13,7 @@ import { SnackbarOrigin, VariantType } from 'notistack';
 import analyticsEnum, { logAnalytics, courseNumAsDecimal } from '$lib/analytics/analytics';
 import trpc from '$lib/api/trpc';
 import { warnMultipleTerms } from '$lib/helpers';
-import { setLocalStorageUserId, setLocalStorageDataCache, removeLocalStorageSessionId } from '$lib/localStorage';
+import { setLocalStorageUserId, setLocalStorageDataCache } from '$lib/localStorage';
 import AppStore from '$stores/AppStore';
 import { scheduleComponentsToggleStore } from '$stores/ScheduleComponentsToggleStore';
 import { useSessionStore } from '$stores/SessionStore';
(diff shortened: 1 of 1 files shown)


MinhxNguyen7:
Can you update the test plan with the (sufficiently detailed)  reproduction (on the current version) so that a reviewer can verify that this fix works?

KevinWu098:
Staging fails @alexespejo 
Screenshot 2025-05-21 at 12 40 39 PM

alexespejo:
> Staging fails @alexespejo Screenshot 2025-05-21 at 12 40 39 PM

It's using the dev backend which is broken right?

KevinWu098:
e> > Staging fails @alexespejo Screenshot 2025-05-21 at 12 40 39 PM
> 
> It's using the dev backend which is broken right?

I think I'm OOTL — why is dev broken? )=

edit: and is someone fixing that

alexespejo:
> I think I'm OOTL — why is dev broken? )=


@MinhxNguyen7

MinhxNguyen7:
Wdym by "dev"?

MinhxNguyen7:
The development mirror is broken right now. ~~It's just a missing env var, if you want to fix it, @KevinWu098.~~ The issue is actually non-trivial

KevinWu098:
Sorry haven't gotten around to fixing the dev env but uh... this fix raises more questions: why is there an error from reloading the page? isn't that the more interesting (and seemingly "root") problem?

alexespejo:
> Sorry haven't gotten around to fixing the dev env but uh ... this fix raises more questions: why is there an error from reloading the page? isn't that the more interesting (and seemingly "root") problem?

It seems like if you refresh before the call to `userData.getUserDataWithSessionToken` completes the error gets caught. I'm not sure if that's how that works though. I'm not able to reproduce it locally anymore since the request completes a lot faster. 

https://github.com/user-attachments/assets/577ebde3-c33c-4aac-aaa7-c554e18046e4

KevinWu098 (approved):

PR #1266, opened 2025-05-20
fix: dotenv config

KevinWu098:
## Issues

Closes #

### DIFF
--- a/apps/backend/scripts/fetch-term-data.ts
+++ b/apps/backend/scripts/fetch-term-data.ts
@@ -13,7 +13,7 @@ import { backendEnvSchema } from 'src/env';
 
 import 'dotenv/config';
 
-const env = backendEnvSchema.parse(process.env);
+const env = aapiEnvSchema.parse(process.env);
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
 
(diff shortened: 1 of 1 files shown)


MinhxNguyen7:
I was going to say, that should not have passed review

MinhxNguyen7 (changes requested):
Can you also change it so that you're importing the AAPI env schema and using it to parse `process.env`?

MinhxNguyen7 (approved):
LGTM!

apps/backend/scripts/fetch-term-data.ts
```diff
@@ -9,8 +9,11 @@ import { dirname, join } from 'node:path';
 import { fileURLToPath } from 'node:url';
 
 import { CalendarTerm } from '@packages/antalmanac-types';
+import { backendEnvSchema } from 'src/env';
 
-const apiKey = process.env.ANTEATER_API_KEY;
+import 'dotenv/config';
+
+const env = backendEnvSchema.parse(process.env);
```
MinhxNguyen7:
```suggestion
const env = aapiEnvSchema.parse(process.env);
```

This is the only one that you need, and it should allow the workflow to run properly

MinhxNguyen7 (changes requested):
Wait

MinhxNguyen7 (approved):

PR #1267, opened 2025-05-20
Refactor detecting new user login

### LINKED ISSUE #1240: New User Handling
This can arguably wait, but this component is way too big and should be broken up into multiple.

_Originally posted by @MinhxNguyen7 in https://github.com/icssc/AntAlmanac/pull/1203#discussion_r2091950228_

alexespejo:
## Summary
The implementation for detecting a new AntAlmanac user was pretty messy for auth v1 and was not changed for v2.

I'm not sure if this PR is necessary as all users likely have already created their accounts. However, auto prompting new users with this dialog could open an opportunity for #1186. Since users can import a ZotCourse ID we could use that to display manual search.  

image

## Issues

Closes #1240

### DIFF
--- a/apps/antalmanac/src/components/Header/Import.tsx
+++ b/apps/antalmanac/src/components/Header/Import.tsx
@@ -18,7 +18,7 @@ import {
     Tooltip,
 } from '@mui/material';
 import { CourseInfo } from '@packages/antalmanac-types';
-import { ChangeEvent, useCallback, useEffect, useState, useRef } from 'react';
+import { ChangeEvent, useCallback, useEffect, useState } from 'react';
 import { Link } from 'react-router-dom';
 
 import {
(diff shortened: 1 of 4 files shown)


MinhxNguyen7:
Can you give a description of how it used to work, along with how it will work with the PR?

alexespejo:
> Can you give a description of how it used to work, along with how it will work with the PR?

The user flow is the same, under the hood we were fetching the user and determining if they have an `NULL` schedule ID and storing their email in localStorage to validate they've created an account. There was also a random `useRef` that served no purpose.

In this PR, we just use the same local storage to flag the client when we sign the user up on the loading page

MinhxNguyen7 (approved):
LGTM!

MinhxNguyen7 (approved):
Seems alright

PR #1273, opened 2025-05-22
Allow dev mirror deployment (temp fix)

### LINKED ISSUE #1272: Fix development mirror deployment
### Solution 
- Temporarily always deploy the staging backend (#1273).
  - This breaks local auth.
- Permanent fix: have the *dev* backend create new oAuth clients that redirect to the URLs requested by the frontend.
  - A singleton (e.g., `GoogleOAuth2ClientsManager`) lazily creates and returns clients, allowing for the wildcard if `STAGE=='dev'`. 
  - IMPORTANT: This must be validated to make sure that the redirect follows `staging-{PR}.antalmanac.com/auth` so as not to be a security vulnerability. Be very careful.
  - The client should only be allowed to send a 4-digit number that we slot into the PR number. 

### Discussion
- The development mirror has not been working since the auth merge.
- The error is missing env vars related to Google Auth. However, the problem is that staging instances need to be redirected to different places, which is currently not possible.
- Therefore, the two alternatives are:
  1) Give every staging instance its own backend. As far as I see, this shouldn't incur noticeable compute costs. However, this means that each PR will need to be added to the Google Auth console.
  2) Have an auth micro-service shared between prod and staging.
- I'm inclined to go with the first because it's simpler. 

CC: @KevinWu098 and @alexanderl19

MinhxNguyen7:
## Summary
- The dev mirror has been broken because it was missing the Google Auth credentials. 
- This is a non-trivial problem because the redirect needs to change depending where the frontend is running.
- The fix for now is to deploy with the production redirect.
  - Auth won't work when running `pnpm run start:aa`.
- This PR also makes the backend deploy for every PR, ensuring that staging works.
  - As long as the staging URL is added to the GCP Console.

## Issues

Part of  #1272 

## Future Followup
- Make the dev mirror generate new oAuth clients for staging instances.

### DIFF
--- a/.github/workflows/deploy_production.yml
+++ b/.github/workflows/deploy_production.yml
@@ -160,6 +160,8 @@ jobs:
               run: pnpm --filter "antalmanac-cdk" backend-development deploy
               env:
                   NODE_ENV: development
+                  DB_URL: ${{ secrets.PROD_DB_URL }}
+                  
                   GOOGLE_CLIENT_ID: ${{ secrets.GOOGLE_CLIENT_ID }}
                   GOOGLE_CLIENT_SECRET: ${{ secrets.GOOGLE_CLIENT_SECRET }}
                   GOOGLE_REDIRECT_URI: https://antalmanac.com/auth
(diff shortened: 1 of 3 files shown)


alexespejo:
I know we only get a small handful of contributors but should we update the docs to specify that pnpm start:aa doesn't work?

MinhxNguyen7:
@alexespejo I've added a note to say that local Google sign-in isn't working

alexespejo (approved):
Lgtm

PR #1275, opened 2025-05-23
feat: redesign search

### LINKED ISSUE #791: RFC: Reorganize search page
### Thoughts
- The search is not the obvious thing your eye goes to when you open the page.
- The top bar buttons, e..g, "SAVE," "LOAD", "THEME", etc. look like they're the most important. They're big, bold, and capitalized. In contrast, the most important thing, the search is plain and small.
- There's no obvious reason why the term selector needs a "Term" label. No one would think "2024 Winter Quarter" is anything else.

### Suggestions
- Remove the "Term" label for the selector. 
- Put the term selector with the manual search button below the search bar.
- Remove the "Search" label above the search bar. The tab is called "SEARCH" anyway. 
    - Have a colored icon or search button to draw the user's eyes to it.
- Make the search bar BIG or give it an outline. Maybe color it as well somehow.

### LINKED ISSUE #1184: Two panels for search
## Description
Instead of the original idea of just making advanced search into a drop-down on the same screen as fuzzy search, we're going to have two separate options, "Quick Search" and "Manual Search", where the latter will have manual search available under a separator.

### Mock-Up
It should look something like below, but with only two options.

Image

xgraceyan:
## Summary
Redesigns search page, and a few other fixes
- Adds custom components for textfield and select inputs
- Refactors "LegacySearch" into "ManualSearch"
- Manual Search no longer accepts Advanced Search parameters when collapsed (There should maybe be some way to inform the user about this)
- Advanced search and manual search params in ...SearchForm/constants.ts for easier typing

image
image

## Test Plan
- Test that each field works (produces correct search results) in manual search
- Test that quick search works normally
- Fill in only advanced search parameters (minimum instructor), then collapse them. Search should not be able to work. Expand advanced search. Ensure the search should now complete
- Fill in parameters in both manual and advanced search. Ensure results that match advanced search parameters should only be displayed when advanced search is expanded
- Ensure there are no UI bugs: all input boxes fit and are easy to fill in on both mobile and desktop view

## Issues
- Closes #1184
- Closes #791

## Future Follow-Up
- #1290


apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearchTextFields.tsx
```diff
@@ -60,7 +54,7 @@ export function AdvancedSearchTextFields() {
             const urlParam = new URLSearchParams(url.search);
```
MinhxNguyen7, line 55: I know you didn't write this function, but, while you're here, could you refactor it a bit?

1) This function is not a change handler. It's a change handler factory. (I.e., it's a function that makes change handlers.) Therefore, it should be called `changeHandlerFactory`.
2) The default case in the switch statement is a silent failure. It should at least `console.error`.

CC: @KevinWu098 because the git blame says you wrote it.

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/ManualSearch/ManualSearchInputAdornment.tsx
```diff
@@ -0,0 +1,27 @@
+import { Box, InputAdornment } from '@mui/material';
+import { grey } from '@mui/material/colors';
+
+import { useThemeStore } from '$stores/SettingsStore';
+
+interface ManualSearchInputAdornmentProps {
+    label: string;
+}
+
+export const ManualSearchInputAdornment = ({ label }: ManualSearchInputAdornmentProps) => {
```
MinhxNguyen7: How charming a name.

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchForm.tsx
```diff
@@ -37,22 +36,27 @@ export const SearchForm = ({ toggleSearch }: SearchFormProps) => {
...
-                                <IconButton onClick={toggleManualSearch}>
-                                    <Tune />
-                                </IconButton>
-                            </Tooltip>
-                        </Box>
                     </Box>
 
                     {!manualSearchEnabled ? (
-                        <FuzzySearch toggleSearch={toggleSearch} toggleShowLegacySearch={toggleManualSearch} />
+                        <FuzzySearch toggleSearch={toggleSearch} toggleShowManualSearch={toggleManualSearch} />
```
MinhxNguyen7: There's a type error here.

MinhxNguyen7 (changes requested):
Looks really nice. Once this goes out, I think you can be proud to to have made the search experience quite a bit more pleasant.

Other than the minor code comments:
1) The fact that, in manual search, all the fields are in a flex box means that they sometimes line-break unintuitively. I strongly think that you should group fields together such that they stay on the same line (and resize as necessary) until the screen gets too small, then they break into one-per-line. Otherwise, our fields' positions vary too much depending on the device width.
2) In dark mode, when a search mode is selected, the text is too close to the highlight background color.  Either the background should be more transparent, or the text needs to be a lighter shade.
3) In dark mode, outlines of the boxes for the search fields are too bold (white/blue) on hover/select. It's not blocking, but it feels slightly off, IMO.

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/constants.ts
```diff
@@ -0,0 +1,25 @@
...
+    'units',
+    'endTime',
+    'startTime',
+    'coursesFull',
+    'building',
+    'room',
+    'division',
+    'excludeRestrictionCodes',
+    'days',
+];
```
KevinWu098: issue: this should be declared `as const` so TS will infer `(typeof ADVANCED_SEARCH_PARAMS)[number]` as a string union and not as strings

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/constants.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export type AdvancedSearchParam = (typeof ADVANCED_SEARCH_PARAMS)[number];
+
+export const MANUAL_SEARCH_PARAMS = [
```
KevinWu098, line 16: repeat: this should be declared `as const`

apps/antalmanac/src/stores/CoursePaneStore.ts
```diff
@@ -1,5 +1,6 @@
 import { create } from 'zustand';
 
+import { MANUAL_SEARCH_PARAMS } from '$components/RightPane/CoursePane/SearchForm/constants';
 import RightPaneStore from '$components/RightPane/RightPaneStore';
 interface CoursePaneStore {
```
KevinWu098, line 6: nit: can we add a newline after the imports

apps/antalmanac/src/components/RightPane/RightPaneStore.ts
```diff
@@ -80,17 +87,40 @@ class RightPaneStore extends EventEmitter {
         this.emit('formDataChange');
     };
 
+    storePrevFormData = () => {
+        this.prevFormData = structuredClone(this.formData);
+        this.emit('formDataChange');
+    };
+
+    restorePrevFormData = () => {
+        if (!this.prevFormData) return;
```
KevinWu098: nit: always brace your conditionals
```suggestion
        if (!this.prevFormData) {
            return;
        }
```

apps/antalmanac/src/components/RightPane/CoursePane/CoursePaneRoot.tsx
```diff
@@ -17,6 +17,13 @@ export function CoursePaneRoot() {
     const { key, forceUpdate, searchFormIsDisplayed, displaySearch, displaySections } = useCoursePaneStore();
 
     const handleSearch = useCallback(() => {
+        const advancedSearchEnabled = useCoursePaneStore.getState().advancedSearchEnabled;
```
KevinWu098: question: why aren't we using the store as a hook? 

e.g. const advancedSearchEnabled = useCoursePaneStore(shallow((store) => store.advancedSearchEnabled))

(the syntax might be slightly off)

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearchTextFields.tsx
```diff
@@ -1,16 +1,9 @@
...
-    Select,
-    Switch,
-    FormControlLabel,
-    type SelectChangeEvent,
-} from '@mui/material';
+import { MenuItem, Box, type SelectChangeEvent } from '@mui/material';
 import { useState, useEffect, useCallback, type ChangeEvent } from 'react';
 
+import { ManualSearchSelect } from '../ManualSearch/ManualSearchSelect';
+import { ManualSearchTextField } from '../ManualSearch/ManualSearchTextField';
```
KevinWu098: nit: prefer absolute imports over relative

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearchTextFields.tsx
```diff
@@ -115,6 +109,7 @@ export function AdvancedSearchTextFields() {
                     break;
                 }
                 default:
+                    console.error(`Could not find advanced search field with name ${name}`);
                     break;
```
KevinWu098: issue: ultimately this is still a silent failure

suggestion: add a utility to lib/utils that encodes this in logic

```ts
export const safeUnreachableCase = <T>(
  v: never,
  retVal?: T,
): T => {
  // if this code is running, v didn't turn out to be `never` after all, so tell TS that
  const castedV = v as unknown

  console.error(`Reached a (safe) unreachable case: ${v}`)
  return retVal
}
```

This utility will ensure, when used in the `default` of a switch statement, that the switch value has been exhausted (all potential types were handled)

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SectionCodeSearchBar.tsx
```diff
@@ -1,8 +1,9 @@
-import { TextField } from '@mui/material';
 import { ChangeEvent, PureComponent } from 'react';
 
 import RightPaneStore from '../../RightPaneStore';
 
+import { ManualSearchTextField } from './ManualSearch/ManualSearchTextField';
```
KevinWu098: nit: absolute imports please

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchForm.tsx
```diff
@@ -26,6 +28,12 @@ export const SearchForm = ({ toggleSearch }: SearchFormProps) => {
         [toggleSearch]
     );
 
+    const toggleSearchMode = (event: React.MouseEvent<HTMLElement>, value: string) => {
+        event.preventDefault();
+        if (!value) return;
+        toggleManualSearch();
+    };
```
KevinWu098, line 37: issue: clicking the currently selected toggle button shouldn't switch to the other option

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchForm.tsx
```diff
@@ -37,22 +45,32 @@ export const SearchForm = ({ toggleSearch }: SearchFormProps) => {
...
+                        color="primary"
+                        value={manualSearchEnabled ? 'manual' : 'quick'}
+                        exclusive
+                        aria-label="Search selection"
+                        sx={{
+                            paddingTop: 1,
+                            '& .MuiToggleButton-root.Mui-selected': {
+                                backgroundColor: isDark ? alpha(DODGER_BLUE, 0.05) : undefined,
+                            },
+                        }}
```
KevinWu098, line 62: suggestion: why not use `grey[800]` / `grey[200]` like you did with the adornment?

apps/antalmanac/src/components/RightPane/RightPaneStore.ts
```diff
@@ -1,13 +1,10 @@
 import { EventEmitter } from 'events';
 
+import { AdvancedSearchParam, ManualSearchParam } from './CoursePane/SearchForm/constants';
```
KevinWu098: nit: absolute import

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/GeSelector.tsx
```diff
@@ -1,6 +1,8 @@
-import { FormControl, InputLabel, MenuItem, Select, type SelectChangeEvent } from '@mui/material';
+import { MenuItem, type SelectChangeEvent } from '@mui/material';
 import { useEffect, useCallback, useState } from 'react';
 
+import { ManualSearchSelect } from './ManualSearch/ManualSearchSelect';
```
KevinWu098: nit: absolute import

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/FuzzySearch.tsx
```diff
@@ -1,10 +1,12 @@
-import { Autocomplete, type AutocompleteInputChangeReason, TextField } from '@mui/material';
+import { Autocomplete, type AutocompleteInputChangeReason } from '@mui/material';
 import type { SearchResult } from '@packages/antalmanac-types';
 import { PureComponent } from 'react';
 import UAParser from 'ua-parser-js';
 
 import RightPaneStore from '../../RightPaneStore';
 
+import { ManualSearchTextField } from './ManualSearch/ManualSearchTextField';
```
KevinWu098: nit: absolute import

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/CourseNumberSearchBar.tsx
```diff
@@ -1,6 +1,7 @@
-import { TextField } from '@mui/material';
 import { ChangeEvent, useCallback, useEffect, useState } from 'react';
 
+import { ManualSearchTextField } from './ManualSearch/ManualSearchTextField';
```
KevinWu098: nit: absolute import

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/ManualSearch/ManualSearchInputAdornment.tsx
```diff
@@ -0,0 +1,27 @@
...
+    const isDark = useThemeStore((store) => store.isDark);
+    return (
+        <InputAdornment position="start" sx={{ marginRight: 0 }}>
```
KevinWu098: suggestion: it may be nice visually for our adornments to share some (optional) minimum width for the manual search, similar to Zotcourse so the search looks cleaner

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchForm.tsx
```diff
@@ -37,22 +45,32 @@ export const SearchForm = ({ toggleSearch }: SearchFormProps) => {
                 }}
             >
                 <Stack spacing={2}>
-                    <Box sx={{ display: 'flex', paddingTop: 1, alignItems: 'center', gap: 1 }}>
+                    <ToggleButtonGroup
```
KevinWu098, line 50: issue: I'm noticing that when manual select is selected, the border between it and quick search disappears

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/DepartmentSearchBar/DepartmentSearchBar.tsx
```diff
@@ -80,7 +82,7 @@ export function DepartmentSearchBar() {
     }, [recentSearches]);
 
     return (
-        <Box sx={{ flexGrow: 1, width: '50%' }}>
+        <Box sx={{ flexGrow: 1, width: '50%', padding: 0 }}>
```
KevinWu098: question: what is this padding style supposed to be overriding?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/DepartmentSearchBar/DepartmentSearchBar.tsx
```diff
@@ -1,6 +1,8 @@
-import { Autocomplete, Box, TextField } from '@mui/material';
+import { Autocomplete, Box } from '@mui/material';
 import { useCallback, useEffect, useState } from 'react';
 
+import { ManualSearchTextField } from '../ManualSearch/ManualSearchTextField';
```
KevinWu098: nit: absolute import

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/ManualSearch/ManualSearchTextField.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+import { ManualSearchInputAdornment } from './ManualSearchInputAdornment';
+
+import { BLUE } from '$src/globals';
+import { useThemeStore } from '$stores/SettingsStore';
+
+interface ManualSearchTextFieldProps {
+    label: string;
+    textFieldProps?: TextFieldProps;
+    autocompleteProps?: TextFieldProps;
```
KevinWu098: issue: why do we have two fields of the same type?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/ManualSearch/ManualSearchTextField.tsx
```diff
@@ -0,0 +1,61 @@
...
+    formControlProps?: FormControlProps;
+    helperText?: string;
```
KevinWu098: issue: isn't this `placeholder` which should belong to `textFieldProps`?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/ManualSearch/ManualSearchInputAdornment.tsx
KevinWu098, line 1: The naming of this component, as well as the ones for select and text input, seems to be incorrectly specific to the manual search even though they're also used for the quick search. Perhaps `SearchAdornment`, `AdornedSelect`, `AdornedTextField` would be better. 

Further, given that these components aren't only used in `ManualSearch` (the directory), they should probably be in their own directory

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/ManualSearch/ManualSearchTextField.tsx
```diff
@@ -0,0 +1,61 @@
...
+    fullWidth?: boolean;
```
KevinWu098: issue: this should be in formControlProps

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/ManualSearch/ManualSearchTextField.tsx
```diff
@@ -0,0 +1,61 @@
+import { FormControl, FormControlProps, TextField, TextFieldProps } from '@mui/material';
+import { grey } from '@mui/material/colors';
+
+import { ManualSearchInputAdornment } from './ManualSearchInputAdornment';
```
KevinWu098: nit: absolute import

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/ManualSearch/ManualSearchTextField.tsx
```diff
@@ -0,0 +1,61 @@
...
+            <TextField
+                fullWidth={fullWidth}
+                size="small"
+                variant="outlined"
+                placeholder={helperText}
+                {...textFieldProps}
+                InputProps={{
+                    ...autocompleteProps?.InputProps,
+                    sx: { paddingX: 0 },
+                    startAdornment: <ManualSearchInputAdornment label={label} />,
```
KevinWu098: question: for accessibility, could you investigate how MUI recommends we ensure that the label (/adornment) is correctly attributed as such? Would `aria-label` on text field or select suffice?

KevinWu098: A related question: does the adornment get read semantically as part of the input component or as a separate component?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/ManualSearch/ManualSearchTextField.tsx
```diff
@@ -0,0 +1,61 @@
...
+                }}
+                inputProps={{
+                    ...autocompleteProps?.inputProps,
+                    sx: {
+                        paddingX: 1,
+                    },
+                }}
+                sx={{
```
KevinWu098: question: Why do we have two sets of input props?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/TermSelector.tsx
```diff
@@ -1,6 +1,8 @@
-import { FormControl, InputLabel, MenuItem, Select, type SelectChangeEvent } from '@mui/material';
+import { MenuItem, type SelectChangeEvent } from '@mui/material';
 import { useCallback, useEffect, useState } from 'react';
 
+import { ManualSearchSelect } from './ManualSearch/ManualSearchSelect';
```
KevinWu098: nit: absolute import

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/GeSelector.tsx
```diff
@@ -54,32 +56,25 @@ export function GeSelector() {
...
+            label="General Education"
+            selectProps={{
+                value: ge,
+                onChange: handleChange,
+                fullWidth: true,
+                variant: 'standard',
+            }}
+            formControlProps={{
+                sx: { flexBasis: '50%', flexGrow: 1 },
+            }}
```
KevinWu098: question: why're we adding this additional style?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/ManualSearch/ManualSearchTextField.tsx
```diff
@@ -0,0 +1,61 @@
...
+                    '& .MuiOutlinedInput-root:hover .MuiOutlinedInput-notchedOutline': {
+                        borderColor: isDark ? grey[400] : grey[600],
+                    },
+                    '& .MuiOutlinedInput-root': {
+                        '&.Mui-focused fieldset': {
+                            borderColor: isDark ? BLUE : undefined,
+                            borderWidth: 1,
+                        },
+                    },
+                }}
```
KevinWu098: issue: these styles don't appear to be functioning. I imagine the active state should be blue, but the hover state is overriding it. Why can't we just omit this?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/ManualSearch/ManualSearchTextField.tsx
```diff
@@ -0,0 +1,61 @@
...
+                {...textFieldProps}
+                InputProps={{
+                    ...autocompleteProps?.InputProps,
+                    sx: { paddingX: 0 },
+                    startAdornment: <ManualSearchInputAdornment label={label} />,
+                }}
+                inputProps={{
+                    ...autocompleteProps?.inputProps,
+                    sx: {
+                        paddingX: 1,
```
KevinWu098: issue: why do we have different styling here?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/ManualSearch/ManualSearchSelect.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { useThemeStore } from '$stores/SettingsStore';
+
+interface ManualSearchSelectProps<T = string | string[]> {
+    label: string;
+    selectProps?: SelectProps<T>;
+    formControlProps?: FormControlProps;
+    helperText?: string;
+    children?: React.ReactNode;
+    fullWidth?: boolean;
+}
```
KevinWu098: issue: similar to text fields, half of these props belong to either `selectProps` or `formControlProps`

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/ManualSearch/ManualSearchSelect.tsx
```diff
@@ -0,0 +1,53 @@
+import { FormControl, FormControlProps, OutlinedInput, Select, SelectProps } from '@mui/material';
+import { grey } from '@mui/material/colors';
+
+import { ManualSearchInputAdornment } from './ManualSearchInputAdornment';
```
KevinWu098: issue: absolute import

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/ManualSearch/ManualSearchSelect.tsx
```diff
@@ -0,0 +1,53 @@
...
+                placeholder={helperText}
+                {...selectProps}
+                sx={{
+                    paddingLeft: 0,
+                    '&:hover .MuiOutlinedInput-notchedOutline': {
+                        borderColor: isDark ? grey[400] : grey[600],
+                    },
+                    '&.Mui-focused .MuiOutlinedInput-notchedOutline': {
+                        borderColor: isDark ? BLUE : undefined,
+                        borderWidth: 1,
```
KevinWu098: issue: these styles seem to be a bit overkill and don't work as expected. why not use the defaults?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearchTextFields.tsx
```diff
@@ -151,65 +146,54 @@ export function AdvancedSearchTextFields() {
...
-                value={units}
-                onChange={handleChange('units')}
-                type="search"
                 helperText="ex. 3, 4, or VAR"
-                style={{ width: 80 }}
+                textFieldProps={{
+                    value: units,
+                    onChange: changeHandlerFactory('units'),
+                    type: 'search',
+                    style: { width: 200 },
```
KevinWu098: nit: use `sx`

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearchTextFields.tsx
```diff
@@ -218,132 +202,147 @@ export function AdvancedSearchTextFields() {
...
-                        checked={building === 'ON'}
-                    />
-                }
+            <ManualSearchSelect
                 label="Online Only"
-                labelPlacement="top"
-                style={{ margin: 0, justifyContent: 'flex-end' }}
-            />
+                selectProps={{
+                    value: building === 'ON' ? 'True' : 'False',
```
KevinWu098: question: why not set the value to be a boolean?

KevinWu098 (changes requested):

apps/antalmanac/src/components/RightPane/RightPaneStore.ts
```diff
@@ -80,17 +87,40 @@ class RightPaneStore extends EventEmitter {
         this.emit('formDataChange');
     };
 
+    storePrevFormData = () => {
+        this.prevFormData = structuredClone(this.formData);
+        this.emit('formDataChange');
```
KevinWu098: question: does this require an event emit?

jotalis:
It seems like there are a lot of MUI style overrides going on to create the custom input components for manual search, particularly with the input adornments.

Perhaps it would be cleaner and easier for you and future devs to customize / maintain if you rebuilt this component using an MUI Label component and TextField component? 

Something like:
```
<div>
    <LeftLabel />
    <TextField />
</div>
```

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchForm.tsx
xgraceyan: This is fixed in the recent commits, I believe the deployment is outdated

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchForm.tsx
xgraceyan: I think the background color being light blue makes sense, unless you mean it should be grey? This was also modified per Minh's feedback of the highlight background color being too dark in dark mode (I assume this was referring to ToggleButton; let me know if I read it wrong)

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchForm.tsx
xgraceyan: Where are you noticing that? I can't seem to find a disappearing border on my end

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/DepartmentSearchBar/DepartmentSearchBar.tsx
xgraceyan: Nothing actually, I must have left it in there and forgot to remove it 🤦 Thanks for catching it

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/ManualSearch/ManualSearchTextField.tsx
xgraceyan: InputProps is for the OutlinedInput wrapping input and inputProps is for the native input element. As far as I know both props needs to pass in both for autocomplete to work

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearchTextFields.tsx
xgraceyan: I tried setting it to boolean initially, but value can only be a string or number

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdornedInputs/AdornedTextField.tsx
```diff
@@ -0,0 +1,52 @@
...
+                        '&:hover:not(.Mui-focused) .MuiOutlinedInput-notchedOutline': {
+                            borderColor: isDark ? grey[500] : grey[600],
+                        },
+                        '&.Mui-focused .MuiOutlinedInput-notchedOutline': {
+                            borderWidth: 1,
+                        },
+                    },
+                    '&&& .MuiInputBase-sizeSmall': {
+                        paddingLeft: 0,
+                    },
```
KevinWu098, line 47: 🔥👾

apps/antalmanac/src/lib/utils.ts
```diff
@@ -28,3 +28,11 @@ export function notNull<T>(value: T): value is NonNullable<T> {
 export function getReferencesOccurring(reference: string[], input?: string | string[] | null): boolean[] {
     return input ? reference.map((reference) => input.includes(reference)) : reference.map(() => false);
 }
+
+export const safeUnreachableCase = <T>(v: never, retVal?: T): T | undefined => {
```
KevinWu098, line 32: 🔥

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/TermSelector.tsx
KevinWu098, line 1: issue: I don't think you need to accept an `isManual` prop here. Just set the value `isManual={true}` (since this will always be true, right?)

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/FuzzySearch.tsx
```diff
@@ -187,22 +186,20 @@ class FuzzySearch extends PureComponent<FuzzySearchProps, FuzzySearchState> {
...
-                sx={{
-                    '& .MuiOutlinedInput-root.MuiInputBase-sizeSmall': {
-                        paddingLeft: 0,
-                    },
-                }}
+                size="small"
                 renderInput={(params) => (
-                    <ManualSearchTextField
-                        autocompleteProps={params}
+                    <AdornedTextField
```
KevinWu098, line 193: issue: this should probably also carry `isManual={true}` right?

suggestion: I think rather than "isManual", it might be better to declare it as some other descriptor, something that indicates this is a special input since its one of the "big" ones. `isRequired`? I honestly don't really know. But I think it makes sense for the fuzzy search and the term selector to be width aligned :)

KevinWu098 (changes requested):
Overall code looks great. Really happy with the progress and the look of the PR.

I only have some minor comments now, as well as some style suggestions:

Screenshot 2025-05-31 at 2 45 48 PM

These "manual" labels are a little wide for my taste — can we shorten them a tad? Similarly, I think the term selector and fuzzy search on the quick search should have the same minWidth, since term is on both pages

KevinWu098 (changes requested):
Oh also, there seems to be a slight UI difference where you can open the department dropdown by clicking the label. Same thing with fuzzy search. Perhaps that's something to do with autocomplete? Can we standardize the behavior to either allow focus on all labels or to disable it on the autocomplete labels?

chrome-capture-2025-5-31 (2)

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/TermSelector.tsx
xgraceyan: Hmm I was thinking the lengths of the input adornment would be shorter on quick search (since there's no need for them to be longer unless it is to remain consistent with manual search). I personally think the extra space is a little too much on quick search but it's not too bad

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/FuzzySearch.tsx
xgraceyan: I renamed it to isAligned for now, I couldn't think of a better name 😅

xgraceyan:
> Oh also, there seems to be a slight UI difference where you can open the department dropdown by clicking the label. Same thing with fuzzy search. Perhaps that's something to do with autocomplete?

Oh yeah, I didn't even notice that. It's fixed now

KevinWu098 (changes requested):
One final nit

Screenshot 2025-06-01 at 12 54 39 AM

I think you'll need to tweak up the width a tad... :p

KevinWu098 (approved):
Fantastic. Banger PR. Well implemented around MUI's jank and a great improvement to our search experience.

CCing @MinhxNguyen7 for any final thoughts

KevinWu098 (dismissed):
Okay, after some further counseling, I think this PR deserves more time (in-person preferably) for us to really perfect the UI and experience (e.g. what should the input hover behavior be with regard to the label?).

So I'm going to put a block on this PR until we chit chat on ~~Wednesday~~ Monday

MinhxNguyen7:
@xgraceyan Can you resolve the comments that have been addressed?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchForm.tsx
xgraceyan: Where is the type error? toggleManualSearch has same type as toggleShowManualSearch as far as I know, I only refactored from Legacy to Manual

xgraceyan:
> @xgraceyan Can you resolve the comments that have been addressed?

Resolved except for one -- not sure where type error on SearchForm is

alexespejo (dismissed):
chore: `Box` component has default `flexDirection`, `gap`, `width` attributes 


ex: 
```html
<Box flexDirection="column" width="100%" gap="2"> </Box>
```

alexespejo (commented):
thought: I think it came up in a meeting but I think it'd be nice to preserve the search mode you're using. Like with local storage

jotalis (dismissed):
Some UI/UX feedback I've gathered from others and my own opinions: https://docs.google.com/document/d/1GJM5UV31lt6Z-d4a0U6Ppwl0BEXonnpCz7sK0hNEsGY/edit?tab=t.0#heading=h.lrkpr38cc9xp

MinhxNguyen7:
There are various important areas of improvement for this feature, ranging from refactoring to use composition to more intuitive user interactions. However, after lengthy discussion, we have decided to merge this and execute the refactor and refinements subsequently. Despite the shortcomings, this PR is still a *strict improvement* over the previous system. The issues raised do not invalidate that. The feature works, and we do not see any risk of usability regression. 

The comments above are important, and @xgraceyan will work on them immediately following this feature merge.

If there's a button  we can press that improves the user experience with no risk and no marginal cost, we should press it. Therefore, let's ship this.

MinhxNguyen7 (approved):
```
                __/___            
          _____/______|           
  _______/_____\_______\_____     
  \              < < <       |    
```

Issue #1278, opened 2025-05-26
Missing query parameters in course pane fetching


MinhxNguyen7:
## Reproduction
- Using manual search, search for GE-2, BIO SCI sections that occur on Wednesday.
- See that some sections are not on Wednesday.
- See that this query returns only sections that occur on Wednesday: anteaterapi.com/v2/rest/websoc?department=BIO+SCI&ge=GE-2&fullCourses=ANY&days=W&quarter=Fall&year=2025.

## Diagnosis
- Running the backend locally and logging the query URL shows that the search waterfalls, where the first query has all of the parameters set, but then there are queries where some parameters are missing.
- The issue is probably some logic in the CourseRenderPane making those queries with `WebSOC.query`/`WebSOC.queryMultiple`.

PR #1279, opened 2025-05-26
Log Websoc Proxy URL

MinhxNguyen7:
## Summary
- Log the query URL that we make to AAPI's WebSOC endpoint.
- This is useful for debugging.

## Issues
- Part of #1278


alexespejo (approved):
Lgtm

PR #1288, opened 2025-06-03
Fix/dev mirror deploymnet

### LINKED ISSUE #1272: Fix development mirror deployment

alexespejo:
## Summary
Creates a Google OAuth class to dynamically change the Redirect URI

## Issues
- Still requires us to manually whitelist every staging URL for testing

Closes #1272

### DIFF
--- a/apps/backend/src/lib/googleOAuth.ts
+++ b/apps/backend/src/lib/googleOAuth.ts
@@ -1,0 +1,59 @@
+import { URL } from 'url';
+import { OAuth2Client } from 'google-auth-library';
+import { backendEnvSchema } from 'src/env';
+import { TRPCError } from '@trpc/server';
+
+export class GoogleOAuth2ClientsManager {
+    /**
+     * Map of redirect origins to OAuth2Clients.
+     */
+    private clients: Map<string, OAuth2Client> = new Map();
+    private readonly clientId: string;
+    private readonly clientSecret: string;
+    private readonly allowDynamicRedirects: boolean;
+    private readonly redirectOrigin: string;
+
+    constructor(_clientId: string, _clientSecret: string) {
+        this.clientId = _clientId;
+        this.clientSecret = _clientSecret;
+
+        const { STAGE, GOOGLE_REDIRECT_URI } = backendEnvSchema.parse(process.env);
+        this.allowDynamicRedirects = STAGE !== 'prod';
+        this.redirectOrigin = GOOGLE_REDIRECT_URI;
+    }
+
+    getClient(redirectOrigin: string): OAuth2Client {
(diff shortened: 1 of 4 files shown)


apps/backend/src/lib/googleOAuth.ts
```diff
@@ -0,0 +1,59 @@
+import { URL } from 'url';
+import { OAuth2Client } from 'google-auth-library';
+import { backendEnvSchema } from 'src/env';
+import { TRPCError } from '@trpc/server';
+
+export class GoogleOAuth2ClientsManager {
```
MinhxNguyen7, line 10: Have a short docstring what this class does.

apps/backend/src/lib/googleOAuth.ts
```diff
@@ -0,0 +1,59 @@
...
+    constructor(_clientId: string, _clientSecret: string) {
+        this.clientId = _clientId;
+        this.clientSecret = _clientSecret;
+
+        const { STAGE, GOOGLE_REDIRECT_URI } = backendEnvSchema.parse(process.env);
+        this.allowDynamicRedirects = STAGE !== 'prod';
+        this.redirectOrigin = GOOGLE_REDIRECT_URI;
+    }
+
+    getClient(redirectOrigin: string): OAuth2Client {
```
MinhxNguyen7, line 35: A docstring here, including what the important conditions are, would be good, too

MinhxNguyen7 (changes requested):
See above

MinhxNguyen7 (approved):
Let's see

PR #1237, opened 2025-05-05
Next.js, SST, Unified Auth Migration

jotalis:
> [!CAUTION]
> Do not merge unless explicitly approved by @andrew-wang0 or @KevinWu098 

## Summary
This PR overhauls the AntAlmanac web framework, deployment infrastructure, and authentication. 

Major changes include:
1. Vite -> Next.js
2. Custom CDK stack -> SST
3. Google Auth -> Unified Auth (OIDC, backed by Google Auth)

`packages/db/.env`:
```
DB_URL="postgresql://postgres:postgres@localhost:5432/antalmanac"
```

`apps/antalmanac/.env`:
```
STAGE=local
MAPBOX_ACCESS_TOKEN=[ASK LEAD]
ANTEATER_API_KEY=[GET AT https://dashboard.anteaterapi.com/]
DB_URL="postgres://postgres:postgres@localhost:5432/antalmanac"
OIDC_CLIENT_ID=antalmanac-dev
OIDC_ISSUER_URL=https://auth.icssc.club
GOOGLE_REDIRECT_URI=http://localhost:3000/auth
```
 

## Issues
Closes #1196 
Closes #1149
Closes #1238 
Closes #1285 
Closes #1286

PR #1294, opened 2025-06-18
fix: Google login from prod

alexespejo:
## Summary
The new dev mirror breaks auth on prod but works on staging and locally.  

When the `.env` `STAGE=prod` the `GoogleOAuth2ClientsManager` throws an error when it should return the `OAuth2Client` with a redirect uri that matches the `GOOGLE_REDIRECT_URI`  

## Test Plan
To mimic prod you have to test locally and make changes in `/backend` to files: `.env` and `/src/index.ts`

`.env`: change `STAGE=local` to `STAGE=prod`

`index.ts`: change line 69 to `if (env.STAGE !== 'local')`

Note: To replicate the prod bug the the above local 

## Issues

Closes #
#1288

### DIFF
--- a/apps/backend/src/lib/googleOAuth.ts
+++ b/apps/backend/src/lib/googleOAuth.ts
@@ -33,11 +33,9 @@ export class GoogleOAuth2ClientsManager {
      * @returns OAuth2Client configured for the given redirect origin
      */
     getClient(redirectOrigin: string): OAuth2Client {
+        // If the STAGE is prod then we will use the value stored from GOOGLE_REDIRECT_URI (i.e. https://antalmanac.com/auth)
         if (!this.allowDynamicRedirects && redirectOrigin !== this.redirectOrigin) {
-            throw new TRPCError({
-                code: 'BAD_REQUEST',
-                message: 'Dynamic redirects are not allowed',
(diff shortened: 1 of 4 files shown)


apps/backend/src/lib/googleOAuth.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * Gets or creates an OAuth2Client for the given redirect origin.
+     *
+     * @param redirectOrigin - The origin URL to create/get a client for
+     * @returns OAuth2Client configured for the given redirect origin
+     */
+    getClient(redirectOrigin: string): OAuth2Client {
+        // If the STAGE is prod then we will use the value stored from GOOGLE_REDIRECT_URI (i.e. https://antalmanac.com/auth)
+        if (!this.allowDynamicRedirects && redirectOrigin !== this.redirectOrigin) {
```
alexespejo, line 37: @MinhxNguyen7 do you remember why we were throwing an error here. It would say `Dynamic redirects are not allowed`

KevinWu098:
Closing as superceded by https://github.com/icssc/AntAlmanac/pull/1237

PR #1296, opened 2025-06-19
fix: auth expiration error

alexespejo:
## Summary
Issue: The code that authenticates the user takes a token arguement and it was passing the expired token but didn't validate the token. However, validating the token is unnecessary here because this bug is only caused when the user first logs in, so the the old token is trivial.

When session expires, prompt user to log back in
image

## Test Plan
1. In the DB, filter for your account in `users` and associated `session` rows
2. Login with your email
3. Manually set the `expiration_date` to some day in the past 
4. Refresh page (should be prompted to log back in)
5. Attempt login
6. Note: Logging in and out normally shouldn't be affected
## Issues

Closes #

### DIFF
--- a/apps/antalmanac/src/components/Header/Load.tsx
+++ b/apps/antalmanac/src/components/Header/Load.tsx
@@ -13,6 +13,7 @@ import {
     DialogContentText,
     DialogTitle,
     TextField,
+    AlertColor,
 } from '@mui/material';
 import { ChangeEvent, PureComponent, useEffect, useState, useCallback } from 'react';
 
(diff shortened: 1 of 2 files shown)


apps/antalmanac/src/routes/AuthPage.tsx
alexespejo, line 1: This code is leftover from "Auth 1" and needs refactoring

alexespejo: For sanity, I feel this can go up just to resolve the problem and I'll work on refactoring today

apps/antalmanac/src/components/Header/Load.tsx
```diff
@@ -177,6 +178,17 @@ class LoadSaveButtonBase extends PureComponent<LoadSaveButtonBaseProps, LoadSave
...
 }
 
+const ALERT_MESSAGES: Record<string, { title: string; severity: AlertColor }> = {
+    SESSION_EXPIRED: {
+        title: 'Your session has expired. Please sign in again.',
+        severity: 'info',
+    },
+    SCHEDULE_IMPORTED: {
+        title: 'This schedule seems to have already been imported!',
+        severity: 'warning',
```
KevinWu098: Tbh this being a warning + the somewhat vague phrasing seems to be an issue for users.

I know this is out of scope, but can we swap this to an info severity and maybe rephrase to be like… “This schedule was previously imported to a Google account. Did you want to sign in with Google?”

KevinWu098 (approved):
I’m on mobile so I can’t test particularly thoroughly

I trust you though

KevinWu098 (approved):

PR #1300, opened 2025-07-11
fix: term selector flex box

KevinWu098:
## Summary

Before
Screenshot 2025-07-10 at 8 41 18 PM

After
Screenshot 2025-07-10 at 8 41 12 PM

## Issues

Closes #

### DIFF
--- a/apps/antalmanac/src/components/Header/Import.tsx
+++ b/apps/antalmanac/src/components/Header/Import.tsx
@@ -371,7 +371,7 @@ export function Import() {
                     {importSource !== ImportSource.AA_USERNAME_IMPORT && (
                         <Stack spacing={1}>
                             <DialogContentText>Make sure you also have the right term selected.</DialogContentText>
-                            <Box sx={{ display: 'flex', gap: 1 }}>
+                            <Box sx={{ display: 'flex' }}>
                                 <TermSelector />
                             </Box>
                         </Stack>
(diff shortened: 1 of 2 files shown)


apps/antalmanac/src/components/Header/Import.tsx
```diff
@@ -369,10 +369,12 @@ export function Import() {
                     )}
 
                     {importSource !== ImportSource.AA_USERNAME_IMPORT && (
-                        <>
+                        <Stack spacing={1}>
                             <DialogContentText>Make sure you also have the right term selected.</DialogContentText>
-                            <TermSelector />
-                        </>
+                            <Box sx={{ display: 'flex', gap: 1 }}>
```
jotalis: question: why is `gap: 1` applied -- doesn't it functionally do nothing here?

apps/antalmanac/src/components/Header/Import.tsx
```diff
@@ -369,10 +369,12 @@ export function Import() {
...
+                                <TermSelector />
+                            </Box>
+                        </Stack>
```
jotalis: note that on most screens, the last option you can see (in this case 2013 Fall Quarter) is actually not the last option (2009 Fall Quarter) because of the `marginThreshold: null` style

this _works_ on the search page's term selector, for the most part, because the dropdown is visibly placed higher, not in the middle of the screen. if you want, it would be nice to quickly fix this issue here as well.

image

jotalis (changes requested):
Looks mostly good to me

apps/antalmanac/src/components/Header/Import.tsx
KevinWu098: Fixed, set the threshold to 4

jotalis (changes requested):
The search dropdown is now broken :p 

image

jotalis (approved):
lgtm

PR #1307, opened 2025-10-04
fix: prevent reactour popover clipping

### LINKED ISSUE #1011: Making the X button on Reactour bigger with padding and better UX?

dsnsgithub:
## Summary
- Added `margin` and `paddingTop` to prevent the step count and the X button from being clipped.

## Test Plan
- Set the dimensions of the viewport to 1920x1080 (this problem persists across all the resolutions I tested, but for reproducibility) 
- Start the Tutorial
- Click to the second step

Before:
image

After:
image

## Issues
Closes https://github.com/icssc/AntAlmanac/issues/1011

### DIFF
--- a/apps/antalmanac/src/App.tsx
+++ b/apps/antalmanac/src/App.tsx
@@ -115,6 +115,8 @@ export default function App() {
                                 borderRadius: 5,
                                 boxShadow: '0 0 10px #000',
                                 padding: 20,
+                                paddingTop: 40,
+                                margin: 20,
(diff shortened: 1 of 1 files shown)


alexespejo (approved):
Sorry this PR got buried! 😅

PR #1308, opened 2025-10-08
Rename Load Button to Sign in and file cleanup

alexespejo:
## Summary
Background: For transitioning users to using OAuth we retained the original AA design (i.e the Load and Save buttons never changed) but since all users at this point are using their google accounts it makes sense to finalize the transition. 

This PR also includes a cleanup of the old `Load.tsx` component which was used for both Loading and Saving schedules but has now just been delegated to Sigining in and loading legacy schedules

 Before
image

After 😎
image

## Test Plan
Sign in and out. 
1. The sign in button should stay to the right of the "Import" button when loading
2. Signing in and out should be the same 
3. Loading old schedules via username should be the same (may not be testable in dev)

## Issues

Closes #


xgraceyan (approved):
Code LGTM, but I think we can rename Signin and Signout to SignIn and SignOut to maintain consistency with e.g. SignInButton.

PR #1310, opened 2025-10-16
Feat: flashing icon to indicate async classes

### LINKED ISSUE #1181: AA Calendar Should Indicate Async
Our calendar tab should have some (small) indicator that there are async classes in one's added

Bennables:
## Summary
There was no way to tell if you had asynchronous classes on antalmanac. 

## Test Plan
Add an async class and see if the UI changes

## Issues

Closes #1181

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -68,5 +68,12 @@ jobs:
 
               run: pnpm build
 
+            - name: Generate termData for tests
+              env:
+                HUSKY: 0
+                NODE_ENV: staging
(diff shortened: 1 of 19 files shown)


vicksey:
Hey @Bennables, I tested this and it works as intended, but it currently never turns off once triggered.
We also talked as a team about how this might not be the right long term solution for notifying users about async courses being added. For now, users can still rely on the MUI snackbar confirmation when an async course is added, so we're going to hold off on merging this. I’ll start looking into cleaner alternatives to display added async courses.

KevinWu098:
Closing as not planned and/or superceded by alternate implementation

PR #1317, opened 2025-11-01
Auto save ScheduleIndex as temporary save data

### LINKED ISSUE #1289: Auto save ScheduleIndex
I noticed this when I use the app. When you're looking at multiple schedules and you close and reopen the app sometimes you notice that the schedule you were previously looking at is not on the page. This is because we only display the last saved schedule not the last opened schedule. Autosave can mask this behavior if the schedule you last opened schedule is also the schedule you last edited. 

Dev notes:
- `scheduleIndex` is a column in the `schedules` table that dictates which schedule displays to the user
- We could store the index locally and only load a different index if the incoming data has been updated (say from another device or browser)

Image

Choollol:
## Summary

Allow temporary save data to be stored locally before eventually being overwritten or permanently saved. Also save `currentScheduleIndex` as temp save data so the last viewed schedule is stored, instead of just the last edited schedule.

## Test Plan

1. Switching the viewed schedule without making any actions that trigger autosave should save the current schedule index locally.
2. Switching schedules without saving then reloading should load the last viewed schedule, not just the last edited schedule.
3. Saving manually should delete temp save data.
4. Autosaving should delete temp save data.
5. Saving data on one tab then reloading on another tab should overwrite and delete temp save data.

## Issues

Closes #1289

### DIFF
--- a/apps/antalmanac/src/actions/AppStoreActions.ts
+++ b/apps/antalmanac/src/actions/AppStoreActions.ts
@@ -18,6 +18,7 @@ import { setLocalStorageUserId, setLocalStorageDataCache } from '$lib/localStora
 import AppStore from '$stores/AppStore';
 import { scheduleComponentsToggleStore } from '$stores/ScheduleComponentsToggleStore';
 import { useSessionStore } from '$stores/SessionStore';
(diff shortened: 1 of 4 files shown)


alexespejo (approved):
LGTM! 🚢 Tested locally

PR #1318, opened 2025-11-01
Rapidly resizing the window duplicates schedules

### LINKED ISSUE #1211: Rapidly resizing the window duplicates schedules
Image

Image

Seems to be a problem with how switching to mobile loads schedule again... we probably have some jank code where we refetch data when we render the mobile UI versus the desktop UI

Choollol:
## Summary

Fix unnecessary `Header` and `Stack` wrapper component duplication. Clear schedules after async-fetching data instead of before to avoid double-filling schedules.

## Test Plan

1. Quickly resizing the UI between mobile and desktop versions (through browser window or developer tools section) should not cause schedules to be reloaded. This can be confirmed at a glance by ensuring the "Schedule loaded" snackbar does not open.

## Issues

Closes #1211

### DIFF
--- a/apps/antalmanac/src/routes/Home.tsx
+++ b/apps/antalmanac/src/routes/Home.tsx
@@ -17,7 +17,6 @@ import { useScheduleManagementStore } from '$stores/ScheduleManagementStore';
 function MobileHome() {
     return (
         <Stack component="main" height="100dvh">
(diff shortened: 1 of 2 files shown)


KevinWu098 (approved):
LGTM, tested locally! Great work :)

PR #1327, opened 2025-11-06
Fix Schedule Section Colors

### LINKED ISSUE #1321: Extremely Light/Dark Colors when creating a schedule
When adding classes that have multiple sections, one of the section colors might randomly be either extremely bright or dark.

To reproduce, I used ICS 6B as an example, but it could be any course with multiple sections.
1. Add both the lecture and a discussion
2. Remove the lecture, re-add the lecture
3. Remove the discussion, re-add the discussion
4. Repeat step 2 and 3 until you see bugged color variation

This bug doesn't only happen under this specific scenario, but this is the easiest way to reproduce.

Image

If AntAlmanac uses purple hues, it can become almost pitch black.

Image
Image

---

This is likely related to https://github.com/icssc/AntAlmanac/pull/1268.

dsnsgithub:
## Summary
Fixes #1321, Reverts https://github.com/icssc/AntAlmanac/commit/68f49dcf0c761a434d5997a5e3e67a91cb2d05a2

Restores/improves schedule section colors.

The root cause of the problem was https://github.com/icssc/AntAlmanac/pull/548:
```js
Math.round((color.l + delta) * 100 % 100)/100
```
when generating a new close color. This would keep pushing the luminance value higher until it would rollover, creating extremely bright and dark section colors.

---

The tweaking of HSL seems hacky, especially considering that luminance does not accurately represent how the human eye perceives light. 

I have changed the color generation so it uses MUI colors only. This will also make it significantly easier to add custom themes in the future.

<table>
  <tr>
    <th>Old</th>
<th>
    
Schedule-old

</th>
  </tr>
  <tr>
     <th>New</th>
<th>
Schedule-new

</th>
  </tr>
</table>

In my testing, one other difference is that if the user picks a custom color, any other sections in the same course that are added will also share the same custom color, instead of a close color. I believe this is for the better as the luminance tweaking inherently creates UI/accessibility problems.

My code also has the added benefit of keeping the color consistent when re-adding the same classes.
## Test Plan
Refer to reproduction in #1321
## Issues

Closes #1321

## Future Followup
https://github.com/icssc/AntAlmanac/pull/1326 is likely not necessary.

### DIFF
--- a/apps/antalmanac/src/stores/scheduleHelpers.ts
+++ b/apps/antalmanac/src/stores/scheduleHelpers.ts
@@ -18,13 +18,13 @@ import {
 import { ScheduleCourse } from '@packages/antalmanac-types';
 
 const colorVariants: Record<string, string[]> = {
-    blue: [blue[300], blue[200], blue[100]],
-    pink: [pink[300], pink[200], pink[100]],
-    purple: [purple[300], purple[200], purple[100]],
-    green: [green[300], green[200], green[100]],
-    amber: [amber[300], amber[200], amber[100]],
-    deepPurple: [deepPurple[300], deepPurple[200], deepPurple[100]],
-    deepOrange: [deepOrange[300], deepOrange[200], deepOrange[100]],
+    blue: [blue[300], blue[200], blue[100], blue[400], blue[500]],
+    pink: [pink[300], pink[200], pink[100], pink[400], pink[500]],
+    purple: [purple[300], purple[200], purple[100], purple[400], purple[500]],
+    green: [green[300], green[200], green[100], green[400], green[500]],
+    amber: [amber[300], amber[200], amber[100], amber[400], amber[500]],
+    deepPurple: [deepPurple[300], deepPurple[200], deepPurple[100], deepPurple[400], deepPurple[500]],
+    deepOrange: [deepOrange[300], deepOrange[200], deepOrange[100], deepOrange[400], deepOrange[500]],
 };
 
 export const colorPickerPresetColors = [
--- a/apps/antalmanac/src/stores/scheduleHelpers.ts
+++ b/apps/antalmanac/src/stores/scheduleHelpers.ts
@@ -1,4 +1,20 @@
-import { amber, blue, deepOrange, deepPurple, green, pink, purple } from '@mui/material/colors';
+import {
+    amber,
+    brown,
+    blue,
+    cyan,
+    deepOrange,
+    deepPurple,
+    green,
+    grey,
+    indigo,
+    lightBlue,
+    pink,
+    purple,
+    red,
+    teal,
+    yellow,
+} from '@mui/material/colors';
 import { ScheduleCourse } from '@packages/antalmanac-types';
 
 export interface HSLColor {
@@ -7,8 +23,26 @@ export interface HSLColor {
     l: number;
 }
 
-const defaultColors = [blue[500], pink[500], purple[500], green[500], amber[500], deepPurple[500], deepOrange[500]];
+const defaultColors = [blue[200], pink[200], purple[200], green[200], amber[200], deepPurple[200], deepOrange[200]];
 
+export const colorPickerPresetColors = [
+    brown[200],
+    red[200],
+    deepOrange[200],
+    amber[200],
+    yellow[200],
+    green[200],
+    teal[200],
+    cyan[100],
+    lightBlue[200],
+    indigo[200],
+    deepPurple[200],
+    pink[200],
+    green[300],
+    grey[600],
+    grey[300],
+    grey[50],
+];
 /**
  * Converts a hex color to HSL
  * Assumes the hex color is in the format #RRGGBB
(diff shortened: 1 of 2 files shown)


apps/antalmanac/src/stores/scheduleHelpers.ts
```diff
@@ -116,9 +150,22 @@ function colorIsContained(color: HSLColor, usedColors: Iterable<HSLColor>, delta
...
 
+/**
+ * Checks if an HSL color is visible and provides enough contrast against the background.
+ *
+ * @param color HSLColor: The color to check.
+ * @param minimum_luminance number: The minimum luminance value to consider the color visible, defaults to 0.2.
+ * @param maximum_luminance number: The maximum luminance value to consider the color visible, defaults to 0.8.
+ * @returns boolean: True if the color is visible enough, false otherwise.
+ */
+function isColorVisible(color: HSLColor, minimum_luminance = 0.2, maximum_luminance = 0.8): boolean {
```
alexespejo: formatting: camel case

alexespejo:
Small nit but LGTM! Congrats on first PR (kinda) – put up even before our first meeting :D

dsnsgithub:
@alexespejo I didn't really like the luminance tweaking that was being done so I changed the color generation to use MUI colors only. It is probably worth taking another look at.

alexespejo (changes requested):
Everything lgtm!

image

apps/antalmanac/src/stores/scheduleHelpers.ts
dsnsgithub: code was removed with later commits, no longer an issue

apps/antalmanac/src/stores/scheduleHelpers.ts
```diff
@@ -1,151 +1,80 @@
...
 import { ScheduleCourse } from '@packages/antalmanac-types';
 
-export interface HSLColor {
-    h: number;
-    s: number;
-    l: number;
-}
-
-const defaultColors = [blue[500], pink[500], purple[500], green[500], amber[500], deepPurple[500], deepOrange[500]];
+const colorVariants: Record<string, string[]> = {
```
alexespejo, line 20: > I’ve never seen it — but I wouldn’t be surprised? I don’t think we should assume it will be at max 3 - Kevin

Honestly I assumed there was a max of 3 but he's not wrong. I think it's best to be dynamic with variants. In the future we might leverage the color picker to support something like async classes (not a great example)

alexespejo (changes requested):

apps/antalmanac/src/stores/scheduleHelpers.ts
dsnsgithub: This code already handles extra section categories by falling back to the default section color, which seems to be the logical choice as it reuses colors when all color families are taken anyway. 

In my opinion, changing the colors of the sections dynamically is not ideal from a UI perspective. My perspective is that dynamically generating close colors by tweaking raw HSL/color values inherently creates UI inconsistencies, accessibility problems, and strays away from MUI. This dynamic solution doesn't fully fix the problem either, as the possible colors are still finite due to a minimum amount of variance between generated colors. By limiting the number of colors that can be used, we can vet the colors to ensure that text is visible while also using colors chosen by the MUI designers to be consistent with the rest of the UI.

It makes sense to me to add more colors as a precaution (even if the code can fallback), but they should fit the MUI theme and be vetted. Dynamic colors also means that it will not only stray away from MUI, but also violate the color choices of any potential themes that we decide to add in the future.


I can add the [400] and maybe even [500] series of colors just in case, and they would be also used when the user adds more than seven classes. Due to the way the code generates colors, the first looped color will be the default color, so it would reuse [300], then [400] and [500].

alexespejo: I'm unopposed to defaulting back to the original colors if there is a scenario where we need more than 3. But I know it was intentional for the colors to desaturate for every subsection. 

cc: @KevinWu098

alexespejo (changes requested):
burh i keep forgetting that i'm supposed to hit submit review

apps/antalmanac/src/stores/scheduleHelpers.ts
dsnsgithub: We have three decent options for color order:
1. Revert to `[300], [200], [100]`
2. `[400], [300], [200], [100]` or `[500], [400], [300], [200], [100]`
3. Keep the same `[300], [200], [100], [400], [500]`

At [500] and at [400] for pink/purple, the text turns white for contrast, which IMO we should try to avoid.
The starting [400] series looks subjectively worse because we lose the pastel colors.
The [50] series of colors is objectively too bright.

I think we should just revert to `[300], [200], [100]` and fallback to [300] for any hypothetical new sections.

alexespejo: Imma @vicksey just for a second opinion. Rotating the colors isn't a bad idea after all they are just hypothetical

alexespejo: Now that I look at it the ~~oops~~ app that I took inspiration from with this issue are consistent with their font colors

alexespejo (approved):

PR #1336, opened 2025-11-15
Feat: TBA sections added dialogue on Calendar

### LINKED ISSUE #1181: AA Calendar Should Indicate Async

vicksey:
## Summary
When a user adds a TBA section to their schedule, a collapsible dialogue is now added to the bottom right of their calendar.
Alternative to #1310 that's less blaring to the user.
image
image

## Test Plan
- [ ] Add/remove TBA time sections

## Issues
 
Closes #1181


apps/antalmanac/src/components/Calendar/AsyncCalendarCard.tsx
```diff
@@ -0,0 +1,134 @@
...
+        zIndex: theme => theme.zIndex.modal - 1,
+        width: '22rem',
+      }}
+    >
+      <Alert
+        icon={<InfoOutlined fontSize="small" />}
+        severity="info"
+        variant="outlined"
+        sx={{
+          bgcolor: 'rgba(255,255,255,0.9)',
```
alexespejo: fix: dark mode
image

apps/antalmanac/src/components/Calendar/AsyncCalendarCard.tsx
```diff
@@ -0,0 +1,134 @@
+import { Collapse, IconButton, Alert, AlertTitle, Box, Typography } from '@mui/material';
```
alexespejo: I think it could be a little more centered vertically with the caret 🙂 
image

apps/antalmanac/src/components/Calendar/AsyncCalendarCard.tsx
```diff
@@ -0,0 +1,134 @@
...
+          width: '100%',
+        }}
+        action={
+        <IconButton size="small" onClick={handleToggleCollapse}>
+            {collapsed ? <ExpandMore fontSize="small" /> : <ExpandMore fontSize="small" sx={{ transform: 'rotate(180deg)' }} />}
+        </IconButton>
+        }
+      >
+        <AlertTitle sx={{ fontSize: '1em' }}>
+          You've added TBA sections:
```
alexespejo: nit: I'm not sure if "TBA sections" is as widely understood as "Async classes." But I'm assuming TBA isn't always async. 

Maybe change to "Async/TBA sections"?

apps/antalmanac/src/lib/localStorage.ts
```diff
@@ -21,6 +21,7 @@ enum LocalStorageKeys {
     importedUser = 'importedUser',
```
alexespejo, line 20: I think this is nice for the UX but you might be able to leverage the `tempSaveData` from #1317

alexespejo (changes requested):
Pretty cool 😄

vicksey:
@alexespejo I resolved all your comments, lmk if theres anything else !

apps/antalmanac/src/components/Calendar/AsyncCalendarCard.tsx
```diff
@@ -0,0 +1,146 @@
...
+    };
+  }, [collapsed, visible]);
+
+  useEffect(() => {
+    const temp = getLocalStorageTempSaveData();
+    if (!temp) return;
+
+    try {
+      const parsed = JSON.parse(temp);
+      setCollapsed(parsed.currentScheduleIndex === scheduleIndex);
```
alexespejo: Bad communication on my part. When I said to leverage the `TempSaveData` I mean't to add a key to it instead of using the `currentScheduleIndex` key. 

Basically you would have two keys `currentScheduleIndex` and `openAsyncTab` (or whatever you want to call it) 

It does create some overhead since you will need to make sure you're only injecting the key value pair into `TempSaveData` so you're not overwriting whatever JSON is in there

apps/antalmanac/src/lib/localStorage.ts
```diff
@@ -21,6 +21,7 @@ enum LocalStorageKeys {
...
     fromLoading = 'fromLoading',
     tempSaveData = 'tempSaveData',
+    tbaAddedCollapsed = 'tbaAddedCollapsed',
```
alexespejo: You removed this right?

alexespejo (changes requested):

apps/antalmanac/src/lib/localStorage.ts
vicksey: oversight, i've removed my bad

apps/antalmanac/src/components/Calendar/AsyncCalendarCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+import { useIsMobile } from '$hooks/useIsMobile';
+
+interface TbaSection {
+  courseTitle: string;
+  sectionCode: string;
+}
+
+export default function AsyncCalendarCard() {
+  const isMobile = useIsMobile();
+  if (isMobile) return null;
```
KevinWu098: issue: unless I'm mistaken, I believe this breaks React's rules of hooks in that it makes the amount of hooks rendered variable between renders

suggestion: move the isMobile check below all of the hooks

```ts
if (!visble || isMobile) {
	return null;
}
```

apps/antalmanac/src/components/Calendar/AsyncCalendarCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+import { getLocalStorageTempSaveData } from '$lib/localStorage';
+import { setTempSaveData } from '$stores/localTempSaveDataHelpers';
+import { useIsMobile } from '$hooks/useIsMobile';
+
+interface TbaSection {
+  courseTitle: string;
+  sectionCode: string;
+}
+
+export default function AsyncCalendarCard() {
```
KevinWu098: issue: this is a TBA card, right? Not an async card? 

edit: I notice the component seems to oscillate between using TBA, Async, or some combination of both — this should be consistent

apps/antalmanac/src/components/Calendar/AsyncCalendarCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+    AppStore.on('removedCoursesChange', handleUpdate);
+    AppStore.on('clearSchedule', handleUpdate);
+    AppStore.on('currentScheduleIndexChange', handleUpdate);
+    return () => {
+      AppStore.off('addedCoursesChange', handleUpdate);
+      AppStore.off('removedCoursesChange', handleUpdate);
+      AppStore.off('clearSchedule', handleUpdate);
+      AppStore.off('currentScheduleIndexChange', handleUpdate);
+    };
+  }, []);
```
KevinWu098: issue: Why do we have a flag that forces a rerender when courses change? My interpretation here is that we want `tbaSections` to be reactive to the value of `AppStore.schedule.getCurrentCourses()`. 

If so, we shouldn't "pipe" reactive events across multiple hooks.

suggestion: we can have this useEffect update the value of tbaSections stored in a `useState` (this means we don't have to go from "course updated" -> "update trigger" -> "update `tbaSections`" and instead just say "course updated" -> "update `tbaSections`"

apps/antalmanac/src/components/Calendar/AsyncCalendarCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+    const handleScreenshot = () => {
+      if (collapsed && visible) {
+        setCollapsed(false);
+      }
+    };
+    AppStore.on('screenshot', handleScreenshot);
+    return () => {
+      AppStore.off('screenshot', handleScreenshot);
+    };
+  }, [collapsed, visible]);
```
KevinWu098: issue: this doesn't seem to actually work?

apps/antalmanac/src/components/Calendar/AsyncCalendarCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+  useEffect(() => {
+    const raw = getLocalStorageTempSaveData();
+    if (!raw) return;
```
KevinWu098: nit: always use braces
```suggestion
    if (!raw) {
    	return;
    }
```

apps/antalmanac/src/components/Calendar/AsyncCalendarCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+      const parsed = JSON.parse(raw);
+      const openIndex = parsed.openAsyncTbaCard;
+
+      setCollapsed(openIndex !== scheduleIndex);
+    } catch {
+      setCollapsed(false);
+    }
+  }, [scheduleIndex]);
+
+  const tbaSections: TbaSection[] = useMemo(() => {
```
KevinWu098: issue: see comment above about `updateTrigger`

apps/antalmanac/src/components/Calendar/AsyncCalendarCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+    if (scheduleIndex == null) return [];
```
KevinWu098: nit: braces

apps/antalmanac/src/components/Calendar/AsyncCalendarCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+    const courses = AppStore.schedule.getCurrentCourses();
+    const sectionsWithTBA: TbaSection[] = [];
+
+    for (const course of courses) {
+      const section = course.section;
+      if (!section) continue;
```
KevinWu098: nit: braces

apps/antalmanac/src/components/Calendar/TbaCalendarCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+          sectionCode: section.sectionCode,
+        });
+      }
+    }
+    return sectionsWithTBA;
+  }, [updateTrigger, scheduleIndex]);
+
+  useEffect(() => {
+    setVisible(tbaSections.length > 0);
+  }, [tbaSections]);
```
KevinWu098: issue: we should not use a `useState` for a value which is wholly derived from another value

suggestion: if a value is derived from another, we can just define it

```ts
const visible = tbaSections.length > 0
```

React automatically recalculates the value of `visible` when `tbaSections` changes, ergo we do not need a `useEffect` to do what React already does for us

apps/antalmanac/src/components/Calendar/TbaCalendarCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+  useEffect(() => {
+    if (visible) {
+      setCollapsed(false);
+      setTempSaveData({
+      currentScheduleIndex: scheduleIndex,
+      openAsyncTbaCard: scheduleIndex,
+    });
```
KevinWu098: chore: formatting

KevinWu098: issue: this is a redundant useEffect. This is, effectively, reactive to the value of visible, which is reactive to the value of `tbaSections`. Therefore, whenever we update the value of `tbaSections`, we can also `setCollapsed` and `setTempSaveData` in the callback which handles `tbaSections`

apps/antalmanac/src/lib/localStorage.ts
```diff
@@ -270,6 +270,7 @@ export function getLocalStoragePWADismissalTime() {
     return window.localStorage.getItem(LSK.pwaDismissalTime);
 }
 
+// Helper functions for temporary data
```
KevinWu098: chore: extraneous comment

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
```diff
@@ -18,6 +18,7 @@ import AppStore from '$stores/AppStore';
 import { useHoveredStore } from '$stores/HoveredStore';
 import { scheduleComponentsToggleStore } from '$stores/ScheduleComponentsToggleStore';
 import { useThemeStore, useTimeFormatStore } from '$stores/SettingsStore';
+import AsyncCalendarCard from './AsyncCalendarCard';
```
KevinWu098: chore: absolute import please

apps/antalmanac/src/components/Calendar/TbaCalendarCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+  }, [updateTrigger, scheduleIndex]);
+
+  useEffect(() => {
+    setVisible(tbaSections.length > 0);
+  }, [tbaSections]);
+
+  useEffect(() => {
+    if (visible) {
+      setCollapsed(false);
+      setTempSaveData({
```
KevinWu098: question: what is temp data? it looks like this is not something that you added in this PR, ergo it means that this localStorage value is also read/written to by another component. this looks like it writes over the value. could you confirm that this will not create issues with other consumers of tempSaveData?

KevinWu098: issue: the name `tempSaveData` implies that it is supposed to be temporary... but this component saves and utilizes the value on subsequent visits to the website

suggestion: create a dedicated localStorage value or figure out another way to store data for this. perhaps... on the `Schedule` object we save to the db?

apps/antalmanac/src/components/Calendar/TbaCalendarCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+      currentScheduleIndex: scheduleIndex,
+      openAsyncTbaCard: scheduleIndex,
+    });
+    }
+  }, [visible, scheduleIndex]);
+
+  const handleToggleCollapse = () => {
+    setCollapsed((prev) => {
+      const next = !prev;
+      setTempSaveData({
```
KevinWu098: question (repeat): what is temp data?

apps/antalmanac/src/components/Calendar/AsyncCalendarCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+      const next = !prev;
+      setTempSaveData({
+        currentScheduleIndex: scheduleIndex,
+        openAsyncTbaCard: next ? null : scheduleIndex,
+      });
+      return next;
+    });
+  };
+
+  if (!visible) return null;
```
KevinWu098: nit: braces

apps/antalmanac/src/components/Calendar/AsyncCalendarCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+  return (
+    <Box
+      sx={{
+        position: 'absolute',
+        bottom: 16,
+        left: 'auto',
+        right: 16,
+        zIndex: (theme) => theme.zIndex.drawer - 1,
+        width: '22rem',
```
KevinWu098: question: why 22 rem? is this specific to some styling of the calendar? is it arbitrary?

apps/antalmanac/src/components/Calendar/TbaCalendarCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+          bgcolor: theme.palette.background.paper,
+          py: 0.5,
+          px: 1,
+          width: '100%',
+        }}
+        action={
+        <IconButton size="small" onClick={handleToggleCollapse}>
+            {collapsed ? <ExpandMore fontSize="small" /> : <ExpandMore fontSize="small" sx={{ transform: 'rotate(180deg)' }} />}
+        </IconButton>
+        }
```
KevinWu098: nit (repeat): formatting

you should check your prettier settings and write an issue if it appears to be a problem with AA and not your configuration

apps/antalmanac/src/components/Calendar/TbaCalendarCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+        icon={<InfoOutlined fontSize="small" />}
+        severity="info"
+        variant="outlined"
+        sx={{
+          bgcolor: theme.palette.background.paper,
+          py: 0.5,
+          px: 1,
+          width: '100%',
+        }}
+        action={
```
KevinWu098: issue: the action button looks misaligned/not respecting the alert padding

Image

apps/antalmanac/src/components/Calendar/AsyncCalendarCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+          px: 1,
+          width: '100%',
+        }}
+        action={
+        <IconButton size="small" onClick={handleToggleCollapse}>
+            {collapsed ? <ExpandMore fontSize="small" /> : <ExpandMore fontSize="small" sx={{ transform: 'rotate(180deg)' }} />}
+        </IconButton>
+        }
+      >
+        <AlertTitle sx={{ fontSize: '1em', my: 'auto'}}>
```
KevinWu098: question: `em`? Although `em` is more correct in some cases, AA tends to use `rem` in most cases iirc. was this intentional?

apps/antalmanac/src/components/Calendar/AsyncCalendarCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+          You've added Async/TBA sections:
+        </AlertTitle>
+
+        <Collapse in={!collapsed} timeout="auto" unmountOnExit>
+          <Box sx={{ mt: 0.25, py: 0.25 }}>
```
KevinWu098: suggestion: is 2 pixels of vertical padding noticeable? seems like we could omit these styles and add a `gap` property to the container if we want to space out the title and the collapse

KevinWu098 (dismissed):

vicksey:
Main Updates:
- Refactored to TbaCalendarCard to stay consistent to websoc
- Screenshot auto opens card
- Removed the finicky logic to save the card's collapse state (idt its necessary & idt many people will be collapsing it since its off on the side)
- Fixed design

apps/antalmanac/src/components/Calendar/TbaCalendarCard.tsx
```diff
@@ -0,0 +1,151 @@
...
+  }, [tbaSections]);
+
+  useEffect(() => {
+    if (visible) {
+      setCollapsed(false);
+    }
+  }, [visible]);
+
+  const handleToggleCollapse = () => {
+    setCollapsed((prev) => {
```
alexespejo:
```suggestion
    setCollapsed((prev) => !prev);
```

apps/antalmanac/src/components/Calendar/TbaCalendarCard.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+interface TbaCalendarCardProps {
+  screenshotTrigger?: number;
+}
+
+export default function TbaCalendarCard({ screenshotTrigger }: TbaCalendarCardProps) {
+  const isMobile = useIsMobile();
+  const [tbaSections, setTbaSections] = useState<TbaSection[]>([]);
+  const [collapsed, setCollapsed] = useState(false);
+  const [visible, setVisible] = useState(false);
```
alexespejo: The visible state is derived directly from `tbaSections.length > 0`. You can eliminate the state and third/fourth useEffect:

```suggestion
// Remove:
// const [visible, setVisible] = useState(false);
// useEffect for setting visible
// useEffect for resetting collapsed

// Replace with:
const visible = tbaSections.length > 0;

useEffect(() => {
  if (tbaSections.length > 0) {
    setCollapsed(false);
  }
}, [tbaSections.length]);```

**Review by alexespejo — CHANGES_REQUESTED** [review]
Some clarity stuff

**Review by alexespejo — CHANGES_REQUESTED** [review]
thought: After consideration it might be a little strange to obstruct the schedule even if on desktop it only obstructs Friday ~8pm-9pm. 

Maybe there is an alternative approach where we can create space at the bottom of the schedule like a footer. So the schedule will scroll normally but the option to view async sections is present. Can also loop in design.

**Comment by alexespejo** [comment]
Can you merge in main I'm curious what the colors look like against the new darker pallete

File: apps/antalmanac/src/components/Calendar/TbaCalendarCard.tsx
Comment (alexespejo) [inline apps/antalmanac/src/components/Calendar/TbaCalendarCard.tsx:1]: issue: funky hover 

Image

File: apps/antalmanac/src/components/Calendar/TbaCalendarCard.tsx [hunk 29]
```diff
@@ -0,0 +1,179 @@
...
+                    width: 48,
+                    height: 48,
+                    borderRadius: '50%',
+                    bgcolor: BLUE,
+                    boxShadow: 2,
+                    cursor: 'pointer',
+                }}
+            >
+                <Typography variant="button" sx={{ fontWeight: 600, letterSpacing: 1, color: 'white' }}>
+                    TBA
```
alexespejo, line 40: thought: different icon? Not sure what an icon would be. I remember we more or less talked about why it's TBA and not like "ASYNC"

apps/antalmanac/src/components/Calendar/TbaCalendarCard.tsx
alexespejo, line 1: thought: hide the menu on mobile? I think everything on mobile is a little too compact

apps/antalmanac/src/components/buttons/Screenshot.tsx
alexespejo, line 1: thought: not sure if it's possible but I wonder if we can remove the tba button from the screenshot 
Image

apps/antalmanac/src/components/Calendar/TbaCalendarCard.tsx
vicksey: I just kept it as TBA since that's the technical term (by like websoc) and its rly short. async could work? its just a tad longer 
image
image

apps/antalmanac/src/components/Calendar/TbaCalendarCard.tsx
vicksey: i think the whole point of the circle was so that it could be compatible for mobile. someone said that its better design practice to like try to make features available for both. also its the same amt of space as the help menu on mobile (we just replaced the functionality)

alexespejo (changes requested):
After a quick talk with designers TBA seems to make the most sense. Sorry for the confusion

apps/antalmanac/src/components/Calendar/TbaCalendarCard.tsx
```diff
@@ -0,0 +1,168 @@
...
+}
+
+function TbaExpandedCard({ tbaSections, onToggle }: { tbaSections: TbaSection[]; onToggle: () => void }) {
+    const theme = useTheme();
+    return (
+        <Alert
+            icon={
+                <InfoOutlined
+                    fontSize="small"
+                    sx={{ color: theme.palette.mode === 'dark' ? theme.palette.common.white : BLUE }}
```
alexespejo: chore: use `useThemeStore` similar to this pattern `    const isDark = useThemeStore((store) => store.isDark);`

apps/antalmanac/src/components/Calendar/TbaCalendarCard.tsx
```diff
@@ -0,0 +1,168 @@
...
+                    <Close fontSize="small" />
+                </IconButton>
+            }
+        >
+            <AlertTitle sx={{ typography: 'subtitle2', my: 'auto' }}>TBA section added:</AlertTitle>
+
+            <Box sx={{ gap: 0.5 }}>
+                {tbaSections.map((section) => (
+                    <Typography
+                        key={`${section.deptCode}-${section.courseNumber}-${section.sectionCode}`}
```
alexespejo, line 95: chore: on mobile maybe we can hide the section code since it'll overflow in a weird way. And mobile context a user wouldn't be conerned with it 

Image

alexespejo (changes requested):

alexespejo (approved):
Shout out to a year old PR. LGTM

PR #1339, opened 2025-11-17
Fix: Mobile UI Start

### LINKED ISSUE #1316: RFC: AntAlmanac Mobile UX on start
When I use AA on my phone it's typically to view the order of my schedule. However, the landing page is the "Search" page and the "schedule loaded" popup annoyingly appears at the bottom blocking the navigation buttons to view my schedule. Maybe this sentiment is shared sentiment 🤷 

Can go about it in different ways
- Put the popup somewhere else
- If the user has schedule data default to the schedule page 
- Locally store whether they had calendar or search open 

Image

AureliaSindhu:
## Summary
- Move mobile snackbars above the bottom navigation bar for better visibility.
- Fix snackbar duration conversion from seconds to milliseconds so they display for the intended time.
- Update mobile initial tab routing so users with saved/loaded schedules land on **Calendar** instead of **Search**.

## Test Plan
- Snackbar Positioning

- Mobile Tab Routing
  - With `userID` in `localStorage`: reload on mobile → **Calendar** selected by default.
  - Without `userID`: reload on mobile → **Search** selected by default.
  - With loaded schedule data: add a course, open new mobile tab → **Calendar** if data persists, otherwise **Search**.

## Issues 
Closes #1316

### DIFF
--- a/apps/antalmanac/src/components/ScheduleManagement/ScheduleManagement.tsx
+++ b/apps/antalmanac/src/components/ScheduleManagement/ScheduleManagement.tsx
@@ -5,6 +5,7 @@ import { useParams } from 'react-router-dom';
 import { ScheduleManagementContent } from '$components/ScheduleManagement/ScheduleManagementContent';
 import { ScheduleManagementTabs } from '$components/ScheduleManagement/ScheduleManagementTabs';
 import { getLocalStorageUserId } from '$lib/localStorage';
+import AppStore from '$stores/AppStore';
 import { paramsAreInURL } from '$stores/CoursePaneStore';
 import { useTabStore } from '$stores/TabStore';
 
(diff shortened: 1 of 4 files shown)


apps/antalmanac/src/components/ScheduleManagement/ScheduleManagement.tsx
```diff
@@ -50,14 +51,48 @@ export function ScheduleManagement() {
...
 
-        if (userId === null || paramsAreInURL()) {
+        if (urlHasManualSearchParams) {
             setActiveTab('search');
-        } else if (isMobile) {
+            return;
+        }
+
+        if (!isMobile) {
+            if (userId === null) {
```
alexespejo: Auth introduces another localStorage key `sessionId` so you can use `getLocalStorageSessionId` to perform a similar check if the user is signed in

alexespejo: We also use a pattern with validateSession inside of  [SessionStore](https://github.com/icssc/AntAlmanac/blob/main/apps/antalmanac/src/stores/SessionStore.ts) if a user is signed in

alexespejo:
bug: When you first add a course section the app jumps to the calendar page. This seems to only happen if you've never saved a schedule before. Having an empty schedule that had previously contained a saved schedule doesn't seem to repro this issue. 

Repro:
Create a new AA account -> add a class section (i.e COMPSCI161 LEC) 

Expected: 
Course should visually be highlighted yellow in the search menu

Actual: 
The page jumps to the calendar screen 

https://github.com/user-attachments/assets/4ce6cca9-8316-4c43-acb7-bade1bad76ca

_Sorry I didn't format this properly in the review_

alexespejo (changes requested):
Hey Lia congrats on your first PR! Small code change and an issue I ran into while testing locally. Lemme know if you need any help

apps/antalmanac/src/components/ScheduleManagement/ScheduleManagement.tsx
AureliaSindhu: fixed!

alexespejo (approved):
LGTM 🚀🤳

PR #1341, opened 2025-11-17
AANTS CDK Changes

IsaacNguyen:
## Summary
Added AANTS stack to CDK 

## Notes
I did this with Minh last year, but we never had the chance to formally test if it worked correctly.

### DIFF
--- a/.github/workflows/deploy_staging.yml
+++ b/.github/workflows/deploy_staging.yml
@@ -57,15 +57,11 @@ jobs:
 
         outputs:
             # A staging backend should only be deployed if either the backend or CDK changed.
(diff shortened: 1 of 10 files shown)


.github/workflows/deploy_staging.yml
```diff
@@ -57,15 +57,11 @@ jobs:
 
         outputs:
             # A staging backend should only be deployed if either the backend or CDK changed.
-            # should-deploy-backend: ${{ steps.changed.outputs.backend == 'true' ||  steps.changed.outputs.cdk == 'true' || steps.changed.outputs.aants=='true' }}
```
KevinWu098, line 60: chore: comment

KevinWu098 (approved):
LGTM but this is one of those things that’s hard to know if it’ll work until it actually runs 😗

PR #1343, opened 2025-11-19
Adds popup notif. (fixes #1183)

### LINKED ISSUE #1183: OAuth Alert on logout
When the user logs out a smal modal indicator would be nice to inform them. 

This issue could also include when a user's session expires (i.e they were logged in some time ago but they've been automatically logged out)

Original message:
              I think everything looks and flows great. The only minor critique I have is that I think there should be better indication of when a user signs out. Maybe just a little popup at the top notifying them that they have _successfully_ been signed out of whatever account they were on. Otherwise, good looks Alex!!

_Originally posted by @IsaacNguyen in https://github.com/icssc/AntAlmanac/pull/1127#discussion_r1975972852_

victorialee13:
## Summary

Improves the sign-out flow by showing a notification when the user signs out of their Google Account.
This update uses the global openSnackbar action to display a top-right toast message including the user's name or email (when available).
The logout notification appears before the session is cleared to ensure it is always visible to the user.

## Test Plan
Manually signed it with test account and clicked 'Sign out'. 
Dropdown menu closes and Snackbar appear top right corner
Message shows one of three choices
  1. account name
  2. account email
  3. 'your account' as fallback
After 800ms delay, session clears through clearSession() and redirects to /.
## Issues

Closes #1183


alexespejo (changes requested):
Congrats on your first PR!! 

This looks like a solid start! This is a smaller change so it's not too big but in the future if you can provide screen shots in the PR that would be good 🙂 

When it comes to user signin notifications we typically use dialogs or popups like this to inform them:
image

We create an [`AlertDialog`](https://github.com/icssc/AntAlmanac/blob/main/apps/antalmanac/src/components/AlertDialog.tsx) component for this particular instance

victorialee13:
Summary
Changes Snackbar Notification which notifies users they have been signed out to instead using the AlertDialong Component

Test Plan
Manually signed it with test account and clicked 'Sign out'.

Screenshot 2025-11-26 135427


Closes https://github.com/icssc/AntAlmanac/issues/1183

apps/antalmanac/src/components/Header/Signout.tsx
```diff
@@ -79,6 +88,18 @@ export function Signout() {
                     <ListItemText>Log out</ListItemText>
                 </MenuItem>
             </Menu>
+            <AlertDialog
+                open={isSignoutDialogOpen}
+                title={`Signed out of ${accountName}`}
+                severity="info"
+                onClose={async () => {
```
alexespejo: We want the dialog to appear after the user has signed out. With the current logic the user has to click the "Confirm" button on this modal to actually log out (i.e if the user refreshes the page before they click "Confirm" they'll still be logged in). 

I would look into the logic we use to determine if a user is logged in. You can also make use of local storage to save whether we're in a "logout" state in order to display the message.

alexespejo: Aside: when we bind functions to event handlers like this we typically define a separate function in the body of the component just to create a separation of concerns :) 
```
function onClick(){}
...
<AlertDialog>
onClose={onClick}
</AlertDialog>
```

alexespejo (changes requested):

apps/antalmanac/src/components/Header/Signout.tsx
```diff
@@ -79,6 +88,18 @@ export function Signout() {
                     <ListItemText>Log out</ListItemText>
                 </MenuItem>
             </Menu>
+            <AlertDialog
+                open={isSignoutDialogOpen}
+                title={`Signed out of ${accountName}`}
```
alexespejo: You don't necessarily have to display the name of the user as the text might wrap with the screen on mobile 

image

apps/antalmanac/src/components/Header/Signout.tsx
victorialee13: In the LocalStorageKeys, added signoutNotice flag and uses it for proper behavior change: Clicking "Log out" immediately clears session and sets signoutNotice.

alexespejo (changes requested):
My fault, I gave bad instructions that are too overkill.

Revert the commits related to the suggested change. 

The flow should be:

"Hit logout" -> invalidates session -> display alert -> "user hits close" -> reload the page 

*the current issue is if you hit logout the user HAS to hit "close" in order to be properly invalidated without being dependent on the user input

apps/antalmanac/src/lib/localStorage.ts
```diff
@@ -96,6 +97,18 @@ export function getLocalStorageSessionId() {
 export function removeLocalStorageSessionId() {
     window.localStorage.removeItem(LSK.sessionId);
 }
+
```
alexespejo, line 100: fix: these changes aren't needed anymore, revert the file changes

alexespejo (changes requested):
Resolve the merge conflicts so the deployment can run

apps/antalmanac/src/components/Header/Header.tsx
```diff
@@ -79,6 +89,14 @@ export function Header() {
                 >
                     NOTE: All changes made to your schedules will be saved to your Google account
                 </AlertDialog>
+                <AlertDialog
+                    open={openSignoutDialog}
+                    title="Signed out successfully"
+                    severity="info"
+                    onClose={handleCloseSignoutDialog}
+                >
+                    You have successfully signed out. Close this dialog to continue browsing AntAlmanac.
```
alexespejo: The message is a little redundant maybe something like, "You can still browse and build schedules, but your changes cannot be saved." 

or maybe no message at all

apps/antalmanac/src/lib/localStorage.ts
alexespejo: fix: this file still needs these changes removed

apps/antalmanac/src/components/Header/Signout.tsx
```diff
@@ -2,48 +2,38 @@ import AccountCircleIcon from '@mui/icons-material/AccountCircle';
...
     const open = Boolean(anchorEl);
-    const handleClick = (event: React.MouseEvent<HTMLButtonElement>) => {
+    const handleClick = (event: MouseEvent<HTMLButtonElement>) => {
         setAnchorEl(event.currentTarget);
     };
     const handleClose = () => {
         setAnchorEl(null);
     };
 
+    const handleLogout = async () => {
```
alexespejo, line 27: It seems like with migration they introduced a new way of loging out a user. We might have to rethink the approach. That's my bad I didn't notice that they created a `logout` procedure 

new logout method
```ts
const handleLogout = async () => {
        if (!session) {
            navigate('/');
            return;
        }

        try {
            const { logoutUrl } = await trpc.userData.logout.mutate({
                sessionToken: session,
                redirectUrl: window.location.origin,
            });
            clearSession();

            window.location.href = logoutUrl;
        } catch (error) {
            console.error('Error during logout', error);
            clearSession();
            navigate('/');
        }
    };
```

apps/antalmanac/src/stores/SessionStore.ts
```diff
@@ -30,10 +30,9 @@ export const useSessionStore = create<SessionState>((set) => {
             const currentSession = getLocalStorageSessionId();
             if (currentSession) {
                 await trpc.auth.invalidateSession.mutate({ token: currentSession });
-                removeLocalStorageSessionId();
-                set({ session: null, sessionIsValid: false });
-                window.location.reload();
+                window.localStorage.removeItem('sessionId');
             }
+            //set({ session: null, sessionIsValid: false });
```
alexespejo: remove comment

lint-staged.config.cjs
```diff
@@ -4,7 +4,7 @@
  * @type {import('lint-staged').Config}
  */
 const config = {
-    '*.?(c|m){js,ts}?(x)': ['prettier --write', 'eslint --quiet --fix'],
+    '*.?(c|m){js,ts}?(x)': ['prettier --write', 'pnpm --filter antalmanac exec eslint --quiet --fix'],
```
alexespejo: question: what is this lol

apps/antalmanac/src/lib/localStorage.ts
```diff
@@ -94,8 +95,16 @@ export function getLocalStorageSessionId() {
     return window.localStorage.getItem(LSK.sessionId);
 }
 
-export function removeLocalStorageSessionId() {
-    window.localStorage.removeItem(LSK.sessionId);
+export function setLocalStorageSignoutNotice(value: string) {
```
alexespejo: if you were able to come up with a solution to not have to use local storage you can remove this

alexespejo (changes requested):
Looks good for the most part. Remove the small changes and you're good to go!

alexespejo (approved):
LGTM

brelieu05 (approved):
lgtm

brelieu05 (approved):
lgtm

PR #1344, opened 2025-11-19
Added Loading Skeleton UI

### LINKED ISSUE #1212: RFC: Add loading skeleton UI
> do we have like a "loading schedule" type of animation? like a skeleton to indicate the schedules are being loaded. I didnt notice that it takes a second for the data to load when i was testing locally
-- Alex E

Good idea! Maybe some loading skeletons for the calendar (some pulsing rectangles), some skeletons for the added page (pulsing rectangles), maybe even pulsing markers for the map page...

Additionally, but not required, could be nice to do a skeleton animation instead of a loading gif for the search...

brelieu05:
## Summary
Created pulsing calendar skeleton which shows up when logged in and when classes/events are loading in.

Skeleton events are just preloaded + hardcoded and is type Custom Event as Alex suggested

## Test Plan
Logged in and saw if appeared on reload

## Issues

Closes #1212 

## Future Followup
- Could create for Added Tab and/or Map Tab. However, didn't do it on this sprint because of comment talking about how no one looks on the right side when loading, which I agree with.
- Can change design/layout of skeleton

### DIFF
--- a/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
+++ b/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
@@ -81,22 +81,14 @@ export const ScheduleCalendar = memo(() => {
 
         const skeletonBlueprints = [
             { dayOffset: 0, startHour: 12, startMinute: 0, endHour: 12, endMinute: 50 },
-            { dayOffset: 0, startHour: 13, startMinute: 30, endHour: 14, endMinute: 50 },
-            { dayOffset: 0, startHour: 15, startMinute: 30, endHour: 16, endMinute: 50 },
 
             { dayOffset: 1, startHour: 11, startMinute: 0, endHour: 12, endMinute: 20 },
-            { dayOffset: 1, startHour: 14, startMinute: 0, endHour: 15, endMinute: 20 },
             { dayOffset: 1, startHour: 17, startMinute: 0, endHour: 18, endMinute: 20 },
 
-            { dayOffset: 2, startHour: 13, startMinute: 30, endHour: 14, endMinute: 50 },
-            { dayOffset: 2, startHour: 15, startMinute: 30, endHour: 16, endMinute: 50 },
-
             { dayOffset: 3, startHour: 11, startMinute: 0, endHour: 12, endMinute: 20 },
-            { dayOffset: 3, startHour: 14, startMinute: 0, endHour: 15, endMinute: 20 },
             { dayOffset: 3, startHour: 17, startMinute: 0, endHour: 18, endMinute: 20 },
 
             { dayOffset: 4, startHour: 9, startMinute: 0, endHour: 9, endMinute: 50 },
-            { dayOffset: 4, startHour: 13, startMinute: 0, endHour: 13, endMinute: 50 },
             { dayOffset: 4, startHour: 15, startMinute: 0, endHour: 15, endMinute: 50 },
         ];
 
@@ -109,7 +101,7 @@ export const ScheduleCalendar = memo(() => {
             end.setHours(blueprint.endHour, blueprint.endMinute, 0, 0);
 
             return {
-                color: '#2563eb',
+                color: '#6d6d6d',
                 start,
                 end,
                 title: '',
--- a/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
+++ b/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
@@ -3,7 +3,7 @@ import './calendar.css';
 
 import { Box, Backdrop, useTheme } from '@mui/material';
 import moment from 'moment';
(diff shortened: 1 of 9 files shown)


alexespejo (changes requested):
you got the right idea!

I think it might be off putting to users since at a glance they might think their schedule got changed to... a heart 

MUI has this [skeleton component](https://mui.com/material-ui/react-skeleton/) that I think might be a good addition. I do like the pulsing animation though. 
_Note: if the MUI component doesn't work than I wouldn't be opposed to some hard coded css animation_

While I do like the heart it may look more _elegant_ if the loading schedule resembled a schedule 🙂 

image

brelieu05:
Changed it so the loading skeleton is its own type instead of checking if event.title == "Loading..." so no user can intentionally create a skeleton event.

Additionally, it's no longer a heart 😭 and its just a copy of a the schedule you sent but in loading form. (I still think you can do something cool with the loading like make a snowflake for winter quarter or a leaf for fall quarter)

I kept everything an "antalmanac" blue so its clear that its a placeholder for loading, but let me know if there's a different color I should use.

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
```diff
@@ -70,7 +78,157 @@ export const ScheduleCalendar = memo(() => {
...
+                    color: '#6d6d6d',
+                    start,
+                    end,
+                    title: '',
+                    isSkeletonEvent: true,
+                } as SkeletonEvent;
+            });
+        }
+
+        const skeletonBlueprintVariations = [
```
alexespejo: Move consts to a separate file

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
```diff
@@ -70,7 +78,157 @@ export const ScheduleCalendar = memo(() => {
         showFinalsSchedule,
     ]);
 
-    const events = getEventsForCalendar();
+    useEffect(() => {
```
alexespejo, line 86: This is a nice add-on. I would also look into the instance where the user adds/deletes an item from the schedule. For consistency I think we should cache the change as well so that it's reflected on the next reload.

Repro: Open schedule -> remove or add a lec or dis -> refresh 

Expected: 
- The loading items should reflect the most recent changes 

Actual:
- Since the changes weren't cached we default to the random preset

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
```diff
@@ -70,7 +78,157 @@ export const ScheduleCalendar = memo(() => {
...
+                }
+            } else if (hasHadEventsRef.current) {
+                deleteTempSaveData();
+                hasHadEventsRef.current = false;
+            }
+        }
+    }, [eventsInCalendar, loadingSchedule]);
+
+    const createSkeletonEvents = useCallback((): SkeletonEvent[] => {
+        const baseDate = new Date(2018, 0, 1);
```
alexespejo: Separate into constants. Use across other instances of this default date

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
```diff
@@ -70,7 +78,157 @@ export const ScheduleCalendar = memo(() => {
...
+        const weekStart = new Date(baseDate);
```
alexespejo: Is this redundant?

alexespejo (changes requested):

brelieu05:
Should be good now, had to make a dedicated cache for the skeleton. lmk if theres a better way of doing it

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
```diff
@@ -70,7 +83,94 @@ export const ScheduleCalendar = memo(() => {
...
+
+        if (savedDataString) {
+            const parsedData = JSON.parse(savedDataString);
+            if (Array.isArray(parsedData) && parsedData.length > 0) {
+                skeletonBlueprints = parsedData;
+            }
+        }
+
+        if (skeletonBlueprints) {
+            return skeletonBlueprints.map((blueprint) => {
```
alexespejo: This logic repeats on line 156

alexespejo (changes requested):

apps/antalmanac/src/lib/localStorage.ts
```diff
@@ -281,3 +282,15 @@ export function getLocalStorageTempSaveData() {
 export function removeLocalStorageTempSaveData() {
     window.localStorage.removeItem(LSK.tempSaveData);
 }
+
```
alexespejo, line 285: we do lwky needa refactor local storage

alexespejo (approved):
LGTM! (what are the odds this breaks). Also hit "Resolve Conversations" when you resolve them in the code lol

PR #1347, opened 2025-11-21
fix: Enrollment History Dates

dsnsgithub:
## Summary

The API returns dates in YYYY-MM-DD, but `new Date(day, month, year)` uses zero indexing. This creates a problem where the day and month are both one ahead.

Instead of breaking the date apart, it can be passed directly into the Date function.
When displaying the date, it now uses the browser's `toLocaleDateString` which accomplishes the same thing as before. As a side effect, it could also provide some level of localization if the user uses a different language.

I removed the ZotTracker link since the data hasn't been updated for a while now.
The API also seems to return "-1", which makes no sense in terms of waitlist numbers so they are now filtered out. "n/a" doesn't appear to be a response that the API gives anymore and I presume it was replaced with "-1".

## Test Plan
Check any course with enrollment history.

Current:
image

Fixed:
image

## Issues

The exact opposite problem of https://github.com/icssc/AntAlmanac/pull/1189

### DIFF
--- a/apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
+++ b/apps/antalmanac/src/components/RightPane/SectionTable/EnrollmentHistoryPopup.tsx
@@ -1,5 +1,5 @@
 import { ArrowBack, ArrowForward } from '@mui/icons-material';
-import { Box, IconButton, Link, Typography, Skeleton, Tooltip, useMediaQuery } from '@mui/material';
+import { Box, IconButton, Typography, Skeleton, Tooltip, useMediaQuery } from '@mui/material';
 import { useState, useEffect, useMemo, useCallback } from 'react';
 import {
     LineChart,
@@ -95,7 +95,6 @@ export function EnrollmentHistoryPopup({ department, courseNumber }: EnrollmentH
         } | ${currEnrollmentHistory.instructors.join(', ')}`;
     }, [courseNumber, department, enrollmentHistory, graphIndex]);
     const isDark = useThemeStore((state) => state.isDark);
-    const encodedDept = useMemo(() => encodeURIComponent(department), [department]);
     const axisColor = isDark ? '#fff' : '#111';
     const tooltipDateColor = '#111';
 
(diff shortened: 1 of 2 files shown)


KevinWu098 (approved):
LGTM, but just so I understand what's going on... in your issue section where you describe #1189... is "opposite" the correct phrasing here? 

My understanding is that #1189 corrected a different problem with the date string, but seemingly introduced a different off-by-one error. And then, this PR resolves that bug?

dsnsgithub:
TLDR: Dates are hard. Opposite might not be the right word for the situation.

Here is some [background](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date):
When constructing a date (`new Date(year, month, day)`):
- `day` acts as normal, starting at `1` and counts normally
- `month` starts from `0` for January and goes to `11` for December
- `year` is normal
The `getDay`, `getMonth`, `getFullYear` also return these special stored numbers.

This is the quirk that has confused people:
> If any parameter overflows its defined bounds, it "carries over". For example, if a monthIndex greater than 11 is passed in, those months will cause the year to increment;

The API (`date`/`dateString`) returns dates in "shortened" ISO 8601 (YYYY-MM-DD), as the string does not include time/timezone information.

The unintuitive indexing has caused problems in the past, but a working version was released with https://github.com/icssc/AntAlmanac/pull/944
```ts
for (const [i, date] of enrollmentHistory.dates.entries()) {
       const d = new Date(date);
       const formattedDate = `${d.getMonth() + 1}/${d.getDate()}/${d.getFullYear()}`;
```

The `date` was passed directly into the Date constructor, but at the time Safari strictly followed ES5, which meant returning `NaN` for any date string that didn't conform perfectly to the full ISO 8601 string. This was later fixed.

In the meantime, the "solution" AntAlmanac implemented for Safari compatibility was splitting the day, month, and year before merging them back again (https://github.com/icssc/AntAlmanac/pull/987).
```ts
for (const [i, dateString] of enrollmentHistory.dates.entries()) {
      const [day = '', month = '', year = ''] = dateString.split('-');
      const date = new Date(Number(day), Number(month), Number(year));
      const formattedDate = `${date.getMonth()}/${date.getDate() - 1}/${date.getFullYear()}`;
```

This doesn't work:
`day` actually represents the year and `year` represents the day, but the code seems to just handle it. 
`2025-11-26` gets parsed into `new Date(2025, 11, 26)` which represents December 26, 2025.
The formatted date will represent it as `11/25/2025`.
It didn't matter that the month was wrong because the number from `getMonth` was the same number as what was entered, except for December.

`2025-12-26` becomes January 26, 2026 (`new Date(2025, 12, 26)`) because of rollover.
This resembles a `month % 12` because it is similar, which is why that was a question that was asked in https://github.com/icssc/AntAlmanac/pull/1189, attempting to fix this problem for good.
```ts
for (const [i, dateString] of enrollmentHistory.dates.entries()) {
       const [day = '', month = '', year = ''] = dateString.split('-');
       const date = new Date(Number(day), Number(month), Number(year));
       const formattedDate = `${date.getMonth() + 1}/${date.getDate()}/${date.getFullYear()}`;
```
This might seem correct, it is the same `formattedDate` string that the previous working solution used, but there is still this lingering bad `Date` construction from before.

---

I used the word "opposite" because under this parsing method, I thought that
```ts
const formattedDate = `${date.getMonth()}/${date.getDate()}/${date.getFullYear()}`;
```
might actually "work", even though it is not optimal, because it is just taking the wrong numbers given to it.
This code actually still doesn't work for the reason I described above, my PR does though.

KevinWu098:
Got it @dsnsgithub -- thanks for the really insightful, detailed write up here

KevinWu098:
Merge at will!

PR #1348, opened 2025-11-24
Endorse PeterPortal

dsnsgithub:
## Summary

Draft PR because these PeterPortal images should be updated,

Currently waiting on https://github.com/icssc/peterportal-client/pull/865.

Old:
image

New:
image

## Test Plan

If we are thinking about potential migration problems, we could unhide the help box even if it was dismissed before, since there is a chance that even if the user does not care about the sample program, they still might be interested in PeterPortal.


dsnsgithub:
Superseded by https://github.com/icssc/AntAlmanac/pull/1377

PR #1349, opened 2025-11-24
feat: Next PWA

xgraceyan:
## Summary
Set up Next.js PWA for AntAlmanac.
Updated `app/manifest.ts` according to [PWA builder report card](https://www.pwabuilder.com/reportcard?site=https://antalmanac.com), addresses at least all warnings and adds other relevant fields
Added safe areas and other visual changes to make it look better on mobile.
mobile_calendar

Also i have no idea why pnpm-lock has 6k line diff lol

## Test Plan
Use either an actual phone or Simulator from XCode tools to navigate to the deployment and add it to home screen as PWA. Test that safe areas are implemented and nothing is being visually blocked, and that the PWA functions as expected and there are no visual issues

I have no idea how this looks on an android phone, if we were planning to also put it on the play store..

## Issues

Closes #

## Future Followup
We should take nicer screenshots with graphics if they are going to be on our App Store listing


KevinWu098:
Serwist looks really cool! Excited for this PWA :))

apps/antalmanac/src/components/Header/Header.tsx
```diff
@@ -44,42 +44,45 @@ export function Header() {
...
-            color="primary"
-            sx={{
-                height: 52,
-                padding: 1,
-                boxShadow: 'none',
-                backgroundColor: BLUE,
-            }}
-        >
-            <Box
+        <>
```
KevinWu098: question: is this fragment necessary?

apps/antalmanac/src/components/Header/Header.tsx
```diff
@@ -44,42 +44,45 @@ export function Header() {
...
+            <AppBar
```
KevinWu098, line 86: Image

Looks like the header elements aren't center aligned

apps/antalmanac/next-env.d.ts
KevinWu098, line 1: issue: should this be in the dif?

KevinWu098 (changes requested):
Looks really good, tested on my iPhone! Some small questions.

Also, I notice that we use svh, safe top, safe bottom — as a sanity check, could you explain (in a github comment, no need to write it into the codebase) why each of those styling elements was added? Is it possible to make it more clear/easier to figure out where we apply PWA-specific styles? My concern is long term maintainability and the brittleness of our styling

xgraceyan:
I'm not that great at CSS so if there is a better solution proposed I'd be happy to implement that instead. I agree the styling is pretty weird lol, I've been trying ways to make it simpler and checking if it looks right on my phone but this is it for now

`Header.tsx` uses safe top to make room for the notch/whatever else is at the top of the phone
`Home.tsx` sets height as svh (content) + safe top to account for the header
`SettingsMenu.tsx` has safe top/bottom padding
`HelpMenu.tsx` also has room for safe bottom
`ScheduleManagementTabs.tsx` has safe bottom padding too -- tbh we couldve just done this in `Home` or something but I figured it would look a little nicer if the tabs looked like they were extending to the bottom of the screen instead of ending and then having a more obvious darker space at the end

This is a bit brittle, but for what I have right now the majority of future changes will probably be in the "main content" so no more PWA specific styles need to be applied there. Anything outside of that, ex help menu, settings drawer will need to account for safe areas.

KevinWu098 (dismissed):

apps/antalmanac/package.json
```diff
@@ -61,7 +61,9 @@
         "leaflet-routing-machine": "^3.2.12",
         "leaflet.locatecontrol": "^0.79.0",
         "moment": "^2.29.4",
+        "moment-timezone": "^0.5.43",
```
KevinWu098: nit: what is this dependency? is this extraneous?

xgraceyan: Uhh yeah mightve been and I kept it during merge conflict just to be safe. I don't see it in main, so i'll just delete it 😛

KevinWu098 (dismissed):

alexespejo (approved):
LGTM :p 🚀

PR #1350, opened 2025-11-25
Feat: Redo Button

### LINKED ISSUE #1152: Redo functionality
## Description
- There's an undo button, which can also be activated with |ctrl|+|z|, but there doesn't seem to be redo functionality.
- I vaguely remember this having been possible, but maybe I was imagining it.

AureliaSindhu:
## Summary

This PR implements redo functionality to complement the existing undo feature in AntAlmanac. The implementation adds:

- **Redo button UI**: A new redo button in the calendar toolbar next to the existing undo button
- **Redo state management**: Added `futureStates` stack in the `Schedules` store to track states that can be redone
- **Redo action handling**: Implemented `advanceState()` method in `Schedules` and `redoAction()` in `AppStore` to handle redo operations
- **Analytics tracking**: Added analytics event tracking for redo actions
- **Action type updates**: Refactored `UndoAction` to `UndoRedoAction` with a `direction` field to support both undo and redo in the action history

## Test Plan
- Add a course to the schedule
- Click undo button - the course should be removed
- Click redo button - the course should be added back
- Verify the redo button is disabled when there's nothing to redo

## Issues
- Fixed old issue (old branch was removed)  
- Previously, only undo functionality existed. Users could undo actions but had no way to redo them if they undid something by mistake.
- The undo/redo system now properly maintains both undo and redo history stacks.

Closes #1152

### DIFF
--- a/apps/antalmanac/src/App.css
+++ b/apps/antalmanac/src/App.css
@@ -21,3 +21,10 @@
         scrollbar-width: none;
     }
 }
+
+/* Ensure snackbars sit above the mobile navigation tabs */
+@media (max-width: 600px) {
+    .notification-snackbar-container {
+        padding-bottom: calc(env(safe-area-inset-bottom, 0px) + 64px) !important;
+    }
+}
--- a/apps/antalmanac/src/App.tsx
+++ b/apps/antalmanac/src/App.tsx
@@ -118,7 +118,7 @@ export default function App() {
                             }),
                         }}
                     >
-                        <SnackbarProvider>
+                        <SnackbarProvider classes={{ containerRoot: 'notification-snackbar-container' }}>
                             <RouterProvider router={ROUTER} />
                         </SnackbarProvider>
                     </TourProvider>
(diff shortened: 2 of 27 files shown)


apps/antalmanac/src/App.css
```diff
@@ -20,4 +20,11 @@
     * {
         scrollbar-width: none;
     }
+}
+
+/* Ensure snackbars sit above the mobile navigation tabs */
```
alexespejo, line 25: Can you resolve the git diff by merging in main. It seems like you branched off of your mobile UI branch :)

alexespejo (changes requested):
Sorry I forgot to hit "submit." Behaves well is there a way to support CTRL + SHIFT + Z since we do support CTRL + Z

alexespejo:
The extra button does create this overflow issue with mobile view. Maybe this PR should handle this. 

Maybe you can hide these buttons behind a drop down like this: 

image

image

apps/antalmanac/src/App.css
AureliaSindhu: oki!

AureliaSindhu:
updates: 
- redo shortcut (command/ ctrl + shift + z) 
- mobileview navbar stacked (fixed overflow issue)

alexespejo (approved):
Looks good

Small detail on mobile. Maybe we can consider putting the undo and redo buttons on the outside of the menu since the user could interact with them the most and it would be awkward to have to open the menu every time. Your call!

image

alexespejo (approved):
Type

PR #1351, opened 2025-11-25
Feat: Add Missing Sections Warning

### LINKED ISSUE #1330: Required sections missing warning
We should add a gentle warning system when a user adds a course but hasn’t selected all the required sections. For example, a user has added a lecture but not the corresponding required discussions or labs.

I was thinking some sort of mild “!” sign or highlighting of the course with missing required sections on the Added Pane.

Image
e.g. this ICS53 is missing a required Lab section

This would be preferred over doing it in the Search Pane, as it would be annoying to have a flickering warning appear while a user is quickly clicking around to see which discussion or lab fits best.

A simple tooltip or light yellow/orange highlight with text like “This course is missing a required section.” would be enough for the user to realize their mistake and go back to the Search Pane.

calebwongg:
## Summary
Added a warning displaying the missing sections for a course on the added courses page.

When a user selects classes for a course, the required sections for a course are checked (e.g., lecture, lab, discussion) have been added. If any required section type is missing, the UI displays a warning so the user can correct their schedule.

## Test Plan

- Add SOME sections of a course to schedule, and check the added courses tab to verify there is a warning for whichever section is missing
- Add all sections of a course to a schedule. In the added courses tab, ensure there are no warnings, and then delete one of the sections to verify a warning pops up. Add the section back to verify the warning disappears 
- Check UI on both light mode and dark mode
## Issues

Closes #1330

## Before
image

## After
image

## Future Followup 

- Could add a hyperlink to the Alert banner that directs the user to the scheduling page for whichever course they have a missing section 
- Could add an alert for if a user schedules two different sections in two different categories (e.g. the user schedules a lecture 'A' but a discussion 'B')

### DIFF
--- a/apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
+++ b/apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
@@ -421,8 +421,20 @@ function AddedSectionsGrid() {
                                 missingLabels.push('Lecture');
                             } else if (section === 'sem') {
                                 missingLabels.push('Seminar');
-                            } else if (section === 'rec') {
-                                missingLabels.push('Recitation');
+                            } else if (section === 'res') {
+                                missingLabels.push('Research');
+                            } else if (section == 'qiz') {
+                                missingLabels.push('Quiz');
+                            } else if (section == 'tap') {
+                                missingLabels.push('Tutorial Assistance Program');
+                            } else if (section == 'col') {
+                                missingLabels.push('Colloquium');
+                            } else if (section == 'act') {
+                                missingLabels.push('Activity');
+                            } else if (section == 'stu') {
+                                missingLabels.push('Studio');
+                            } else if (section == 'tut') {
+                                missingLabels.push('Tutorial');
                             }
                         }
 
--- a/apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
+++ b/apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
@@ -1,5 +1,5 @@
 import { Box, Chip, Paper, SxProps, TextField, Tooltip, Typography } from '@mui/material';
-import { AACourse, WebsocSectionType } from '@packages/antalmanac-types';
+import { AACourse } from '@packages/antalmanac-types';
 import { usePostHog } from 'posthog-js/react';
 import { useCallback, useEffect, useMemo, useState } from 'react';
 
@@ -7,6 +7,7 @@ import { ColumnToggleDropdown } from '../CoursePane/CoursePaneButtonRow';
 import SectionTableLazyWrapper from '../SectionTable/SectionTableLazyWrapper';
 
 import CustomEventDetailView from './CustomEventDetailView';
+import { getMissingSections } from './getMissingSections';
 
 import { updateScheduleNote } from '$actions/AppStoreActions';
 import { ClearScheduleButton } from '$components/buttons/Clear';
@@ -30,26 +31,11 @@ const buttonSx: SxProps = {
     pointerEvents: 'auto',
 };
 
-interface CourseWithTerm extends AACourse {
+export interface CourseWithTerm extends AACourse {
     term: string;
 }
 
 const NOTE_MAX_LEN = 5000;
-
-//Checks added courses for missing sections
-const checkCompleteSections = (userCourses: CourseWithTerm): string[] => {
-    //Get required types from stored section types
-    const requiredTypes = userCourses.sectionTypes
-        ? new Set([...userCourses.sectionTypes].map((t) => t.toLowerCase()))
-        : new Set<WebsocSectionType>();
-
-    //Get the section types the user has added
-    const userTypes = new Set(userCourses.sections.map((section) => section.sectionType.trim().toLowerCase()));
-    //Compare types the user added with the required types
-    const missingTypes = [...requiredTypes].filter((type) => !userTypes.has(type));
-
-    return missingTypes;
-};
 
 function getCourses() {
     const currentCourses = AppStore.schedule.getCurrentCourses();
@@ -370,35 +356,7 @@ function AddedSectionsGrid() {
                 {courses.length < 1 ? NoCoursesBox : null}
                 <Box display="flex" flexDirection="column" gap={1}>
                     {courses.map((course) => {
-                        const missing = checkCompleteSections(course);
-                        const missingSections = missing.map((section) => {
-                            switch (section) {
-                                case 'dis':
-                                    return 'Discussion';
-                                case 'lab':
-                                    return 'Lab';
-                                case 'lec':
-                                    return 'Lecture';
-                                case 'sem':
-                                    return 'Seminar';
-                                case 'res':
-                                    return 'Research';
-                                case 'qiz':
-                                    return 'Quiz';
-                                case 'tap':
-                                    return 'Tutorial Assistance Program';
-                                case 'col':
-                                    return 'Colloquium';
-                                case 'act':
-                                    return 'Activity';
-                                case 'stu':
-                                    return 'Studio';
-                                case 'tut':
-                                    return 'Tutorial';
-                                default:
-                                    return section;
-                            }
-                        });
+                        const missingSections = getMissingSections(course);
 
                         return (
                             <Box key={course.deptCode + course.courseNumber + course.courseTitle}>
--- a/apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
+++ b/apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
@@ -38,40 +38,21 @@ const NOTE_MAX_LEN = 5000;
 
 //Checks added courses for missing sections
 const checkCompleteSections = (userCourses: CourseWithTerm): string[] => {
-    try {
-        console.log('aaaaaa user course section Types: ', userCourses.sectionTypes);
-        //Get required types from stored section types
-        const requiredTypes = userCourses.sectionTypes
-            ? new Set([...userCourses.sectionTypes].map((t) => t.toLowerCase()))
-            : new Set<string>();
-        console.log('Required Types:', requiredTypes);
+    //Get required types from stored section types
+    const requiredTypes = userCourses.sectionTypes
+        ? new Set([...userCourses.sectionTypes].map((t) => t.toLowerCase()))
+        : new Set<string>();
 
-        //Get the section types the user has added
-        const userTypes = new Set(userCourses.sections.map((section) => section.sectionType.trim().toLowerCase()));
-        console.log('User Types:', userTypes);
-        //Compare types the user added with the required types
-        const missingTypes = [...requiredTypes].filter((type) => !userTypes.has(type));
-        console.log('Missing Types:', missingTypes);
+    //Get the section types the user has added
+    const userTypes = new Set(userCourses.sections.map((section) => section.sectionType.trim().toLowerCase()));
+    //Compare types the user added with the required types
+    const missingTypes = [...requiredTypes].filter((type) => !userTypes.has(type));
 
-        return missingTypes;
-    } catch (error) {
-        console.error(error);
-        return [];
-    }
+    return missingTypes;
 };
 
 function getCourses() {
     const currentCourses = AppStore.schedule.getCurrentCourses();
-    console.log('Current Courses:', currentCourses);
-    console.log('Current Courses Length:', currentCourses.length);
-    console.log(
-        'Current Courses Section Types:',
-        currentCourses.map((c) => c.sectionTypes)
-    );
-    console.log(
-        'Current Courses Section Types Details:',
-        currentCourses.map((c) => [...c.sectionTypes])
-    );
 
     const formattedCourses: CourseWithTerm[] = [];
 
--- a/apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
+++ b/apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
@@ -408,31 +408,31 @@ function AddedSectionsGrid() {
                     {courses.map((course) => {
                         const courseKey = `${course.deptCode}${course.courseNumber}`;
                         const missing = missingTypes[courseKey] || [];
-                        const missingLabels = [];
+                        const missingSections = [];
 
                         for (const section of missing) {
                             if (section === 'dis') {
-                                missingLabels.push('Discussion');
+                                missingSections.push('Discussion');
                             } else if (section === 'lab') {
-                                missingLabels.push('Lab');
+                                missingSections.push('Lab');
                             } else if (section === 'lec') {
-                                missingLabels.push('Lecture');
+                                missingSections.push('Lecture');
                             } else if (section === 'sem') {
-                                missingLabels.push('Seminar');
+                                missingSections.push('Seminar');
                             } else if (section === 'res') {
-                                missingLabels.push('Research');
+                                missingSections.push('Research');
                             } else if (section == 'qiz') {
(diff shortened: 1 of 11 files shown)


apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
```diff
@@ -155,6 +156,12 @@ function SectionTable(props: SectionTableProps) {
                 />
             </Box>
 
+            {missingSections && missingSections.length > 0 && (
+                <Alert severity="warning" sx={{ mb: 1 }}>
+                    Missing required sections: {missingSections.join(', ')}
+                </Alert>
```
KevinWu098, line 161: issue: this is a styling concern, but `<Alert />` on dark mode looks pretty gnarly... 

suggestion: when working through feature work, definitely make sure to test on both light and dark mode!

apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
```diff
@@ -155,6 +156,12 @@ function SectionTable(props: SectionTableProps) {
                 />
             </Box>
 
+            {missingSections && missingSections.length > 0 && (
```
KevinWu098: nit: `missingSections?.length` expresses the same condition

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -385,7 +470,7 @@ export function AddedCoursePane() {
             setSkeletonMode(AppStore.getSkeletonMode());
         };
 
-        console.log('Opened added ourse');
+        console.log('Opened added course');
```
KevinWu098: chore: remove console.log

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -362,6 +446,7 @@ function AddedSectionsGrid() {
                                     allowHighlight={false}
                                     analyticsCategory={analyticsEnum.addedClasses}
                                     scheduleNames={scheduleNames}
+                                    missingSections={missingLabels}
```
KevinWu098: nit: if it's convenient, we should name variables the same as props (and vice versa).

apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
```diff
@@ -1,5 +1,6 @@
 import { Assessment, ShowChart as ShowChartIcon } from '@mui/icons-material';
 import { Box, Paper, Table, TableCell, TableContainer, TableHead, TableRow, useMediaQuery } from '@mui/material';
+import { Alert } from '@mui/material';
```
KevinWu098: issue: your linter should be warning you that you're importing twice from the same source

suggestion: move Alert into the correct import line and figure out what's up with your linter

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -317,6 +352,25 @@ function AddedSectionsGrid() {
...
+                const missingSections = await checkCompleteSections(course);
+                missing[courseKey] = missingSections;
+            }
+
+            setMissingTypes(missing);
+        };
+        if (courses.length > 0) {
+            checkCourses();
+        }
+    }, [courses]);
```
KevinWu098: issue: there's a couple things that are tricky with the approach here. One glaring issue is that this useEffect runs every time `courses` mutates. In other words, anytime any edit is made, we, via `checkCompleteSections`, query WebSOC for _every_ single course that has been added.

suggestion: utilize a different approach to determine if sections are missing data (see my other comments)

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -354,6 +408,36 @@ function AddedSectionsGrid() {
...
+                            } else if (section == 'col') {
+                                missingLabels.push('Colloquium');
+                            } else if (section == 'act') {
+                                missingLabels.push('Activity');
+                            } else if (section == 'stu') {
+                                missingLabels.push('Studio');
+                            } else if (section == 'tut') {
+                                missingLabels.push('Tutorial');
+                            }
+                        }
```
KevinWu098: nit: if we're going to write a bunch of if statements, this would be better as a utility. the problem here is in readability of the component. 

suggestion: as much as possible, we should not add logic in the "render component" part of the JSX (the stuff in the `return`)

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -290,6 +324,7 @@ function AddedSectionsGrid() {
     const [courses, setCourses] = useState(getCourses());
     const [scheduleNames, setScheduleNames] = useState(AppStore.getScheduleNames());
     const [scheduleIndex, setScheduleIndex] = useState(AppStore.getCurrentScheduleIndex());
+    const [missingTypes, setMissingTypes] = useState<Record<string, string[]>>({});
```
KevinWu098: issue: see below/my other comments, but we shouldn't be storing missingTypes at a "global" (relative to this component) level.

suggestion: one good way to approach the React model of components is to really embrace and internalize that React is: `ui = fn(state)`. Which is to say, the UI is derived from the state. While this is rather plain to say out loud, here's how it can be applied to guide your approach: 

If we have the state _too_ high in the component tree, then more UI is dependent on the state. Since missingTypes is specific to each course (the UI of ICS 32 doesn't care what the UI of ICS 46 is), it would make more sense for the `state` which determines the UI to be "closer" to the ui itself. When ICS 46 changes, ICS 32 would prefer not to rerender, since it is not dependent on ICS 46.

In short, even if we decided to proceed with your present approach, we would want to refactor missingTypes to be calculated for each individual course, rather than altogether.

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -49,6 +81,8 @@ function getCourses() {
                 needleCourse.courseTitle === course.courseTitle
         );
 
+        //If the course already exists (if the user is already taking a section of this course),
+        //then group them together so they display on top of each other
```
KevinWu098: chore: seemingly extraneous / newly added comment

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -36,6 +37,37 @@ interface CourseWithTerm extends AACourse {
...
+//Checks added courses for missing sections
+const checkCompleteSections = async (userCourses: CourseWithTerm): Promise<string[]> => {
+    try {
+        const websocParams = {
+            department: userCourses.deptCode,
+            courseNumber: userCourses.courseNumber,
+            term: userCourses.term,
+        };
+
+        const fullCourseData = await WebSOC.query(websocParams);
```
KevinWu098: issue: I've discussed this across a few of the comments below, but this approach is not very performant for our goal.

For one, every time `courses` changes, we trigger this method which queries for EVERY course the user has added. This, intuitively, is inefficient since ICS 32 does not care about changes to ICS 46. So, we should not query for it again.

KevinWu098 (changes requested):
Congrats on your first feature-y PR! This will definitely be a really nice QOL boost for users who are newer to AA and Websoc.

That being said, you have the unfortunate fate of having a PR reviewed by a horribly picky reviewer (me). 

I go into more detail across the comments which I have left below, but the present approach is one that we would like to refactor to be more performant and to be more idiomatic. The overarching challenge presently is that the `useEffect` implementation makes more requests than necessary and that there is another approach which does not require any websoc requests!

Right now, you are making additional websoc requests to check for section types. However, AA already fetches course data from WebSOC on page load. This means that, when we load in, we _already_ have the information that you're querying for in this PR. I would recommend taking a look at the logic of `flattenSOCObject`, where we extend a `WebsocSection` into an `AASection` by giving it additional properties (e.g. color). 

```js
        school.departments.forEach((dept) => {
            accumulator.push(dept);

            dept.courses.forEach((course) => {
                    (section as AASection).color = courseColors[section.sectionCode];
                }
                accumulator.push(course as AACourse);
            });
        });
```

Additionally, make use of our Websoc types! `WebsocSectionType`

packages/types/src/websoc.ts
```diff
@@ -1,13 +1,16 @@
-import { WebsocSection, WebsocCourse } from '@packages/anteater-api-types';
+import { WebsocSection, WebsocCourse, WebsocSectionType } from '@packages/anteater-api-types';
 
 type AASectionExtendedProperties = {
     color: string;
 };
 
+export type { WebsocSectionType };
```
KevinWu098: question: is there a reason why we're reexporting the type and not just importing `WebsocSectionType` directly where needed? in most cases, we do not want to pipe around imports

apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
```diff
@@ -155,6 +155,11 @@ function SectionTable(props: SectionTableProps) {
                 />
             </Box>
 
+            {missingSections?.length > 0 && (
+                <Alert severity="warning" sx={{ mb: 1, backgroundColor: '#fff4e5ff', color: '#856404ff' }}>
```
KevinWu098: issue: we shouldn't hardcode styles like this. the reason being that it's much harder maintain in the future — what I would advise is to consider updating the MUI theme in `Theme.tsx` (ideally using MUI color tokens: https://mui.com/material-ui/customization/palette/)

KevinWu098: thought: another, perhaps more idiomatic, way of doing this would be to first submit a separate PR updating the styling of the Alert, _then_ updating this branch. this would be better for separating concerns, but it will delay this PR for however long it takes you to restyle the alert (which shouldn't take very long, fwiw)

apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
```diff
@@ -1,5 +1,5 @@
 import { Assessment, ShowChart as ShowChartIcon } from '@mui/icons-material';
-import { Box, Paper, Table, TableCell, TableContainer, TableHead, TableRow, useMediaQuery } from '@mui/material';
+import { Box, Paper, Table, TableCell, TableContainer, TableHead, TableRow, useMediaQuery, Alert } from '@mui/material';
```
KevinWu098: nit: something must be wrong with AntAlmanac's husky settings... but anyways `Alert` should be the first import in this list (it's alphabetical)

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -36,6 +36,21 @@ interface CourseWithTerm extends AACourse {
 
 const NOTE_MAX_LEN = 5000;
 
+//Checks added courses for missing sections
```
KevinWu098: nit: comments should be reserved for cases where code is not immediately readable. in other words, your code should, ideally, always be very readable and self-documenting. saying that this method "checks for missing sections" is redundant if the code itself is descriptive

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -36,6 +36,21 @@ interface CourseWithTerm extends AACourse {
...
+const checkCompleteSections = (userCourses: CourseWithTerm): string[] => {
+    //Get required types from stored section types
+    const requiredTypes = userCourses.sectionTypes
```
KevinWu098: issue: correct me if I'm wrong, but isn't `userCourses.sectionTypes` already a `Set`?

```
type AACourseExtendedProperties = {
    sections: AASection[];
    sectionTypes: Set<WebsocSectionType>;
};
```

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -36,6 +36,21 @@ interface CourseWithTerm extends AACourse {
 
 const NOTE_MAX_LEN = 5000;
 
+//Checks added courses for missing sections
+const checkCompleteSections = (userCourses: CourseWithTerm): string[] => {
```
KevinWu098: nit: the name of this is not super clear, right? since the result of the function is to return the missingSections? `getMissingSections` might be more apt. You may also want to move this logic into its own component, e.g. <MissingRequiredSectionsAlert />

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -354,6 +370,36 @@ function AddedSectionsGrid() {
                 {courses.length < 1 ? NoCoursesBox : null}
                 <Box display="flex" flexDirection="column" gap={1}>
                     {courses.map((course) => {
+                        const missing = checkCompleteSections(course);
+                        const missingSections = missing.map((section) => {
+                            switch (section) {
```
KevinWu098: thought: this would be better in its own component

see above comment

KevinWu098 (changes requested):
Good work! A couple more changes and I think we'll be good to go

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -30,7 +31,7 @@ const buttonSx: SxProps = {
     pointerEvents: 'auto',
 };
 
-interface CourseWithTerm extends AACourse {
+export interface CourseWithTerm extends AACourse {
```
calebwongg, line 34: Not sure if its okay to export this type here or if it should be exported from /types/src/websoc.ts

apps/antalmanac/src/components/RightPane/AddedCourses/getMissingSections.ts
```diff
@@ -0,0 +1,41 @@
+import { CourseWithTerm } from './AddedCoursePane';
+
+export const getMissingSections = (userCourses: CourseWithTerm): string[] => {
+    //Get required types from stored section types
```
KevinWu098: nit: too many comments

apps/antalmanac/src/components/RightPane/AddedCourses/getMissingSections.ts
```diff
@@ -0,0 +1,41 @@
+import { CourseWithTerm } from './AddedCoursePane';
```
KevinWu098: chore: please use absolute imports e.g. `@directory/file`, not the relative ones e.g. `../../../`

apps/antalmanac/src/components/RightPane/AddedCourses/getMissingSections.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export const getMissingSections = (userCourses: CourseWithTerm): string[] => {
+    //Get required types from stored section types
+    const requiredTypes = userCourses.sectionTypes;
+    //Get the section types the user has added
```
KevinWu098: repeat: comment

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
KevinWu098: This is good!

KevinWu098 (changes requested):
A little bit more nits, sorry

KevinWu098 (approved):
LGTM 🔥

PR #1353, opened 2025-12-03
AANTS Frontend Changes

IsaacNguyen:
## Summary
AANTS frontend changes (notifications menu at top, notification menu for each course, and unsubscribe page)

## Media/Screenshots

image
image
image
image

## Test Plan
- [x] Still connects properly with backend (propagates notifications to db)
- [x] Works with current Auth

## Notes
- Unsubscribe page is highkey chopped, definitely gonna need a redesign for that
- The notifications button removes the button that allows you to change the color of a course
   - Real estate on the course rows is limited
   - Kevin made it this way but I think we should find a way to do both without making it look clunky


apps/antalmanac/src/components/RightPane/AddedCourses/Notifications/NotificationsDialog.tsx
```diff
@@ -0,0 +1,83 @@
...
+                    onClick={handleOpen}
+                    size="small"
+                    disabled={disabled}
+                >
+                    <Notifications />
+                </IconButton>
+            </Tooltip>
+
+            <Dialog open={open} onClose={handleClose} fullWidth>
+                <DialogTitle>Manage Notifications</DialogTitle>
```
alexespejo, line 74: UX: It takes a second for the notifications to propagate so I would have a small loading animation . Additionally, if the user doesn't have any notifications set I would have a "You don't have any notifications enabled" type of message. 

image

PR #1352, opened 2025-12-03
fix: Persist Search Input on Loss of Focus

dsnsgithub:
## Summary
Reopening https://github.com/icssc/AntAlmanac/pull/1302.
Branch of the PR was deleted, created a new branch.

> When using the fuzzy search bar, clicking away would cause the input to clear.
> 
> Steps to reproduce:
> 
> 1. Select "Search"
> 2. Select "Quick Search"
> 3. In the search bar (Where it says 'Search for courses, departments, GEs...'), enter something
> 4. Click anywhere outside of the search bar
> 5. The input disappears

## Test Plan

> Steps to test:
> 
> 1. Select Search
> 2. Select Quick Search
> 3. In the search bar, enter something
> 4. Click anywhere outside the search bar
> 5. The input stays, and the previous search automatically comes up

## Issues

Closes #1301

PR #1355, opened 2025-12-03
fix: Prevent Old Search Results on Loss of Focus

dsnsgithub:
## Summary
Old search results persist after blur.

Regression created by my PR #1352, apologies once more.

## Test Plan
- Query two characters in search, load results
- Delete query
- Focus search box
- Old search results still appear


KevinWu098 (approved):
Haven't reviewed it on the deployment -- but I trust you 💖

PR #1357, opened 2025-12-03
Schedule Link Sharing

### LINKED ISSUE #911: Suggestion: View Friends' Schedules
# Hello AntAlmanac Developers.

## Recently I've come across an epiphany. 

When signing up for classes, a lot of the times you want to see what classes your friends are taking. Wouldn't it be cool to have, for example, hover over a class, and see if you have any friends taking that class, or just a friends list to show their respective class list?

Please let me know your thoughts.

### LINKED ISSUE #1346: Sharable Course Schedule Links

brelieu05:
## Summary
Implemented Schedule Link Sharing

- User can share their schedule with a link
  - Currently, the sharing param is the db schedules UUID
- When the user opens the share link, they see the schedule, but it's only temporary and if they refresh it will disappear
  - User can press the normal save button to add it as a permanent schedule
  - It loads the user's schedule first -> adds + switches to shared schedule
- If invalid share link, snackbar will notify user so
 

image
image
image
image
image

## Issues
How should I handle the multiple snackbars? Just one after everything loads or keep two to indicate when account schedules have loaded and when shared schedule has loaded? since they're separate things

Closes #1346


apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -262,6 +262,38 @@ export const importScheduleWithUsername = async (username: string) => {
...
+
+        scheduleComponentsToggleStore.setState({ openScheduleSelect: true, openLoadingSchedule: false });
+
+        changeCurrentSchedule(currentSchedules.scheduleIndex);
+
+        return { imported: true, error: null };
+    }
+
+    scheduleComponentsToggleStore.setState({ openLoadingSchedule: false });
+    return { imported: false, error: new Error('Failed to load shared schedule') };
```
KevinWu098: nit: this would be more readable inverted

if (!isScheduleLoaded) {
	blah blah
	blah blah
}

// success logic

apps/antalmanac/src/components/dialogs/ShareSchedule.tsx
```diff
@@ -0,0 +1,139 @@
+import { ContentCopy, Check } from '@mui/icons-material';
```
alexespejo: I think the "Share" button we use is typically associated with a dialog like this but to the average user the details of the link are unnecessary. I would just default to a copy link icon

apps/antalmanac/src/components/Calendar/Toolbar/ScheduleSelect/ScheduleSelect.tsx
```diff
@@ -221,6 +222,7 @@ export function SelectSchedulePopover() {
                                         </Box>
 
                                         <Box display="flex" alignItems="center" gap={0.5}>
+                                            <ShareScheduleButton index={index} disabled={skeletonMode} />
```
alexespejo: Unsaved schedules won't have an ID so won't be shareable and you'll get an error. There are different ways you can approach it. 

1) You could just throw a warning that you need to save the schedule in order to save it 

2) Restructure the way you format the URL to not depend on the schedule id 
- You could scrape the section codes and just parse them in the client when you load the shared schedule. Might be a little bloated and buggy
- Could add a nice user experience by not having that small barrier (though not sure if it's worth the engineering time)

apps/antalmanac/src/backend/lib/rds.ts
```diff
@@ -308,6 +308,67 @@ export class RDS {
         await tx.insert(customEvents).values(dbCustomEvents);
     }
 
+    /**
```
alexespejo, line 357: Instead of fetching the schedule ID you can include it in the payload when we initially fetch the schedule data scheout the `Schedules` type in [`schedule.ts`](https://github.com/icssc/AntAlmanac/blob/main/packages/types/src/schedule.ts)

KevinWu098 (dismissed):
I haven't read the code deeply, but I'll share a few thoughts/feedback before doing a deeper read later

1. Feature is definitely cool
2. the mergeShortCourseSchedules breaks because it adds a `(1)` so adding a shared schedule "Foo" becomes "Foo (1)" if the receiving user has another schedule named "Foo"
3. Anytime we ask someone to sign in, we should make it easy for them — give them a button
4. We should be more clear that this is a static copy
5. Relating to #4... do we have to be a static copy?  🌚

alexespejo (changes requested):

brelieu05:
Created new and separate Shared Schedule page
- User can add to schedules list
- Fixed shared schedule naming
- Notify user if link is invalid or expired
- Links will expire if the original sender resaves and thus changes id

alexespejo (changes requested):
1) I hate to be that guy but is there a way for you to consolidate the git diff so it only shows the file changes you made. If it's too much of a hassle it's whatevs we'll figure it out. 

2) Also can you restore the merge conflicts (resolving them might resolve the git diff issue) and may fix the staging deployment which seems to be breaking

image

brelieu05:
holay molay my bad

brelieu05:
With some cursor shenanigans, I think I was able to fix it

Also when are we playing aram

alexespejo:
Resolve merge conflict so the deployment can run, gracias

brelieu05:
I finally figured out what was causing all tests to not run, it has been fixed and staging works 🫶

alexespejo (changes requested):
Haven't looked at code too indepth yet but some bugs 🐞:

1) Unsaved changes 
Repro: Opening a shared link -> editing the schedule (removing/adding classes) -> "refresh" -> "accept unsaved changes" -> the changes persist 

The user either should be locked out of changing the shared schedule or lock unsaved changes from triggering 

2) Design

    1) Small detail I think it would be nice to preview the schedule name in the drop down button so that there isn't just some empty space 
    2) Maybe bold the name of the schedule being imported (e.g. "Viewing Shared Schedule: **2026 Winter**)
image



3) Sharing a deleted schedule id "breaks" the user's data 

Repro: have two users open (1 and 2) -> (1) copy share link -> (1) edit and save the schedule -> (2) open the shared link -> recieve error message -> hit "go home" -> the user's data should appear empty 

Refreshing the page will fix the schedule but it would be offputting the the user to think their data has been wiped

apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
```diff
@@ -93,6 +94,8 @@ export const CourseCalendarEvent = ({ selectedEvent, scheduleNames, closePopover
     const paperRef = useRef<HTMLDivElement>(null);
     const quickSearch = useQuickSearch();
     const { isMilitaryTime } = useTimeFormatStore();
+    const location = useLocation();
+    const isSharedSchedulePage = location.pathname.startsWith('/share/');
```
alexespejo: chore: these two lines appear 11 times in the code base. I'd be good if there's a way to just make it a utility 

```ts
const location = useLocation();
const isSharedSchedulePage = location.pathname.startsWith('/share/');
 ```

File: apps/antalmanac/src/components/Calendar/Toolbar/ScheduleSelect/schedule-select-buttons/ShareScheduleButton.tsx [hunk 6]
```diff
@@ -0,0 +1,59 @@
...
+}
+
+export function ShareScheduleButton({ index, disabled }: ShareScheduleButtonProps) {
+    const [copied, setCopied] = useState(false);
+    const { sessionIsValid } = useSessionStore();
+    const { enqueueSnackbar } = useSnackbar();
+
+    const handleCopy = useCallback(async () => {
+        if (!sessionIsValid) {
+            enqueueSnackbar('Please sign in to share schedules', { variant: 'error' });
```
alexespejo: chore: we have an `AppStore` utility `AppStore.openSnackbar()I`

alexespejo: thought: if the user is not signed in we can just disable the button with a tooltip instead of providing a notification

apps/antalmanac/src/components/Header/Signin.tsx
```diff
@@ -275,18 +279,35 @@ export const Signin = () => {
...
+                                if (validSession) {
+                                    updateSession(sessionToken);
+                                }
+                            })
+                            .catch((err) => {
+                                console.error('Error validating session:', err);
+                            });
+                    }
+                } else {
+                    // Normal behavior: load schedules
```
alexespejo: nit: comment

apps/antalmanac/src/components/Header/Signin.tsx
```diff
@@ -275,18 +279,35 @@ export const Signin = () => {
             const sessionID = getLocalStorageSessionId();
 
             if (savedUserID != null || sessionID !== null) {
-                void loadScheduleAndSetLoadingAuth(savedUserID ?? '', true);
+                if (isSharedSchedulePage) {
+                    const sessionToken = getLocalStorageSessionId() ?? '';
```
alexespejo: issue: `sessionID` is defined on line 279

apps/antalmanac/src/components/Header/Signin.tsx
```diff
@@ -275,18 +279,35 @@ export const Signin = () => {
             const sessionID = getLocalStorageSessionId();
 
             if (savedUserID != null || sessionID !== null) {
-                void loadScheduleAndSetLoadingAuth(savedUserID ?? '', true);
+                if (isSharedSchedulePage) {
```
alexespejo: nit: since we abstract what the loadSchedule inside of `loadScheduleAndSetLoadingAuth` maybe provide a short comment explaining why we just validate the session and not load the data

apps/antalmanac/src/routes/SharedSchedulePage.tsx
```diff
@@ -0,0 +1,381 @@
...
+                            borderColor: 'divider',
+                        }}
+                    >
+                        <Stack direction="row" justifyContent="space-between" alignItems="center">
+                            <Typography variant="h6" component="h1">
+                                Viewing Shared Schedule: <strong>{scheduleName}</strong>
+                            </Typography>
+                            <Stack direction="row" spacing={1} alignItems="center">
+                                <Button variant="contained" onClick={handleAddToMySchedules}>
+                                    Add to My Schedules
```
alexespejo: issue: responsive design 

image

apps/antalmanac/src/components/dialogs/ShareSchedule.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+    const shareUrl = scheduleId ? `${window.location.origin}/share/${scheduleId}` : '';
+
+    const handleCopy = useCallback(async () => {
+        if (!shareUrl) return;
+
+        try {
+            await navigator.clipboard.writeText(shareUrl);
+            setCopied(true);
+            enqueueSnackbar('Link copied to clipboard!', { variant: 'success' });
```
alexespejo: chore: `AppStore.openSnackbar`

apps/backend/src/routers/userData.ts
```diff
@@ -0,0 +1,240 @@
+import { UserSchema } from '@packages/antalmanac-types';
```
alexespejo: issue: I got hella confused when I saw this. The backend was moved to `apps/antalmanac/src/backend`. I these routes should exist in there

brelieu05: yeah the routes I added are in both backends since I made these changes before the merge. Do you want me to remove the routes I made from the old backend?

apps/antalmanac/src/stores/AppStore.ts
```diff
@@ -418,6 +432,8 @@ class AppStore extends EventEmitter {
 
     loadSkeletonSchedule(savedSchedule: ScheduleSaveState) {
         this.schedule.setSkeletonSchedules(savedSchedule.schedules);
+        // Set the current schedule index to match the saved schedule index
```
alexespejo: nit: remove comments

apps/antalmanac/src/components/ScheduleManagement/ScheduleManagement.tsx
```diff
@@ -50,6 +53,12 @@ export function ScheduleManagement() {
             return;
         }
 
+        // Don't set search tab when on shared schedule page
```
alexespejo: nit: comments

apps/antalmanac/src/actions/ActionTypesStore.ts
```diff
@@ -165,6 +169,12 @@ class ActionTypesStore extends EventEmitter {
     }
 
     async loadScheduleFromUnsavedActions() {
+        // Don't load unsaved actions when viewing a shared schedule (skeleton mode)
```
alexespejo: nit: comment

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -183,6 +191,16 @@ export async function autoSaveSchedule(
             },
         });
 
+        // Update schedule IDs in AppStore after saving
```
alexespejo: comment

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -183,6 +191,16 @@ export async function autoSaveSchedule(
...
+        if (result && 'scheduleIds' in result && Array.isArray(result.scheduleIds)) {
+            const schedules = AppStore.getSchedules();
+            for (let i = 0; i < Math.min(schedules.length, result.scheduleIds.length); i++) {
+                schedules[i].id = result.scheduleIds[i];
+            }
+            // Emit events to notify components of the ID updates
```
alexespejo: comment

apps/antalmanac/src/components/dialogs/ShareSchedule.tsx
```diff
@@ -0,0 +1,123 @@
+import { ContentCopy, Check } from '@mui/icons-material';
```
alexespejo: issue: if unused remove file

alexespejo (changes requested):
Fix comments and I'll approve

apps/antalmanac/src/routes/SharedSchedulePage.tsx
```diff
@@ -0,0 +1,413 @@
+import { Add, Close } from '@mui/icons-material';
+import { useMediaQuery, useTheme, Stack, Alert, Button, Box, Typography, IconButton } from '@mui/material';
+import { LocalizationProvider } from '@mui/x-date-pickers';
+import { AdapterDateFns } from '@mui/x-date-pickers/AdapterDateFnsV2';
+import type { ScheduleSaveState, ShortCourseSchedule, ShortCourse } from '@packages/antalmanac-types';
```
alexespejo: issue: consolidate imports

apps/antalmanac/src/routes/SharedSchedulePage.tsx
```diff
@@ -0,0 +1,413 @@
...
+        };
+
+        loadSharedSchedule();
+
+        return () => {
+            setOpenLoadingSchedule(false);
+        };
+    }, [scheduleId, setOpenLoadingSchedule]);
+
+    const handleExitSharedSchedule = useCallback(async () => {
```
alexespejo: issue: This function repeats a lot of logic with `handleAddToMySchedules` and `handleGoHome`. We should try to containerize our logic into reusable functions.

alexespejo (changes requested):
Couple comments. Approach seems sane.

alexespejo (changes requested):
I believe you have some merge conflicts

brelieu05:
my b, didnt see it until after requesting. im on it rn

brelieu05:
I'm getting issues merging main into my branch, specifically with ESLint regarding AANTS with the error "Unable to resolve path to module 'aws-lambda'". I just pushed the specific files with conflict edits with sharedSchedule.

Currently still figuring out how to fix the merge

alexespejo:
> I'm getting issues merging main into my branch, specifically with ESLint regarding AANTS with the error "Unable to resolve path to module 'aws-lambda'". I just pushed the specific files with conflict edits with sharedSchedule.
> 
> Currently still figuring out how to fix the merge

I believe in you ✊ 

image

brelieu05:
what are all these deployment errors that don't have to do with my sharedSchedule commits

alexespejo:
Oh ts worked lol?

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -137,6 +137,14 @@ export const saveSchedule = async (
                     },
                 });
 
+                if (result && 'scheduleIds' in result && Array.isArray(result.scheduleIds)) {
```
Choollol: Might wanna move this logic into a `setSchedules` function in `AppStore`, especially since it's duplicated in `autoSaveSchedule` too.

Also could consider changing the condition to `Array.isArray(result?.scheduleIds)`, just a little more concise.

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -186,6 +194,14 @@ export async function autoSaveSchedule(providerID: string, options: AutoSaveSche
             },
         });
 
+        if (result && 'scheduleIds' in result && Array.isArray(result.scheduleIds)) {
```
Choollol: See [#r2772150367](https://github.com/icssc/AntAlmanac/pull/1357/changes#r2772150367)

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -204,13 +220,20 @@ export const mergeShortCourseSchedules = (
 ) => {
     const existingScheduleNames = new Set(currentSchedules.map((s: ShortCourseSchedule) => s.scheduleName));
     const cacheSchedule = incomingSchedule.map((schedule: ShortCourseSchedule) => {
-        let scheduleName = schedule.scheduleName;
-        if (existingScheduleNames.has(schedule.scheduleName)) {
-            scheduleName = scheduleName + '(1)';
+        let scheduleName = `${importMessage}${schedule.scheduleName}`;
+        let counter = 1;
```
Choollol: Small suggestion: Could move this logic into a separate util file and use it both here and in `Schedules.ts`. Might be better to use `Set` by default (?), so we could change `getNextScheduleName` to match.

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -283,6 +310,59 @@ export const importScheduleWithUsername = async (username: string) => {
...
+        id: undefined,
+        scheduleName: sharedSchedule.scheduleName,
+        scheduleNote: sharedSchedule.scheduleNote || '',
+        courses: sharedSchedule.courses,
+        customEvents: sharedSchedule.customEvents,
+    };
+
+    const currentSchedules = AppStore.schedule.getScheduleAsSaveState();
+
+    mergeShortCourseSchedules(currentSchedules.schedules, [incomingSchedule], '(shared)-');
```
Choollol: Might want to put the `(shared)-` prefix in a shared constant since it's also used in [`SharedSchedulePage`](https://github.com/icssc/AntAlmanac/pull/1357/changes#diff-bd0e9717702784099989654cb10966852a29b32bcb4796bc95acbde91d23682dR275).

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -283,6 +310,59 @@ export const importScheduleWithUsername = async (username: string) => {
...
+
+            await autoSaveSchedule(accounts.providerAccountId, {
+                userInfo: {
+                    email: users.email,
+                    name: users.name,
+                    avatar: users.avatar,
+                },
+            });
+        } catch (err) {
+            console.error('Failed to auto-save after importing shared schedule:', err);
```
Choollol: Should we also open a snackbar or something similar to tell the user something went wrong?

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -283,6 +310,59 @@ export const importScheduleWithUsername = async (username: string) => {
...
+        }
+    }
```
Choollol, line 344: Is there anything specific that should happen if the user's session isn't valid?

apps/antalmanac/src/components/Calendar/Toolbar/ScheduleSelect/schedule-select-buttons/ShareScheduleButton.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { useSessionStore } from '$stores/SessionStore';
+
+interface ShareScheduleButtonProps {
+    index: number;
+    disabled?: boolean;
+}
+
+export function ShareScheduleButton({ index, disabled }: ShareScheduleButtonProps) {
+    const [copied, setCopied] = useState(false);
+    const { sessionIsValid } = useSessionStore();
```
Choollol:
```suggestion
    const sessionIsValid = useSessionStore(state => state.sessionIsValid);
```

So state changes only cause rerenders when needed.

apps/antalmanac/src/components/Calendar/Toolbar/CalendarToolbar.tsx
```diff
@@ -159,7 +161,7 @@ export const CalendarToolbar = memo((props: CalendarPaneToolbarProps) => {
                             color={showFinalsSchedule ? 'primary' : 'inherit'}
                             onClick={handleToggleFinals}
                             id={showFinalsSchedule ? 'finals-button-pressed' : 'finals-button'}
-                            disabled={skeletonMode}
+                            disabled={skeletonMode && !isSharedSchedulePage}
```
Choollol: Small suggestion: Extract this into a shared variable for mobile and non-mobile.

apps/antalmanac/src/components/Calendar/Toolbar/CalendarToolbar.tsx
```diff
@@ -291,12 +294,14 @@ export const CalendarToolbar = memo((props: CalendarPaneToolbarProps) => {
                         </IconButton>
                     </Tooltip>
 
-                    <ClearScheduleButton
-                        size="medium"
-                        fontSize="small"
-                        skeletonMode={skeletonMode}
-                        analyticsCategory={analyticsEnum.calendar}
-                    />
+                    {!isSharedSchedulePage && (
```
Choollol: Small question: Should this disable the button rather than not render it at all? It would keep the layout consistent with the normal calendar, but not a big issue.

apps/antalmanac/src/components/Calendar/CourseCalendarEvent.tsx
```diff
@@ -152,22 +154,24 @@ export const CourseCalendarEvent = ({ selectedEvent, scheduleNames, closePopover
...
-                                logAnalytics(postHog, {
-                                    category: analyticsEnum.calendar,
-                                    action: analyticsEnum.calendar.actions.DELETE_COURSE,
-                                });
-                            }}
-                        >
-                            <Delete fontSize="inherit" />
-                        </IconButton>
-                    </Tooltip>
+                    {!isSharedSchedulePage && (
```
Choollol: Similar to https://github.com/icssc/AntAlmanac/pull/1357/changes#r2775833753

apps/antalmanac/src/components/Map/Map.tsx
```diff
@@ -224,9 +226,19 @@ export default function CourseMap() {
 
     const onBuildingChange = useCallback(
         (building?: ExtendedBuilding | null) => {
-            navigate(`/map?location=${building?.id}`);
+            if (isSharedSchedulePage) {
```
Choollol: This code works fine, but issue is related: When clicking the location of a class from the popover in the calendar (`MapLink.tsx`), it goes out of the shared schedule page and some buggy stuff happens.

Image

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -215,25 +238,50 @@ function ScheduleNoteBox() {
 }
 
 function SkeletonSchedule() {
-    const [skeletonSchedule, setSkeletonSchedule] = useState(AppStore.getCurrentSkeletonSchedule());
+    const getScheduleData = () => {
+        const skeletonSchedule = AppStore.getCurrentSkeletonSchedule();
+        if (!skeletonSchedule.courses || skeletonSchedule.courses.length === 0) {
+            const regularCourses = AppStore.schedule.getCurrentCourses();
+            if (regularCourses.length > 0) {
+                return {
```
Choollol, line 252: Small suggestion: Could give this some type safety by typing the return as `ShortCourseSchedule` and adding `id: undefined` to the returned object below.

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -215,25 +238,50 @@ function ScheduleNoteBox() {
...
             AppStore.off('skeletonScheduleChange', updateSkeletonSchedule);
             AppStore.off('currentScheduleIndexChange', updateSkeletonSchedule);
+            AppStore.off('addedCoursesChange', updateSkeletonSchedule);
+            AppStore.off('customEventsChange', updateSkeletonSchedule);
         };
     }, []);
 
     const sectionsByTerm: [string, string[]][] = useMemo(() => {
-        const result = skeletonSchedule.courses.reduce(
+        const courses = skeletonSchedule?.courses || [];
```
Choollol: Small question: When would `skeletonSchedule` be nullish? I'm just asking because null/undefined aren't part of its type. Did you run into problems with this in your own testing?

apps/antalmanac/src/components/ScheduleManagement/ScheduleManagementTab.tsx
```diff
@@ -12,12 +13,77 @@ interface ScheduleManagementTabProps {
...
+            e.preventDefault();
+        }
     };
 
+    if (isSharedSchedulePage) {
+        const tabComponent = (
+            <Tab
+                id={tab.id}
+                icon={tab.icon}
+                iconPosition={isMobile ? 'top' : 'start'}
```
Choollol: Extract shared props between this component and the typical non-shared-page component (at the very bottom). Mostly `iconPosition` and `sx`, I think the others are fine.

apps/antalmanac/src/hooks/useIsSharedSchedulePage.tsx
```diff
@@ -0,0 +1,10 @@
+import { useLocation } from 'react-router-dom';
+
```
Choollol, line 2: We could have an extra layer of abstraction and create an additional `useIsReadOnlyView` hook or something similar that uses `useIsSharedSchedulePage`. This way, we wouldn't need to add conditions everywhere if we add an extra feature that needs to disable deleting/editing.

apps/antalmanac/src/routes/SharedSchedulePage.tsx
```diff
@@ -0,0 +1,414 @@
...
+        >
+            <Stack direction="column">
+                <ScheduleCalendar />
+            </Stack>
+            <Stack direction="column" ref={scheduleManagementRef}>
+                <ScheduleManagement />
+            </Stack>
+        </Split>
+    );
+}
```
Choollol: This is a lot of duplication with the `MobileHome` and `DesktopHome` components. Maybe we could just import those components and use them here?

apps/antalmanac/src/routes/SharedSchedulePage.tsx
```diff
@@ -0,0 +1,414 @@
...
+
+export function SharedSchedulePage() {
+    const theme = useTheme();
+    const navigate = useNavigate();
+    const { scheduleId } = useParams<{ scheduleId: string }>();
+    const { sessionIsValid } = useSessionStore();
```
Choollol:
```suggestion
    const sessionIsValid = useSessionStore((state) => state.sessionIsValid);
```

apps/antalmanac/src/routes/SharedSchedulePage.tsx
```diff
@@ -0,0 +1,414 @@
...
+    const { setOpenLoadingSchedule } = scheduleComponentsToggleStore();
```
Choollol:
```suggestion
    const setOpenLoadingSchedule = scheduleComponentsToggleStore((state) => state.setOpenLoadingSchedule);
```

apps/antalmanac/src/routes/SharedSchedulePage.tsx
```diff
@@ -0,0 +1,414 @@
...
+                await AppStore.loadSchedule(emptySchedule);
+            }
+
+            setOpenLoadingSchedule(false);
+            navigate('/');
+        } catch (err) {
+            console.error('Error exiting shared schedule:', err);
+            setOpenLoadingSchedule(false);
+            navigate('/');
+        }
```
Choollol: Can move `setOpenLoadingSchedule(false);` and `navigate('/');` outside the try-catch.

apps/antalmanac/src/routes/SharedSchedulePage.tsx
```diff
@@ -0,0 +1,414 @@
...
+        };
+    }, [scheduleId, setOpenLoadingSchedule]);
+
+    const handleExitSharedSchedule = useCallback(async () => {
+        try {
+            setOpenLoadingSchedule(true);
+
+            if (AppStore.getSkeletonMode()) {
+                AppStore.exitSkeletonMode();
+            }
```
Choollol: Tiny nitpick: These can be moved outside the try-catch.

apps/antalmanac/src/routes/SharedSchedulePage.tsx
```diff
@@ -0,0 +1,414 @@
...
+
+                if (AppStore.getSkeletonMode()) {
+                    AppStore.exitSkeletonMode();
+                }
+
+                const sharedSchedule = await trpc.userData.getSharedSchedule.query({
+                    scheduleId,
+                });
+
+                const scheduleSaveState = {
```
Choollol: Tiny nitpick: Could type this as `ScheduleSaveState`.

apps/antalmanac/src/routes/SharedSchedulePage.tsx
```diff
@@ -0,0 +1,414 @@
...
+                    }
+                }
+            } else {
+                const defaultTerm = getDefaultTerm();
+                const emptyScheduleData = {
+                    scheduleName: defaultTerm.shortName.replaceAll(' ', '-'),
+                    courses: [] as ShortCourse[],
+                    customEvents: [] as RepeatingCustomEvent[],
+                    scheduleNote: '',
+                } as ShortCourseSchedule;
```
Choollol: I think 
```ts
const emptyScheduleData: ShortCourseSchedule = {
    scheduleName: defaultTerm.shortName.replaceAll(' ', '-'),
    courses: [] as ShortCourse[],
    customEvents: [] as RepeatingCustomEvent[],
    scheduleNote: '',
    id: undefined,
};
```
would be safer.

Could we also replace defaultTerm.shortName.replaceAll(' ', '-') with AppStore.getDefaultScheduleName()?

Optional suggestion: This has a lot of logic that's shared with `Schedules.ts`'s constructor, you could extract it into a utility function.

apps/antalmanac/src/routes/SharedSchedulePage.tsx
```diff
@@ -0,0 +1,414 @@
...
+                await AppStore.loadSchedule(scheduleSaveState as ScheduleSaveState);
+            }
+        } else {
+            const defaultTerm = getDefaultTerm();
+            const emptyScheduleData = {
+                scheduleName: defaultTerm.shortName.replaceAll(' ', '-'),
+                courses: [] as ShortCourse[],
+                customEvents: [] as RepeatingCustomEvent[],
+                scheduleNote: '',
+            } as ShortCourseSchedule;
```
Choollol: I think 
```ts
const emptyScheduleData: ShortCourseSchedule = {
    scheduleName: defaultTerm.shortName.replaceAll(' ', '-'),
    courses: [] as ShortCourse[],
    customEvents: [] as RepeatingCustomEvent[],
    scheduleNote: '',
    id: undefined,
};
```
would be safer.

Also, I think `defaultTerm.shortName.replaceAll(' ', '-')` could be replaced with `AppStore.getDefaultScheduleName()`?

Optional suggestion: This has a lot of logic that's shared with `Schedules.ts`'s constructor, you could extract it into a utility function.

apps/antalmanac/src/routes/SharedSchedulePage.tsx
```diff
@@ -0,0 +1,414 @@
...
+                    courses: [] as ShortCourse[],
+                    customEvents: [] as RepeatingCustomEvent[],
+                    scheduleNote: '',
+                } as ShortCourseSchedule;
+                const emptySchedule: ScheduleSaveState = {
+                    schedules: [emptyScheduleData],
+                    scheduleIndex: 0,
+                };
+                await AppStore.loadSchedule(emptySchedule);
+            }
```
Choollol: This is a lot of logic that's shared with `handleLoadSchedule`, maybe could replace this with a call to `handleLoadSchedule`.

apps/antalmanac/src/routes/SharedSchedulePage.tsx
```diff
@@ -0,0 +1,414 @@
...
+        if (!scheduleId) {
+            return;
+        }
+
+        try {
+            setOpenLoadingSchedule(true);
+
+            if (AppStore.getSkeletonMode()) {
+                AppStore.exitSkeletonMode();
+            }
```
Choollol: This looks like it could be extracted into a reusable function, maybe `beginLoadingSchedule` or something.

apps/antalmanac/src/routes/SharedSchedulePage.tsx
```diff
@@ -0,0 +1,414 @@
...
+        }
+    }, [scheduleId, sessionIsValid, navigate, setOpenLoadingSchedule, handleLoadSchedule]);
+
+    const handleGoHome = useCallback(async () => {
+        try {
+            setOpenLoadingSchedule(true);
+
+            if (AppStore.getSkeletonMode()) {
+                AppStore.exitSkeletonMode();
+            }
```
Choollol: See https://github.com/icssc/AntAlmanac/pull/1357/changes#r2776575421

apps/antalmanac/src/routes/SharedSchedulePage.tsx
```diff
@@ -0,0 +1,414 @@
...
+
+            if (sessionIsValid) {
+                const sessionToken = useSessionStore.getState().session;
+                if (!sessionToken) {
+                    throw new Error('No session token available');
+                }
+
+                await handleLoadSchedule(sessionToken);
```
Choollol: This also looks like it could be extracted into a separate function, maybe loadSessionSchedule or something like that.

apps/antalmanac/src/routes/SharedSchedulePage.tsx
```diff
@@ -0,0 +1,414 @@
...
+                AppStore.exitSkeletonMode();
+            }
+
+            const sessionToken = useSessionStore.getState().session;
+
+            if (!sessionToken) {
+                throw new Error('No session token available');
+            }
+
+            await handleLoadSchedule(sessionToken);
```
Choollol: See https://github.com/icssc/AntAlmanac/pull/1357/changes#r2776579369

apps/antalmanac/src/routes/SharedSchedulePage.tsx
```diff
@@ -0,0 +1,414 @@
...
+                if (!loadSuccess) {
+                    throw new Error('Failed to load shared schedule');
+                }
+                if (AppStore.getCurrentScheduleIndex() !== 0) {
+                    AppStore.changeCurrentSchedule(0);
+                }
+                AppStore.skeletonMode = true;
+                AppStore.emit('skeletonModeChange');
+                setScheduleName(sharedSchedule.scheduleName);
+                setOpenLoadingSchedule(false);
```
Choollol: See https://github.com/icssc/AntAlmanac/pull/1357/changes#r2776065343, though this applies to only `setOpenLoadingSchedule(false);`.

apps/antalmanac/src/routes/SharedSchedulePage.tsx
```diff
@@ -0,0 +1,414 @@
...
+                        'success',
+                        `Shared schedule "${sharedSchedule.scheduleName}" added to your schedules!`
+                    );
+                } else {
+                    await importSharedScheduleById(scheduleId);
+                }
+            }
+
+            setOpenLoadingSchedule(false);
+            navigate('/');
```
Choollol: See https://github.com/icssc/AntAlmanac/pull/1357/changes#r2776065343

apps/antalmanac/src/routes/SharedSchedulePage.tsx
```diff
@@ -0,0 +1,414 @@
...
+            const sessionToken = useSessionStore.getState().session;
+
+            if (!sessionToken) {
+                throw new Error('No session token available');
+            }
+
+            await handleLoadSchedule(sessionToken);
+
+            setOpenLoadingSchedule(false);
+            navigate('/');
```
Choollol: See https://github.com/icssc/AntAlmanac/pull/1357/changes#r2776065343

apps/antalmanac/src/routes/SharedSchedulePage.tsx
```diff
@@ -0,0 +1,414 @@
...
+    const navigate = useNavigate();
+    const { scheduleId } = useParams<{ scheduleId: string }>();
+    const { sessionIsValid } = useSessionStore();
+    const { setOpenLoadingSchedule } = scheduleComponentsToggleStore();
+    const [error, setError] = useState<string | null>(null);
+    const [scheduleName, setScheduleName] = useState<string | null>(null);
+    const hasAttemptedLoadRef = useRef(false);
+    const currentScheduleIdRef = useRef<string | undefined>(scheduleId);
+
+    const isMobileScreen = useMediaQuery(theme.breakpoints.down('sm'));
```
Choollol:
```suggestion
    const isMobileScreen = useIsMobile();
```
and `const theme = useTheme();` (above) can be deleted

apps/antalmanac/src/routes/SharedSchedulePage.tsx
```diff
@@ -0,0 +1,414 @@
...
+export function SharedSchedulePage() {
+    const theme = useTheme();
+    const navigate = useNavigate();
+    const { scheduleId } = useParams<{ scheduleId: string }>();
+    const { sessionIsValid } = useSessionStore();
+    const { setOpenLoadingSchedule } = scheduleComponentsToggleStore();
+    const [error, setError] = useState<string | null>(null);
+    const [scheduleName, setScheduleName] = useState<string | null>(null);
+    const hasAttemptedLoadRef = useRef(false);
+    const currentScheduleIdRef = useRef<string | undefined>(scheduleId);
```
Choollol: Small nitpick: Might be just personal preference, but I think adding some whitespace between "categories" of hooks is more readable, e.g. adding empty lines between misc hooks, store hooks, `useState`, and `useRef`.

apps/antalmanac/src/routes/SharedSchedulePage.tsx
```diff
@@ -0,0 +1,414 @@
...
+                    </Button>
+                </Stack>
+            </LocalizationProvider>
+        );
+    }
+
+    return (
+        <LocalizationProvider dateAdapter={AdapterDateFns}>
+            <PatchNotes />
+            <InstallPWABanner />
```
Choollol: We should find a way to de-duplicate these components (and `NotificationSnackbar` and `HelpMenu` below) since they're copy-pasted from `Home`. Maybe a wrapper component or something?

apps/antalmanac/src/routes/SharedSchedulePage.tsx
```diff
@@ -0,0 +1,414 @@
...
+                    </Box>
+                )}
+                {isMobileScreen ? <MobileSharedSchedule /> : <DesktopSharedSchedule />}
+            </Stack>
+
+            <NotificationSnackbar />
+            <HelpMenu />
+        </LocalizationProvider>
+    );
+}
```
Choollol, line 32: I think a lot of this component could be separated out into another component, maybe `SharedScheduleBanner.tsx`. It might also help with performance since it wouldn't be rerendering the entire page when state changes.
Here's what a mockup of the new `SharedSchedulePage` could look like:
```ts
export function SharedSchedulePage() {
    const isMobileScreen = useIsMobile();

    const [error, setError] = useState<string | null>(null);

    if (error) {
        return (
            <LocalizationProvider dateAdapter={AdapterDateFns}>
                <Stack component="main" height="100dvh" justifyContent="center" alignItems="center" spacing={2}>
                    <SharedScheduleBanner error={error} setError={setError} />
                </Stack>
            </LocalizationProvider>
        );
    }

    return (
        <LocalizationProvider dateAdapter={AdapterDateFns}>
            <PatchNotes />
            <InstallPWABanner />

            <Stack component="main" height="100dvh">
                <Header />
                <SharedScheduleBanner error={error} setError={setError} />
                {isMobileScreen ? <MobileSharedSchedule /> : <DesktopSharedSchedule />}
            </Stack>

            <NotificationSnackbar />
            <HelpMenu />
        </LocalizationProvider>
    );
}
```

And the new component could look something like this:
```ts
const SharedScheduleBanner = ({error, setError}: Props) => {
    const theme = useTheme();
    const navigate = useNavigate();
    const { scheduleId } = useParams<{ scheduleId: string }>();
    const { sessionIsValid } = useSessionStore();
    const { setOpenLoadingSchedule } = scheduleComponentsToggleStore();
    const [scheduleName, setScheduleName] = useState<string | null>(null);
    const hasAttemptedLoadRef = useRef(false);
    const currentScheduleIdRef = useRef<string | undefined>(scheduleId);

    const isMobileScreen = useMediaQuery(theme.breakpoints.down('sm'));

    useEffect(() => {
        if (currentScheduleIdRef.current !== scheduleId) {
            hasAttemptedLoadRef.current = false;
            currentScheduleIdRef.current = scheduleId;
        }

        if (hasAttemptedLoadRef.current) {
            return;
        }

        const loadSharedSchedule = async () => {
            if (!scheduleId) {
                setError('Invalid schedule ID');
                setOpenLoadingSchedule(false);
                hasAttemptedLoadRef.current = true;
                return;
            }

            hasAttemptedLoadRef.current = true;

            try {
                setOpenLoadingSchedule(true);

                removeLocalStorageUnsavedActions();

                if (AppStore.getSkeletonMode()) {
                    AppStore.exitSkeletonMode();
                }

                const sharedSchedule = await trpc.userData.getSharedSchedule.query({
                    scheduleId,
                });

                const scheduleSaveState = {
                    schedules: [
                        {
                            id: sharedSchedule.id,
                            scheduleName: sharedSchedule.scheduleName,
                            scheduleNote: sharedSchedule.scheduleNote || '',
                            courses: sharedSchedule.courses,
                            customEvents: sharedSchedule.customEvents,
                        },
                    ],
                    scheduleIndex: 0,
                };

                const loadSuccess = await AppStore.loadSchedule(scheduleSaveState);

                if (!loadSuccess) {
                    throw new Error('Failed to load shared schedule');
                }
                if (AppStore.getCurrentScheduleIndex() !== 0) {
                    AppStore.changeCurrentSchedule(0);
                }
                AppStore.skeletonMode = true;
                AppStore.emit('skeletonModeChange');
                setScheduleName(sharedSchedule.scheduleName);
                setOpenLoadingSchedule(false);
                setError(null);
            } catch (err) {
                console.error('Error loading shared schedule:', err);
                setError('Failed to load shared schedule. It may not exist or may have been deleted.');
                setOpenLoadingSchedule(false);
            }
        };

        loadSharedSchedule();

        return () => {
            setOpenLoadingSchedule(false);
        };
    }, [scheduleId, setOpenLoadingSchedule]);

    const handleExitSharedSchedule = useCallback(async () => {
        try {
            setOpenLoadingSchedule(true);

            if (AppStore.getSkeletonMode()) {
                AppStore.exitSkeletonMode();
            }

            if (sessionIsValid) {
                const sessionToken = useSessionStore.getState().session;

                if (sessionToken) {
                    const userDataResponse = await trpc.userData.getUserDataWithSession.query({
                        refreshToken: sessionToken,
                    });
                    const scheduleSaveState =
                        (userDataResponse as { userData?: ScheduleSaveState }).userData ?? userDataResponse;

                    if (scheduleSaveState) {
                        await AppStore.loadSchedule(scheduleSaveState as ScheduleSaveState);
                    }
                }
            } else {
                const defaultTerm = getDefaultTerm();
                const emptyScheduleData = {
                    scheduleName: defaultTerm.shortName.replaceAll(' ', '-'),
                    courses: [] as ShortCourse[],
                    customEvents: [] as RepeatingCustomEvent[],
                    scheduleNote: '',
                } as ShortCourseSchedule;
                const emptySchedule: ScheduleSaveState = {
                    schedules: [emptyScheduleData],
                    scheduleIndex: 0,
                };
                await AppStore.loadSchedule(emptySchedule);
            }

            setOpenLoadingSchedule(false);
            navigate('/');
        } catch (err) {
            console.error('Error exiting shared schedule:', err);
            setOpenLoadingSchedule(false);
            navigate('/');
        }
    }, [navigate, sessionIsValid, setOpenLoadingSchedule]);

    const handleLoadSchedule = useCallback(async (sessionToken?: string) => {
        if (sessionToken) {
            const userDataResponse = await trpc.userData.getUserDataWithSession.query({
                refreshToken: sessionToken,
            });
            const scheduleSaveState =
                (userDataResponse as { userData?: ScheduleSaveState }).userData ?? userDataResponse;

            if (scheduleSaveState) {
                await AppStore.loadSchedule(scheduleSaveState as ScheduleSaveState);
            }
        } else {
            const defaultTerm = getDefaultTerm();
            const emptyScheduleData = {
                scheduleName: defaultTerm.shortName.replaceAll(' ', '-'),
                courses: [] as ShortCourse[],
                customEvents: [] as RepeatingCustomEvent[],
                scheduleNote: '',
            } as ShortCourseSchedule;
            const emptySchedule: ScheduleSaveState = {
                schedules: [emptyScheduleData],
                scheduleIndex: 0,
            };
            await AppStore.loadSchedule(emptySchedule);
        }
    }, []);

    const handleAddToMySchedules = useCallback(async () => {
        if (!scheduleId) {
            return;
        }

        try {
            setOpenLoadingSchedule(true);

            if (AppStore.getSkeletonMode()) {
                AppStore.exitSkeletonMode();
            }

            if (sessionIsValid) {
                const sessionToken = useSessionStore.getState().session;
                if (!sessionToken) {
                    throw new Error('No session token available');
                }

                await handleLoadSchedule(sessionToken);
                await importSharedScheduleById(scheduleId);
            } else {
                const currentSchedules = AppStore.schedule.getScheduleAsSaveState();
                const sharedSchedule = await trpc.userData.getSharedSchedule.query({ scheduleId });
                const currentSchedule = currentSchedules.schedules[currentSchedules.scheduleIndex];

                if (currentSchedule && currentSchedule.scheduleName === sharedSchedule.scheduleName) {
                    const prefixedName = `(shared)-${sharedSchedule.scheduleName}`;
                    AppStore.renameSchedule(currentSchedules.scheduleIndex, prefixedName);
                    openSnackbar(
                        'success',
                        `Shared schedule "${sharedSchedule.scheduleName}" added to your schedules!`
                    );
                } else {
                    await importSharedScheduleById(scheduleId);
                }
            }

            setOpenLoadingSchedule(false);
            navigate('/');
        } catch (err) {
            console.error('Error adding schedule to account:', err);
            setOpenLoadingSchedule(false);
            if (AppStore.getSkeletonMode()) {
                AppStore.exitSkeletonMode();
            }
            navigate('/');
        }
    }, [scheduleId, sessionIsValid, navigate, setOpenLoadingSchedule, handleLoadSchedule]);

    const handleGoHome = useCallback(async () => {
        try {
            setOpenLoadingSchedule(true);

            if (AppStore.getSkeletonMode()) {
                AppStore.exitSkeletonMode();
            }

            const sessionToken = useSessionStore.getState().session;

            if (!sessionToken) {
                throw new Error('No session token available');
            }

            await handleLoadSchedule(sessionToken);

            setOpenLoadingSchedule(false);
            navigate('/');
        } catch (err) {
            console.error('Error loading user data:', err);
            setOpenLoadingSchedule(false);
            navigate('/');
        }
    }, [navigate, setOpenLoadingSchedule, handleLoadSchedule]);

    if (error) {
        return (
            <>
                <Alert severity="error">{error}</Alert>
                <Button variant="contained" onClick={handleGoHome}>
                    Go Home
                </Button>
            </>
        );
    }

    return (
        <Box
            sx={{
                px: 2,
                py: isMobileScreen ? 0.5 : 1,
                backgroundColor: 'background.paper',
                borderBottom: 1,
                borderColor: 'divider',
            }}
        >
            <Stack
                direction={'row'}
                justifyContent="space-between"
                alignItems={'center'}
                spacing={isMobileScreen ? 1 : 0}
            >
                <Typography
                    variant="h6"
                    component="h1"
                    sx={{
                        fontSize: isMobileScreen ? '1rem' : undefined,
                        overflow: 'hidden',
                        textOverflow: 'ellipsis',
                        display: 'flex',
                        alignItems: 'center',
                    }}
                >
                    Viewing Shared Schedule:{' '}
                    <strong style={{ whiteSpace: 'nowrap', marginLeft: '0.25rem' }}>{scheduleName}</strong>
                </Typography>
                <Stack direction="row" spacing={1} alignItems="center" sx={{ flexShrink: 0 }}>
                    {isMobileScreen ? (
                        <IconButton
                            aria-label="Add to My Schedules"
                            onClick={handleAddToMySchedules}
                            size="small"
                            sx={{
                                backgroundColor: 'primary.dark',
                                color: 'white',
                                borderRadius: 2,
                                '&:hover': {
                                    backgroundColor: 'primary.main',
                                },
                            }}
                        >
                            <Add />
                        </IconButton>
                    ) : (
                        <Button variant="contained" onClick={handleAddToMySchedules} size="large">
                            Add to My Schedules
                        </Button>
                    )}
                    <IconButton
                        aria-label="Exit shared schedule"
                        onClick={handleExitSharedSchedule}
                        color="inherit"
                        size={isMobileScreen ? 'medium' : 'large'}
                    >
                        <Close />
                    </IconButton>
                </Stack>
            </Stack>
        </Box>
    );
};
```
Note that `setError(null);` is moved down in `loadSharedSchedule` to prevent infinite rerenders.

You don't have to make these exact changes of course, just giving you a potential starting spot.

Choollol (changes requested):
Sorry for the comment barrage haha they're mostly small suggestions. There are a few bigger/more general issues though. I'll try to list them here:
1. https://github.com/icssc/AntAlmanac/pull/1357/changes#r2775883412
2. https://github.com/icssc/AntAlmanac/pull/1357/changes#r2776030275
3. https://github.com/icssc/AntAlmanac/pull/1357/changes#r2776665863
4. Do we want to log analytics?
5. Something I noticed is that saving a schedule changes its ID, which means the share schedule link becomes invalid. I have no idea how we would get around this, maybe we don't have to, but I wanted to bring it up.

apps/antalmanac/src/actions/AppStoreActions.ts
brelieu05: Can you elaborate on Set by default? Since existingScheduleNames is already a Set?

apps/antalmanac/src/components/Calendar/Toolbar/ScheduleSelect/ScheduleSelect.tsx
```diff
@@ -116,11 +119,14 @@ export function SelectSchedulePopover() {
     }, []);
 
     const scheduleMappingToUse = skeletonMode ? skeletonScheduleMapping : scheduleMapping;
+    const displayName = isSharedSchedulePage
+        ? AppStore.getScheduleNames()[currentScheduleIndex] || scheduleMappingToUse[currentScheduleIndex]?.name
+        : scheduleMappingToUse[currentScheduleIndex]?.name;
```
Choollol: When logged in, the schedule name on the schedule select button shows the wrong schedule name:

Image

apps/antalmanac/src/components/Calendar/Toolbar/CalendarToolbar.tsx
Choollol: Actually, question: Should the app be marked as being in skeleton mode when on the shared schedule page? Since there's a schedule loaded, it feels like it isn't in skeleton mode.

brelieu05:
> Do we want to log analytics? 

CC: @alexespejo 

>Something I noticed is that saving a schedule changes its ID, which means the share schedule link becomes invalid. I have no idea how we would get around this, maybe we don't have to, but I wanted to bring it up.

It's an intended feature as we want it as a one-time copy only rather than a true share.

apps/antalmanac/src/components/buttons/MapLink.tsx
```diff
@@ -10,14 +10,18 @@ interface MapLinkProps {
 
 export const MapLink = ({ buildingId, room }: MapLinkProps) => {
     const { setActiveTab } = useTabStore();
+    const { pathname } = useLocation();
+    const isSharedPath = pathname.startsWith('/share/');
```
Choollol: This looks like duplicated logic from `useIsSharedSchedulePage`. If it's possible that this code and `useIsSharedSchedulePage` would need to change independently of each other, then it's fine to leave as is. If it's for the exact same purpose, then replacing this with a call to `useIsSharedSchedulePage` would be better.

apps/antalmanac/src/components/buttons/MapLink.tsx
```diff
@@ -10,14 +10,18 @@ interface MapLinkProps {
...
+
+    const toLocation = isSharedPath ? `${pathname}?location=${buildingId}` : `/map?location=${buildingId}`;
```
Choollol: Should we instead check if we're on any page that isn't the home page instead? It would let us not have to change anything if we add a new page later. I'm not too familiar with this area though, so final call is up to you.

apps/antalmanac/src/components/Calendar/Toolbar/CalendarToolbar.tsx
```diff
@@ -198,16 +198,16 @@ export const CalendarToolbar = memo((props: CalendarPaneToolbarProps) => {
             >
                 <Box display="flex" flexDirection="row" gap={0.5}>
                     <Box display="flex" flexWrap="nowrap" alignItems="center" gap={0.5}>
-                        <IconButton onClick={handleUndo(postHog)} disabled={skeletonMode}>
+                        <IconButton onClick={handleUndo(postHog)} disabled={isReadonlyView}>
```
Choollol: Could you explain why this button (and the others) aren't disabled during skeleton mode anymore? I'm not too familiar with skeleton mode and when/how it works, but to me it seems like they should still be disabled.

apps/antalmanac/src/components/Calendar/Toolbar/CalendarToolbar.tsx
```diff
@@ -284,24 +284,22 @@ export const CalendarToolbar = memo((props: CalendarPaneToolbarProps) => {
...
-                            size="medium"
-                            fontSize="small"
-                            skeletonMode={skeletonMode}
-                            analyticsCategory={analyticsEnum.calendar}
-                        />
-                    )}
+                    <ClearScheduleButton
+                        size="medium"
+                        fontSize="small"
+                        skeletonMode={isReadonlyView}
```
Choollol: https://github.com/icssc/AntAlmanac/pull/1357#discussion_r2779628556
> Could you explain why this button (and the others) aren't disabled during skeleton mode anymore? I'm not too familiar with skeleton mode and when/how it works, but to me it seems like they should still be disabled.

apps/antalmanac/src/components/Calendar/Toolbar/CalendarToolbar.tsx
```diff
@@ -198,16 +198,16 @@ export const CalendarToolbar = memo((props: CalendarPaneToolbarProps) => {
...
                         </IconButton>
-                        <IconButton onClick={handleRedo(postHog)} disabled={skeletonMode}>
+                        <IconButton onClick={handleRedo(postHog)} disabled={isReadonlyView}>
                             <RedoIcon fontSize="small" />
                         </IconButton>
                         <CustomEventDialog key="custom" scheduleNames={AppStore.getScheduleNames()} />
 
                         <Tooltip title="More options">
-                            <IconButton onClick={handleMenuOpen} disabled={skeletonMode}>
+                            <IconButton onClick={handleMenuOpen} disabled={isReadonlyView}>
```
Choollol: There are some actions in this menu that can still be done during readonly mode, like download calendar and taking a screenshot. Maybe we could disable the edit actions in the menu rather than disable the entire menu?

btw don't worry about this one, I'll make this change.

Choollol (changes requested):
Left some questions about the new changes! I'll also be opening another PR with some more fixes/readonly-disabling.

Some important stuff:
1. The wrong schedule is showed when logged in. The schedule select popover also shows the user's personal schedules. I can try to fix this, but you might know more about it than I do.
2. How should the course popover quick search be handled? Right now, it takes the user to their home page which is fine, but the popover's position gets messed up. If this should be allowed, I can fix this in my PR.
Image 

3.
>It's an intended feature as we want it as a one-time copy only rather than a true share.

👍 We should make sure users know this, maybe through a note in the shared schedule error display?

alexespejo:
> CC: @alexespejo
> 
> >
Shore

apps/antalmanac/src/components/buttons/MapLink.tsx
```diff
@@ -12,9 +11,9 @@ interface MapLinkProps {
 export const MapLink = ({ buildingId, room }: MapLinkProps) => {
     const { setActiveTab } = useTabStore();
     const { pathname } = useLocation();
-    const isSharedPath = useIsSharedSchedulePage();
 
-    const toLocation = isSharedPath ? `${pathname}?location=${buildingId}` : `/map?location=${buildingId}`;
+    const isHomePage = pathname === '/' || pathname == '/added' || pathname == '/map';
```
Choollol, line 15: Optional: Could put this into a new custom hook, but since this logic is used only here, I think it's fine for now.

apps/antalmanac/src/components/SharedScheduleBanner.tsx
```diff
@@ -0,0 +1,316 @@
...
+                const sessionToken = useSessionStore.getState().session;
+                if (sessionToken) {
+                    const userDataResponse = await trpc.userData.getUserDataWithSession.query({
+                        refreshToken: sessionToken,
+                    });
+
+                    if (userDataResponse?.userData) {
+                        await AppStore.loadSchedule(userDataResponse.userData);
+                    }
+                }
```
Choollol, line 190: Optional: Tiny bit of duplication with `handleLoadSchedule`, but if it's stable enough to be fine then I'm okay with it.

Choollol (dismissed):
LGTM!

Left a couple of suggestions, but they're super tiny and optional.

apps/antalmanac/src/components/SharedScheduleBanner.tsx
brelieu05: Yeah this was to fix the bug regarding the schedule name mismatch

alexespejo (changes requested):
Couple of things 

issue: We should have AANTS disabled
Image

issue: Adding schedules doesn't seem to work..? When I open a shared schedule link I'm signed out. And when I hit "add to schedule" it signs me back in but it doesn't import it 

Image

Choollol:
> issue: Adding schedules doesn't seem to work..? When I open a shared schedule link I'm signed out. And when I hit "add to schedule" it signs me back in but it doesn't import it

Whoops that's my bad. I'll fix it in a new PR along with disabling the notification buttons.

Also, side question, what's the reasoning behind disabling AANTS in readonly view?

alexespejo:
> Also, side question, what's the reasoning behind disabling AANTS in readonly view?

I think it was bc I was treating it as if you were in readonly you were also signed out. So if the user is signed in then it's chill to enable AANTS

alexespejo (dismissed):
Send it

Choollol (dismissed):
does this work

Choollol:
I think @KevinWu098 needs to take away the requesting change status?

KevinWu098:
> I think @KevinWu098 needs to take away the requesting change status?

CC @alexespejo

alexespejo:
Note: Holding off on merging this during registration. It's a long time but I don't think it's worth risking some random bug breaking things 

CC: @brelieu05 @Choollol

alexespejo:
@cubic

apps/antalmanac/src/backend/routers/userData.ts
```diff
@@ -306,6 +306,41 @@ const userDataRouter = router({
...
+        }
+        return schedule;
+    }),
+
+    /**
+     * Gets the schedule ID for a schedule by userId and schedule name.
+     * @param input - An object containing the userId and scheduleName.
+     * @returns The schedule ID, or throws NOT_FOUND if not found.
+     */
+    getScheduleIdByName: procedure
```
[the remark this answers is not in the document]
alexespejo: > Anyone who knows (or guesses) a userId and scheduleName

Will likely never happen

KevinWu098:
Closing as superseded by #1511

PR #1365, opened 2025-12-08
fix: Calendar Download on Firefox Android

dsnsgithub:
## Summary

Firefox Android uses the MIME type to determine file extension.

## Test Plan
Still works on Chrome, Firefox, Safari
Now works on Firefox Android


alexespejo (approved):
🔥 🦊

PR #1376, opened 2025-12-17
feat: Missing Sections Warning

### LINKED ISSUE #1330: Required sections missing warning

calebwongg:
## Summary
Added a warning displaying the missing sections for a course on the added courses page.

When a user selects classes for a course, the required sections for a course are checked (e.g., lecture, lab, discussion) have been added. If any required section type is missing, the UI displays a warning so the user can correct their schedule.

## Test Plan
- Add SOME sections of a course to schedule, and check the added courses tab to verify there is a warning for whichever section is missing
- Add all sections of a course to a schedule. In the added courses tab, ensure there are no warnings, and then delete one of the sections to verify a warning pops up. Add the section back to verify the warning disappears
- Check error handling for users importing saved courses
- Check UI on both light mode and dark mode

## Issues

Closes #1330

## Before
image

## After
image

## Future Followup
- Could add an alert for if a user schedules two different sections in two different categories (e.g. the user schedules a lecture 'A' but a discussion 'B')

## Notes
- Missing sections are not checked for users who are importing courses that were saved to their profile (The only way to do it was to query websoc again when the user signs in). Hopefully should not be a problem since the scheduling period for next quarter is mostly over.


packages/types/src/websoc.ts
```diff
@@ -8,6 +8,7 @@ export type AASection = WebsocSection & AASectionExtendedProperties;
 
 type AACourseExtendedProperties = {
     sections: AASection[];
+    sectionTypes?: Set<WebsocSectionType>;
```
KevinWu098: question: why is this field optional now?

apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
```diff
@@ -154,6 +155,11 @@ function SectionTable(props: SectionTableProps) {
                 />
             </Box>
 
+            {missingSections?.length > 0 && (
+                <Alert severity="warning" sx={{ mb: 1 }}>
+                    Missing required sections: {missingSections.join(', ')}
+                </Alert>
```
KevinWu098: nit (style): doesn't the text look misaligned in the screenshot? worth looking into to see if this is a styling bug or if it's actually the MUI look

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
```diff
@@ -51,6 +58,14 @@ const flattenSOCObject = (SOCObject: WebsocAPIResponse): (WebsocSchool | WebsocD
                 for (const section of course.sections) {
                     (section as AASection).color = courseColors[section.sectionCode];
                 }
+
+                const sectionTypes = new Set<WebsocSectionType>();
+                course.sections.forEach((section) => {
+                    sectionTypes.add(section.sectionType);
+                });
+
+                (course as AACourse).sectionTypes = sectionTypes;
```
KevinWu098: question: I see we have this bit of code in `websoc.ts` and here. why are there two instances of the same code snippet? what do they do / why do we need them both?

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -7,6 +7,7 @@ import { ColumnToggleDropdown } from '../CoursePane/CoursePaneButtonRow';
 import SectionTableLazyWrapper from '../SectionTable/SectionTableLazyWrapper';
 
 import CustomEventDetailView from './CustomEventDetailView';
+import { getMissingSections } from './getMissingSections';
```
KevinWu098: chore: since you're here anyways, you should fix these imports and make them absolute

KevinWu098 (changes requested):
Some misc. comments — I'll take another look once they're addressed and once the merge conflicts with `main` are resolved

KevinWu098:
Is this ready for review? @calebwongg

packages/types/src/websoc.ts
calebwongg: Some AACourses that have not been enriched yet don't have the section types property if the user added/saved them to their schedule before this feature is merged. I made it optional to handle errors so the missing sections check doesn't try to check and access an attribute that doesn't exist

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
calebwongg: This snippet of code is in the getCourseInfo endpoint which is called when a user adds courses via import. I couldn't figure out a way to use this code snippet only once - I had to use it again in the flattenSocObject func in the CourseRenderPane. The websoc query from the regular course search calls the getOne endpoint which just returns the raw websoc query data. The query from the imports calls the getCourseInfo endpoint, and I couldn't find an overlap to avoid this redundant code

Would love to hear if you have any alternate ideas on this implementation :D

apps/backend/src/generated/searchData.ts
KevinWu098, line 1: issue: this should not be in the diff

apps/backend/src/generated/termData.ts
KevinWu098, line 1: issue: this should not be in the diff

packages/types/src/websoc.ts
KevinWu098: Is... that right? What does it mean that they wouldn't have it. We don't save these values into the database, right? It's fetched client-side on load. 

What I'm trying to say is, in what case would sectionTypes not be present if the user has loaded the page?

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
KevinWu098: Eh, we could move it to `lib` as a helper, but I don't really mind the code duplication. Works for me 🤷‍♂️

apps/antalmanac/src/app/Theme.tsx
```diff
@@ -151,6 +151,14 @@ export default function AppThemeProvider(props: Props) {
                             variant: 'standard',
                         },
                     },
+                    MuiAlert: {
+                        styleOverrides: {
+                            standardWarning: {
+                                backgroundColor: appTheme === 'dark' ? '#fff4e5ff' : '#FFEA99',
+                                color: appTheme === 'dark' ? '#856404ff' : '#302800ff',
```
KevinWu098: nit (non-blocking): the yellow seems a little light on dark mode

packages/types/src/websoc.ts
KevinWu098: Okay, upon further thinking, I can see that we don't apply the missing warning _at all_ when we load the page in because we 1. don't enrich `sectionTypes` when we call `fromScheduleSaveState` and 2. don't query the whole course (only the specific section codes).

Although 1. is something that we _should do_, it's moot because the information to do so (2.) simply doesn't exist (we don't store the course itself to query by). 

For now, I can accept that say... this gap in the feature is "intended", but I think that this would be a good exercise for you to also file an issue that can serve as a follow up to patch this gap

KevinWu098 (changes requested):
Looking really good. Mostly minor comments, except one longer set of comments regarding `sectionTypes` on load that (thankfully) is non-blocking.

However, I believe this PR introduces a bug when we undo an action on AA, which crashes the page. You can reproduce the error easily. I believe this may stem from our use of `Set`, which doesn't serialize well, so we may need to use an array as the type instead (this should be easy via gpt)

Sorry for all the back and forth!

KevinWu098:
Also, just curious @calebwongg. What is the added fix / change in this PR that makes it not break like the previous attempt?

calebwongg:
> Also, just curious @calebwongg. What is the added fix / change in this PR that makes it not break like the previous attempt?

Making section types optional prevents the original reason why it broke:

sectionTypes?: Set<WebsocSectionType>;

When a user initially loaded in their saved classes, section types were not enriched when the page was loaded in, so my getMissingSections was looking for an attribute that didn't exist.

KevinWu098:
> > Also, just curious @calebwongg. What is the added fix / change in this PR that makes it not break like the previous attempt?
> 
> Making section types optional prevents the original reason why it broke:
> 
> sectionTypes?: Set;
> 
> When a user initially loaded in their saved classes, section types were not enriched when the page was loaded in, so my getMissingSections was looking for an attribute that didn't exist.

hm... that's really interesting since you'd imagine that there would be a type error... perhaps related to https://github.com/icssc/AntAlmanac/issues/1372?

calebwongg:
> > > Also, just curious @calebwongg. What is the added fix / change in this PR that makes it not break like the previous attempt?
> > 
> > 
> > Making section types optional prevents the original reason why it broke:
> > sectionTypes?: Set;
> > When a user initially loaded in their saved classes, section types were not enriched when the page was loaded in, so my getMissingSections was looking for an attribute that didn't exist.
> 
> hm... that's really interesting since you'd imagine that there would be a type error... perhaps related to #1372?

Just did some further digging/clarifying... 

I was wrong this was actually NOT what caused the break...it was the line `sectionTypes: courseDetails.sectionTypes,` in `AppStoreActions.ts`

I added section types in the declaration for courseDetails in `backend/routers/websoc.ts`, which is what now prevents the crash
      
`               for (const course of department.courses) {
                    
                    const sectionTypesSet = new Set<WebsocSectionType>();
                    course.sections.forEach((section) => {
                        sectionTypesSet.add(section.sectionType);
                    });

                    const sectionTypes = [...sectionTypesSet];

                        courseInfo[section.sectionCode] = {
                            courseDetails: {
                                deptCode: department.deptCode,
                                courseNumber: course.courseNumber,
                                courseComment: course.courseComment,
                                prerequisiteLink: course.prerequisiteLink,
                                sectionTypes: sectionTypes,
                            },
                            section: section,
                        };
                    }
`

packages/types/src/websoc.ts
calebwongg: I can see the gap you're talking about. I will put up an issue. I also removed having the sectionTypes as optional because I realized its not needed

KevinWu098 (approved):
LGTM — thanks for sticking with this!

PR #1378, opened 2025-12-21
AANTS SST

IsaacNguyen:
## Summary
Merged NextJS + SST migration into AANTS and migrated the AANTS CDK code to SST

## Test Plan

- [ ] AANTS actually deploys and runs 🤣🤣🤣

## Issues
image

I kept getting this evil error that was saying the `.eslintignore` file in the root directory was breaking because of a new version of .eslint, but that same file exists on main, so I don't know why it's breaking. Fiddled with changing the version of .eslint in the AANTS folder to no avail. Couldn't figure it out so Kevin just told me to uncomment the pre-commit hook.


sst.config.ts
```diff
@@ -39,5 +40,20 @@ export default $config({
                 NEXT_PUBLIC_BASE_URL: domain,
             },
         });
+
+        const aantsLambda = new sst.aws.Function('AantsLambda', {
+            handler: 'apps/aants/src/lambda.handler',
+            timeout: '20 seconds', // TODO: Test how long AANTS takes to run and change accordingly
```
KevinWu098: chore: when writing `TODO`, a good practice is to also post-pend it with who is writing the todo

e.g. `TODO (@KevinWu098): do something`

sst.config.ts
```diff
@@ -21,6 +21,7 @@ export default $config({
     },
     async run() {
         const domain = getDomain();
+        const dbUrl = $app.stage === 'production' ? process.env.PROD_DB_URL : process.env.DEV_DB_URL;
```
KevinWu098: issue: it's actually incorrect that we distinguish between `PROD_DB_URL` and `DEV_DB_URL`. To an environment, it should just be `DB_URL`

suggestion: In the respective deploy files in `.github`, rather than: `DEV_DB_URL: ${{ secrets.DEV_DB_URL }}`, it should be `DB_URL: ${{ secrets.DEV_DB_URL }}` and then fix appropriately here in `sst.config.ts`

sst.config.ts
```diff
@@ -39,5 +40,20 @@ export default $config({
...
+            timeout: '20 seconds', // TODO: Test how long AANTS takes to run and change accordingly
+            memory: '256 MB',
+            environment: {
+                DB_URL: dbUrl,
+                NODE_ENV: $app.stage === 'production' ? 'production' : 'development',
+            },
+        });
+
+        new sst.aws.Cron('NotificationCronRule', {
+            schedule: 'rate(5 minutes)', // AANTS runs every 5 minutes - TODO: Might change in future
```
KevinWu098: repeat: todo attribution

packages/types/src/notification.ts
```diff
@@ -0,0 +1,16 @@
+export type NotificationStatus = {
+    openStatus: boolean;
+    waitlistStatus: boolean;
+    fullStatus: boolean;
+    restrictionStatus: boolean;
+};
```
KevinWu098: question: Is there a technical reason for why we describe the course statuses (open, waitlist, full) as mutually exclusive booleans? why is it not typed as something like

`courseStatus: 'open' | 'waitlist' | 'full'`

packages/types/src/notification.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export type Notification = {
+    term: string;
+    sectionCode: string;
+    courseTitle: string;
+    sectionType: string;
+    notificationStatus: NotificationStatus;
```
KevinWu098: question: is this a "status" or a "reason"?

packages/types/src/notification.ts
```diff
@@ -0,0 +1,16 @@
...
+    lastUpdated: string;
+    lastCodes: string;
```
KevinWu098, line 15: question: why do we have last codes but not "last status"?

KevinWu098: Ah... `lastUpdated` is... more accurately the last course status... probably worth fixing!

packages/types/src/notification.ts
KevinWu098: Ah. I should've read the DB schema first... i think the naming using `Status` is unclear.

packages/db/src/schema/subscription.ts
```diff
@@ -1,8 +1,6 @@
-import { integer, pgEnum, pgTable, primaryKey, text } from 'drizzle-orm/pg-core';
+import { integer, boolean, pgTable, primaryKey, text } from 'drizzle-orm/pg-core';
 
-import { users } from './auth';
-
-export const subscriptionTargetStatus = pgEnum('subscription_target_status', ['OPEN', 'WAITLISTED']);
```
KevinWu098, line 5: question: why did we opt not to continue using an enum?

apps/cdk/src/stacks/aants.ts
KevinWu098, line 1: chore: this file should be removed, right?

apps/antalmanac/src/lib/enrollmentHistory.ts
```diff
@@ -56,8 +56,9 @@ export class DepartmentEnrollmentHistory {
 
     async find(courseNumber: string): Promise<EnrollmentHistory[] | null> {
         const cacheKey = this.department + courseNumber;
-        return (DepartmentEnrollmentHistory.enrollmentHistoryCache[cacheKey] ??=
-            await this.queryEnrollmentHistory(courseNumber));
+        return (DepartmentEnrollmentHistory.enrollmentHistoryCache[cacheKey] ??= await this.queryEnrollmentHistory(
```
KevinWu098, line 59: thought: wowie, an actual usage of the nullish coalescing assignment operator

apps/antalmanac/src/backend/routers/notifications.ts
```diff
@@ -0,0 +1,51 @@
...
+import { procedure, router } from '../trpc';
+
+const NotificationStatusSchema = z.object({
+    openStatus: z.boolean(),
+    waitlistStatus: z.boolean(),
+    fullStatus: z.boolean(),
+    restrictionStatus: z.boolean(),
+});
+
+const NotificationSchema = z.object({
```
KevinWu098, line 13: thought: i imagine there's a reason we're redefining the schemas in zod versus reusing the drizzle type (if so, please reply and let me know) — but i do hope (and imagine) there is a better way of doing this

but i believe we do this in other places in the codebase, so a refactor would be out of scope within this PR imo

apps/antalmanac/src/backend/lib/rds.ts
```diff
@@ -299,6 +300,8 @@ export class RDS {
      * deduplicating by section code and term.
      * */
     private static async upsertCourses(tx: Transaction, scheduleId: string, courses: ShortCourse[]) {
+        await tx.delete(coursesInSchedule).where(eq(coursesInSchedule.scheduleId, scheduleId));
```
KevinWu098: issue (question): why is this being added here?? i would be very concerned if we were not correctly clearing and updating schedules on `main`

apps/antalmanac/src/backend/routers/notifications.ts
```diff
@@ -0,0 +1,51 @@
...
+            await Promise.all(
+                input.notifications.map((notification) => RDS.upsertNotification(db, input.id, notification))
+            );
+        }),
+
+    updateNotifications: procedure.input(z.object({ notification: NotificationSchema })).mutation(async ({ input }) => {
+        await RDS.updateAllNotifications(db, input.notification);
+    }),
+
+    deleteNotification: procedure
```
KevinWu098, line 40: nit: `id` is vague. `userId` would be better

apps/antalmanac/src/backend/routers/notifications.ts
```diff
@@ -0,0 +1,51 @@
...
+        .input(z.object({ id: z.string(), notification: NotificationSchema }))
+        .mutation(async ({ input }) => {
+            await RDS.deleteNotification(db, input.notification, input.id);
+        }),
+
+    deleteAllNotifications: procedure.input(z.object({ id: z.string() })).mutation(async ({ input }) => {
```
KevinWu098: repeat: `id` is vague

apps/antalmanac/src/backend/routers/notifications.ts
```diff
@@ -0,0 +1,51 @@
...
+    term: z.string(),
+    sectionCode: z.string(),
+    courseTitle: z.string(),
+    sectionType: z.string(),
+    lastUpdated: z.string(),
+    lastCodes: z.string(),
+    notificationStatus: NotificationStatusSchema,
+});
+
+const notificationsRouter = router({
```
KevinWu098, line 23: repeat: let's be verbose and say `userId` rather than `id`. in places where we have multiple objects being referred to, err on the side of verbosity

apps/antalmanac/src/backend/lib/rds.ts
```diff
@@ -648,4 +651,117 @@ export class RDS {
...
+     * @param notification - The notification object to upsert.
+     * @returns A promise that upserts the notification associated with a userId.
+     */
+    static async upsertNotification(db: DatabaseOrTransaction, userId: string, notification: Notification) {
+        return db.transaction((tx) =>
+            tx
+                .insert(subscriptions)
+                .values({
+                    userId,
+                    sectionCode: Number(notification.sectionCode),
```
KevinWu098: question: why do we convert it to a number?

apps/aants/tsconfig.json
KevinWu098, line 1: question: does this match the tsconfig in `apps/antalmanac`?

apps/aants/README.md
KevinWu098, line 1: Image

apps/aants/src/lambda.ts
```diff
@@ -0,0 +1,12 @@
+import { scanAndNotify } from '.';
+
+export async function handler() {
+    // run the script
+    await scanAndNotify();
+
+    // return response
```
KevinWu098: nit: 🤓 comments seem redundant

apps/aants/src/env.ts
KevinWu098, line 1: thought (non-blocking): `t3 env` would be great to implement in a future pr

apps/aants/scripts/createTemplate.ts
```diff
@@ -0,0 +1,76 @@
+/*
+ *   To run this script, use 'pnpm run template'
+ */
+
+import { SESv2Client, CreateEmailTemplateCommand } from '@aws-sdk/client-sesv2';
+
+const client = new SESv2Client({ region: 'us-east-2' });
```
KevinWu098, line 7: question: why `us-east-2`?

apps/aants/scripts/build.mjs
KevinWu098, line 1: question: is this file necessary still with sst?

apps/aants/src/helpers/notificationDispatch.ts
```diff
@@ -0,0 +1,163 @@
...
+    restrictionCodes: string;
+    deptCode: string;
+    courseNumber: string;
+    courseTitle: string;
+    quarter: string;
+    year: string;
+}
+
+const BATCH_SIZE = 450;
+const client = new SESv2Client({ region: 'us-east-2' });
```
KevinWu098, line 20: repeat: `us-east-2`?

apps/aants/src/helpers/notificationDispatch.ts
```diff
@@ -0,0 +1,163 @@
...
+
+        notification = notification.replace(/\n/g, '<br>');
+
+        const time = getFormattedTime();
+
+        const bulkEmailEntries = users
+            .filter((user) => user.email !== null)
+            .map((user) => ({
+                Destination: {
+                    ToAddresses: [user.email!],
```
KevinWu098: issue: never, ever, ever write a non-null assertion `!`. this is the devil and we shall not have it in our codebase

suggestion:

```
const bulkEmailEntries = users
  .filter((user): user is User & { email: string } => user.email !== null)
  .map((user) => ({
    Destination: {
      ToAddresses: [user.email],
    },
  }));
```

Using a type predicate is (ultimately) still an assertion, but it makes clear what our assertion is and why we assert it to be so

KevinWu098 (changes requested):
> [!WARNING]
> I switched the base branch from `aants-merge` to `main` to review the code more easily. If you intended to merge it into `aants-merge`, changing the base branch may screw up the review comments, so do so after we resolve them

Added some thoughts mostly on code quality. The core logic and implementation make sense to me!

apps/aants/scripts/createTemplate.ts
IsaacNguyen: image

`us-east-2` (Ohio) is the region we have SES setup on and the one my AWS account was delegated to. Not sure if this is correct/incorrect but it's just the way it was setup for me

apps/aants/src/helpers/notificationDispatch.ts
IsaacNguyen: Same reason as above
image

apps/aants/scripts/createTemplate.ts
KevinWu098: CC @andrew-wang0 any clue?

packages/db/src/schema/subscription.ts
IsaacNguyen: It's been a minute but I remember it raising a bunch of type errors that I couldn't figure out how to resolve. I asked Minh about it, but he couldn't figure them out either, so we just went back to making the column `text`. Not sure where it was causing things to break, but I think that's why

KevinWu098: I'd give it another shot — drizzle definitely has some bugginess with generating the migration file but it shouldn't be anything a Google search can't resolve

packages/types/src/notification.ts
IsaacNguyen: The best way to put it is that `NotificationStatus` is an object that encapsulates the types of class statuses that this specific notification is subscribed to and will trigger updates for. I don't know if I can call it a "reason" or a "status", but I had a hard time settling on the nomenclature for this idea of what a notification is notifying for. If you have any suggestions, I am happy to change things around.

apps/antalmanac/src/backend/lib/rds.ts
IsaacNguyen: I don't think I added this haha. My guess would be that this is old code that got removed from main a while ago, but it's been sitting on my old branch of AANTS when I ported things over. I'll go ahead and get rid of it.

apps/antalmanac/src/backend/routers/notifications.ts
IsaacNguyen: At least in this case, I think it's because the shape of the API call from WebSoc differs from what we want to store in the database. For instance, WebSoc returns a `term` attribute, whereas in the database for notifications, it made more sense to store a `year` and `quarter` separately.

apps/antalmanac/src/backend/lib/rds.ts
IsaacNguyen: I believe it's because the section code from WebSoc comes in as a `string`, and it conflicts with the typing in the database

apps/aants/tsconfig.json
IsaacNguyen: Not for all parts. Is it supposed to? This was the default output from when I generated the `aants` folder

packages/db/src/schema/subscription.ts
IsaacNguyen: OH I think I remember this was causing issues when we scraped WebSoc and there was a conflict between the zod schema we were using and the pgEnum in the database. I remember it happening because of some code that exists in the AANTS frontend changes. I'll put a pin in this and fix it once those changes are approved and merged in so I can test everything holistically.

apps/aants/tsconfig.json
KevinWu098: The default `tsconfig` is pretty sparse, but I suppose for this PR it will suffice! Tracking the issue now in #1379

apps/antalmanac/src/backend/lib/rds.ts
KevinWu098: :3 don't we decide what the type should be in the database? The reason why I'm being pedantic about this bit is that we have quite a lot of nasty type errors in the codebase deriving from not being able to decide if sectionCode is a number or a string, so I'd like to nip it in the bud for AANTS

My recommendation is probably to match AnteaterAPI which iirc uses strings (you may need to fact check me here)

packages/types/src/notification.ts
KevinWu098: Hm... well I think we should make an effort to be more verbose then... perhaps let's say... `notifyOn`? idk... `notificationReason`... ?? 

Something that isn't "status", as status implies this is the status of a notification (i.e. something more transient/dynamic) rather than what I think this object is supposed to represent, which is moreso a configuration for when to notify. It's not a "status" like "loading" or "completed", hence we should try to disambiguate with another name

apps/antalmanac/src/backend/lib/rds.ts
IsaacNguyen: Okay sounds good!

alexespejo (approved):
✊

PR #1380, opened 2025-12-23
fix: sat display

### LINKED ISSUE #1358: Saturday is not correctly displayed on the Calendar
When we have saturday courses, our Calendar component does not display them correctly. Our configuration in `CalendarRoot` is probably wrong.

Image

While working on this issue, please note that the calendar, by default, is defaulted to some date in 2018 iirc. However, when we render finals, we update the calendar date to be correct based on the term we're in. This may or may not be related

AureliaSindhu:
## Summary

Fixed the calendar week layout so Saturday appears after Friday instead of before Sunday. Changed the week start day from Saturday (`dow: 6`) to Sunday (`dow: 0`). 

- **Normal schedules:** Week starts on Sunday, displays Su → M → Tu → W → Th → F → Sa
- **Finals schedule:** Week starts on Saturday, displays Sa → Su → M → Tu → W → Th → F

**Before:** Sa | Su | M | Tu | W | Th | F  
**After:** Su | M | Tu | W | Th | F | Sa

Screenshot 2026-01-09 at 12 06 08 PM
Screenshot 2026-01-09 at 12 06 13 PM

## Test Plan
   - [ ] Drama 198 (Fall 2025) — Saturday and Sunday classes appear in correct columns
   - [ ] Math 2B (Fall 2025) — Saturday finals appear correctly in finals view
   - [ ] Verify day order: Su, M, Tu, W, Th, F, Sa (normal) and Sa, Su, M, Tu, W, Th, F (finals)

## Issues
Saturday was displayed for the initial week. 

Closes #1358


KevinWu098 (changes requested):
This looks like it does work for regular weeks, but on finals weeks, the week begins on Saturday. You can see this on the schedule, Fall 2025 exams are `Dec 6–12` (Dec 6 is a Saturday). 

In updating the start of the week, we have regressed in this area. Although it's outside the scope of the original issue, we should A. fix this and B. cannot proceed with this PR as it breaks finals functionality (which is more critical than the occasional Saturday class)

Your branch:
Image

Production:
 
Image

We may be able to dynamically update the start of the week based on if we're in finals viewing mode.

AureliaSindhu:
sorry i forgot to push my changes

but I created a separate locale (en-us-finals) that starts the week on Saturday, stored separately from the regular en-us locale. So it now should detects when finals start on Saturday and dynamically switches to this Saturday-start locale only during finals viewing mode.

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
```diff
@@ -75,7 +75,7 @@ export function CustomEventDialog(props: CustomEventDialogProps) {
 
     const handleOpen = useCallback(() => {
         setOpen(true);
-        setScheduleIndices(AppStore.schedule.getIndexesOfCustomEvent(props.customEvent?.customEventID ?? -1));
+        setScheduleIndices(AppStore.schedule.getIndexesOfCustomEvent(Number(props.customEvent?.customEventID ?? -1)));
```
KevinWu098, line 78: question: what does this change do?

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
```diff
@@ -28,16 +28,25 @@ import { scheduleComponentsToggleStore } from '$stores/ScheduleComponentsToggleS
...
+//  * CALENDAR_VIEWS will set the correct day range
+ * Start week on Sunday so Saturday appears after Friday.
+ * This ensures the standard week layout: Su, M, Tu, W, Th, F, Sa
  */
+// Normal schedules: Su ... Sa (Sa rightmost)
 // eslint-disable-next-line import/no-named-as-default-member
 moment.updateLocale('en-us', {
     week: {
-        dow: 6,
+        dow: 0, // Sunday = 0, Monday = 1, ..., Saturday = 6
```
KevinWu098, line 41: issue: It looks like, on the staging deploy, Saturday is the first day of the week for me on the regular view

Image

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
```diff
@@ -28,16 +28,25 @@ import { scheduleComponentsToggleStore } from '$stores/ScheduleComponentsToggleS
...
     },
 });
 
+// Finals locale: week starts Saturday (Sa ... Fr)
+moment.defineLocale('en-us-finals', {
```
KevinWu098, line 46: question: I see the other `defineLocale` has an eslint ignore. You should merge main (which has fixed eslint settings), and then see if either A. we need a eslint ignore on this line of code OR B. we can remove the warning from the other line

KevinWu098 (changes requested):
Really great work! Some minor questions regarding logic/fixed.

Additionally, I'd suggest that, when writing a PR, you include screenshots of the new behavior working (or a short video). Also, for your test plan, including courses which trigger certain cases (e.g. saturday classes, saturday final) etc would be helpful for your reviewer!

Some examples of what I used:
1. Drama 198 (Fall 2025) —  Saturday and Sunday classes
2. Math 2B (Fall 2025) — Saturday finals

apps/antalmanac/src/components/Calendar/Toolbar/CustomEventDialog/CustomEventDialog.tsx
AureliaSindhu: It makes sure customEventID is always a number before the lookup. Before, if it came in as a string like "123", the lookup sometimes failed since the store uses numeric IDs.

KevinWu098:
Is this ready for review?

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
AureliaSindhu: Screenshot 2026-01-09 at 12 06 08 PM
Screenshot 2026-01-09 at 12 06 13 PM

fixed!

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
```diff
@@ -28,17 +28,27 @@ import { scheduleComponentsToggleStore } from '$stores/ScheduleComponentsToggleS
...
 
 /*
- * Always start week on Saturday for finals potentially on weekends.
- * CALENDAR_VIEWS will set the correct day range
+//  * Always start week on Saturday for finals potentially on weekends.
+//  * CALENDAR_VIEWS will set the correct day range
+ * Start week on Sunday so Saturday appears after Friday.
+ * This ensures the standard week layout: Su, M, Tu, W, Th, F, Sa
  */
+// Normal schedules: Su ... Sa (Sa rightmost)
```
KevinWu098: nit: we should be using the block quote not regular quotes, right?

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
```diff
@@ -248,6 +258,23 @@ export const ScheduleCalendar = memo(() => {
...
+    const finalsStartsOnSaturday = showFinalsSchedule && finalsDate.getDay() === 6;
+
+    const culture = finalsStartsOnSaturday ? 'en-us-finals' : 'en-us';
+
+    const calendarLocalizer = useMemo(() => {
+        moment.locale(culture);
+        return momentLocalizer(moment);
+    }, [culture]);
+
+    const calendarView = showFinalsSchedule
```
KevinWu098: issue: this seems overly complex. can we simplify the ternary?

KevinWu098 (changes requested):
A few nits. Also, make sure to update your PR description with a demo of changes (before/after), as well as update your test plan with example courses

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
AureliaSindhu: yup, updated it!

KevinWu098 (changes requested):
hm. sorry to throw you into review again. seems like the full week is always shown for finals, even if there's no saturday/sunday final.

is this intended behavior?

KevinWu098 (approved):

KevinWu098:
Good work! Clicking the merge button is left to you :)

PR #1384, opened 2025-12-31
Add static data to the gitignore

alexespejo:
## Summary
When you run `get-data` the fetched static data should be ignored

## Test Plan
Run `pnpm get-data`: the the static data in `/backend` should not be tracked

## Issues

Closes #

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,6 @@ Desktop.ini
 
 # sst
 .sst
+
+# generated
+/apps/backend


KevinWu098 (changes requested):
Err, because we no longer have an `apps/backend` (it's now apps/antalmanac/src/backend`), I believe the remaining generated files in that specific repository are just remnants of past code.

If the goal is removing the static data altogether, which lives at `src/generated/` now, that is gitignored already at in `apps/antalmanac/.gitignore`

alexespejo:
Oh I see my mistake

PR #1385, opened 2026-01-01
Fix: Duplicate users created on sign up

alexespejo:
## Summary
There is a race condition when you first create your AA account two user rows are created. 

Local dev:
image

## Test Plan
1) Signin to AA with an email that's not in the DB 

Expected: Only 1 row should appear in the db for that user and all functionality should be the same 

## Issues

Closes #


KevinWu098 (approved):
Because it's a race condition, I don't know how to test it absolutely — but I trust that it works.

However, I think using refs to prevent race conditions is a bit of an antipattern (i.e. why do we have a race condition in the first place?). Still, I think this is fine... maybe a `TODO (@alexespejo): investigate race condition source` is warranted

PR #1387, opened 2026-01-04
AAxPP Shared sidebar

### LINKED ISSUE #1386: AA x PP Sidebar and ICSSC Footer
Port over the code from PeterPortal's [`branding-changes`](https://github.com/icssc/peterportal-client/blob/branding-changes/site/src/shared-components/SideNav.tsx) branch

xgraceyan:
> [!CAUTION]
> Do not merge until both AA and PP are ready

## Summary
Adds a shared sidebar to be used between AA and PP. Had to move some stuff (mostly colors) around so it can be used in AntAlmanac.

image
image

## Test Plan
- Ensure design is consistent with PeterPortal's sidebar
- Ensure sidebar looks good in both light and dark mode

## Issues

Closes #1386 partially (w/ about box)

### DIFF
--- a/apps/antalmanac/src/shared-components/SideNav.tsx
+++ b/apps/antalmanac/src/shared-components/SideNav.tsx
@@ -1,12 +1,11 @@
-import EventNoteIcon from '@mui/icons-material/EventNote';
-import RouteIcon from '@mui/icons-material/Route';
+import { EventNote, Route } from '@mui/icons-material';
 import { Paper, Stack, SxProps, Typography, useTheme } from '@mui/material';
 import Link from 'next/link';
 
 import { useIsMobile } from '$hooks/useIsMobile';
 import { useThemeStore } from '$stores/SettingsStore';
 
-const SideNav = () => {
+export const SideNav = () => {
     const isMobile = useIsMobile();
     const isDark = useThemeStore((store) => store.isDark);
     const theme = useTheme();
@@ -30,14 +29,14 @@ const SideNav = () => {
             <Stack direction="column" alignItems="center" sx={{ gap: '16px', paddingTop: '8px' }}>
                 <Link href="https://antalmanac.com" style={{ textDecoration: 'none' }}>
                     <Stack direction="column" alignItems="center">
-                        <EventNoteIcon fontSize="medium" />
+                        <EventNote fontSize="medium" />
                         <Typography fontSize={11}>Scheduler</Typography>
                     </Stack>
                 </Link>
 
                 <Link href="/" style={{ textDecoration: 'none' }}>
                     <Stack direction="column" alignItems="center" sx={{ color: isDark ? 'white' : 'black' }}>
-                        <RouteIcon fontSize="medium" />
+                        <Route fontSize="medium" />
                         <Typography fontSize={11}>Planner</Typography>
                     </Stack>
                 </Link>
@@ -45,5 +44,3 @@ const SideNav = () => {
         </Paper>
     );
 };
-
-export default SideNav;
--- a/apps/antalmanac/src/shared-components/SideNav.tsx
+++ b/apps/antalmanac/src/shared-components/SideNav.tsx
@@ -1,0 +1,50 @@
+import EventNoteIcon from '@mui/icons-material/EventNote';
+import RouteIcon from '@mui/icons-material/Route';
+import { Paper, Stack, SxProps, Typography, useTheme } from '@mui/material';
+import Link from 'next/link';
+
+import { useIsMobile } from '$hooks/useIsMobile';
+import { useThemeStore } from '$stores/SettingsStore';
+
+const SideNav = () => {
+    const isMobile = useIsMobile();
+    const isDark = useThemeStore((store) => store.isDark);
+    const theme = useTheme();
+
+    if (isMobile) {
+        return null;
+    }
+
+    const paperStyleOverrides: SxProps = {
+        height: '100%',
+        width: 64,
+        zIndex: 300,
+        top: 0,
+        left: 0,
+        borderRight: `1px solid ${theme.palette.divider}`,
+        borderRadius: 0,
+        flexShrink: 0,
+    };
+
+    return (
+        <Paper elevation={0} sx={paperStyleOverrides}>
+            <Stack direction="column" alignItems="center" sx={{ gap: '16px', paddingTop: '8px' }}>
+                <Link href="https://antalmanac.com" style={{ textDecoration: 'none', color: 'primary' }}>
+                    <Stack direction="column" alignItems="center">
+                        <EventNoteIcon fontSize="medium" />
+                        <Typography fontSize={11}>Scheduler</Typography>
+                    </Stack>
+                </Link>
+
+                <Link href="/" style={{ textDecoration: 'none' }}>
+                    <Stack direction="column" alignItems="center" sx={{ color: isDark ? 'white' : 'black' }}>
+                        <RouteIcon fontSize="medium" />
+                        <Typography fontSize={11}>Planner</Typography>
+                    </Stack>
+                </Link>
+            </Stack>
+        </Paper>
+    );
+};
+
+export default SideNav;
--- a/apps/antalmanac/src/shared-components/SideNav.tsx
+++ b/apps/antalmanac/src/shared-components/SideNav.tsx
@@ -28,7 +28,7 @@ const SideNav = () => {
     return (
         <Paper elevation={0} sx={paperStyleOverrides}>
             <Stack direction="column" alignItems="center" sx={{ gap: '16px', paddingTop: '8px' }}>
-                <Link href="https://antalmanac.com" style={{ textDecoration: 'none', color: 'primary' }}>
+                <Link href="https://antalmanac.com" style={{ textDecoration: 'none' }}>
                     <Stack direction="column" alignItems="center">
                         <EventNoteIcon fontSize="medium" />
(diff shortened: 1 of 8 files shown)


apps/antalmanac/src/shared-components/SideNav.tsx
```diff
@@ -0,0 +1,49 @@
...
+        top: 0,
+        left: 0,
+        borderRight: `1px solid ${theme.palette.divider}`,
+        borderRadius: 0,
+    };
+
+    return (
+        <Paper elevation={0} sx={paperStyleOverrides}>
+            <Stack direction="column" alignItems="center" sx={{ gap: '16px', paddingTop: '8px' }}>
+                <Link href="https://antalmanac.com" style={{ textDecoration: 'none' }}>
```
KevinWu098: issue: should this be antalmanac.com? why not "/"?

apps/antalmanac/src/shared-components/SideNav.tsx
```diff
@@ -0,0 +1,49 @@
...
+                        <RouteIcon fontSize="medium" />
+                        <Typography fontSize={11}>Planner</Typography>
+                    </Stack>
+                </Link>
+            </Stack>
+        </Paper>
+    );
+};
+
+export default SideNav;
```
KevinWu098: nit: we should prefer not using `default` exports when we can just export the component directly

apps/antalmanac/src/shared-components/SideNav.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { useThemeStore } from '$stores/SettingsStore';
+
+const SideNav = () => {
+    const isMobile = useIsMobile();
+    const isDark = useThemeStore((store) => store.isDark);
+    const theme = useTheme();
+
+    if (isMobile) {
+        return null;
+    }
```
KevinWu098, line 18: question: have we figured out how we're going to handle navigation between the apps on mobile?

apps/antalmanac/src/shared-components/SideNav.tsx
```diff
@@ -0,0 +1,49 @@
+import EventNoteIcon from '@mui/icons-material/EventNote';
+import RouteIcon from '@mui/icons-material/Route';
```
KevinWu098:
```suggestion
import { EventNoteIcon, RouteIcon } from '@mui/icons-material';
```
Let's just have one import

KevinWu098 (changes requested):
Nice work. it would be good to link to the relevant source from PeterPortal, where we're getting this code.

Some comments below too

apps/antalmanac/src/shared-components/SideNav.tsx
xgraceyan: The side nav taken from peterportal just had this link to antalmanac.com. I lowkey have no idea if we agreed to have antalmanac as the root, or /scheduler, or what, but I'll link this to the root

apps/antalmanac/src/shared-components/SideNav.tsx
xgraceyan: Not sure, peterportal had it return null on mobile; we can maybe talk about this next meeting

apps/antalmanac/src/shared-components/SideNav.tsx
KevinWu098: Looking at the markup, planner is antalmanac, so for us, we would want to route to "/". Below, in the other button, rather than routing to "/", we'd want to route to peterpotal, right?

xgraceyan: Wait planner is antalmanac? I'm pretty sure we were Scheduler 😭

KevinWu098: Err yes, sorry planner is PP, yes. But regardless, the links should be corrected

xgraceyan: Should have fixed the links, just pointed the other one to peterportal.org lol

KevinWu098:
@xgraceyan Looks like Caden put up PeterPortal's version of how they're doing cross-app switching, which is in the mobile sidebar. I think we can implement it similarly. Can do it this PR, or in another PR pointing towards this one

https://github.com/icssc/peterportal-client/pull/917

xgraceyan:
@KevinWu098 Okay, i've added something similar here, let me know what you think:
image

KevinWu098:
> @KevinWu098 Okay, i've added something similar here, let me know what you think: image

How does it look at full width? rather than constrained

xgraceyan:
> How does it look at full width? rather than constrained
image
Like this? Only change is on hover

KevinWu098:
> > How does it look at full width? rather than constrained
> 
> image Like this? Only change is on hover

I think it looks good :). bit hard to tell since the image is cropped, admittedly, but i think it works!

CadenLee2:
This looks great! I have a few things to note:

1. Currently on [Planner](https://staging-917.peterportal.org/), we only show the link in the hamburger/profile menu if you're on mobile (it's hidden on desktop). Our rationale was that we want users to know the sidebar is the preferred form of navigation (as opposed to getting in the habit of going into the menu). I think it's up to you if you want to replicate that behavior or not.

> > How does it look at full width? rather than constrained
>
> Like this? Only change is on hover

2. I agree, this looks good

3. Also, I noticed from testing that the current Scheduler and Planner mobile breakpoints are different (~760px vs 840px), which definitely messes up the switching experience if you're on a screen in that range. Unfortunately, Planner in its current state can't reduce the mobile breakpoint by much. Would it be reasonable for us to compromise at some shared breakpoint like 800px?

4. Finally, Planner has a base Branding Changes branch and merging all merge-related changes into it before we finally merge to main (including not just small things like styling, but also things like updating the SST config). This is somewhat a question for @KevinWu098 , but is the plan to make this PR function in the same way? Because of the scope, I think it might make more sense to follow the pattern of having a new Branding Changes branch and merge this PR into that first, so that it can all be tested and merged to main at once when the time comes.

Edit: fixed numbers

KevinWu098:
> This looks great! I have a few things to note:
> 
> 
> 
> 1. Currently on [Planner](https://staging-917.peterportal.org/), we only show the link in the hamburger/profile menu if you're on mobile (it's hidden on desktop). Our rationale was that we want users to know the sidebar is the preferred form of navigation (as opposed to getting in the habit of going into the menu). I think it's up to you if you want to replicate that behavior or not.
> 
> 
> 
> > > How does it look at full width? rather than constrained
> 
> >
> 
> > Like this? Only change is on hover
> 
> 
> 
> 2. I agree, this looks good
> 
> 
> 
> 3. Also, I noticed from testing that the current Scheduler and Planner mobile breakpoints are different (~760px vs 840px), which definitely messes up the switching experience if you're on a screen in that range. Unfortunately, Planner in its current state can't reduce the mobile breakpoint by much. Would it be reasonable for us to compromise at some shared breakpoint like 800px?
> 
> 
> 
> 4. Finally, Planner has a base Branding Changes branch and merging all merge-related changes into it before we finally merge to main (including not just small things like styling, but also things like updating the SST config). This is somewhat a question for @KevinWu098 , but is the plan to make this PR function in the same way? Because of the scope, I think it might make more sense to follow the pattern of having a new Branding Changes branch and merge this PR 
into that first, so that it can all be tested and merged to main at once when the time comes.
> 
> 
> 
> Edit: fixed numbers

1. This makes sense to me
2. :) 
3. I'm not too sure about the breakpoints, but this should be workable (IIRC AA's breakpoint is pegged to the tailwind breakpoints, so it'b nice to deviate... but we can figure this out on a follow up PR?)
4. Yes, I agree

apps/antalmanac/src/shared-components/SideNav.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <Paper elevation={0} sx={paperStyleOverrides}>
+            <Stack direction="column" alignItems="center" sx={{ gap: '16px', paddingTop: '8px' }}>
+                <Link href="/" style={{ textDecoration: 'none' }}>
+                    <Stack direction="column" alignItems="center">
+                        <EventNote fontSize="medium" />
+                        <Typography fontSize={11}>Scheduler</Typography>
+                    </Stack>
+                </Link>
+
+                {/* TODO: Fix PP link after merge */}
```
KevinWu098: nit (non-blocking): When writing TODOs, it is good practice to append the writer's name (or whomever is responsible) to the TODO (ex: `TODO (@KevinWu098)`. Serves to self-document who should deal with it

KevinWu098 (approved):
LGTM. I've gone ahead and created a branch, `scheduler`, which I've pointed this PR towards. 

Merge at will, but do make sure to log an issue for 1. the mobile breakpoints (see @CadenLee2's comments) and 2. updating the links. All of our changes related to the merge will go into this `scheduler` dev branch

KevinWu098:
@xgraceyan 
SCR-20260105-kmio

SCR-20260105-kmhr

The icons and font are different between the sidebar, could we get this fixed in this PR?

xgraceyan:
> @xgraceyan SCR-20260105-kmio
> 
> SCR-20260105-kmhr
> The icons and font are different between the sidebar, could we get this fixed in this PR?

Should be the same size now, I think our base font sizes are different which caused the inconsistency. Ideally I think our MUI config should be the same. Not sure what the difference in font is

KevinWu098:
Feel free to merge once you've created the appropriate issues from above! @xgraceyan

PR #1395, opened 2026-01-06
chore: add next-env.d.ts to .gitignore

dsnsgithub:
## Summary
This file is auto generated and does not need to be committed. 

## Test Plan
`next-env.d.ts` is created regardless if added to repo.


alexespejo (approved):
LGTM!

Issue #1258, opened 2025-05-19
Eslint and type errors should block commits


KevinWu098:
We've been too lax in allowing type errors and lint errors to hang around in the codebase. 

A PR resolving this issue should:
1. aim to resolve as many lint and type errors as possible (clearly flagging where they did not / could not)
2. Reenable type errors throwing and lint errors throwing

PR #1397, opened 2026-01-06
fix: eslint

### LINKED ISSUE #1371: Fix eslint problems

dsnsgithub:
## Summary
Fixes all eslint errors and warnings.

Special thanks to Warp for fixing the missing `posthog` dependency across 10 files for me 😄.

For https://github.com/icssc/AntAlmanac/commit/7793e95cf663014f85ea5aed71ca1a90d1a36fbf, I added comments to ignore eslint warnings because of https://github.com/icssc/AntAlmanac/commit/9a4dff5dbaf493a7f44aec7e5d4f0cabfcdde716.

Next.js had the option to lint during builds (removed in Next 16), but this was ignored in the config, so there is no need for it anymore. ESLint should block ~~commits~~ PRs, but that is a [different issue](https://github.com/icssc/AntAlmanac/issues/1258).

## Test Plan
`pnpm dev` no longer outputs:
```
⚠ `eslint` configuration in next.config.mjs is no longer supported. See more info here: https://nextjs.org/docs/app/api-reference/cli/next#next-lint-options
⚠ Invalid next.config.mjs options detected:
⚠     Unrecognized key(s) in object: 'eslint'
⚠ See more info here: https://nextjs.org/docs/messages/invalid-next-config
 ```
 
 `pnpm build` works.
 
 `pnpm lint` does not fail like before.
 
 `pnpm test` still succeeds.
 
 
None of my commits should have changed functionality, at most adding `useCallback` in https://github.com/icssc/AntAlmanac/commit/4d95bc03c5b647426f2b24b81be6a7e842557d51 to satisfy eslint.

## Issues

Closes #1371

### CODE REVIEW

**Comment by dsnsgithub** [comment]
In terms of the remaining eslint warnings, we could either use comments to ignore them if the usage of `any` is justified or find a better type to replace it, it depends on what you want.

**Comment by KevinWu098** [comment]
> In terms of the remaining eslint warnings, we could either use comments to ignore them if the usage of `any` is justified or find a better type to replace it, it depends on what you want.

The database types are kind of bunk rn, so IIRC that's why I just slapped an `any` down. if it can be typed properly via drizzle types, that would be the goal. this is potentially better resolved in a dedicated "fix database drizzle type usage" PR. For Jay's code, does `never` work?

**Comment by dsnsgithub** [comment]
> > In terms of the remaining eslint warnings, we could either use comments to ignore them if the usage of `any` is justified or find a better type to replace it, it depends on what you want.
> 
> The database types are kind of bunk rn, so IIRC that's why I just slapped an `any` down. if it can be typed properly via drizzle types, that would be the goal. this is potentially better resolved in a dedicated "fix database drizzle type usage" PR. For Jay's code, does `never` work?

`never` doesn't work because `attributes` is assigned.

```ts
const { attributes, isDragging, listeners, setNodeRef, setActivatorNodeRef, transform } = useSortable({
    id,
});

const context = useMemo(
    () => ({
        attributes,
        listeners,
        ref: setActivatorNodeRef,
    }),
    [attributes, listeners, setActivatorNodeRef]
);
```


I like 

```ts
interface Context {
    attributes: DraggableAttributes | {};
    listeners: DraggableSyntheticListeners;
    ref(node: HTMLElement | null): void;
}
```

since `DraggableAttributes` is what useSortable gave `attributes`.

I am just wondering if `any` was chosen for a particular reason.

dsnsgithub:
I will just add this change for now, it can be changed if needed.

ESLint isn't a fan of `{}`, changed it to `null` in my commit.

dsnsgithub:
> > In terms of the remaining eslint warnings, we could either use comments to ignore them if the usage of `any` is justified or find a better type to replace it, it depends on what you want.
> 
> The database types are kind of bunk rn, so IIRC that's why I just slapped an `any` down. if it can be typed properly via drizzle types, that would be the goal. this is potentially better resolved in a dedicated "fix database drizzle type usage" PR. For Jay's code, does `never` work?

https://github.com/icssc/AntAlmanac/pull/1398 should fix the database types and remove `any`.

Merging it alongside this PR will eliminate all ESLint errors and warnings.

alexespejo:
prepare for the wildest pr review ever

alexespejo (approved):
LGTM

PR #1398, opened 2026-01-06
fix: typescript problems

### LINKED ISSUE #1372: Fix typescript problems

Choollol:
## Summary

Fix most of the type errors in the project.

## Test Plan

1. Ensure that there are no type errors in the project (except for places like generated files, e.g. validator.ts)
2. Also make sure the changes haven't made any breaking changes
- Download-schedule button
- Custom events
- Add schedule (default-generated name)
- Settings menu styles are the same
- Color picker
- AdvancedSearchTextFields has aria-labelledby prop
- Auth and map

## Issues

Closes #1372


dsnsgithub:
~~Can you sync your fork to the `main` branch (along with your fork's branch)? It should fix the staging deploy.~~
Edit: I merged `main` for you, don't worry about the staging failure, it isn't your fault.


Thanks for the PR! It is going to be (and already is) a great help during future development.

dsnsgithub:
Fixed the remaining Next.js type errors and added `pnpm typecheck` to help in the future.

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -309,16 +313,26 @@ export const loadSchedule = async (
...
-                    AppStore.loadSkeletonSchedule(scheduleSaveState);
+                    error = true;
+                }
+
+                if (error) {
                     openSnackbar(
                         'error',
                         `Network error loading course information for "${providerId}". 	              
-                        If this continues to happen, please submit a feedback form.`
+                            If this continues to happen, please submit a feedback form.`
```
dsnsgithub: nit: I know the whitespace with the snackbar is pretty jank, but is this necessary?

Choollol: Whoops, that might have been Prettier auto-formatting when I saved. Do you want me to undo that change?

dsnsgithub: Sure, the snackbar strings seem to already include a weird mix of whitespace that should be dealt with at a later time, for now just leave it as it was.

Choollol: Just pushed a commit that fixes it!

dsnsgithub: Resolve this conversation and I will re-review :fire:

apps/antalmanac/src/backend/lib/rds.ts
```diff
@@ -622,7 +619,7 @@ export class RDS {
                 .then((res) => {
                     return { users: res[0].users, accounts: res[0].accounts };
                 })
-        );
+        ) as Promise<{ users: User; accounts: Account }>;
```
dsnsgithub: This solves the typescript errors but we shouldn't have to do this because we will be losing type safety. Was this the only way to make it work?

dsnsgithub (changes requested):
From my initial testing, there are no breaking changes and functionality remains intact. 

Hopefully we can iron out the "bunk" DB types.

apps/antalmanac/src/backend/lib/rds.ts
Choollol: If we let typescript deduce the type, we get a bunch of type errors that have to do with null not being able to be assigned to optional properties (which can only take undefined). The functions with those optional properties in their parameter types mostly have inline type declarations, so we could add null to each of those. There is one usage in Signout.tsx that uses a schema-inferred type though (exported from user.ts), and that doesn't have null as an option.

I can think of a couple options, we could either have a unified User/UserInfo type with the added nulls, or we could change the schema which might be risky. Or we could keep the type cast which does lose some safety.

dsnsgithub: My concern is that `User` type is different from the type of the DB user, in that the `User` type has a `userData` key that is added after.

For instance:

The DBuser type is passed under `{ userInfo: users }`, where `users` is actually the DBuser
https://github.com/icssc/AntAlmanac/blob/be4291cb38ec1859ff5b3d6fc64bc285597843d4/apps/antalmanac/src/actions/AppStoreActions.ts#L29

https://github.com/icssc/AntAlmanac/blob/be4291cb38ec1859ff5b3d6fc64bc285597843d4/apps/antalmanac/src/actions/ActionTypesStore.ts#L147

If `users` was actually the `User` type, `userData` would be there under `userInfo.userData`, but it shouldn't and isn't suppose to be there.

https://github.com/icssc/AntAlmanac/blob/be4291cb38ec1859ff5b3d6fc64bc285597843d4/apps/antalmanac/src/actions/AppStoreActions.ts#L129-L138





To move between the DB user and the normal user which **we should not do** because none of the functions are expecting `User` (thanks https://deepwiki.com/icssc/AntAlmanac):
```ts
static async getUserAndAccountBySessionToken(  
      db: DatabaseOrTransaction,  
      refreshToken: string  
  ): Promise<{ users: User; accounts: Account }> {  
      return db.transaction(async (tx) => {  
          const result = await tx  
              .select()  
              .from(sessions)  
              .innerJoin(users, eq(sessions.userId, users.id))  
              .innerJoin(accounts, eq(users.id, accounts.userId))  
              .where(eq(sessions.refreshToken, refreshToken))  
              .execute();  
    
          if (result.length === 0) {  
              throw new Error('Session not found');  
          }  
    
          const dbUser = result[0].users;  
          const dbAccount = result[0].accounts;  
    
          // Transform database user to application User type  
          const appUser: User = {  
              id: dbUser.id,  
              userData: await RDS.getUserDataByUid(tx, dbUser.id).then(data => data?.userData || { schedules: [], scheduleIndex: 0 }),  
              name: dbUser.name || undefined,  
              email: dbUser.email || undefined,  
              avatar: dbUser.avatar || undefined,  
              imported: dbUser.imported || undefined,  
          };  
    
          return { users: appUser, accounts: dbAccount };  
      });  
  }
```

---

Edit: This AI code snippet from DeepWiki has some bugs:
`dbUser.imported || undefined` will return undefined when `dbUser.imported` is false. 
The use of `|| undefined` is also not great because of potential bugs with falsy values.

I shouldn't have even included it, just thought it would be helpful to illustrate the differences between the DB user type and the normal User type.

--- 

I think the best course of action might be using the types given by Drizzle (maybe adding the null/undefined checks that you said with the DB types ~~similar to the snippet above~~) and making sure the functions that consume it have accurate types.

dsnsgithub: cc: @KevinWu098 thoughts?

Choollol: Just trying to understand what you're proposing at the end of your reply, are you saying we could do something similar to the snippet but without making `users` a `User` and just add the null checks, then change the types that the other functions expect?

That sounds good to me, just one small problem in `Signout.tsx` that keeps a state variable of type `User`. We can probably change this though, since it only uses the `avatar` and `name` properties, so I agree with the idea of returning the DB type.

dsnsgithub: Sorry for being unclear, it is pretty much the first option that you proposed in your previous comment.

Leave the DB types alone:

```ts
static async getUserAndAccountBySessionToken(db: DatabaseOrTransaction, refreshToken: string) {
    return db.transaction((tx) =>
        tx
            .select()
            .from(sessions)
            .innerJoin(users, eq(sessions.userId, users.id))
            .innerJoin(accounts, eq(users.id, accounts.userId))
            .where(eq(sessions.refreshToken, refreshToken))
            .execute()
            .then((res) => {
                return { users: res[0].users, accounts: res[0].accounts };
            })
    );
}
```

Add the null checks like you said for the optional types.

Fix `Signout.tsx` types, it should not be expecting `User` type if it is calling this function.

KevinWu098: I'm pretty OOTL here, but as long as things compile and continue to work, i trust y'all :3

dsnsgithub: > I'm pretty OOTL here, but as long as things compile and continue to work, i trust y'all :3

I think if you were "ITL", you would wonder how things **are** still compiling and continuing to work 🤣.
I didn't even know you could make a Next.js build with TypeScript/ESLint errors until now, it usually blocks you.

This PR should help resolve these mysteries though (thanks @Choollol!) 😄

Choollol: Just pushed another commit that adds the null options! That should (hopefully) be the last of the type errors, at least for this function.

apps/antalmanac/src/lib/tourExampleGeneration.ts
```diff
@@ -202,6 +202,7 @@ export function sampleClassFactory({
     finalExam,
 }: sampleClassOptions): ScheduleCourse {
```
dsnsgithub, line 203: Anteater API changed their schema/types in the time between creating this PR and reviewing it (https://github.com/icssc/anteater-api/pull/272). 

The existing code is extremely fragile as is, breaking and creating a TypeScript error as soon as the API modifies their schema in any way. `pnpm install` should not be able to create TypeScript errors and this implementation needs to be changed when type checking is enforced.

apps/antalmanac/src/lib/tourExampleGeneration.ts
dsnsgithub, line 222: If there are better solutions, let me know, but for now we can add `isCancelled` to fix the type error and I will approve the PR on the next review.

```diff
@@ -219,6 +219,7 @@ export function sampleClassFactory({
                 sectionEnrolled: '500',
                 totalEnrolled: '500',
             },
+            isCancelled: false,
             numNewOnlyReserved: '0',
             numOnWaitlist: '99',
             numRequested: '0',
```

Choollol: Done!

Choollol: If there's a better way to do this, to avoid type errors like this when database schemas change, please let me know too! I'm not sure how to solve this either and it'd probably be good to know.

dsnsgithub (approved):
LGTM!

PR #1406, opened 2026-01-07
ci: sst deploy environment

KevinWu098:
## Summary
This PR adds a unique staging environment to the deploy workflow. Previously, new deploys to `staging` (static) name would overwrite one another as far as github was concerned (the urls would stay deployed though

Bug was introduced in #1237

## Issues

Closes #


alexespejo (approved):
LGTM!

PR #1412, opened 2026-01-09
fix: SST locking on simultaneous deployment

dsnsgithub:
## Summary
Add concurrency to deployment workflows.
Prevent SST from locking due to concurrent deployments. Only one instance of the workflow for a given stage can run.

Since SST can presumably simultaneously deploy as long as the stage is different, I have accounted for this.

I chose `cancel-in-progress: false` for the SST workflows, which is actually the default behavior for concurrency groups (and slightly redundant), because I want to emphasize that I don't want the workflow cancelling while running `sst deploy`.

## Issues

Part of https://github.com/icssc/AntAlmanac/commit/580d1f293197efa9e58178e71c8a8fc1d2e7c301 (https://github.com/icssc/AntAlmanac/pull/1392) is probably unnecessary.

### DIFF
--- a/.github/workflows/deploy_production.yml
+++ b/.github/workflows/deploy_production.yml
@@ -4,6 +4,10 @@ on:
     push:
         branches: [main]
 
+concurrency:
+    group: sst-production-deploy
+    cancel-in-progress: false
+
 env:
     AWS_REGION: ${{ secrets.AWS_REGION }}
 
--- a/.github/workflows/deploy_production.yml
+++ b/.github/workflows/deploy_production.yml
@@ -5,7 +5,7 @@ on:
         branches: [main]
 
 concurrency:
-    group: sst-production-deploy
+    group: sst-production
     cancel-in-progress: false
 
 env:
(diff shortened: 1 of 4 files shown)


KevinWu098 (approved):
Looks good to me (although I don't have much experience writing workflow scripts), trusting you on this one!

Does it make sense to also apply this to the `destroy` workflow?

dsnsgithub:
Now that I think about it, yes. It might be possible to have conflict, maybe closing the PR, then quickly reopening the PR. `sst remove` and `sst deploy` could clash, then if I close the PR once again we could then have `sst remove`, `sst deploy`, and another `sst remove` all running at the same time. I haven't tested SST commands like this and SST docs only specify `sst deploy` as locking, but to be safe I will just add `destroy` to the same staging group.

dsnsgithub:
I was so confused why it was cancelling, the reason is only the unrelated test workflow is cancelling. The `test` workflow has `cancel-in-progress: true`. I chose `cancel-in-progress: false` for the sst workflows, which is actually the default behavior for concurrency groups (and slightly redundant), because I want to emphasize that I don't want the workflow cancelling while running `sst deploy`.


This will skip workflows that haven't started yet ("pending") and are no longer needed instead of running them, which is perfect for our use case since you will only see the final commit in staging anyway.

KevinWu098 (approved):

PR #1413, opened 2026-01-10
feat: AntAlmanac Scheduler

KevinWu098:
> [!CAUTION]
> Do not merge this PR unless explicitly authorized by leadership (Alex, Kevin, Caden)

## Summary
Staging environment for AntAlmanac Scheduler

## Issues

Closes #

PR #1414, opened 2026-01-11
feat: move sidebar/drawer functionality into profile popover

### LINKED ISSUE #1407: AA x PP Move Sidebar to Popover Menu
We want to unify the layouts between AntAlmanac and PeterPortal. We've decided to remove our sidebar and mimic PeterPortal's dropdown menu that appears when clicking on the user's profile. We will port all of the features in the sidebar into the drop down (themes, time, experimental features) 

Reference [PeterPortal's PR](https://github.com/icssc/peterportal-client/pull/923) for how they handle this menu

### PeterPortal menu

Image

calebwongg:
## Summary
Unified the layouts between AntAlmanac and PeterPortal. Removed our sidebar to mimic PeterPortal's dropdown menu that appears when clicking on the user's profile .Ported all of the features from the sidebar into the drop down (themes, time, experimental features)
image
image
image
## Test Plan
Theme tabs work, experimental features work, and log out works
## Issues

Closes #1407


KevinWu098:
Haven't seen the deployment yet, but just from a cursory glance, i'm noticing that the signed in button doesn't quite match the PP implementation -> 

Screenshot 2026-01-11 at 3 43 52 AM

apps/antalmanac/src/components/Header/Signout.tsx
```diff
@@ -79,22 +80,51 @@ export function Signout() {
...
+                            border: '1px solid',
+                            borderColor: 'background.default',
+                        },
+                    },
                 }}
             >
-                <MenuItem onClick={handleLogout}>
+                <SettingsMenu user={user} />
+
+                <Divider style={{ marginTop: '10px', marginBottom: '12px' }}/>
```
calebwongg: I eyeballed A LOT of padding and font sizes, please let me know if there is a better way to style these / if they should be fixed. All of the font sizes using h1, h2, etc. are copied from peterportal but the font sizes using rem are all pretty much eyeballed

apps/antalmanac/src/components/Header/SettingsMenu.tsx
```diff
@@ -52,114 +50,141 @@ function ThemeMenu() {
...
+                            sx={{
+                                flex: 1,
+                                display: 'flex',
+                                alignItems: 'center',
+                                justifyContent: 'center',
+                                gap: 0.5,
+                                padding: '8px 12px',
+                                cursor: 'pointer',
+                                fontWeight: 'bold',
+                                fontSize: '1.1rem',
```
calebwongg: Peterportal has hardcoded font sizes? I wasn't too sure what to use

apps/antalmanac/src/components/Header/SettingsMenu.tsx
```diff
@@ -52,114 +50,144 @@ function ThemeMenu() {
...
+                            key={tab.value}
+                            onClick={() => handleThemeChange(tab.value as 'light' | 'dark' | 'system')}
+                            sx={{
+                                flex: 1,
+                                display: 'flex',
+                                alignItems: 'center',
+                                justifyContent: 'center',
+                                minwidth: 0,
+                                overflow: 'hidden',
+                                textOverflow: 'ellipsis',
```
calebwongg: `minwidth: 0,
overflow: 'hidden',
textOverflow: 'ellipsis',`

There SHOULDN'T be any issues where overflow handling is needed but i added it just in case? Let me know if I should delete it

apps/antalmanac/src/components/Header/Signout.tsx
```diff
@@ -78,23 +80,53 @@ export function Signout() {
...
+                    horizontal: 'right',
+                }}
+                transformOrigin={{
+                    vertical: 'top',
+                    horizontal: 'right',
+                }}
+                slotProps={{
+                    paper: {
+                        sx: {
+                            width: {xs: 200, sm: 260, md: 330},
```
calebwongg: Peterportal popover has a fixed width which works because the elements within the popover are also fixed. 

In one of my above comments I talked about having a dynamic font size (I used 1.1rem), so I added this responsiveness. Again, not sure if we also want to have our popover size fixed. Also, I just chose sizes that "looked good",  but the sizes are kind of just randomly chosen

apps/antalmanac/src/components/Header/Signout.tsx
Awesome-E: If it helps, PeterPortal's root font size is `16px`, which means that every `1rem` should be `16px`.

apps/antalmanac/src/components/Header/Signout.tsx
Awesome-E: I don't think this necessarily has to be pixel-perfect since it requires an additional click to open on both sides. The reason PeterPortal's is fixed instead of going down below 300px is because long usernames or emails are more likely to cause overflow when we fit the content that way.

If it looks good with a long enough name and username, I don't think this is needs to be tweaked that more more from a visual standpoint. The only thing I would consider, if not outsize effort, is the color of the buttons since on Planner's side we use the dark blue.

apps/antalmanac/src/components/Header/Signout.tsx
```diff
@@ -67,7 +69,7 @@ export function Signout() {
                 aria-haspopup="true"
                 aria-expanded={open ? 'true' : undefined}
                 onClick={handleClick}
-                sx={{ width: 'fit-content' }}
+                sx={{ width: 'fit-content', borderRadius: 4 }}
```
CadenLee2: To fix the difference that Kevin pointed out, you can check out the shared component here: https://github.com/icssc/peterportal-client/pull/923/changes (`site/src/shared-components/ProfileMenuButtons.tsx`)

This component isn't necessarily 1-1 compatible with AA (e.g. you will need to add in the avatar/account circle icon code, and may need to restructure some things) but can be used for reference. Ideally you can copy the shared component from the PeterPortal repo into the AntAlmanac shared components folder, then make necessary adjustments, use it here and work from there.

apps/antalmanac/src/components/Header/SettingsMenu.tsx
```diff
@@ -220,69 +250,62 @@ function ExperimentalMenu() {
...
-    }, []);
-
-    const handleDrawerClose = useCallback(() => {
-        setDrawerOpen(false);
-    }, []);
-
-    return (
-        <>
-            <IconButton onClick={handleDrawerOpen} color="inherit" size="large" style={{ padding: '4px' }}>
-                <MenuRounded />
```
alexespejo, line 316: issue: When the user is signed out we should allow them to modify some of the settings like themes, time, etc. PeterPortal keeps the hamburger menu which will open the popover. 

Refernce to how PP handles signed out users: 
Image

sicn4rf:
image
image

issue: the new popover menu is missing these buttons that used to exist at the bottom of the sidebar menu

apps/antalmanac/src/components/Header/Signout.tsx
calebwongg: I'm not sure why it happened, because peter portal appears to be using the mui primary main color. I used this exact same color for AntAlmanac, but it still looks different.

KevinWu098:
Noting that the time setting is not an experimental feature

alexespejo:
Resolve the merge conflict and the git diff seems off I think it should resolve the random files in the git diff too

alexespejo:
I can't lie gang i think that added more to the git diff. Sorry I should have specified that the merge conflict is coming from the `scheduler` branch not `main`

calebwongg:
> I can't lie gang i think that added more to the git diff. Sorry I should have specified that the merge conflict is coming from the `scheduler` branch not `main`

Yeah my bad im working on it right now

apps/antalmanac/src/components/Header/ProfileMenuButtons.tsx
```diff
@@ -0,0 +1,66 @@
...
+import Image from 'next/image';
+import MenuIcon from '@mui/icons-material/Menu';
+import { User } from '@packages/antalmanac-types';
+
+interface ProfileMenuButtonsProps {
+    user: User | null;
+    handleOpen: (event: React.MouseEvent<HTMLElement>) => void;
+    handleSettingsOpen: (event: React.MouseEvent<HTMLElement>) => void;
+}
+const ProfileMenuButtons: FC<ProfileMenuButtonsProps> = ({ user, handleOpen, handleSettingsOpen }) => {
```
alexespejo: nit: prefer exporting non-default components. 
```suggestion
export const ProfileMenuButtons: FC<ProfileMenuButtonsProps> = ({ user, handleOpen, handleSettingsOpen }) => {
```

KevinWu098 (changes requested):
*squint* i have no idea why this branch has so many diffs from other PRs that aren't getting resolved by merging in the upstream...

that aside, things are looking really solid.

Image

I have two styling related points.
1. the logout button looks like it has no padding
2. the donate/about/feedback buttons look like they're `justify-around` rather than `justify-between` (at least visually, i haven't read the code). put another way, there's space to the left and right of that button group so it doesn't align correctly.

After these are addressed, and if the diff could get cleaned up (perhaps this requires a rebase?), i'll dive into the code

KevinWu098:
Other side nit is that the exact positioning of the popover != planner's positioning, but I think that's also downstream of the font/spacing theme differences. just wanted to note it (though no action required presently)

CadenLee2:
> the exact positioning of the popover != planner's positioning, but I think that's also downstream

Agree, this is nearly impossible to notice unless you're explicitly attempting to compare the two.

Some small styling notes include:
1. The primary blue color is indeed different for dark theme between AA and PP currently. (AA's primary gets set to "dodger blue" in dark theme, which is different from the brand darker blue. Searching AA files for `import { BLUE } from '$src/globals';` should show some examples of using the darker blue, which should be an easy change)
2. Hovering the "about" button shows that it's a shorter height than the other two buttons. This is a small detail but ideally should be fast to fix:
image


Otherwise, looks like most of the other issues mentioned in the above conversation are fixed! Note that one idea for the buttons is using a footer like this, though I'd consider that out of scope:
image

KevinWu098 (changes requested):
Image

Notes:
1. Smth happening with overflow on the added page, and only on the added page.
2. When we're loading in, there's no loader on the sign in button anymore — is this intended?

calebwongg:
> Image
> Notes:
> 
> 1. Smth happening with overflow on the added page, and only on the added page.
> 2. When we're loading in, there's no loader on the sign in button anymore — is this intended?

The overflow bug seems to be universal on the scheduler branch. I tested it with the deployment from #1413 and the same issue persists. It also breaks when you switch to the added tab on mobile. 

I did not intend on removing the loader ... will add it back.

apps/antalmanac/src/app/Theme.tsx
```diff
@@ -30,7 +30,7 @@ const lightTheme: PaletteOptions = {
 
 const darkTheme: PaletteOptions = {
     primary: {
-        main: DODGER_BLUE,
+        main: BLUE,
```
KevinWu098: issue: this is a styling regression — we intentionally use the lighter DODGER_BLUE (AA theme) 

Before
Image

After
Image

suggestion: if you're aiming to update the colors in the popover, I would simply apply { BLUE } in the component itself

apps/antalmanac/next.config.mjs
```diff
@@ -7,6 +7,9 @@ const nextConfig = {
     eslint: {
         ignoreDuringBuilds: true,
     },
+    images: {
+        unoptimized: true,
+    },
```
KevinWu098, line 17: question: is this part of the diff? what is the purpose, if it is?

apps/antalmanac/src/components/Header/Header.tsx
```diff
@@ -65,11 +64,12 @@ export function Header() {
...
 
                 <Stack direction="row" sx={{ alignItems: 'center' }}>
-                    <Import key="studylist" />
                     <Save />
+                    <Import key="studylist" />
                     {sessionIsValid ? <Signout /> : <Signin />}
-                    <AppDrawer key="settings" />
                 </Stack>
+                
+
```
KevinWu098: issue: this diff seems extraneous — we do in fact want the ordering to be Import, Save, Menu

see #1410

KevinWu098 (commented):
I need to read the thick of it @profilemenubuttons, but here's a few things i caught

apps/antalmanac/next.config.mjs
calebwongg: Adding this lets us use `'next/image'` to display the user avatars. Next automatically optimizes images but it only works for certain allowed domains for external images/images not in the public folder. The user avatar was something like a `lh3.googleusercontent.com` which was unallowed, so adding this lets us render the image. Peterportal has the same config in their project as well.

I used the nextjs `Image `component because this is consistent with the way peterportal displays their user avatars. Before we were using mui `Avatar`, which I honestly think is a better option for this specific project, because when I use `Image` we need to check to make sure the user avatar isn't undefined with `{avatar ? (` and `{user.avatar ? (`. We need this check because the declaration for a User in `'@packages/antalmanac-types'` declares the user avatar as `'avatar?': 'string',`.

We display the users avatar two times. Once in the header and once in in the popover:
```
 return (
        <Button sx={profileButtonStyles} onClick={handleOpen} variant="text" color="inherit">
            {avatar ? (
                <Image src={avatar} alt={name ?? 'User avatar'} width={24} height={24} style={profilePicStyles} />
            ) : (
                <AccountCircleIcon sx={{ width: 24, height: 24 }} />
            )}
            <MenuIcon />
        </Button>
    );

```
and 

```
<Box sx={{ display: 'flex', alignItems: 'center', gap: 2, mb: 2 }}>
             {user.avatar ? (
                 <Image
                    src={user.avatar}
                    alt={user.name ?? 'User'}
                    width={50}
                    height={50}
                    style={{ borderRadius: '50%', objectFit: 'cover' }}
                />
            ) : (
                <AccountCircleIcon sx={{ width: 50, height: 50 }} />
            )}

```

To reference peter portal's code for displaying their user avatar, it looks like: 
```  <div className="navbar-profile">
      <button className="profile-button" onClick={handleClick}>
        <Image src={picture} alt={name} className="navbar-profile-pic" width={36} height={36} />
      </button>
```
and 
```
<div>
      <div className="profile-popover-header">
        <Image src={picture} alt={name} width="50" height="50" />
        <div>
          <h1 title={name}>{name}</h1>
          <h2 title={email}>{email}</h2>
        </div>
```
        
Theirs looks cleaner and they don't need the check because their `User` type definition is as followed: 
```
export interface User {
  /**
   * google id
   */
  id: string;
  email: string;
  name: string;
  picture: string;
}
```

So using mui `Avatar` would probably be better for us, but it would make more inconsistencies between us and peterportal. Let me know what you think!

calebwongg:
> I need to read the thick of it @profilemenubuttons, but here's a few things i caught

For reference, profilemenubuttons is almost identical to the component from Peter portal.

KevinWu098 (approved):
In the interest of time, I pushed some minor refactors (minor in terms of logic changed which is ~= 0) for readability. Otherwise, this PR LGTM.

I'll note for posterity that I think 
1. the large amount of custom styling is a code smell, but I can live with it
2. the time/theme components do look... not so good with the `BLUE`... but I am going to defer further action on this (perhaps in late w3 we can take another pass at color choices)
3. we probably should look at something like [MUI toggle button](https://mui.com/material-ui/react-toggle-button/) when we refactor this

CC @CadenLee2 for sign off on appearance

CadenLee2 (approved):
> 1. the large amount of custom styling is a code smell, but I can live with it
> 2. the time/theme components do look... not so good with the BLUE... but I am going to defer further action on this (perhaps in late w3 we can take another pass at color choices)
> 3. we probably should look at something like [MUI toggle button](https://mui.com/material-ui/react-toggle-button/) when we refactor this

I agree with all of your points, particularly that we should merge this now then touch up later if/when we have time.

1. Hopefully would be fixed as we start to align our themes later on
2. Yep, at least on Planner side in dark theme the selected background is blue but non-selected text is our lighter blue. This is also related partly to theming so isn't something that needs to be changed rn
3. Mui toggle button is a good idea. I noticed that the current Scheduler profile menu buttons don't have the MUI ripple effect, which makes it inconsistent, and on the Planner side we definitely could replace our homegrown selectors with toggle buttons at some point

But yes, these all can be addressed in future PRs

PR #1416, opened 2026-01-12
staging: test

dsnsgithub:
> [!CAUTION]
> Do not merge this PR, this PR is only to provide staging for https://github.com/icssc/AntAlmanac/pull/1398.

Issue #1375, opened 2025-12-15
Delete stale, orphaned infra


KevinWu098:

KevinWu098:
Eh, an ongoing chud task

PR #1417, opened 2026-01-12
ci: staging

dsnsgithub:
> [!CAUTION]
> Do not merge this PR, this PR is only a staging test. 

Staging is failing, see https://github.com/icssc/AntAlmanac/pull/1416#issuecomment-3736546030

Seems to be failing at PeterPortal too.


KevinWu098:
Hm -- should be an easy fix, albeit manual one. Loosely tracked in #1375, since this is more of an org level problem with orphaned resources than an AA one.

Will reply here once I've deleted a few KV stores

KevinWu098:
> Hm -- should be an easy fix, albeit manual one. Loosely tracked in #1375, since this is more of an org level problem with orphaned resources than an AA one.
> 
> Will reply here once I've deleted a few KV stores

Went in and deleted 6 KV stores manually and requested a service quota increase with AWS — should give a little bit of breathing room. thanks for flagging

PR #1421, opened 2026-01-14
Fix: Refactored backend to fetch course data with schedule

### LINKED ISSUE #1059: Fetch grades info with schedule
## Description

- Right now, we fetch the user schedule data, then the course data, resulting in slow schedule load times.
- Now, there's no reason to do that since we're proxying AAPI anyway. We should just combine the data in the backend.

vicksey:
## Summary
Refactored the backend to return a hydrated schedule (schedule + course data in one response) instead of requiring the frontend to make n separate websoc requests per course.

Note: The legacy ScheduleSaveState type is kept for backward compatibility (fallback scenarios, skeleton loading, imports, db saving)

## Test Plan
- [ ] Log in to an account with saved schedules
- [ ] Import schedule
- [ ] Load legacy?
- [ ] Make sure custom events load correctly

## Issues

Closes #1059


alexespejo:
Sorry for ghosting this. I'm looking through codebase and stuff to compose my review ✍️

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -97,6 +99,33 @@ export function isEmptySchedule(schedules: ShortCourseSchedule[]) {
     return true;
 }
 
+const isHydratedSaveState = (
+    saveState: HydratedScheduleSaveState | ScheduleSaveState
```
alexespejo: question: I've been trying to trace the code where the new `HydratedScheduleSaveState` is used. Do you keep the `ScheduleSaveState` to support legacy data? I think it would make sense to refactor the original data type rather than injecting a new optional type ontop of it.

apps/antalmanac/src/backend/lib/websoc-service.ts
```diff
@@ -0,0 +1,79 @@
+import type { WebsocAPIResponse, CourseInfo, WebsocSectionType } from '@packages/antalmanac-types';
```
alexespejo: issue: The code seems to come from [`websoc.ts`](https://github.com/icssc/AntAlmanac/blob/vz/fetch-grades-w-schedules/apps/antalmanac/src/backend/routers/websoc.ts). Try and consolidate the functionality across the two so we're not repeating code

apps/antalmanac/src/backend/routers/userData.ts
```diff
@@ -37,6 +38,78 @@ const saveGoogleSchema = type({
...
 });
 
+/**
+ * Hydrates schedule courses with full course information from WebSOC.
+ * Transforms ShortCourse[] to ScheduleCourse[] by fetching course details.
+ */
+async function hydrateScheduleCourses(schedules: ShortCourseSchedule[]): Promise<Array<{
+    scheduleName: string;
+    courses: ScheduleCourse[];
+    customEvents: any[];
```
alexespejo: issue: will prob throw an error

alexespejo (changes requested):
Sorry for the very delayed review. it took me a while to figure out how stuff works :p

apps/antalmanac/src/actions/AppStoreActions.ts
vicksey: ok i've looked more into it:
- the short schedule format is what's being stored to an account/user in the db (its called ScheduleSaveState throughout our codebase lol)
- the short state is used for skeleton fallback when hydration fails, so essentially legacy workflow will occur if fails
- converting from hydrated to short on the frontend is easy anyways since we're just trimming data
(there are two functions that do it toShortScheduleSaveState() in AppStoreActions.ts and toShortState() in AppStore.ts)
- The hydrated state is mostly a performance/efficiency thing so that we aren't spamming websoc requests for each course like we used to which essentially addresses the issue

i mean it'd be nice if we could figure out to get hydrated schedules to work everywhere, but i don't think its feasible with our codebase and itd just be unnecessary for our db.

i've addressed the two other small comments, but lmk what you think! :D

apps/antalmanac/src/backend/routers/userData.ts
```diff
@@ -78,9 +162,53 @@ const userDataRouter = router({
...
+     */
+    getUserDataHydrated: procedure.input(userInputSchema.assert).query(async ({ input }) => {
+        if ('userId' in input) {
+            const userData = await RDS.getUserDataByUid(db, input.userId);
+
+            if (!userData?.userData?.schedules) {
+                return userData;
+            }
+
+            const hydratedSchedules = await hydrateScheduleCourses(userData.userData.schedules);
```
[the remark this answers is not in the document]
vicksey: not worth the extra function its only a few lines

apps/antalmanac/src/backend/routers/userData.ts
```diff
@@ -78,9 +165,64 @@ const userDataRouter = router({
...
+     */
+    getUserDataHydrated: procedure.input(userInputSchema.assert).query(async ({ input }) => {
+        if ('userId' in input) {
+            const userData = await RDS.getUserDataByUid(db, input.userId);
+
+            if (!userData?.userData?.schedules) {
+                return userData;
+            }
+
+            try {
```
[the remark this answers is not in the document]
vicksey: its fine its so short

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -79,7 +80,7 @@ export const openSnackbar = (
     AppStore.openSnackbar(variant, message, duration, position, style);
 };
 
-export function isEmptySchedule(schedules: ShortCourseSchedule[]) {
+export function isEmptySchedule(schedules: ShortCourseSchedule[] | ScheduleSaveState['schedules']) {
```
alexespejo, line 83: question: `ShortCourseSchedule[]`  and `ScheduleSaveState['schedules']` seem to have the same shape?

packages/types/src/schedule.ts
```diff
@@ -44,7 +44,17 @@ export const ScheduleSaveStateSchema = type({
     schedules: arrayOf(ShortCourseScheduleSchema),
     scheduleIndex: 'number',
 });
-export type ScheduleSaveState = typeof ScheduleSaveStateSchema.infer;
+export type ShortScheduleSaveState = typeof ScheduleSaveStateSchema.infer;
+
+export type ScheduleSaveState = {
+    schedules: Array<{
+        scheduleName: string;
+        courses: (ShortCourse | ScheduleCourse)[];
```
alexespejo, line 52: I'm still weary about mixing the types just afraid it will create unforeseen complications in prod

alexespejo, line 52: I'm still weary of mixing the types in our code base. I'm afraid of unforseen complications in prod that can come up

alexespejo (changes requested):
I want to use the dev feature in #1565 to test the changes on staging. Maybe we can also benchmark the performance improvements

apps/antalmanac/src/actions/AppStoreActions.ts
vicksey: ScheduleSaveState['schedules'] has ScheduleCourse which has the course data

packages/types/src/schedule.ts
vicksey: It's more of a fallback in case websoc resp fails and also allows for the same type to be used to save into db. Like I could technically get rid of it and seperate the two types, bc idt aa will load without websoc. I'm more weary having just a state that must relies on websoc resp but if you think its better, than I can just remove it the mixed type, seperate the two. There's just no way for me to not keep the legacy type bc of our db, skeleton logic, etc.

KevinWu098:
Closing, stale

PR #1422, opened 2026-01-14
Load departments from Anteater API

### LINKED ISSUE #1333: Use Anteater API for departments
Currently, AntAlmanac hardcodes the list of departments and their names (for search filters) as a large constant. Instead, Anteater API's [recently merged /websoc/departments](https://github.com/icssc/anteater-api/pull/252) endpoint, which provides this exact functionality, can be used.

## Existing code

This is problematic because it may become inconsistent with actual UCI catalogue data over time, and also may include decades-outdated or irrelevant departments (which Anteater API allows us to exclude).

https://github.com/icssc/AntAlmanac/blob/3d11879127c42ac27e8db5b1e2eb298c07c492c8/apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/DepartmentSearchBar/constants.ts#L1-L10

## Implementation suggestions
Consider following a method like [PeterPortal's redux-based global caching system](https://github.com/icssc/peterportal-client/pull/831) for this kind of data to cache the API response on the frontend. The idea is:
- Create a slice to store departments
- Create a hook to load the data from the API once, and update the global state
- Call that hook once, and elsewhere in the app, get the departments from the state

xgraceyan:
## Summary
Uses departments fetched from [Anteater API](https://anteaterapi.com/reference#tag/websoc/GET/v2/rest/websoc/departments) instead of the hardcoded `SearchForm/DepartmentSearchBar/constants.ts`

Departments are fetched only once and stored in `DepartmentsStore.ts`
Department search bar calls `useDepartments` hook to get departments

## Test Plan
- Ensure departments are stored in the Zustand store and fetched from there on subsequent renders
- Ensure departments search bar properly loads all departments from the last 10 years (set as 10 for now)
- Ensure departments search bar functions as expected

## Issues

Closes #1333

### DIFF
--- a/apps/antalmanac/src/backend/routers/websoc.ts
+++ b/apps/antalmanac/src/backend/routers/websoc.ts
@@ -3,7 +3,7 @@ import type {
     CourseInfo,
     WebsocCourse,
     WebsocSectionType,
-    WebsocAPIDepartmentResponse,
+    WebsocAPIDepartmentsResponse,
 } from '@packages/antalmanac-types';
 import { z } from 'zod';
 
@@ -92,7 +92,7 @@ const queryWebSocDepartments = async () => {
         },
     });
     const data = await response.json();
-    return data.data as WebsocAPIDepartmentResponse;
+    return data.data as WebsocAPIDepartmentsResponse;
 };
 
 function combineWebsocResponses(responses: WebsocAPIResponse[]) {
--- a/apps/antalmanac/src/backend/routers/websoc.ts
+++ b/apps/antalmanac/src/backend/routers/websoc.ts
@@ -5,6 +5,7 @@ import type {
     WebsocSectionType,
     WebsocAPIDepartmentsResponse,
 } from '@packages/antalmanac-types';
+import { TRPCError } from '@trpc/server';
 import { z } from 'zod';
 
 import { procedure, router } from '../trpc';
@@ -92,6 +93,12 @@ const queryWebSocDepartments = async () => {
         },
     });
     const data = await response.json();
+    if (!data || !data.data) {
+        throw new TRPCError({
+            code: 'INTERNAL_SERVER_ERROR',
+            message: 'Departments API returned no data',
+        });
+    }
     return data.data as WebsocAPIDepartmentsResponse;
 };
 
--- a/apps/antalmanac/src/backend/routers/websoc.ts
+++ b/apps/antalmanac/src/backend/routers/websoc.ts
@@ -1,8 +1,15 @@
-import type { WebsocAPIResponse, CourseInfo, WebsocCourse, WebsocSectionType } from '@packages/antalmanac-types';
+import type {
+    WebsocAPIResponse,
+    CourseInfo,
+    WebsocCourse,
+    WebsocSectionType,
+    WebsocAPIDepartmentResponse,
+} from '@packages/antalmanac-types';
 import { z } from 'zod';
 
 import { procedure, router } from '../trpc';
 
+const DEPARTMENT_YEAR_RANGE = 10;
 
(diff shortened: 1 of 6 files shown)


dsnsgithub:
FYI: There is a new linting workflow that will fail on ESLint warnings.

After syncing to `main` and running `pnpm --filter antalmanac lint`:
```
apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/DepartmentSearchBar/DepartmentSearchBar.tsx
  75:9  warning  React Hook useCallback has a missing dependency: 'options'. Either include it or remove the dependency array  react-hooks/exhaustive-deps

✖ 1 problem (0 errors, 1 warning)

ESLint found too many warnings (maximum: 0).
```
 
Nothing serious, but flagging as the workflow might fail on other PRs.

apps/antalmanac/src/backend/routers/websoc.ts
```diff
@@ -74,6 +82,19 @@ const queryWebSoc = async ({ input }: { input: Record<string, string> }) => {
...
+const queryWebSocDepartments = async () => {
+    const minYear = new Date().getFullYear() - DEPARTMENT_YEAR_RANGE;
+    const url = `https://anteaterapi.com/v2/rest/websoc/departments?since=${minYear}`;
+
+    const response = await fetch(url, {
+        headers: {
+            ...(process.env.ANTEATER_API_KEY && { Authorization: `Bearer ${process.env.ANTEATER_API_KEY}` }),
+        },
+    });
+    const data = await response.json();
```
vicksey, line 95: Validate that the response and data came through to prevent errors going through

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/DepartmentSearchBar/DepartmentSearchBar.tsx
```diff
@@ -27,6 +29,12 @@ const parseLocalStorageRecentlySearched = (): string[] => {
 };
 
 export function DepartmentSearchBar() {
+    const { departments } = useDepartments();
+
+    const departmentsWithAll = departments ? { ...DEFAULT_DEPARTMENTS, ...departments } : DEFAULT_DEPARTMENTS;
```
vicksey, line 34: Optional suggestion: In the slight chance it fails to fetch should we still default to the hard-coded DEPARTMENT_MAP? 
e.g.     const departmentsWithAll = departments ? { ...DEFAULT_DEPARTMENTS, ...departments } : DEPARTMENT_MAP;

I know it kind of defeats the purpose... I'm just saying it bc when I initially opened the staging instance I was having network issues and the dropdown was empty. (also if we ever have future offline mobile). If we do though we should manually update the constant.

apps/antalmanac/src/stores/DepartmentsStore.ts
```diff
@@ -0,0 +1,27 @@
...
+    loadDepartments: async () => {
+        if (get().departments != null) {
+            return;
+        }
+        try {
+            const departmentsData = await WebSOC.getDepartments();
+            const departmentsMap: Record<string, string> = {};
+            for (const dept of departmentsData) {
+                departmentsMap[dept.deptCode] = `${dept.deptCode}: ${dept.deptName}`;
+            }
```
vicksey: nit: ues Object.fromEntries

const departmentsMap = Object.fromEntries(
departmentsData.map( dept => [dept.deptCode, `${dept.deptCode}: ${dept.deptName}`]
));

vicksey (commented):
Besides my few small comments, everything else looks good! Slay PR 😄💃🥳

vicksey (approved):
LGTM! 🚀

PR #1424, opened 2026-01-14
deps: mui v7

### LINKED ISSUE #1374: Update MUI to latest versions
MUI 5 is not fully compatible with the latest versions of React (19). We should update asap

AureliaSindhu:
## Summary
Upgraded Material UI from **v5 → v7** and aligned related MUI packages. Also fixed a MUI v7 breaking change caused by deep imports (restricted by package `exports`).

**MUI Core Packages (Updated to v7):**
- `@mui/material`: `^5.11.7` → `^7.3.7`
- `@mui/icons-material`: `^5.11.0` → `^7.3.7`
- `@mui/material-nextjs`: `^5.0.0` → `^7.3.7`
- `@mui/lab`: `^5.0.0-alpha.118` → `7.0.1-beta.21`
- `@mui/system`: added `^7.3.7` (required by MUI v7 dependency layout)

**Code changes:**
- Removed deep import `@mui/material/styles/createTypography` for `CSSProperties` and used `React.CSSProperties` instead (deep imports are blocked in v7).

## Test Plan
- Install deps: `pnpm install`
- Type check: `pnpm typecheck` (or `cd apps/antalmanac && pnpm typecheck`)
- Dev: `pnpm dev`
  - Verify components (grid, date/ time pickers/ theme switching)

## Issues
Closes #1374


KevinWu098:
Cool PR. will this be a chain of PRs from 5 -> 6 -> 7 etc?

KevinWu098 (changes requested):
Couple nits before i actually review this:

1. title is innaccurate, and probably should be `deps:` rather than `update:` (which is not a conventional commit prefix iirc)
2. PR did not deploy

.gitignore
```diff
@@ -5,6 +5,9 @@
 ## Packages installed by NPM. https://docs.npmjs.com/about-packages-and-modules
 node_modules/
 
+## pnpm local store. https://pnpm.io/npmrc#store-dir
+.pnpm-store/
```
KevinWu098: question: why is this in the diff now?

KevinWu098 (approved):
LGTM

.gitignore
AureliaSindhu: there was a cahce on my end, so i added so it won't track the changes. But i think we don't need it now, do u want me to remove it?

KevinWu098: yeah, let's simplify the diff

KevinWu098 (approved):

PR #1426, opened 2026-01-17
ci: add posthog key to prod

### LINKED ISSUE #1425: ci/cd: missing PostHog key
`NEXT_PUBLIC_PUBLIC_POSTHOG_KEY` is not set in production, so PostHog is not initialized.

https://github.com/icssc/AntAlmanac/blob/0fed92242787b8a58dc97920c3f0428ba663bd09/apps/antalmanac/src/providers/PostHog.tsx#L8-L21

Image

KevinWu098:
## Summary
Adds `NEXT_PUBLIC_PUBLIC_POSTHOG_KEY` to repo secrets (manually) and SST

## Issues

Closes #1425

### DIFF
--- a/.github/workflows/deploy_production.yml
+++ b/.github/workflows/deploy_production.yml
@@ -18,6 +18,7 @@ env:
     OIDC_CLIENT_ID: ${{ secrets.OIDC_CLIENT_ID }}
     OIDC_ISSUER_URL: ${{ secrets.OIDC_ISSUER_URL }}
     GOOGLE_REDIRECT_URI: ${{ secrets.GOOGLE_REDIRECT_URI }}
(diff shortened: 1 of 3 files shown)


dsnsgithub (changes requested):
I think you also need to add it in the workflows as well. PostHog keys are going to be visible to the client so secrets aren't necessary, but they don't really hurt either. The old Vite system used variables instead.

https://github.com/icssc/AntAlmanac/blob/899f851dd8da0722d393756d0440b3effa8b7221/.github/workflows/deploy_staging.yml#L13-L22

KevinWu098:
> I think you also need to add it in the workflows as well. PostHog keys are going to be visible to the client so secrets aren't necessary, but they don't really hurt either. The old Vite system used variables instead.
> 
> https://github.com/icssc/AntAlmanac/blob/899f851dd8da0722d393756d0440b3effa8b7221/.github/workflows/deploy_staging.yml#L13-L22

good catch

dsnsgithub (approved):
LGTM!

PR #1428, opened 2026-01-17
feat: add schedule selector to mobile Added Pane (#1177)

### LINKED ISSUE #1177: Allow schedule switching on mobile Added Pane
On mobile, where we only have the singular tabs, we should have the ability to switch the current schedule in the added pane (e.g. a schedule select popover)

ronaldw07:
## Summary
Adds a schedule selector to the mobile Added Pane, allowing mobile users to switch between schedules. The selector only appears on mobile viewports and reuses the existing `SelectSchedulePopover` component.
## Test Plan
1. Open AntAlmanac on localhost:3000
2. Resize browser window to mobile width (<600px) or use device emulation
3. Navigate to the "Added" courses tab
4. Verify schedule selector button appears next to schedule name (e.g., "Schedule 1 (0 Units)")
5. Click the selector and verify you can switch between schedules
6. Resize to desktop width and verify selector does NOT appear (desktop uses toolbar selector)
## Issues
Closes #1177

### DIFF
--- a/apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
+++ b/apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
@@ -1,4 +1,5 @@
 import { updateScheduleNote } from '$actions/AppStoreActions';
+import { SelectSchedulePopover } from '$components/Calendar/Toolbar/ScheduleSelect/ScheduleSelect';
 import { ClearScheduleButton } from '$components/buttons/Clear';
 import { CopyScheduleButton } from '$components/buttons/Copy';
 import { EmptyState } from '$components/EmptyState';
@@ -7,6 +8,7 @@ import { getMissingSections } from '$components/RightPane/AddedCourses/getMissin
 import { NotificationsDialog } from '$components/RightPane/AddedCourses/Notifications/NotificationsDialog';
 import { ColumnToggleDropdown } from '$components/RightPane/CoursePane/CoursePaneButtonRow';
 import SectionTable from '$components/RightPane/SectionTable/SectionTable';
+import { useIsMobile } from '$hooks/useIsMobile';
 import analyticsEnum, { logAnalytics } from '$lib/analytics/analytics';
 import { clickToCopy } from '$lib/helpers';
 import AppStore from '$stores/AppStore';
@@ -296,7 +298,7 @@ function AddedSectionsGrid() {
     const [courses, setCourses] = useState(getCourses());
     const [scheduleNames, setScheduleNames] = useState(AppStore.getScheduleNames());
     const [scheduleIndex, setScheduleIndex] = useState(AppStore.getCurrentScheduleIndex());
-
+    const isMobile = useIsMobile();
     useEffect(() => {
         const handleCoursesChange = () => {
             setCourses(getCourses());
@@ -350,7 +352,10 @@ function AddedSectionsGrid() {
                 <NotificationsDialog buttonSx={buttonSx} />
             </Box>
             <Box sx={{ marginTop: 7 }}>
-                <Typography variant="h6">{`${scheduleName} (${scheduleUnits} Units)`}</Typography>
+                <Box sx={{ display: 'flex', alignItems: 'center', gap: 1 }}>
+                    <Typography variant="h6">{`${scheduleName} (${scheduleUnits} Units)`}</Typography>
+                    {isMobile && <SelectSchedulePopover />}
+                </Box>
                 {courses.length === 0 && (
                     <EmptyState
                         Icon={MenuBook}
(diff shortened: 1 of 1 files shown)


apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -290,7 +292,7 @@ function AddedSectionsGrid() {
     const [courses, setCourses] = useState(getCourses());
     const [scheduleNames, setScheduleNames] = useState(AppStore.getScheduleNames());
     const [scheduleIndex, setScheduleIndex] = useState(AppStore.getCurrentScheduleIndex());
-
+    const isMobile = useIsMobile();
```
Choollol, line 302: Add some empty lines before and after this line to keep it readable. In my experience, the convention is usually to keep different types of hook-calls separated, though it varies a lot case-by-case. In this case you'd probably want to keep that whitespace though.

Choollol (changes requested):
Looks like it works to me, no breaking changes that I found! However, the formatting of the schedule title does get shifted around a little since the schedule selector takes up some space. You might want to double check with the team for the design of this UI.
Example:
image

ronaldw07:
Does anyone have any thoughts or comments on this for the design especially?

chrissl9:
This formatting might result from a user changing their font size in their mobile settings. To resolve this issue, a suggestion would be to separate the schedule selector and title into 2 separate rows. I believe that this would improve the crowding issues in both situations where the schedule title wraps around and how it looks normally.

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -350,7 +354,14 @@ function AddedSectionsGrid() {
...
             </Box>
             <Box sx={{ marginTop: 7 }}>
-                <Typography variant="h6">{`${scheduleName} (${scheduleUnits} Units)`}</Typography>
+                {isMobile ? (
+                    <Box sx={{ display: 'flex', flexDirection: 'column', gap: 1 }}>
+                        <Typography variant="h6">{`${scheduleName} (${scheduleUnits} Units)`}</Typography>
+                        <SelectSchedulePopover />
+                    </Box>
+                ) : (
+                    <Typography variant="h6">{`${scheduleName} (${scheduleUnits} Units)`}</Typography>
```
Choollol: There's a bit of duplication here that could get confusing in the future. You could either move the schedule title out of the containing `Box`, which would always render the container, or you could assign the title to a variable and use that instead.

@alexespejo Do you have any opinions on which option is better? Or if it's even necessary to keep this duplication out.

One other thing to mention is that `SelectSchedulePopover`'s position is hardcoded and the formatting is a little messed up when on the mobile view.

Alex, what are your thoughts on setting `ScheduleSelectPopover`'s anchor point based on the button rather than the screen? It could look something like this:
```ts
export function SelectSchedulePopover() {
  const anchorElementRef = useRef(null);
  ...
  <Button
    ref={anchorElementRef}
    ...
  </Button>
  ...
  <Popover
    open={openScheduleSelect}
    anchorReference="anchorEl"
    anchorEl={anchorElementRef.current}
    ...
  </Popover>
```

Choollol (changes requested):

Choollol:
There hasn't been any activity for over a week, so I'll be closing this PR.

ronaldw07:
Could we wait for alex's input still?

Choollol:
Sure, that's my bad! Looking at it again, could you explain why the entire `Box` is being conditionally rendered rather than just `SelectSchedulePopover`?

Also, once https://github.com/icssc/AntAlmanac/pull/1357 is merged, you can pull from main since it fixes the issue with the popover's position.

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
alexespejo: Sorry for the delay I've been getting caught up behind a lot of different things. But regarding the duplication it's marginal but I am curious what the `Box` component is for.

ronaldw07:
updated the conditional ternary to just render the typography once with {isMobile && <SelectSchedulePopover />} below it, so the box wrapper is gone since it was only there for stacking. the popover anchoring is already handled via anchorElementRef in the current main, so pulling from main resolves that too.

Choollol:
LGTM! Once https://github.com/icssc/AntAlmanac/pull/1357 is merged, I'll approve this PR and merge it.

ronaldw07:
hi, just rebased to fix the conflicts. any updates on this?

Choollol:
Thanks for fixing the merge conflicts! I'll go ahead and approve and merge this since it looks good.

Choollol (approved):

PR #1433, opened 2026-01-18
fix: show missing section warnings on schedule load

### LINKED ISSUE #1399: Show missing section warnings for saved courses on page load
When AntAlmanac first loads and the user is already signed in, courses restored from `fromScheduleSaveState ` may fail to show missing section warnings, even if the course requires additional sections (e.g. lab, discussion).

In fromScheduleSaveState, the API call only fetches data for the specific section codes that were saved. This means:
- We do not fetch the full course data
- We do not know all required section types for the course
- Therefore, we cannot determine which required sections are missing

Example:
A course requires Lecture + Lab + Discussion, and the saved schedule contains only Lecture + Lab. On initial load, the missing section warning will only appear for lecture/lab inconsistencies (if any), but no warning will be shown for the missing discussion

ronaldw07:
## Summary
Fixes #1399 - Show missing section warnings for saved courses on page load

When AntAlmanac loads and restores courses from `fromScheduleSaveState`, the app now detects and displays warnings for courses that are missing required sections (e.g., lab, discussion).

## Changes
- **Added `checkForMissingSections()` method to AppStore** that:
  - Groups courses by course ID (deptCode + courseNumber + term)
  - Compares required section types (from `sectionTypes` array) with enrolled section types
  - Identifies missing required sections
  - Displays a warning snackbar listing all affected courses
- **Integrated into schedule loading**: Call `checkForMissingSections()` after successful schedule load in `loadSchedule()`
- **Extracted `sectionTypeToName()` utility function**: Created a reusable utility to convert section type codes to readable names
- **Refactored `getMissingSections.ts`**: Updated to use the shared utility function, eliminating code duplication

## How It Works
The `getCourseInfo` API endpoint (called by `fromScheduleSaveState`) already returns the full `sectionTypes` array containing all available section types for a course. After a schedule loads successfully, `checkForMissingSections()` compares the required section types with the sections the user has enrolled in and displays warnings for any missing required sections.

## Example Output
If a user has saved a schedule with only Lecture + Lab sections of a course that requires Lecture + Lab + Discussion, they'll see:

```
1 course is missing required sections:
COMPSCI 161 (2024 Fall): Missing Discussion
```

For multiple courses:
```
2 courses are missing required sections:
COMPSCI 161 (2024 Fall): Missing Discussion
PHYSICS 7C (2024 Fall): Missing Lab
```

## Test plan
- [ ] Load a saved schedule with courses missing required sections
- [ ] Verify warning snackbar appears on page load
- [x] Verify warning message correctly identifies missing sections
- [x] Verify existing missing section warnings in the AddedCourses pane still work
- [x] Verify no warnings appear for courses with all required sections

### DIFF
--- a/apps/antalmanac/src/stores/AppStore.ts
+++ b/apps/antalmanac/src/stores/AppStore.ts
@@ -26,6 +26,7 @@ import type {
     AddScheduleAction,
 } from '$actions/ActionTypesStore';
 import type { CalendarEvent, CourseEvent } from '$components/Calendar/CourseCalendarEvent';
+import { sectionTypeToName } from '$lib/utils';
 import { Schedules } from '$stores/Schedules';
 import { useTabStore } from '$stores/TabStore';
 import { deleteTempSaveData, loadTempSaveData, setTempSaveData } from '$stores/localTempSaveDataHelpers';
@@ -418,6 +419,9 @@ class AppStore extends EventEmitter {
         this.emit('currentScheduleIndexChange');
         this.emit('scheduleNotesChange');
 
+        // Check for missing sections after loading schedule
+        this.checkForMissingSections();
+
         return true;
     }
 
@@ -502,6 +506,62 @@ class AppStore extends EventEmitter {
     updateScheduleNote(newScheduleNote: string, scheduleIndex: number) {
         this.schedule.updateScheduleNote(newScheduleNote, scheduleIndex);
         this.emit('scheduleNotesChange');
+    }
+
+    /**
+     * Check for missing required sections in loaded courses and display warnings.
+     * This is called after loading a schedule from saved state to detect courses
+     * that are missing required sections (e.g., lab, discussion).
+     */
+    checkForMissingSections() {
+        const coursesWithMissingSections: { courseId: string; missingSections: string[] }[] = [];
+
+        // Group courses by unique course (deptCode + courseNumber + term)
+        const courseGroups = new Map<string, ScheduleCourse[]>();
+
+        for (const course of this.schedule.getCurrentCourses()) {
+            const courseId = `${course.deptCode} ${course.courseNumber} (${course.term})`;
+            if (!courseGroups.has(courseId)) {
+                courseGroups.set(courseId, []);
+            }
+            courseGroups.get(courseId)!.push(course);
+        }
+
+        // Check each course group for missing sections
+        for (const [courseId, courses] of courseGroups) {
+            // All courses in the group should have the same sectionTypes
+            const sectionTypes = courses[0].sectionTypes;
+            if (!sectionTypes || sectionTypes.length === 0) {
+                continue;
+            }
+
+            // Get all section types the user has enrolled in
+            const enrolledSectionTypes = new Set(courses.map((c) => c.section.sectionType));
+
+            // Find missing section types
+            const missingSectionTypes = sectionTypes.filter((type) => !enrolledSectionTypes.has(type));
+
+            if (missingSectionTypes.length > 0) {
+                // Map section type codes to readable names
+                const missingSections = missingSectionTypes.map(sectionTypeToName);
+
+                coursesWithMissingSections.push({ courseId, missingSections });
+            }
+        }
+
+        // Display warning if there are courses with missing sections
+        if (coursesWithMissingSections.length > 0) {
+            const warningMessages = coursesWithMissingSections.map(
+                ({ courseId, missingSections }) => `${courseId}: Missing ${missingSections.join(', ')}`
+            );
+
+            const message =
+                coursesWithMissingSections.length === 1
+                    ? `1 course is missing required sections:\n${warningMessages[0]}`
+                    : `${coursesWithMissingSections.length} courses are missing required sections:\n${warningMessages.join('\n')}`;
+
+            this.openSnackbar('warning', message, 10);
+        }
     }
 
     termsInSchedule = (term: string) =>
--- a/apps/antalmanac/src/stores/AppStore.ts
+++ b/apps/antalmanac/src/stores/AppStore.ts
@@ -26,7 +26,6 @@ import type {
     AddScheduleAction,
 } from '$actions/ActionTypesStore';
 import type { CalendarEvent, CourseEvent } from '$components/Calendar/CourseCalendarEvent';
-import { sectionTypeToName } from '$lib/utils';
 import { Schedules } from '$stores/Schedules';
 import { useTabStore } from '$stores/TabStore';
 import { deleteTempSaveData, loadTempSaveData, setTempSaveData } from '$stores/localTempSaveDataHelpers';
@@ -419,9 +418,6 @@ class AppStore extends EventEmitter {
         this.emit('currentScheduleIndexChange');
         this.emit('scheduleNotesChange');
 
-        // Check for missing sections after loading schedule
-        this.checkForMissingSections();
-
         return true;
     }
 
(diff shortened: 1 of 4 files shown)


calebwongg (changes requested):
Haven't looked too deep into the code yet, but upon just checking to see if the feature works, we still aren't displaying the missing sections warning when the user loads in their schedule
image
This is the what my added courses pane looks like right when I load in my classes, and if the user saved an incomplete schedule, we have no missing sections warning. We want to have the warning display right when the schedule loads in. Feel free to text me on discord (calbe__) if you have any further questions!

apps/antalmanac/src/stores/AppStore.ts
```diff
@@ -504,6 +508,62 @@ class AppStore extends EventEmitter {
...
+            }
+        }
+
+        // Display warning if there are courses with missing sections
+        if (coursesWithMissingSections.length > 0) {
+            const warningMessages = coursesWithMissingSections.map(
+                ({ courseId, missingSections }) => `${courseId}: Missing ${missingSections.join(', ')}`
+            );
+
+            const message =
```
calebwongg: We want to keep the warning consistent, so instead of the snackbar we want to use the mui alert. You shouldn't need to code in the alert because we already have it in SectionTable.tsx: 
```
            {missingSections?.length > 0 && (
                <Alert
                    severity="warning"
                    sx={{
                        mb: 1,
                        '& .MuiAlert-message': {
                            display: 'flex',
                            alignItems: 'center',
                        },
                    }}
                >
                    Missing required sections: {missingSections.join(', ')}
                </Alert>
            )}
```

calebwongg: Look into how the missing sections attributes flow through the code. We store the missing sections as a property on `CourseWithTerm `objects and check that value when the section table mounts. Feel free to reach out if you have questions

calebwongg (changes requested):
issue: Again haven't looked into the code, but now we are staying consistent with the mui alert and there is some sort of missing sections comparison going on upon schedule load (which is good). 

However, the comparison doesn't work as it thinks that some classes require sections that don't exist for that class. 
image
An example is that EECS 31 doesn't require a lab or seminar. The warning should only be displaying for the lab

apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -591,6 +592,49 @@ export class Schedules {
...
+                    const key = `${term}-${courseInfo.courseDetails.deptCode}-${courseInfo.courseDetails.courseNumber}`;
+                    uniqueCourses.add(key);
+                }
+            }
+
+            // Fetch full course data for each unique course
+            const fullCourseRequests = Array.from(uniqueCourses).map(async (courseKey) => {
+                const [term, deptCode, courseNumber] = courseKey.split('-');
+                try {
+                    const fullData = await WebSOC.query({ term, deptCode, courseNumber });
```
calebwongg: nit: websoc lets you query with an array of section codes. Its redundant here to have an individual query for each section code. We would rather only need one query using an array of section codes as the query parameter.

ronaldw07: oh I see. I will try to fix it

calebwongg: sounds good thank you !!!

ronaldw07: let me know if this works!

calebwongg (changes requested):
issue: Still didn't get into the bulk of the code, but some functionality is still a little buggy! You can see here that I'm getting an alert for a missing lab when MATH 3A has no lab section: 

Image 

Another thing worth looking into also is that auth is substantially slower when loading in the app. It also seems a bit slower switching between the 'SEARCH' pane and the 'ADDED' pane (this could just be placebo). I'm suspecting it might still be because of redundant API queries, so try and look into it and see what you can do. If you run into any problems, or if you have any questions on how to implement this fix feel free to reach out! Please test before marking ready for review!

ronaldw07:
Fixed the false positive warnings. The previous dept-wide query was returning courses from other departments with the same course number (e.g. PHYSICS 3A and SOC SCI 3A contaminating MATH 3A's sectionTypes with Lab). Now queries each course individually by (term, deptCode, courseNumber) and filters results to only the matching department and course number. Tested locally - MATH 3A with only lecture correctly shows 'Missing: Discussion' and no false Lab warning.

apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -591,6 +592,47 @@ export class Schedules {
 
             await Promise.all(websocRequests);
 
+            // Fetch complete sectionTypes for each unique course
+            // When getCourseInfo is called with specific sectionCodes, WebSOC only returns those sections,
+            // so sectionTypes is incomplete. We need to fetch all sections of each course to get complete sectionTypes.
+            const completeSectionTypesDict = new Map<string, CourseDetails['sectionTypes']>();
```
calebwongg: nit: utilize WebSocSectionType

apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -591,6 +592,47 @@ export class Schedules {
...
+                    const key = `${term}|${courseInfo.courseDetails.deptCode}|${courseInfo.courseDetails.courseNumber}`;
+                    uniqueCourses.add(key);
+                }
+            }
+
+            // Fetch sectionTypes for each unique (term, deptCode, courseNumber) individually
+            const fullCourseRequests = Array.from(uniqueCourses).map(async (courseKey) => {
+                const [term, deptCode, courseNumber] = courseKey.split('|') as [string, string, string];
+                try {
+                    const fullData = await WebSOC.query({ term, deptCode, courseNumber });
```
calebwongg: Querying websoc twice is definitely redundant, but then I'm also not sure if there is another way to do it. I guess just something to keep in mind

apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -591,6 +592,47 @@ export class Schedules {
...
+
+                    const sectionTypesSet = new Set<CourseDetails['sectionTypes'][number]>();
+                    for (const school of fullData.schools) {
+                        for (const dept of school.departments) {
+                            if (dept.deptCode !== deptCode) continue;
```
calebwongg: nit: this extra layer of safety is redundant and not needed. We never really verify the api for the other times when we call it

apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -591,6 +592,47 @@ export class Schedules {
...
+                            for (const course of dept.courses) {
+                                if (course.courseNumber !== courseNumber) continue;
```
calebwongg: nit: this extra layer of safety is redundant and not needed.

apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -591,6 +592,47 @@ export class Schedules {
...
+            // When getCourseInfo is called with specific sectionCodes, WebSOC only returns those sections,
+            // so sectionTypes is incomplete. We need to fetch all sections of each course to get complete sectionTypes.
+            const completeSectionTypesDict = new Map<string, CourseDetails['sectionTypes']>();
+            const uniqueCourses = new Set<string>();
+
+            // Extract unique (term, deptCode, courseNumber) combinations
+            // Use '|' as separator since term contains spaces and courseNumber may contain letters
+            for (const [term, courseInfoMap] of courseInfoDict) {
+                for (const [, courseInfo] of Object.entries(courseInfoMap)) {
+                    const key = `${term}|${courseInfo.courseDetails.deptCode}|${courseInfo.courseDetails.courseNumber}`;
```
calebwongg: instead of building a synthetic string key (${term}|${dept}|...), we should probably use a nested Map structure: Map<Term, Map<Dept, Map<CourseNum, SectionTypes>>>. this way we aren't building our own keys

calebwongg (commented):
feature looks good ! some small changes though, the main one being to change the idea of building our own keys with '|'

calebwongg (changes requested):

ronaldw07:
addressed all the nits from Feb 25 (hopefully) — switched to a nested Map<Term, Map<Dept, Map<CourseNum, SectionTypes>>> instead of building synthetic string keys, used WebsocSectionType directly for the set type, and removed the redundant dept/courseNumber filter checks inside the loop.

PR #1437, opened 2026-01-19
Fix: Use custom event ID as key instead of title

### LINKED ISSUE #1436: Added courses pane custom event duplicate React key error
When two custom events have the same name, an error pops up in the console since their names are used as the keys when mapping custom events to rendered components.
Example:

Image

Fix: Use custom events' IDs instead, since they're actually unique. (I think?)

Image

Image

Choollol:
## Summary

Fix a bug where using custom event titles as keys when `.map()`ing causes an error when multiple custom events have the same title.

## Test Plan

1. Create two custom events with the same name (in the same schedule)
2. Navigate to the Added tab and ensure no errors pop up in the console

## Issues

Closes #1436


KevinWu098 (approved):
LGTM

PR #1438, opened 2026-01-20
feat: Add automated workflow to detect new academic quarters

### LINKED ISSUE #1319: Fuzzy search data does not automatically propagate with the quarter
Currently, Winter 2026 correctly surfaces as an available quarter on ⁠AA, but the app still doesn’t roll over to the new quarter automatically because the fuzzy search (quick search) data is generated at build time and won’t update until the app is redeployed.

Notes: 
Possible solutions

1. Write a microservice to poll for a new SocAvailable quarter to redeploys the backend and rebuild the cache 
- if this can go through github actions, that would be cool, but likely this will need to be a cron (AWS EventBridge)
- AAPI does gather SOC quarter schedule for the year which we can use to dictate when to trigger the redeployment  

2. Temporary resolution set a reminder on your phone to trigger the build manually ```git commit -m "ci: trigger" --allow-empty```

sicn4rf:
## Summary
Adds a GH actions workflow that automatically detects when new academic quarters become available in AnteaterAPI and triggers a rebuild. 

## Test Plan
Tested three scenarios on branch `fp/auto-quarter-rollover`:

1. Bootstrap - File doesn't exist:
- Workflow created `.github/deployed_terms.json` from AnteaterAPI
- Committed and pushed successfully

2. New quarter detection:
- Manually removed the latest term in `deployed_terms.json` to simulate the mismatch
- Workflow detected the change and updated the file
- Committed and pushed the file with the appropriate message

3. No changes (most common case):
- Workflow detected terms match
- Skipped commit step
- No unnecessary commits

## Issues
Closes #1319

### DIFF
--- a/.github/workflows/check_new_quarter.yml
+++ b/.github/workflows/check_new_quarter.yml
@@ -41,7 +41,23 @@ jobs:
                     exit 1
                 fi
 
-                
+                LATEST_TERM=$(echo "$API_RESPONSE" | jq -r '.data[0].longName')
+
+                if [ ! -f .github/deployed_terms.json ]; then
+                    echo "deployed_terms.json does not exist"
+                    echo "Creating initial deployed_terms.json..."
+                    echo "$API_RESPONSE" > .github/deployed_terms.json
+                    echo "update_needed=true" >> $GITHUB_OUTPUT
+                    echo "new_term=$LATEST_TERM" >> $GITHUB_OUTPUT
+                    echo "reason=initial deployed terms tracking file not generated" >> $GITHUB_OUTPUT
+
+                    exit 0
+                fi
+
+        
+
+
+
 
 
 
--- a/.github/workflows/check_new_quarter.yml
+++ b/.github/workflows/check_new_quarter.yml
@@ -66,7 +66,20 @@ jobs:
                     echo "reason=new quarter available, trigger rebuild" >> $GITHUB_OUTPUT
                     echo "$API_RESPONSE" > .github/deployed_terms.json
 
-        
+            - name: Commit and push updated terms
+              if: steps.check.outputs.update_needed == 'true'
+              run: |
+                git config user.name "github-actions[bot]"        
+                git config user.email "github-actions[bot]@users.noreply.github.com"        
+
+                git add .github/deployed_terms.json
+
+                git commit -m "ci: ${{ steps.check.outputs.reason }}
+New term: ${{ steps.check.outputs.new_term }}
+Updated by automated workflow on $(date -u +%Y-%m-%d)
+                
+                git push origin main
+                echo "Updated deployed_terms.json and triggered rebuild"
 
 
 
--- a/.github/workflows/check_new_quarter.yml
+++ b/.github/workflows/check_new_quarter.yml
@@ -52,9 +52,11 @@ jobs:
                       echo "deployed_terms.json does not exist"
                       echo "Creating initial deployed_terms.json..."
                       echo "$API_RESPONSE" > .github/deployed_terms.json
-                      echo "update_needed=true" >> $GITHUB_OUTPUT
-                      echo "new_term=$LATEST_TERM" >> $GITHUB_OUTPUT
-                      echo "reason=initial deployed terms tracking file not generated" >> $GITHUB_OUTPUT
+                      {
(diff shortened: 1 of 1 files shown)


.github/workflows/check_new_quarter.yml
```diff
@@ -0,0 +1,92 @@
+name: Check for New Quarter and Update Deployed Terms if needed
```
dsnsgithub: nit: Title Case

```suggestion
name: Check for New Quarter and Update Deployed Terms If Needed
```

.github/workflows/check_new_quarter.yml
```diff
@@ -0,0 +1,92 @@
...
+              uses: actions/checkout@v4
+              with:
+                  token: ${{ secrets.GITHUB_TOKEN }}
+
+            - name: Configure Git
+              run: |
+                  git config user.name "github-actions[bot]"
+                  git config user.email "github-actions[bot]@users.noreply.github.com"
+
+            - name: Check for new Terms
```
dsnsgithub: nit: consistent capitalization

```suggestion
            - name: Check for new terms
```

.github/workflows/check_new_quarter.yml
```diff
@@ -0,0 +1,92 @@
...
+    workflow_dispatch:
+
+env:
+    ANTEATER_API_KEY: ${{ secrets.ANTEATER_API_KEY }}
+
+permissions:
+    contents: write
+
+jobs:
+    check-and-reschedule:
```
dsnsgithub: nit: remnant of old code, doesn't seem to reschedule anymore.

```suggestion
    update-quarters:
```

.github/workflows/check_new_quarter.yml
```diff
@@ -0,0 +1,92 @@
...
+        runs-on: ubuntu-latest
+        steps:
+            - name: Checkout
+              uses: actions/checkout@v4
```
dsnsgithub:
```suggestion
              uses: actions/checkout@v6
```

nit: checkout@v6 is technically the newest, but it is practically identical to v4. Still worth using if possible.

.github/workflows/check_new_quarter.yml
```diff
@@ -0,0 +1,92 @@
+name: Check for New Quarter and Update Deployed Terms if needed
+
+on:
+    schedule:
+        - cron: '0 * * * *'
```
dsnsgithub, line 5: > The schedule event can be delayed during periods of high loads of GitHub Actions workflow runs. High load times include the start of every hour. If the load is sufficiently high enough, some queued jobs may be dropped. To decrease the chance of delay, schedule your workflow to run at a different time of the hour.

https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#schedule

In my testing in a different repo, this seems to cause the workflow to get dropped every time. I think GitHub scheduled workflows is broken right now, regardless it doesn't seem particularly reliable.

dsnsgithub (changes requested):
None of the nits are merge blocking, I was especially nit-picky in this review. Feel free to resolve the comments on those if you feel they are overbearing.

I wasn't able to get the scheduled workflow working on my private repo, but I don't think it is your fault since workflows in other repos with similar cron also don't seem to run correctly. If there is a solution or I am doing something wrong, please let me know.

dsnsgithub (approved):
The workflow works, but at the time of writing GitHub scheduled actions has a bug where it does not trigger unless the default branch name is `master`.

For now, I will approve the PR because no future changes are needed, but keep in mind the workflow will not automatically run until GitHub makes some changes on their side.

If there are any potential changes that can help alleviate this problem, feel free to request a re-review, but I trust that this PR will be merged when the bug is fixed.


Further information:
https://github.com/actions/runner/issues/4210
https://github.com/orgs/community/discussions/185081
https://github.com/orgs/community/discussions/185024

dsnsgithub:
Merge at will!

> There was a recently enabled feature flag on the server side that causes this issue for schedule workflow. The feature flag has been disabled, push any new change (ex: a new empty line in Readme.md) to the repository should fix those broken schedule workflows. Sorry for the break. 🙇 - https://github.com/actions/runner/issues/4210#issuecomment-3800490971

PR #1441, opened 2026-01-21
fix: add missing query parameters when manual searching by GE

### LINKED ISSUE #1278: Missing query parameters in course pane fetching

calebwongg:
## Summary
When using manual search and searching by GE, the api query was missing parameters. Searching by GE and e.g. instructor, day, start time, etc would not work. Now, these parameters are included in the api query.
image

## Test Plan
Manual search by GE + some other filter. Make sure all parameters are in the query url

## Issues

Closes #1278

### DIFF
--- a/apps/antalmanac/src/components/RightPane/SectionTable/GEDataFetchProvider.tsx
+++ b/apps/antalmanac/src/components/RightPane/SectionTable/GEDataFetchProvider.tsx
@@ -24,6 +24,16 @@ class GeDataFetchProvider extends PureComponent<SectionTableProps> {
             ge: 'ANY',
             courseNumber: this.props.courseDetails.courseNumber,
             courseTitle: this.props.courseDetails.courseTitle,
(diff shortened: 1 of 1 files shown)


brelieu05 (approved):
LGTM

PR #1443, opened 2026-01-21
fix: tutorial box

### LINKED ISSUE #1415: Tutorial steps are mispositioned
Something (z-index related, perhaps? ref-related, perhaps?) is causing the first few steps of the Tutorial to not position on the right elements.

Image
Image
Image
Image

----

1 and 2 are positioned to presumable (0,0), and 3 and 4 are positioned such that the tutorial box is cut off OR the tutorial box cuts off the element it's highlighting

AureliaSindhu:
## Summary
Fix the tutorial tour popover getting cut off-screen by adjusting the tour popover margin and setting better step positions (including centering the welcome step). Also updates the search step selector to #fuzzy-search so it anchors correctly.

Screenshot 2026-01-23 at 12 40 44 PM
Screenshot 2026-01-23 at 12 40 50 PM
Screenshot 2026-01-23 at 12 40 55 PM

## Test Plan
- Start the app and launch the tutorial tour through the help button.
- Navigate through the tour steps, especially:
	- Step 2 (search)
	- Map + map pane steps
	- Final save/sign-in step
- Verify the popover does not stick in the top-left during transitions.

## Issues
- Changes were made in the padding, it's either the top will get cut a little bit or the bottom ones. 
- To fix the fallback -> If a selector truly never becomes measurable, the helper times out after ~8s and logs an info message.

Closes #1415

### DIFF
--- a/apps/antalmanac/src/lib/TutorialHelpers.tsx
+++ b/apps/antalmanac/src/lib/TutorialHelpers.tsx
@@ -70,7 +70,8 @@ export function namedStepsFactory(goToStep: (step: number) => void): Record<Tour
 
     return {
         welcome: {
-            selector: '#root',
+            selector: '#__next',
+            position: 'center',
             content: (
                 <>
                     Welcome to AntAlmanac!
@@ -87,21 +88,23 @@ export function namedStepsFactory(goToStep: (step: number) => void): Record<Tour
             },
         },
         searchBar: {
-            selector: '#searchBar',
+            selector: '#fuzzy-search',
             content: 'You can search for your classes here!',
             action: () => {
                 markTourHasRun();
                 setActiveTab('search');
             },
-            mutationObservables: ['#searchBar'],
+            mutationObservables: ['#fuzzy-search'],
         },
         importButton: {
             selector: '#import-button',
             content: 'Quickly add your classes from WebReg or Zotcourse!',
+            position: 'bottom',
         },
         calendar: {
             selector: '.rbc-time-view', // Calendar.
             content: 'See the classes in your schedule!',
+            position: 'bottom',
             action: () => {
                 addSampleClasses();
                 const finalsButtonPressed = document.getElementById('finals-button-pressed');
@@ -190,6 +193,7 @@ export function namedStepsFactory(goToStep: (step: number) => void): Record<Tour
                     That&apos;s it 🎉 Good luck with your classes!
                 </>
             ),
+            position: 'bottom',
             action: tourActionFactory(() => goToNamedStep(TourStepName.saveAndLoad), {
                 selector: '#load-save-container',
                 eventType: 'click',
--- a/apps/antalmanac/src/lib/TutorialHelpers.tsx
+++ b/apps/antalmanac/src/lib/TutorialHelpers.tsx
@@ -70,7 +70,7 @@ export function namedStepsFactory(goToStep: (step: number) => void): Record<Tour
 
     return {
         welcome: {
-            selector: '#__next',
+            selector: '#root',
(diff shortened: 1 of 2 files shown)


Choollol (commented):
There's a tiny bug on the `calendar` step (step 4) where if the viewport is less than 995 pixels wide, it'll jitter like it's trying to go to the bottom then move toward the top left. I found that changing its selector to `#calendar-root` (same as the `showFinals` step) fixes it, but I'll leave the final decision to you since you know more about this area.

The `mapPane` step (step 10) also gets stuck in the top left when the map isn't open before, reproducible by going from 8 to 9 to 10. Might be a bug with Reactour, but I'm not sure what the exact problem is. Again, you know more than I do so up to you what to do!

Edit: The `mapPane` bug only happens the first time step 10 is reached, probably because the map is rendered or something? So not the biggest issue since not many people are going to be replaying the tour like that, but just something I noticed.

AureliaSindhu:
## Changes

- Step 4: The jitter was caused by targeting .rbc-time-view, which shifts during responsive layout. I changed it to target the more stable #calendar-root and added resize/mutation observables so Reactour re-measures as the layout changes.
- Step 10: It was stuck because setActiveTab('map') doesn’t navigate to /map, so #map-pane isn’t mounted the first time. Made the step click #map-tab when needed and added a small delayed re-measure so Reactour repositions once the map mounts.

Choollol (dismissed):
Looks good!

Steps 3 and 11 clip a tiny bit into the top of the screen, but imo it's unnoticeable. If someone else has a stronger opinion, feel free to add on, but I'll go ahead and approve.

apps/antalmanac/src/lib/TutorialHelpers.tsx
```diff
@@ -179,8 +184,23 @@ export function namedStepsFactory(goToStep: (step: number) => void): Record<Tour
         mapPane: {
             selector: '#map-pane',
             content: 'Click on a day to see your route!',
-            action: () => setActiveTab('map'),
-            mutationObservables: ['#map-pane'],
+            action: () => {
+                setActiveTab('map');
+
+                if (!window.location.pathname.includes('/map')) {
```
alexespejo, line 244: issue: The behavior of stickingh in the left corner of the screen seems to be unresolved? Both pieces of code just have a delay while waiting for the map to load.

alexespejo (changes requested):
Left one comment. Correct me if I misinterpreted the purpose of the comment change

apps/antalmanac/src/lib/TutorialHelpers.tsx
AureliaSindhu: hmm yeaa, i can change the tour step to wait until #map-pane exists and has a non-zero bounding box before advancing/re-measuring, instead of using setTimeout, that should eliminate the top-left fallback, does that sounds better?

AureliaSindhu:
Fixed the top-left “(0,0)” tour fallback by making the tour wait until the step target is actually mounted.

alexespejo (changes requested):
Still not able to see the difference? "Is it a works on my machine" problem lol. You're saying it should stay in the same position as Step 9 before moving down to Step 10?

alexespejo (changes requested):
> Fixed the top-left “(0,0)” tour fallback by making the tour wait until the step target is actually mounted.

Still don't seem to see the difference? "Is it a works on my machine" problem lol. The intended behavior is it should stay in the Step 9 position then jump to step 10?

alexespejo (approved):
LGTM

Image

PR #1447, opened 2026-01-22
AANTS SQS Implementation

IsaacNguyen:
## Summary
Changed the flow of AANTS to respect the SES email sending limit (14 email/sec).

### Before:
- AANTS would detect a change
- Creates batched email requests by classes and sends them out 
     - Didn't respect the sending limit of 14 emails/sec and batch limit of 50 emails

### Now:
- AANTS detects a change
- Creates individual email requests and sends to AWS SQS
- AWS SQS keeps all email requests in it
- SQS will take 14 of these emails at a time over a period of 1 second to process
- SQS sends this batch of emails to the new Email Processing Lambda which uses SES to send the actual emails

NOTE: I limited the Email Processing Lambda to ONLY do one execution at a time. This way it respects the email sending limit.

## Test Plan
- [ ] See if it works lol

## Issues
To be determined


sst.config.ts
```diff
@@ -43,27 +43,68 @@ export default $config({
...
+                    ],
+                    resources: [emailQueue.arn],
+                },
             ],
         });
 
+        emailQueue.subscribe(emailProcessorLambda.arn, {
+            batch: {
+                size: 14, // Match SES rate limit 
+                window: '1 second', // Collect messages for up to 1 second
```
KevinWu098: question: idk how AWS calculates their rate limits — do we want to give ourselves some breathing room and say... 14 emails every 1.25s?

apps/aants/src/emailProcessor.ts
```diff
@@ -0,0 +1,55 @@
...
+import { SQSEvent, SQSBatchResponse, SQSBatchItemFailure } from 'aws-lambda';
+import { EmailRequest } from './helpers/emailQueue';
+
+const sesClient = new SESv2Client({});
+
+/**
+ * Processes SQS records containing individual email requests.
+ * 
+ * @param records - Array of SQS records, each containing one email request
+ * @returns Array of failed message IDs for partial batch response
```
jotalis: this comment seems to be inaccurate? we are not returning the partial batch response and this means any emails that fail processing just get removed from the queue.

is this intended behavior for the mvp? ideally, we'd probably want to retry sending emails and then after n retries, move the failed emails to a DLQ for us to manually inspect the failed emails

apps/aants/src/emailProcessor.ts
```diff
@@ -0,0 +1,55 @@
...
+
+        return {
+            statusCode: 200,
+            body: JSON.stringify('Success'),
+        };
+
+    } catch (error) {
+        console.error('Error processing email records:', error);
+        throw error;
+    }
```
jotalis: if we're swallowing errors in `processEmailRecords`, isn't the try and catch in handler somewhat redundant?

here, we'd ideally retrieve the failed record messageIds as item identifiers from `processEmailRecords` and then instead of a try catch, just have the handler return a batch of the failures as `batchItemFailures`


see [src](https://docs.aws.amazon.com/lambda/latest/dg/example_serverless_SQS_Lambda_batch_item_failures_section.html)
```tsx
// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
// SPDX-License-Identifier: Apache-2.0
import { SQSEvent, SQSBatchResponse, Context, SQSBatchItemFailure, SQSRecord } from 'aws-lambda';

export const handler = async (event: SQSEvent, context: Context): Promise<SQSBatchResponse> => {
    const batchItemFailures: SQSBatchItemFailure[] = [];

    for (const record of event.Records) {
        try {
            await processMessageAsync(record);
        } catch (error) {
            batchItemFailures.push({ itemIdentifier: record.messageId });
        }
    }

    return {batchItemFailures: batchItemFailures};
};

async function processMessageAsync(record: SQSRecord): Promise<void> {
    if (record.body && record.body.includes("error")) {
        throw new Error('There is an error in the SQS Message.');
    }
    console.log(`Processed message ${record.body}`);
}

```

apps/aants/src/emailProcessor.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+/**
+ * Lambda handler for processing SQS messages containing individual email requests.
+ * 
+ * @param event - The SQS event containing up to 14 records (one email each)
+ * @returns Partial batch response indicating which messages failed
+ */
+export async function handler(event: SQSEvent){
```
jotalis: would also be nice to type the responses of `handler` and `processEmailRecords`

apps/aants/src/helpers/notificationDispatch.ts
```diff
@@ -104,65 +103,65 @@ async function sendNotification(
...
+            userId: '',
+            quarter: '',
+            year: '',
+            stagingPrefix: stagingPrefix,
+        });
+
+        // Send each email as a separate SQS message
+        await Promise.all(
+            usersWithEmail.map((user) =>
+                queueEmail(queueUrl, {
```
jotalis: thought: it might be slightly better for network volume to batch send to SQS here, but the scale we're operating at here is basically negligible

apps/aants/src/helpers/notificationDispatch.ts
```diff
@@ -104,65 +103,65 @@ async function sendNotification(
...
+            instructor: '',
+            days: '',
+            hours: '',
+            time: '',
+            sectionCode: '',
+            userId: '',
+            quarter: '',
+            year: '',
+            stagingPrefix: stagingPrefix,
+        });
```
jotalis: question: why are we constructing `templateData` and passing it to `EmailRequest`? it doesn't seem like we are using it when we process and send the email commands?

```tsx
async function processEmailRecords(records: SQSEvent['Records']) {
    for (const record of records) {
        try {
            const emailRequest: EmailRequest = JSON.parse(record.body);

            const command = new SendEmailCommand({
                FromEmailAddress: emailRequest.FromEmailAddress,
                Destination: emailRequest.Destination,
                Content: {
                    Template: {
                        TemplateName: emailRequest.TemplateName,
                        TemplateData: emailRequest.ReplacementTemplateData,
                    },
                },
            });

            await sesClient.send(command);
        } catch (error) {
            console.error(`Failed to send email for record ${record.messageId}:`, error);
        }
    }
}
```

apps/aants/src/index.ts
jotalis, line 176: why are we batching course codes here?

apps/aants/src/emailProcessor.ts
IsaacNguyen: Oh yeah sorry about that. I was looking into doing something with retrying emails but got rid of it because I wasn't sure if that was something I should do and forgot to get rid of this comment. I'll go ahead and implement it and a DLQ if that's the correct practice for this kind of thing

apps/aants/src/helpers/notificationDispatch.ts
IsaacNguyen: I didn't really get to test out my code too much because I just wanted to put up the PR and I got busy after, but I just realized there's a `SendTemplatedEmailCommand` which needs the `TemplateData`. I'll refactor the code this weekend to use this

apps/aants/src/index.ts
IsaacNguyen: I batched course codes here because Andrew said that WebSoc could only handle a request with around 450 course codes at a time. I've been noticing that AANTS hasn't been correctly scraping all the classes, so I need to look into that this weekend

KevinWu098: Are we querying websoc directly (?) — wouldn't we just hit the AAPI which doesn't have these problems? (although perhaps we could run into load issues...?) 

but i guess my point being that the 450 course codes on websoc... doesn't apply to us here if we're hitting aapi

apps/aants/src/emailProcessor.ts
jotalis: yeah DLQs are pretty standard

apps/aants/src/index.ts
jotalis: i think so via @icssc/libwebsoc-next -- does AAPI incrementally update their cache / db for this data instead of running scheduled job that retrieves all data at once? if they update incrementally, then you could directly fetch from AAPI at a rate even lower than 5 min to send emails for those incrementally updated classes, ultimately reducing how many emails you'd have to send out each batch. probably wont hit or encounter this scalability problem anytime soon though

IsaacNguyen: I'm honestly unsure about how AAPI fetches their data—I could ask them. However, I can explain why we scrape WebSoc directly. Last year, Minh and Dante had a long discussion about a possible race condition between AANTS and AAPI. Basically, at least last year, AAPI would scrape its data from WebSoc every minute or something like that. There becomes a problem that is AANTS is polling AAPI, there is some kind of race condition where AANTS won't properly see that a class status has updated. I honestly can't remember the specifics, but it was a hot point of contention following an ICSSC meeting where the two argued for almost 2 hours. I just remember the consensus being that scraping WebSoc directly made the most sense.

alexespejo (approved):
Seems legit

```
                   |    |    |
                  )_)  )_)  )_)
                 )___))___))___)\
                )____)____)_____)\\
              _____|____|____|____\\\__
    ---------\                         /---------
      ^^^^^   ^^^^^^^^^^^^^^^^^^^^^^^^
        ^^^^        ^^^^      ^^^    ^^
              ^^^^        ^^^
```

KevinWu098:
Hold uppp, I wanna get eyes on this before merging into AANTS merge

apps/aants/src/helpers/notificationDispatch.ts
```diff
@@ -101,68 +103,49 @@ async function sendNotification(
         notification = notification.replace(/\n/g, '<br>');
```
jotalis: can we improve the readability of this code? also why do we have to do `notification = notification.replace(/\n/g, '<br>');` ? can we not directly put <br> in the string creation?

a cleaner implementation of this might look like this:
```tsx
const parts = [];

if (statusChanged) {
    const status = currentStatus === 'Waitl' ? 'WAITLISTED' : currentStatus;
    parts.push(`The class is now <strong>${status}</strong>`);
}

if (codesChanged) {
    parts.push(`The class now has restriction codes <strong>${restrictionCodes}</strong>`);
}

const notification = parts.map(p => `- ${p}`).join('<br>');
```

apps/aants/src/helpers/emailQueue.ts
```diff
@@ -0,0 +1,34 @@
...
+            QueueUrl: queueUrl,
+            MessageBody: JSON.stringify(emailRequest),
+        });
+
+        await sqsClient.send(command);
+    } catch (error) {
+        console.error('Error queueing email to SQS:', error);
+        throw error;
+    }
+}
```
jotalis, line 36: instead of passing in queueUrl as a param, can we just define it in the file at the top? it's slightly cleaner and centralizes things

```tsx
const QUEUE_URL = process.env.EMAIL_QUEUE_URL;
```
then update the function and the places where you call the function:


```suggestion
/**
 * Sends a single email request to SQS queue.
 * Each email becomes its own SQS message for individual processing.
 *
 * @param emailRequest - The email data to be queued
 * @returns Promise that resolves when the message is successfully sent to the queue
 */
export async function queueEmail(emailRequest: EmailRequest): Promise<void> {
    try {
        const command = new SendMessageCommand({
            QueueUrl: QUEUE_URL,
            MessageBody: JSON.stringify(emailRequest),
        });

        await sqsClient.send(command);
    } catch (error) {
        console.error('Error queueing email to SQS:', error);
        throw error;
    }
}
```

apps/aants/src/emailProcessor.ts
jotalis, line 1: nice

apps/aants/src/index.ts
```diff
@@ -90,35 +104,94 @@ function processSchool(school: WebsocSchool, quarter: string, year: string) {
```
jotalis, line 99: these return Promises, can we make the function explicitly async for clarity and to ensure we properly catch errors

apps/aants/src/index.ts
```diff
@@ -90,35 +104,94 @@ function processSchool(school: WebsocSchool, quarter: string, year: string) {
...
-        process.exit(0);
+        console.error(
+            '[ERROR] Error in managing subscription:',
+            error instanceof Error ? error.message : String(error)
+        );
+        throw error;
     }
 }
+
+scanAndNotify();
```
jotalis: would this be problematic? when we import this file in `lambda.ts` won't this line run, meaning `scanAndNotify()` runs twice? if you want to keep the pnpm run aants thing u can scope it


```suggestion
if (require.main === module) {
    scanAndNotify();
}
```

jotalis (changes requested):
haven't tested locally but code looks good! just a few small nits

also i assume we've already requested an AWS production account / quota increase for SES?

apps/aants/src/index.ts
KevinWu098: The inner functions should also be awaited, I believe

KevinWu098 (commented):
Looks good, and I trust @alexespejo and @jotalis 's judgment here, so I haven't inspected super closely.

Only note I have is on why we have so many additional formatting diffs? I recall you had trouble with getting it to pass, yes, but shouldn't that have been a problem on `aants-merge` too?

/shrug ultimately i don't really care but if the diff can be made cleaner that'b dope

IsaacNguyen:
> haven't tested locally but code looks good! just a few small nits
> 
> also i assume we've already requested an AWS production account / quota increase for SES?

Yeah we have AWS production. In terms of a quota increase, that's something that @andrew-wang0 would have to do

IsaacNguyen:
> Looks good, and I trust @alexespejo and @jotalis 's judgment here, so I haven't inspected super closely.
> 
> Only note I have is on why we have so many additional formatting diffs? I recall you had trouble with getting it to pass, yes, but shouldn't that have been a problem on `aants-merge` too?
> 
> /shrug ultimately i don't really care but if the diff can be made cleaner that'b dope

I'm pretty sure every file I reformatted was because it was failing the new eslint workflow though. Like I'm not quite sure if I can undo those changes. It's not a problem on `aants-merge` because it's currently failing the eslint workflow I think...

image

jotalis (approved):
🔥

KevinWu098 (approved):

alexespejo (approved):
Send it

PR #1450, opened 2026-01-23
fix: improve fuzzy search keywords

### LINKED ISSUE #1282: Improve search keywords
- E.g. "sociology"

sicn4rf:
## Summary
Add department `name` field to fuzzy search keys so searches like "sociology" or "computer science" match departments by their full name, not just abbreviations. Previously just uses `['id', 'alias']`, that's why "sociol" "cs" "compsci" works but not the full word "sociology".

## Test Plan
1. Search "sociology" -> SOCIOL department now appears in results
2. Search "computer science" -> COMPSCI department now appears in results
3. Search by alias/id "CS" and "COMPSCI" still works.

Before:
image

After:
image

## Issues

Closes #1282

### DIFF
--- a/apps/antalmanac/src/backend/routers/search.ts
+++ b/apps/antalmanac/src/backend/routers/search.ts
@@ -92,7 +92,7 @@ const searchRouter = router({
                 matchedSections.length === MAX_AUTOCOMPLETE_RESULTS
                     ? []
                     : fuzzysort.go(query, searchData.departments, {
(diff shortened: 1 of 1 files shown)


alexespejo (approved):
LGTM. It feels awkward that so many departments will appear even if users don't necessarily search like that but I'll approve on the argument that users rarely type full dept names to find a course

PR #1453, opened 2026-01-23
fix: Improve search relevancy

### LINKED ISSUE #1449: Fuzzy search results should prioritize courses offered this quarter
The fuzzy search applies a limit of 12 results before sorting by availability. This means searching "CS" returns mostly "Not Offered", since it just takes the first 12 matches regardless of whether it's relevant to the quarter or not. Even though many CS courses are actually offered this quarter, the autocomplete result is dominated not offered courses. 
Image

sicn4rf:
## Summary
Fix fuzzy search to prioritize relevant offered courses in the current quarter.

**Changes**:
- *Courses*: Pre-compute a hashset of offered courses for O(1) lookup time. Increase the limit of course matches to 100, sort them by availability, then slice down to 12 (or however much space is left)
- *Departmets*: Add a threshold of 0.8 to filter out weak matches (e.g., "econ" no longer shows "Molecular Biology and Biochemistry")

Previously, search results were dominated by "Not Offered" courses and irrelevant department matches.

## Test Plan
1. Search "cs":
- Shows 3 relevant departments and 9 offered courses.

2.  Search "econ":
- Shows only the "Economics" department, and 11 offered econ courses. No irrelevant departments

### Before
image
image

### After
image
image

## Issues

Closes #1449

### DIFF
--- a/apps/antalmanac/src/backend/routers/search.ts
+++ b/apps/antalmanac/src/backend/routers/search.ts
@@ -39,14 +39,8 @@ const toGESearchResult = (key: GECategoryKey): [string, SearchResult] => [
 
 const toMutable = <T>(arr: readonly T[]): T[] => arr as T[];
 
-const isCourseOffered = (
-    department: string,
-    courseNumber: string,
-    termSectionCodes: Record<string, SectionSearchResult>
-): boolean => {
-    return Object.values(termSectionCodes).some((section) => {
-        return section.department === department && section.courseNumber === courseNumber;
-    });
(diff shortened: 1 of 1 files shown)


jotalis (approved):
lgtm, v nice speedup and quality of life ux improvement 👍 

only thing i would comment on is potentially tweaking the threshold number to be slightly less restrictive; maybe run some quick experiments on a larger variety of queries to see where to move the needle but 0.7 seems fine for now

PR #1456, opened 2026-01-24
Fix: Disable HelpMenu container's pointer events

### LINKED ISSUE #1445: HelpMenu element blocking clicks
The help menu (in the bottom right) blocks clicks even when it's collapsed. Probably shouldn't happen

Choollol:
## Summary

Fixes a bug where the HelpMenu would block pointer events even when the `SpeedDial` was closed.

## Test Plan

Confirm that the area just above the help menu button isn't blocking cursor interaction
Example:
1. Go to map tab
2. Check that hovering cursor over area above help menu button keeps the cursor a grab pointer, and that dragging moves the map

## Issues

Closes #1445

### DIFF
--- a/apps/antalmanac/src/components/HelpMenu/HelpMenu.tsx
+++ b/apps/antalmanac/src/components/HelpMenu/HelpMenu.tsx
@@ -8,7 +8,6 @@ import {
     type SvgIconProps,
     Tooltip,
     IconButton,
-    SxProps,
 } from '@mui/material';
 import { useCallback, useState } from 'react';
 
@@ -29,11 +28,6 @@ export type HelpMenuAction = {
     disableOnMobile?: boolean;
     onClick: VoidFunction;
 } | null;
-
-/** When adding a new item, add this to its `sx` prop */
-const commonChildItemStyles: SxProps = {
-    pointerEvents: 'auto',
-};
 
 export function HelpMenu() {
     const isMobile = useIsMobile();
@@ -76,6 +70,9 @@ export function HelpMenu() {
                 right: 8,
                 zIndex: theme.zIndex.fab,
                 pointerEvents: 'none',
+                '& > *:not(.speedDial)': {
+                    pointerEvents: 'auto',
+                },
             })}
             spacing={1}
             alignItems="center"
@@ -90,6 +87,7 @@ export function HelpMenu() {
 
             <SpeedDial
                 ariaLabel="Help Menu"
+                className="speedDial"
                 icon={
                     <Tooltip title="Help Menu" placement="left">
                         <Box
@@ -135,7 +133,6 @@ export function HelpMenu() {
                     size="large"
                     onClick={handleOpenAutoSaveWarning}
                     sx={{
-                        ...commonChildItemStyles,
                         bgcolor: isDark ? 'warning.dark' : 'warning.main',
                         color: 'white',
                         height: '4rem',
--- a/apps/antalmanac/src/components/HelpMenu/HelpMenu.tsx
+++ b/apps/antalmanac/src/components/HelpMenu/HelpMenu.tsx
@@ -29,6 +29,8 @@ export type HelpMenuAction = {
     onClick: VoidFunction;
 } | null;
 
+const SPEEDDIAL_CLASS_NAME = 'speedDial';
+
 export function HelpMenu() {
     const isMobile = useIsMobile();
     const [open, setOpen] = useState(false);
@@ -70,7 +72,7 @@ export function HelpMenu() {
                 right: 8,
                 zIndex: theme.zIndex.fab,
                 pointerEvents: 'none',
-                '& > *:not(.speedDial)': {
+                [`& > *:not(.${SPEEDDIAL_CLASS_NAME})`]: {
                     pointerEvents: 'auto',
                 },
             })}
@@ -87,7 +89,7 @@ export function HelpMenu() {
 
             <SpeedDial
                 ariaLabel="Help Menu"
-                className="speedDial"
+                className={SPEEDDIAL_CLASS_NAME}
                 icon={
                     <Tooltip title="Help Menu" placement="left">
                         <Box
--- a/apps/antalmanac/src/components/HelpMenu/HelpMenu.tsx
+++ b/apps/antalmanac/src/components/HelpMenu/HelpMenu.tsx
@@ -8,6 +8,7 @@ import {
     type SvgIconProps,
     Tooltip,
     IconButton,
+    SxProps,
 } from '@mui/material';
 import { useCallback, useState } from 'react';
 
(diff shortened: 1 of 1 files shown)


apps/antalmanac/src/components/HelpMenu/HelpMenu.tsx
```diff
@@ -29,6 +30,11 @@ export type HelpMenuAction = {
     onClick: VoidFunction;
 } | null;
 
+/** When adding a new item, add this to its `sx` prop */
+const commonChildItemStyles: SxProps = {
+    pointerEvents: 'auto',
+};
```
KevinWu098: issue: i don't really understand why we'd extract out specifically the pointer events...? Why not the rest? I admit i don't think we should be doing this

KevinWu098 (changes requested):

apps/antalmanac/src/components/HelpMenu/HelpMenu.tsx
Choollol: It's because the container needs to have `pointerEvents: 'none'` which propagates to its children, so I added this to make it clearer that any new additions need to have this style in order to be cursor-interactable.

I thought about a couple other possible solutions, like using a HOC, but this seemed like the best option to me. If you have any suggestions, let me know!

KevinWu098: But you use it in only one place? Sorry if im missing something but this seems really odd to me.

There are other styles passed to the children, it's very unorthodox that we're doing this as a wholly separate sx object

Choollol: It's only being used in one place right now, but if anything gets added to the component, like another warning alert or something, it'll have pointer events disabled by default. Whoever adds it might be really confused for quite a while before figuring out that they should be setting pointer events to auto, which is what I want to prevent with this shared variable.

KevinWu098:
```ts
                pointerEvents: 'none',
                '& > *': { pointerEvents: 'auto' },
```

~~Thoughts?~~ man nvm i thought this was gonna work, i hate this component

KevinWu098: Putting aside the shared variable business, which still seems like a code smell, it also seems downstream of another code smell which is that we have to manage the pointer events (?). Seems really painful...

Is there no way to... reduce the height? or not render it based on `open`? or something? i dont have great visibility here ofc, but I don't really love having to mess with pointer events

Choollol: Rendering the speed dial items based on `open` breaks the fancy animation. I tried having two open states, but it makes the elements shift a bit since the items are wider than the open/close button.

Do you know if it's possible to make the container's height smaller than its children? I couldn't find a way to shrink the height while keeping the speed dial items in the container's flow.

Choollol: > ```ts
>                 pointerEvents: 'none',
>                 '& > *': { pointerEvents: 'auto' },
> ```
> 
> ~Thoughts?~ man nvm i thought this was gonna work, i hate this component

How do you feel about adding this?
```ts
'& > *:not(.speedDial)': {
    pointerEvents: 'auto',
},
```
and `className="speedDial"` to the `SpeedDial` component?
Seems to work, though if anyone adds a new component that needs to change `pointerEvents`, they might be confused for a bit.

KevinWu098: Works for me!

Choollol: Done!

KevinWu098 (approved):
I'm gonna approve because this is pedantic, but I believe the correct thing to do here is to actually just use the stable MUI className

```ts
                [`& > *:not(.MuiSpeedDial-root)`]: {
                    pointerEvents: 'auto',
                },
```

Rather than defining our own stuff and adding more complexity, I think it's perfectly reasonable that devs on a MUI codebase should understand that that className is from MUI. Seems cleaner, no additional custom-defined class name, more "MUI"

Choollol:
> I'm gonna approve because this is pedantic, but I believe the correct thing to do here is to actually just use the stable MUI className
> 
> ```ts
>                 [`& > *:not(.MuiSpeedDial-root)`]: {
>                     pointerEvents: 'auto',
>                 },
> ```
> 
> Rather than defining our own stuff and adding more complexity, I think it's perfectly reasonable that devs on a MUI codebase should understand that that className is from MUI. Seems cleaner, no additional custom-defined class name, more "MUI"

That makes sense to me! One concern I have though is that it's not immediately apparent why only the speed dial is ignored, since many readers probably won't know that speed dial uses pointer events to handle enabling/disabling. The added complexity is easier to understand imo, but I can also get some others' opinions on this!

KevinWu098:
> > I'm gonna approve because this is pedantic, but I believe the correct thing to do here is to actually just use the stable MUI className
> > ```ts
> >                 [`& > *:not(.MuiSpeedDial-root)`]: {
> >                     pointerEvents: 'auto',
> >                 },
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Rather than defining our own stuff and adding more complexity, I think it's perfectly reasonable that devs on a MUI codebase should understand that that className is from MUI. Seems cleaner, no additional custom-defined class name, more "MUI"
> 
> That makes sense to me! One concern I have though is that it's not immediately apparent why only the speed dial is ignored, since many readers probably won't know that speed dial uses pointer events to handle enabling/disabling. The added complexity is easier to understand imo, but I can also get some others' opinions on this!

/shrug up to you here — adding a comment is perhaps sufficient here. In any case, just remember that this is a < 10 line change, don't get too bogged down over it!

Choollol:
> > > I'm gonna approve because this is pedantic, but I believe the correct thing to do here is to actually just use the stable MUI className
> > > ```ts
> > >                 [`& > *:not(.MuiSpeedDial-root)`]: {
> > >                     pointerEvents: 'auto',
> > >                 },
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Rather than defining our own stuff and adding more complexity, I think it's perfectly reasonable that devs on a MUI codebase should understand that that className is from MUI. Seems cleaner, no additional custom-defined class name, more "MUI"
> > 
> > 
> > That makes sense to me! One concern I have though is that it's not immediately apparent why only the speed dial is ignored, since many readers probably won't know that speed dial uses pointer events to handle enabling/disabling. The added complexity is easier to understand imo, but I can also get some others' opinions on this!
> 
> /shrug up to you here — adding a comment is perhaps sufficient here. In any case, just remember that this is a < 10 line change, don't get too bogged down over it!

👍I'll go with your suggestion and add a comment explaining a little more.

KevinWu098 (approved):

PR #1460, opened 2026-01-25
1271 importexport schedules as file

### LINKED ISSUE #1271: Import/Export schedules as file
## Description
- There should be a way to move data from prod to staging to facilitate testing.
    - This would enable beta testers to copy over their actual schedule from prod to staging and vice-versa, better incentivizing them to test staging.
- The simplest way to do this without sharing data between prod and staging (which should be avoided at all costs) is to dump the entire schedule as a file.
    - JSON would be the obvious choice. We just have to remember to do *some* validity checking instead of blindly casting.

## Related Issues
- Enables #1241

brelieu05:
## Summary

Added new import type in “Import Schedule” Dialog file
- User can drag and drop or click to browse file
- Validates schedule and makes sure no malformed data in json (checks all necessary keys)
- Before merging to user’s schedules, allows user to select which schedule to import

Added export button (couldn't find an appropriate opposite to the import icon in mui-icons)
- User can select which schedules to export
- Exports schedules as json

https://streamable.com/rbna9b

image
image
image
image
image

## Test Plan
- Tested with malformed and empty jsons to ensure validation worked
- Imported and exported using different schedules, including schedules with custom events

## Issues
- Appropriate icon for export
- Where to put dev tools on sidebar

Closes #1271


apps/antalmanac/src/components/Header/SettingsMenu.tsx
```diff
@@ -215,6 +225,23 @@ function ExperimentalMenu() {
                 </Box>
                 <Switch color={'primary'} value={autoSave} checked={autoSave} onChange={handleAutoSaveChange} />
             </Box>
+
+            <Box style={{ display: 'flex', justifyContent: 'space-between', width: '1' }}>
+                <Box style={{ display: 'flex', alignItems: 'center', gap: 4 }}>
+                    <Typography variant="h6" style={{ display: 'flex', alignItems: 'center', alignContent: 'center' }}>
```
alexespejo: It feels awkward to have a dedicated toggle to show one button.

alexespejo (commented):
I think we should look back at this post merge. I think we can decide on a way to consolidate a settings/special menu for features we don't want to fully expose to users

There will be a conflict since we're getting rid of the sidebar. You can decide on whether to close this PR

alexespejo:
Merge in main. Main no longer has the sidenav so you'll prob have to reorient the ui

brelieu05:
Instead of having a whole switch just for a button, I just added on to the existing import button.

image
image


image
image

alexespejo (changes requested):
issue: I don't think importing the exported schedule works. I'm also not a big fan of having a dedicated toggle just to enable this feature it feels kinda awkward even if it's not meant for normal users. 

thought: What if we expand this PR and create a "DEV mode" where we can basically hide special features like this behind a toggle. Power users can also use this feature if they really wanted to. 

CC @KevinWu098

alexespejo:
This is still a low prio task and we'll pivot the change to a Dev Mode feature #1545

Issue #1367, opened 2025-12-12
Refactor Auth


KevinWu098:
From #1237 

```ts
        // ! TODO @KevinWu098
        // ! Auth uses hardcoded migration logic to handle cases in which stale userIDs
        // ! still contain non OIDC google ids. This is not correct and needs to be fixed.
        // ! Auth and operations upon users and accounts should not depend on localStorage. This is a hack.
```

YKawesome:
Is this causing a sticky login issue? I accidentally signed in with my personal account and I can't seem to find a way to switch accounts even after clearing cache/cookies :(

Also (perhaps unrelated), if you cancel out of the OAuth flow (don't allow ICSSC to access your acct or whatever) you get an api error (or whatever it's called, where you get like "oath rejected" as a json response).

KevinWu098:
> Is this causing a sticky login issue? I accidentally signed in with my personal account and I can't seem to find a way to switch accounts even after clearing cache/cookies :(
> 
> Also (perhaps unrelated), if you cancel out of the OAuth flow (don't allow ICSSC to access your acct or whatever) you get an api error (or whatever it's called, where you get like "oath rejected" as a json response).

Ah hm, I didn't think the logout thing was affecting people in prod. It's a one-liner fix I can put up, thanks for flagging.

The OAuth flow cancelling is a little more tricky, I think, but I can open an issue for it in https://github.com/icssc/auth

YKawesome:
Thanks Kevin :)

KevinWu098:
Largely resolved by #1611 and #1744, with further work (localStorage) being tracked in #1809

PR #1463, opened 2026-01-27
fix: logout url redirect

KevinWu098:
## Summary

Seems like this regressed some (brief) time ago, as this logic 100000% was a part of #1237. Fixed after being raised via #1367 

## Issues

Closes #


KevinWu098:
Fixed, CC @ykawesome

PR #1469, opened 2026-01-30
feat: implement dynamic responsive resizing using ghost measurement

### LINKED ISSUE #1442: Use dropdown menu in Calendar Toolbar
Image

On mobile, some of the toolbar is moved to a dropdown. We should also do this on a tablet/small laptop screen size so it doesn't wrap

sicn4rf:
## Summary
Fixed toolbar wrapping issues by switching from mobile detection to available-space detection. The toolbar now dynamically condenses based on real-time width measurements.

- Added `useElementSize` hook for reactive DOM measurements
- Added a hidden "ghost" toolbar to measure the required width
- Added logic to condense the toolbar to a dropdown menu when `availableWidth < requiredWidth`
- Fixed `react-split` snapping behavior in `Home.tsx` for smoother resizing

Why?
I explicitly moved away from using a static magic number threshold. The original wrapping bug was caused by the previous hardcoded threshold that didn't anticipate new features (like the Redo button), causing the content to overflow and wrap.

By implementing a dynamic measurement system, the toolbar is now ***future-proof*** (i hope.. ?). We can add as many buttons as we want or add text to the left-side buttons, and the layout will automatically calculate the new required qidth and condense only when necessary, eliminating the risk of wrapping.

## Test Plan
1. Drag the split-pane divider to the left to narrow the view.
2. Verify the toolbar switches to the dropdown menu seamlessly when buttons run out of space.
3. Drag the split-pane to the right to verify the buttons revert to the full view when enough space is available.
4. Choose a long name for the schedule so the button is slightly longer.
5. Drag the split-pane again to narrow the view.
6. Verify it still doesn't wrap and condenses at the right point.
7. Reduce the size of my window and narrow my window size
8. Verify the toolbar switches to the dropdown menu seamlessly when buttons run out of space.

## Issues

Closes #1442


alexespejo (changes requested):
Remind me if we wanted to enclose all of the buttons into the drop down menu on super small screens? 

Merge in main the breakpoints subtly changed  

Image

sicn4rf:
> Remind me if we wanted to enclose all of the buttons into the drop down menu on super small screens?
> 
> Merge in main the breakpoints subtly changed
> 
> Image

It wasn't in the initial plan, we simply reused the condensed toolbar that we used strictly for mobile screens but applied it to small laptop/ipad screens as well. In the screenshot you included, it was when you did `cmd + '+'` to  zoom in but I believe if you tried to slide the center bar again, it'll revert the wrap and lock you into a new minimum width. I think this is just because it doesn't re-render on zoom.

alexespejo:
> > Remind me if we wanted to enclose all of the buttons into the drop down menu on super small screens?
> > Merge in main the breakpoints subtly changed
> > Image
> 
> It wasn't in the initial plan, we simply reused the condensed toolbar that we used strictly for mobile screens but applied it to small laptop/ipad screens as well. In the screenshot you included, it was when you did `cmd + '+'` to zoom in but I believe if you tried to slide the center bar again, it'll revert the wrap and lock you into a new minimum width. I think this is just because it doesn't re-render on zoom.

Yeah the sc was with a window with a shrunken width. It's alright if it doesn't collapse all of the buttons. Just wondering if it's behavior we want to adopt

alexespejo:
Changes seem chill. Resolve the merge conflicts

sicn4rf:
### Update on final implementation before merging
We looked deeper into the initial dynamic javascript implementation and concluded that it was unnecessarily complex for a component that rarely changes. The final iteration used css container queries to automatically condense from its full view when its container width is less than or equal to 500px. 

We do this by always having both versions present on screen, but simply altering their `display: ` property between `flex` and `none` depending on the width of its parent container. 

This handles both mobile screens and the desktop split pane resizing without the need of extra listeners.

sicn4rf:
> Changes seem chill. Resolve the merge conflicts

Merge conflicts have been resolved @alexespejo

alexespejo (approved):
LGTM 🇵🇭

PR #1474, opened 2026-02-01
fix: calendar left header gutter flickering while width is set

### LINKED ISSUE #1444: Calendar Header has hydration (?) problems
Image

We can see, when refreshing the page, that the header on the calendar (with the days of week) flickers as it adjusts its width. This should not happen and the calendar should be sized correctly always

Choollol:
## Summary

Fix a bug where rerendering the calendar would make the header flicker as the left gutter's width was being set.

## Test Plan

1. Switch between finals/normal schedule and confirm that the header doesn't flicker

## Issues

Closes #1444

### DIFF
--- a/apps/antalmanac/src/components/Calendar/calendar.css
+++ b/apps/antalmanac/src/components/Calendar/calendar.css
@@ -31,7 +31,8 @@
     border: 0;
 }
 
-.rbc-time-slot {
+.rbc-time-slot,
+.rbc-time-header-gutter {
     min-width: 42px;
 }
 


KevinWu098 (commented):
the mobile heading looks kinda weird now, right?

Image

Choollol:
> the mobile heading looks kinda weird now, right?
> Image

Looks the same as prod to me
image
image

KevinWu098:
> > the mobile heading looks kinda weird now, right?
> > Image
> 
> Looks the same as prod to me image image

oh wow, never noticed 

https://tenor.com/view/bahaha-lol-hd-gif-minion-minion-laugh-gif-2154867417577880306

KevinWu098 (approved):

KevinWu098:
side request — i think at some point i got rid of the white borders on desktop and forgot to check mobile. does the calendar look weird without left and right borders (*on mobile)?

Choollol:
> side request — i think at some point i got rid of the white borders on desktop and forgot to check mobile. does the calendar look weird without left and right borders (*on mobile)?

Hmm, the header's right border does look a tiny bit out of place.
image
The left side looks fine though!

PR #1476, opened 2026-02-02
fix: anchor schedule select popover's position to button

### LINKED ISSUE #1475: Fix schedule select popover position
Make `ScheduleSelectPopover`'s position based on the button instead of being hard coded.

Choollol:
## Summary

Set `ScheduleSelectPopover`'s position to be anchored to the button instead of having a hardcoded position.

## Test Plan

1. Click the schedule select and ensure the popover opens in the correct position
2. Locally, temporarily set `openScheduleSelect` to default to true and ensure that reloading the page opens the schedule select popover in the correct position

## Issues

Closes #1475


KevinWu098:
you'll need to merge main

apps/antalmanac/src/components/Calendar/Toolbar/ScheduleSelect/ScheduleSelect.tsx
```diff
@@ -78,6 +81,10 @@ export function SelectSchedulePopover() {
         setCurrentScheduleIndex(AppStore.getCurrentScheduleIndex());
     }, []);
 
+    useEffect(() => {
```
alexespejo: issue: looking at it a little closer I don't think you need useState/useEffect here. Refs are set automatically after first render, and the Popover reads `anchorEl` lazily when it opens, so you can just pass `anchorEl={anchorElementRef.current}` directly.

alexespejo (changes requested):
Little code quality thing. Works well testing locally!

apps/antalmanac/src/components/Calendar/Toolbar/ScheduleSelect/ScheduleSelect.tsx
Choollol: Oh yeah, I missed that the popover doesn't need it immediately. Changed!

alexespejo (approved):
LGTM

PR #1480, opened 2026-02-02
RFC hide legacy schedule input in sign in dialog 

alexespejo:
## Summary
Since it's been ~1 year since auth shipped all active users should be using google auth so we should start to sunset the remnants of the legacy schedules. 

Request for comment (RFC) bc I'm not sure what the verbiage should be :) 

image

## Issues

Closes #


Choollol:
Maybe we could indicate that the legacy user IDs are outdated? Something like "Have schedules saved to an old user ID?"

Choollol (approved):
LGTM!

PR #1484, opened 2026-02-04
Map fix clean

### LINKED ISSUE #1458: Lock Map to UCI squares
Image

On prod, we only render a small area around UCI. users can still scroll beyond it, resulting in gray areas. mapbox should provide a way to limit the viewable area

victorialee13:
## Summary
Limits the map rendering viewable area to campus buildings so that users don't scroll into loading gray boxes
## Test Plan
Open map and drag map to go up, down, left, right. Limited in viewing further outside of buildings area. 
Tested at diff zoom levels.

Video demo: https://drive.google.com/file/d/1iAwy1Pj8OGUAgkgI4sgL_iQVLcfZqi7Y/view?usp=sharing
## Issues
One thing is that the map is able to go much more north then other directions because one of the buildings Non-OSHPD Central Plan is so far away north.

Closes #1458

### DIFF
--- a/apps/antalmanac/src/components/Map/Map.tsx
+++ b/apps/antalmanac/src/components/Map/Map.tsx
@@ -32,7 +32,7 @@ const WORK_WEEK = ['All', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri'];
 const FULL_WEEK = ['All', 'Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat'];
 const weekendIndices = [0, 6];
 const CAMPUS_CENTER: LatLngTuple = [33.6459, -117.842717];
-const CAMPUS_BOUND_DELTA = 0.04;
+const CAMPUS_BOUND_DELTA = 0.01;
 const CAMPUS_BOUNDS: [LatLngTuple, LatLngTuple] = [
     [CAMPUS_CENTER[0] - CAMPUS_BOUND_DELTA, CAMPUS_CENTER[1] - CAMPUS_BOUND_DELTA],
     [CAMPUS_CENTER[0] + CAMPUS_BOUND_DELTA, CAMPUS_CENTER[1] + CAMPUS_BOUND_DELTA],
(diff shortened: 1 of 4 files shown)


apps/antalmanac/src/components/Map/Map.tsx
```diff
@@ -299,21 +305,31 @@ export default function CourseMap() {
...
             id="map-pane"
         >
-            <MapContainer ref={map} center={[33.6459, -117.842717]} zoom={16} style={{ height: '100%' }}>
+            <MapContainer
+                ref={map}
+                center={CAMPUS_CENTER}
+                zoom={16}
+                style={{ height: '100%' }}
+                maxBounds={CAMPUS_BOUNDS}
+                maxBoundsViscosity={1}
```
alexespejo, line 331: issue: I'm unfamiliar with what classes are offered outside of ring road. But I don't think the map should snap back to the center if you try to drag around.

alexespejo (changes requested):
There might be a way to clean up the git diff with the extra commit lol. I thought it would disappear

apps/antalmanac/src/components/Map/Map.tsx
```diff
@@ -31,6 +31,12 @@ const ATTRIBUTION_MARKUP =
 const WORK_WEEK = ['All', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri'];
 const FULL_WEEK = ['All', 'Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat'];
 const weekendIndices = [0, 6];
+const CAMPUS_CENTER: LatLngTuple = [33.6459, -117.842717];
+const CAMPUS_BOUND_DELTA = 0.001;
```
alexespejo:
```suggestion
const CAMPUS_BOUND_DELTA = 0.04;
```

suggestion: increase this value to somewhere between 0.009-0.04 maybe? Play around with it see what gives the nicest amount of space

alexespejo (changes requested):
Rest seems chill

alexespejo (approved):
Seem cool

PR #1486, opened 2026-02-05
Show only dates in finals calendar headers on mobile

### LINKED ISSUE #1485: Mobile finals calendar column header fix
The headers for each day on the finals schedule overlaps on smaller screens/mobile:

Image

Change it to just show the date since day of the week is already implied:

Image

Choollol:
## Summary

Fixes headers overlapping when viewing the finals calendar on mobile view/small screens.

## Test Plan

1. View finals calendar on mobile/a small screen and check that headers only show dates (excluding days of week) and don't overlap
2. Ensure that the headers for other views haven't changed

## Issues

Closes #1485

### DIFF
--- a/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
+++ b/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
@@ -16,6 +16,7 @@ import { CalendarEventPopover } from '$components/Calendar/CalendarEventPopover'
 import type { CalendarEvent, CourseEvent, SkeletonEvent } from '$components/Calendar/CourseCalendarEvent';
 import { CalendarToolbar } from '$components/Calendar/Toolbar/CalendarToolbar';
 import { skeletonBlueprintVariations } from '$components/Calendar/skeletonBlueprintVariations';
+import { useIsMobile } from '$hooks/useIsMobile';
 import {
     getLocalStorageSkeletonBlueprint,
     removeLocalStorageSkeletonBlueprint,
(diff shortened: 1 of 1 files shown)


alexespejo (changes requested):
thought: In PR #1469 we dynamically change the Calendar controls. Maybe that's a behavior we can adopt here too? – when the calendar window itself we should mimic this mobile/small screen behavior. 

suggestion: maybe in this PR we can format these values to remove the leading 0 (03/20 - > 3/20) 
Image

Choollol:
> thought: In PR #1469 we dynamically change the Calendar controls. Maybe that's a behavior we can adopt here too? – when the calendar window itself we should mimic this mobile/small screen behavior.
> 
> suggestion: maybe in this PR we can format these values to remove the leading 0 (03/20 - > 3/20) Image

Changed the leading zeroes!

Could you explain a little more about what you mean/would like to see with your mobile/small screen thought?

alexespejo:
> > thought: In PR #1469 we dynamically change the Calendar controls. Maybe that's a behavior we can adopt here too? – when the calendar window itself we should mimic this mobile/small screen behavior.
> > suggestion: maybe in this PR we can format these values to remove the leading 0 (03/20 - > 3/20) Image
> 
> Changed the leading zeroes!
> 
> Could you explain a little more about what you mean/would like to see with your mobile/small screen thought?

We have a breakpoint in the calendar panel that will contract the buttons in the menu if it's too small. I was thinking similar behavior could be used here but now that I'm thinking about it. I don't think it's small enough to be a big deal

https://github.com/user-attachments/assets/541963ca-c37f-48bf-928c-f3aebcf7b61f

Choollol:
> We have a breakpoint in the calendar panel that will contract the buttons in the menu if it's too small. I was thinking similar behavior could be used here but now that I'm thinking about it. I don't think it's small enough to be a big deal

Ohh, I see, like when the user is resizing in real time. It wouldn't matter here I think, since when the viewpoint gets small enough, the calendar switches to the mobile view and back to the normal course schedule again (non-finals calendar). Unless you were thinking something else?

alexespejo (approved):
LGTM 🥔

PR #1494, opened 2026-02-08
fix: cache generated folder and update only when necessary

dsnsgithub:
## Summary
Uses GitHub actions cache to cache existing term data, while ensuring that the current term data is up to date.

This should make staging and deployment builds significantly faster and reduce the risk of flaky API requests or rate limits.

## Test Plan
There should be no functional changes.


KevinWu098:
How does one handle partial data? Still trying to root cause this morning's scrape, but it appears that not all data was available at the same time.

KevinWu098:
Can we... perhaps just do something easier and only refetch terms which are < 2 years old (or do not exist)? Will cut down the time just about as much, and adds less complexity.

Not saying that this approach wouldn't be appropriate, but I am thinking more and more about how we can err on the side of less complexity, as the codebase continues to grow (but our average contributor tenure remains bounded)

dsnsgithub:
> How does one handle partial data? Still trying to root cause this morning's scrape, but it appears that not all data was available at the same time.

I made sure that it always checks the current term regardless of if it is cached or not, assuming the partial data would be the current term.

One solution, since AAPI/WebSoc doesn't release all courses at exactly the same time, would be to use @sicn4rf's workflow to run `pnpm --filter antalmanac get-data` periodically. It would update both cache and production. 

With partial data, we could try to rerun the production workflow instead of making a bunch of empty commits? It might be worth waiting for all data to arrive before re-deploying instead of trying to deploy at exactly 12:00am.

> Can we... perhaps just do something easier and only refetch terms which are < 2 years old (or do not exist)? Will cut down the time just about as much, and adds less complexity.
> 
> Not saying that this approach wouldn't be appropriate, but I am thinking more and more about how we can err on the side of less complexity, as the codebase continues to grow (but our average contributor tenure remains bounded)

Currently, it fetches terms that don't exist in the cache and refetches the current term. I can cut complexity by deciding to use the cache without checking current terms, relying on another workflow to update it. I am not sure refetching terms <2 years old is less complex, as you still need the cache for the older terms.

KevinWu098:
Ah hm, I forgot about the cache not being committed to Git 🤷‍♂️

okay I'm withdrawing my horse from this race -- one last note that summer terms are available at the same time together (although they should be small enough that one scrape is fine?)

KevinWu098:
@cubic

KevinWu098 (dismissed):
Let's get this merge conflict resolved and merge! 🚀

dsnsgithub:
@cubic

KevinWu098 (approved):

PR #1504, opened 2026-02-12
AANTS Staging Key Fix

IsaacNguyen:
## Summary
- We had an issue where multiple AANTS staging instances send duplicate emails
- To fix this, based on the staging branch that creates a notification, we will only allow that staging instance's cron job send notifications
   - In that way, only one staging instance is able to send the notification, avoiding duplicates
- This PR will also make sure every existing subscription gets filled with `production` when we add this new column `environment`
image

## Issues

Closes #

### DIFF
--- a/apps/aants/src/helpers/subscriptionData.ts
+++ b/apps/aants/src/helpers/subscriptionData.ts
@@ -50,13 +50,15 @@ async function getUpdatedClasses(
  */
 async function getSubscriptionSectionCodes(): Promise<TermGrouping | undefined> {
     try {
+        const stage = process.env.STAGE;
         const result = await db
             .selectDistinct({
                 sectionCode: subscriptions.sectionCode,
                 quarter: subscriptions.quarter,
                 year: subscriptions.year,
             })
-            .from(subscriptions);
+            .from(subscriptions)
+            .where(eq(subscriptions.environment, stage));
 
         // group together by year and quarter
         const groupedByTerm = result.reduce((acc: TermGrouping, { quarter, year, sectionCode }) => {
@@ -172,10 +174,12 @@ async function getUsers(
 
         const statusColumn = statusColumnMap[status];
 
+        const stage = process.env.STAGE;
         const baseConditions = [
             eq(subscriptions.year, year),
             eq(subscriptions.quarter, quarter),
             eq(subscriptions.sectionCode, sectionCode),
+            eq(subscriptions.environment, stage),
         ];
 
         let notificationCondition;
--- a/apps/antalmanac/src/backend/lib/rds.ts
+++ b/apps/antalmanac/src/backend/lib/rds.ts
@@ -669,9 +669,16 @@ export class RDS {
      * @param db - The database or transaction object to use for the operation.
      * @param userId - The ID of the user for whom we're upserting a notification.
      * @param notification - The notification object to upsert.
+     * @param environment - "production" on production; staging instance + number on staging (e.g. "staging-1337").
      * @returns A promise that upserts the notification associated with a userId.
      */
-    static async upsertNotification(db: DatabaseOrTransaction, userId: string, notification: Notification) {
+    static async upsertNotification(
+        db: DatabaseOrTransaction,
+        userId: string,
+        notification: Notification,
+        environmentValue?: string | null
+    ) {
+        const environment = environmentValue ?? '';
         return db.transaction((tx) =>
             tx
                 .insert(subscriptions)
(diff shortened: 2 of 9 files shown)


apps/aants/src/helpers/subscriptionData.ts
```diff
@@ -1,10 +1,19 @@
...
 import { db } from '../../../../packages/db/src/index';
 import { users } from '../../../../packages/db/src/schema/auth/user';
 import { subscriptions } from '../../../../packages/db/src/schema/subscription';
 
+function stagingInstanceCondition(): SQL {
+    const stage = process.env.STAGE;
+    // if STAGE is not set, return subscriptions with stagingInstance null (production)
+    if (stage === undefined || stage === '') {
+        return isNull(subscriptions.stagingInstance);
+    }
```
KevinWu098: issue: I don't like this. production should never be the fallback. this would mean that, in undefined cases, we'd (potentially) be spamming production. we should always err on the side of not impacting production in unexpected cases

apps/antalmanac/src/backend/lib/rds.ts
```diff
@@ -669,9 +669,16 @@ export class RDS {
      * @param db - The database or transaction object to use for the operation.
      * @param userId - The ID of the user for whom we're upserting a notification.
      * @param notification - The notification object to upsert.
+     * @param stagingInstance - Null on production; staging instance + number on staging (e.g. "staging-1337").
```
KevinWu098: issue: production should just be 'production'

packages/db/migrations/meta/_journal.json
```diff
@@ -1,62 +1,62 @@
...
+            "idx": 6,
+            "version": "7",
+            "when": 1763425968486,
+            "tag": "0006_cuddly_hulk",
+            "breakpoints": true
+        },
+        {
+            "idx": 7,
+            "version": "7",
+            "when": 1770856409376,
```
KevinWu098: issue: why does it seem like this is modifying an existing migration instead of creating a new migration?

KevinWu098 (changes requested):

packages/db/migrations/meta/_journal.json
IsaacNguyen: I was having this problem where it wasn't creating a new migration when I added the `stagingInstance` column after running `pnpm db:migrate`, so I just edited the most recent one which involves dropping the whole `subscriptions` table and adding in the new column. I can try to figure out how to decouple this change into its own migration

apps/antalmanac/src/backend/lib/rds.ts
IsaacNguyen: I just thought it would be misleading since `production` isn't technically a `stagingInstance` but I'll go ahead and change it

KevinWu098: environment, perhaps? or some other term that refers to 'local' | 'staging' | 'production' (but i think environment is semantically correct)

packages/db/migrations/meta/_journal.json
KevinWu098: what was the prior migration? if its related to this, then whatever i can live with it

IsaacNguyen: The prior migration was just dropping the `subscription` table and recreating it (it's written like that because we made another `subscriptions` table a while ago and it wouldn't let me add a bunch of columns to match the schema we have now), so adding in this column was just adding an extra column to it. 
image

KevinWu098: ok sure. how are you handling deleting old staging notifs?

apps/antalmanac/src/backend/lib/rds.ts
IsaacNguyen: Okay I'll go update it

packages/db/migrations/meta/_journal.json
IsaacNguyen: I just went into the dev DB deleted them

sst.config.ts
```diff
@@ -57,6 +57,7 @@ export default $config({
                 GOOGLE_REDIRECT_URI: `https://${domain}/auth`,
                 NEXT_PUBLIC_BASE_URL: domain,
                 NEXT_PUBLIC_PUBLIC_POSTHOG_KEY: process.env.NEXT_PUBLIC_PUBLIC_POSTHOG_KEY,
+                STAGE: $app.stage,
```
KevinWu098, line 60: question: have you smoke tested this? i just want to make sure this reads "staging-1333" and not just "staging"

IsaacNguyen: Yes! The NULL one is from making the subscription on another staging instance, but that won't happen again once I make it NOT NULL
image

KevinWu098 (approved):
Looks good — bit curious why this is deployed to staging-shared though. /shrug whatever!

PR #1509, opened 2026-02-17
fix: router transition error

dsnsgithub:
## Summary

Switch to normal `a` elements instead of attempting to client-side transition between two separate Next.js apps.

## Test Plan
Issue can be replicated by throttling internet speed, then attempting to move between scheduler and planner.
Multiple attempts at using the switcher will compound before falling back to a simple redirect, leading to extremely slow loading times.

image

## Issues

Fixes user complaint:
https://github.com/user-attachments/assets/7a437443-4098-43f3-99c8-c5096319ba7c

Fixes 
image
when transitioning between AntAlmanac scheduler and planner.

### DIFF
--- a/apps/antalmanac/src/components/Header/AppSwitcher.tsx
+++ b/apps/antalmanac/src/components/Header/AppSwitcher.tsx
@@ -1,0 +1,144 @@
+import { EventNote, Route, UnfoldMore } from '@mui/icons-material';
+import {
+    Button,
+    ButtonGroup,
+    CircularProgress,
+    ListItemIcon,
+    ListSubheader,
+    MenuItem,
+    MenuList,
+    Popover,
+    Typography,
+} from '@mui/material';
+import { useState } from 'react';
+
+import { Logo } from '$components/Header/Logo';
+import { BLUE } from '$src/globals';
+
+type AppSwitcherProps = {
+    isMobile: boolean;
+};
+
+export function AppSwitcher({ isMobile }: AppSwitcherProps) {
+    const [anchorEl, setAnchorEl] = useState<HTMLElement | null>(null);
+    const [plannerLoading, setPlannerLoading] = useState(false);
+
+    const platform = window.location.pathname.split('/')[1] === 'planner' ? 'Planner' : 'Scheduler';
+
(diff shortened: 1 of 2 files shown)


Choollol:
Current changes look good! I am wondering though, maybe there's a way we can make the switcher even more robust? We could separate it into its own component (which I think we should do anyway) and disable the button after the user clicks once to switch to the planner. Any thoughts on this?

dsnsgithub:
> Current changes look good! I am wondering though, maybe there's a way we can make the switcher even more robust? We could separate it into its own component (which I think we should do anyway) and disable the button after the user clicks once to switch to the planner. Any thoughts on this?

Separating the switcher into its own component makes sense to me, even if it slightly expands the scope of the PR. Disabling the switcher after clicking on it might confuse the user more, as they might think the planner isn't available at the moment 🤷‍♂️.

Choollol:
> > Current changes look good! I am wondering though, maybe there's a way we can make the switcher even more robust? We could separate it into its own component (which I think we should do anyway) and disable the button after the user clicks once to switch to the planner. Any thoughts on this?
> 
> Separating the switcher into its own component makes sense to me, even if it slightly expands the scope of the PR. Disabling the switcher after clicking on it might confuse the user more, as they might think the planner isn't available at the moment 🤷‍♂️.

What if we showed a loading spinner along with disabling the button?

dsnsgithub:
> > > Current changes look good! I am wondering though, maybe there's a way we can make the switcher even more robust? We could separate it into its own component (which I think we should do anyway) and disable the button after the user clicks once to switch to the planner. Any thoughts on this?
> > 
> > 
> > Separating the switcher into its own component makes sense to me, even if it slightly expands the scope of the PR. Disabling the switcher after clicking on it might confuse the user more, as they might think the planner isn't available at the moment 🤷‍♂️.
> 
> What if we showed a loading spinner along with disabling the button?

Sure, the switcher component is now separate and there is now a loading spinner. Since the switcher is a normal link, now there is both feedback from the browser tab loading and the button.

image

Choollol (approved):
LGTM!

PR #1511, opened 2026-02-19
feat: add friends and schedule linking

### LINKED ISSUE #911: Suggestion: View Friends' Schedules

brelieu05:
## Summary       
                                                                                           
 Builds off of the sharable schedule links system from #1357.         
                                                                               
  - Adds a full friends system: send/accept/decline/remove friend requests by email, block/unblock users, and 
  view a friend's shared schedule via a direct link.                                                          
  - Adds per-schedule sharing toggles so users can choose which of their schedules are visible to friends.    
  - Refactors all friends mutations to verify identity via server-issued session tokens instead of            
  client-supplied `providerAccountId`, closing an impersonation/spoofing vector.                              

<img width="338" height="193" alt="image" src="https://github.com/user-attachments/assets/ccb48088-a9cd-4b95-8a49-fffaf4734746" />
<img width="442" height="265" alt="image" src="https://github.com/user-attachments/assets/9a47aa10-71a4-455f-a857-db3712535215" />
<img width="1665" height="996" alt="image" src="https://github.com/user-attachments/assets/f9eb3621-9cdb-4116-8ef9-7d93a56507a8" />

                                                                                                              
## Features                                                                                                 
                                                                                                              
  ### Friends system                                                                                          
  - **Friend requests** — send by email, accept or decline incoming requests, prevent duplicate requests in
  either direction
  - **Unfriend / block / unblock** — full relationship lifecycle with a blocked state that prevents
  re-requesting
  - **View friend's schedule** — navigate to a friend's shared schedule via direct link; non-friends are      
  blocked from viewing                                                                                        
  - **Friends menu** — popover in the header with tabs for Friends, Requests, and Blocked. Skeleton loading   
  state while data fetches.                                                                                   
  - **Sign-in prompt** — clicking Friends while logged out opens a sign-in dialog instead                     
  - **Per-schedule sharing toggle** — each schedule row gets a `People`/`PeopleOutline` icon to control       
  whether friends can see it; defaults to visible for all existing schedules                                  
                                                                                                              
## DB changes                                                                                               
                                                                                                              
  - `packages/db/src/schema/friendship.ts` — new `friendships` table (`requesterId`, `addresseeId`, `status:  
  PENDING | ACCEPTED | BLOCKED`, timestamps)                                                                  
  - `packages/db/src/schema/schedule/schedule.ts` — added `sharedWithFriends boolean NOT NULL DEFAULT true`   
  - `packages/db/migrations/0008_old_misty_knight.sql` — friendships table + self-friend constraint           
  - `packages/db/migrations/0009_friendships_no_self_friend.sql` — check constraint preventing self-friendship
  - `packages/db/migrations/0010_schedule_shared_with_friends.sql` — new column on schedules     

## Test plan                                                                                                
                                                                                                              
  - [ ] Send a friend request by email — recipient sees it in the Requests tab                                
  - [ ] Accept the request — both users appear in each other's Friends tab                                    
  - [ ] Decline a request — row removed, no ACCEPTED state                                                    
  - [ ] Block a user — removed from friends, blocked in both directions                                       
  - [ ] Try to friend yourself — should return an error                                                       
  - [ ] Send a duplicate friend request — should show a contextual message, not crash                         
  - [ ] Toggle a schedule to hidden — friend visiting your link no longer sees that schedule                  
  - [ ] Hide all schedules — friend sees a warning banner instead of an error                                 
  - [ ] Non-friend tries to visit your schedule link — blocked  

## Issues

Closes #

<!-- [Optional]
## Future Followup
-->


apps/antalmanac/src/components/SharedScheduleBanner.tsx
```diff
@@ -0,0 +1,402 @@
...
+                }
+
+                const schedules = userData.userData.schedules;
+                const scheduleIndex = 0;
+                const loadSuccess = await AppStore.loadSchedule({ schedules, scheduleIndex });
+
+                if (!loadSuccess) {
+                    throw new Error('Failed to load friend schedules');
+                }
+                changeCurrentSchedule(scheduleIndex);
```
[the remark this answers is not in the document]
brelieu05: Keeping redundancy because load schedule does not load the 0th indexed schedule ever.

brelieu05:
@cubic-dev-ai

apps/antalmanac/src/backend/routers/friends.ts
```diff
@@ -0,0 +1,276 @@
...
+ * Router for handling friend-related operations.
+ */
+const friendsRouter = router({
+    /**
+     * Send a friend request by recipient email.
+     */
+    sendFriendRequestByEmail: procedure
+        .input(z.object({ requesterId: z.string(), email: z.string().email() }))
+        .mutation(async ({ input }) => {
+            const requesterId = input.requesterId;
```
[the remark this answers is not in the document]
brelieu05: @cubic-dev-ai

apps/antalmanac/src/backend/lib/rds.ts
```diff
@@ -413,6 +420,72 @@ export class RDS {
...
+    }
+
+    /**
+     * Retrieves a schedule by schedule ID. All schedules are publicly accessible via their ID.
+     *
+     * @param db - The database or transaction object to use for the query.
+     * @param scheduleId - The unique identifier of the schedule.
+     * @returns A promise that resolves to a ShortCourseSchedule object, or null if the schedule is not found.
+     */
+    static async getSharedScheduleById(
```
alexespejo: nit: Although this function is meant for the "shared schedules" feature I think a name like `getSchedulesByUserId`

apps/antalmanac/src/backend/routers/friends.ts
```diff
@@ -0,0 +1,334 @@
...
+    if (!session) {
+        throw new TRPCError({ code: 'UNAUTHORIZED', message: 'Invalid or expired session.' });
+    }
+    return session.userId;
+}
+
+/**
+ * Router for handling friend-related operations.
+ */
+const friendsRouter = router({
```
alexespejo, line 47: nit: add docstrings for the procedures

apps/antalmanac/src/backend/routers/friends.ts
```diff
@@ -0,0 +1,334 @@
...
+import { TRPCError } from '@trpc/server';
+import { and, eq, gt, or } from 'drizzle-orm';
+import { z } from 'zod';
+
+import { procedure, router } from '../trpc';
+
+/**
+ * Resolves a session token to an internal userId, verifying the session is valid and not expired.
+ */
+async function resolveSessionToUserId(sessionToken: string): Promise<string> {
```
alexespejo: Separate this method into `rds.ts`. Commonly in the code base we access the user/accounts using `getUserDataWithSession` or `getUserAndAccountBySessionToken`. You can either add this as a utility and rename as `getUserIdBySessionToken` or reuse the existing functions

apps/antalmanac/src/backend/routers/friends.ts
```diff
@@ -0,0 +1,334 @@
...
+const friendsRouter = router({
+    /**
+     * Send a friend request by recipient email.
+     */
+    sendFriendRequestByEmail: procedure
+        .input(z.object({ sessionToken: z.string(), email: z.string().email() }))
+        .mutation(async ({ input }) => {
+            const requesterId = await resolveSessionToUserId(input.sessionToken);
+
+            const [addressee] = await db.select().from(users).where(eq(users.email, input.email)).limit(1);
```
alexespejo: issue: `getUserByEmail` exists in `rds.ts`

apps/antalmanac/src/backend/routers/friends.ts
```diff
@@ -0,0 +1,334 @@
...
+            }
+
+            if (addressee.id === requesterId) {
+                throw new TRPCError({
+                    code: 'BAD_REQUEST',
+                    message: 'You cannot friend yourself.',
+                });
+            }
+
+            const [existing] = await db
```
alexespejo: issue: for separation of concerns we prefer to abstract drizzle code into the `RDS` class in `rds.ts`. We break this pattern in the code base but we should try our best to uphold the pattern

apps/antalmanac/src/backend/routers/friends.ts
```diff
@@ -0,0 +1,334 @@
...
+                const theyRequestedYou = existing.requesterId === addressee.id && existing.addresseeId === requesterId;
+                throw new TRPCError({
+                    code: 'BAD_REQUEST',
+                    message: theyRequestedYou
+                        ? 'This user has already sent you a friend request. Check your Requests tab to accept.'
+                        : 'A friend request is already pending with this user.',
+                });
+            }
+
+            return await db
```
alexespejo: abstract the ORM code

apps/antalmanac/src/backend/routers/friends.ts
```diff
@@ -0,0 +1,334 @@
...
+import { and, eq, gt, or } from 'drizzle-orm';
+import { z } from 'zod';
+
+import { procedure, router } from '../trpc';
+
+/**
+ * Resolves a session token to an internal userId, verifying the session is valid and not expired.
+ */
+async function resolveSessionToUserId(sessionToken: string): Promise<string> {
+    const [session] = await db
```
alexespejo: in rds.ts we have `getCurrentSession`

apps/antalmanac/src/backend/routers/friends.ts
```diff
@@ -0,0 +1,334 @@
...
+            const requesterId = await resolveSessionToUserId(input.sessionToken);
+
+            if (requesterId === input.addresseeId) {
+                throw new TRPCError({
+                    code: 'BAD_REQUEST',
+                    message: 'You cannot friend yourself.',
+                });
+            }
+
+            const [existing] = await db
```
alexespejo: issue: make reusable (if the code block above does the same thing)

apps/antalmanac/src/backend/routers/friends.ts
```diff
@@ -0,0 +1,334 @@
...
+
+    /**
+     * Accept a friend request.
+     */
+    acceptFriendRequest: procedure
+        .input(z.object({ sessionToken: z.string(), requesterId: z.string() }))
+        .mutation(async ({ input }) => {
+            const addresseeId = await resolveSessionToUserId(input.sessionToken);
+
+            return await db
```
alexespejo: abstract

apps/antalmanac/src/backend/routers/friends.ts
```diff
@@ -0,0 +1,334 @@
...
+                        eq(friendships.status, 'PENDING')
+                    )
+                )
+                .returning();
+        }),
+
+    /**
+     * Get all friends for a user as user objects.
+     */
+    getFriends: procedure.input(z.object({ userId: z.string() })).query(async ({ input }) => {
```
alexespejo: :p create functions for `getFriendshipsSent` and `getFriendshipsRecieved` or whatever seems fitting

apps/antalmanac/src/backend/routers/friends.ts
alexespejo, line 1: Move drizzle code to RDS. I was gonna comment on all of them but I got lazy

alexespejo (changes requested):

brelieu05:
@cubic-dev-ai review

brelieu05:
@cubic-dev-ai review

brelieu05:
@cubic-dev-ai review

brelieu05:
@cubic-dev-ai review

apps/antalmanac/src/hooks/useIsSharedSchedulePage.tsx
alexespejo, line 1: I think it's redundant to have both of `useIsReadonlyView` and `useIsSharedSchedulePage`

apps/antalmanac/src/components/ColorPicker.tsx
```diff
@@ -83,10 +85,11 @@ const ColorPicker = memo(function ColorPicker({
         <>
             <Tooltip title="Change Color">
                 <IconButton
-                    sx={{ color: currColor, padding: 0.5 }}
+                    sx={{ color: currColor, padding: '8px' }}
```
alexespejo: magic number

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody/SectionTableBodyCells/action-cell/DeleteAndNotifications.tsx
```diff
@@ -0,0 +1,69 @@
...
+    courseDetails: CourseDetails;
+}
+
+/**
+ * Sections added to a schedule, can be recolored or deleted.
+ */
+export const DeleteAndNotifications = memo(({ ...props }: DeleteAndNotificationsProps) => {
+    const initialized = useNotificationStore(useShallow((state) => state.initialized));
+    const theme = useTheme();
+    const isMobile = useMediaQuery(theme.breakpoints.down('sm'));
```
alexespejo: Use the `useMobile` hooke

brelieu05:
@cubic-dev-ai review

apps/antalmanac/src/backend/routers/friends.ts
```diff
@@ -0,0 +1,183 @@
+import { db } from '@packages/db/src';
+import { TRPCError } from '@trpc/server';
+import { z } from 'zod';
+
+import { RDS } from '../lib/rds';
```
alexespejo: absolute import

apps/antalmanac/src/backend/routers/friends.ts
```diff
@@ -0,0 +1,183 @@
...
+import { protectedProcedure, router } from '../trpc';
```
alexespejo: absolute import

apps/antalmanac/src/components/Header/Friends/BlockedUserCard.tsx
```diff
@@ -0,0 +1,37 @@
+import { Box, Button } from '@mui/material';
+
+import { FriendIdentity } from './FriendIdentity';
```
alexespejo: change for all relative imports. Apply this to the rest of the files that import from the `Header/Friends` directory 

```suggestion
import { FriendCard } from '$components/Header/Friends/FriendCard';```

File: apps/antalmanac/src/components/SharedScheduleBanner.tsx [hunk 18]
```diff
@@ -0,0 +1,473 @@
...
+                            <span>
+                                <IconButton
+                                    aria-label="Add Friend"
+                                    onClick={handleAddFriend}
+                                    disabled={addFriendSent}
+                                    size={isMobileScreen ? 'small' : 'medium'}
+                                    sx={{
+                                        bgcolor: 'primary.main',
+                                        color: 'white',
+                                        '&:hover': { bgcolor: 'primary.dark' },
```
alexespejo, line 401: We should use our AA blue
Image

sicn4rf:
question: after viewing a friend's schedule, when going back to your schedule, the selected schedule prior to viewing a friends schedule doesn't persist. Is this intentional?

e.x: i have schedules titled: Spring1, Spring2, and Spring3. I'm currently looking at Spring3. i click "Friends" and "View" 
Caleb Wong's schedule. then I exit out of this view, and return to my own schedule, and it puts me in "Spring1".

apps/antalmanac/src/components/SharedScheduleBanner.tsx
```diff
@@ -0,0 +1,473 @@
...
+            setFetchedFriendName(null);
+            setAddFriendSent(false);
+            setIsAlreadyFriend(true);
+        }
+
+        if (hasAttemptedLoadRef.current) {
+            return;
+        }
+
+        const loadContent = async () => {
```
alexespejo: nit: name to something more explicit `loadFriendsData`

apps/antalmanac/src/components/HomeLayout.tsx
alexespejo, line 1: If this is a complete duplicate of the code is on `Home.tsx` you should use it in `Home` and rename the component to like `HomeLayout` or something like that

apps/antalmanac/src/components/SharedScheduleBanner.tsx
```diff
@@ -0,0 +1,473 @@
...
+    const loadFriendSchedules = useCallback(
+        async (userId: string) => {
+            try {
+                beginLoadingSchedule();
+                removeLocalStorageTempSaveData();
+
+                const userData = await trpc.userData.getFriendUserData.query({ userId });
+
+                if (!userData?.userData?.schedules?.length) {
+                    setWarning("This friend hasn't shared any schedules with you.");
```
alexespejo: I think having a full warning page has too much friction I think a warning toast would be sufficient 

Additionally, the "Go Back" button doesn't seem to work

alexespejo (changes requested):
Other stuff 

We should disable the search in the calendar item and don't forget about the short cut. Note: don't grey it out maybe just display it as text without the icon 
Image

On mobile the page defaults to the Added tab. I think it's a little awkward at least when viewing friends it would make more sense to start on the calendar tab

Image

Future Issues: there are plenty of features I think we could explore with this feature we can push off to post release: 
- Autocomplete search (as seen in the original design)
- Nicer Friends view: I think it feels weird that we just disable every other feature. In the future I could see us hiding and showing features specific to the friends view. I believe this was once mentioned in a meeting

apps/antalmanac/src/components/SharedScheduleBanner.tsx
```diff
@@ -0,0 +1,473 @@
...
+        scheduleId,
+        friendName,
+        sessionIsValid,
+        navigate,
+        setOpenLoadingSchedule,
+        beginLoadingSchedule,
+        postHog,
+    ]);
+
+    const handleGoHome = useCallback(async () => {
```
alexespejo: issue: this is duplicate of `handleExitSharedSchedule`

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -425,5 +485,23 @@ export function AddedCoursePane() {
...
+        AppStore.on('addedCoursesChange', updateCourseCounts);
+        AppStore.on('skeletonScheduleChange', updateCourseCounts);
+        return () => {
+            AppStore.off('addedCoursesChange', updateCourseCounts);
+            AppStore.off('skeletonScheduleChange', updateCourseCounts);
+        };
+    }, []);
+
+    const shouldShowAddedSectionsGrid =
+        !skeletonMode || (skeletonMode && courseCounts.hasRegular && !courseCounts.hasSkeleton);
```
alexespejo: issue: dupe !skeletonMode

brelieu05:
> question: after viewing a friend's schedule, when going back to your schedule, the selected schedule prior to viewing a friends schedule doesn't persist. Is this intentional?
> 
> e.x: i have schedules titled: Spring1, Spring2, and Spring3. I'm currently looking at Spring3. i click "Friends" and "View" Caleb Wong's schedule. then I exit out of this view, and return to my own schedule, and it puts me in "Spring1".

@sicn4rf I made this intentional because restoring the previous schedule was causing confusion, it’s easy to lose track of which one you were on after viewing a friend’s schedule. Resetting to a “fresh state” felt clearer and more predictable imo.

KevinWu098:
4k lines

packages/db/src/schema/friendship.ts
```diff
@@ -0,0 +1,56 @@
+import { pgTable, text, timestamp, pgEnum, primaryKey, index, check } from 'drizzle-orm/pg-core';
+import { sql } from 'drizzle-orm';
+
+import { users } from './auth/user';
```
KevinWu098, line 4: nit: give me absolute imports or give me death

apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -58,17 +58,14 @@ export class Schedules {
...
         const scheduleNames = this.getScheduleNames();
         scheduleNames.splice(scheduleIndex, 1);
-        let nextScheduleName = newScheduleName;
-        let counter = 1;
-
-        while (scheduleNames.includes(nextScheduleName)) {
-            nextScheduleName = `${newScheduleName}(${counter++})`;
-        }
-        return nextScheduleName;
+        return getNextScheduleName(newScheduleName, new Set(scheduleNames));
```
KevinWu098, line 61: issue: this is confusing on initial read and frankly I'm surprised this isn't illegal in TS, having a method be the same name as an import. 

suggestion: can we rename one or the other `getNextScheduleName`?

apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -58,17 +58,14 @@ export class Schedules {
...
     }
 
-    getDefaultScheduleName() {
-        return getDefaultTerm().shortName.replaceAll(' ', '-');
+    /**
+     * Get the backend schedule ID for a schedule, if available.
+     */
+    getScheduleId(scheduleIndex: number) {
+        return this.schedules[scheduleIndex]?.id;
```
KevinWu098, line 68: suggestion: We should add a tsconfig rule which marks unchecked index access as possibly undefined, but in the meanwhile, `this.schedules.at(scheduleIndex)?.id` should type it appropirately

apps/antalmanac/src/stores/AppStore.ts
```diff
@@ -68,8 +68,23 @@ class AppStore extends EventEmitter {
...
-    getDefaultScheduleName() {
-        return this.schedule.getDefaultScheduleName();
+    getSchedules() {
+        return this.schedule.getSchedules();
+    }
+
+    setSchedules(scheduleIds: (string | undefined)[] | undefined) {
+        if (Array.isArray(scheduleIds)) {
+            const schedules = this.getSchedules();
+            const numSchedules = Math.min(schedules.length, scheduleIds.length);
```
KevinWu098: question: explain why we compare the provided schedule ids with `getSchedules`?

apps/antalmanac/src/stores/AppStore.ts
```diff
@@ -68,8 +68,23 @@ class AppStore extends EventEmitter {
...
+            for (let i = 0; i < numSchedules; i++) {
+                schedules[i].id = scheduleIds[i];
+            }
```
KevinWu098: question: so we don't check if a scheduleId is undefined before assigning it to one of own schedules? is this safe?

KevinWu098 (changes requested):
initial small batch of comments

apps/antalmanac/src/stores/scheduleHelpers.ts
```diff
@@ -116,6 +119,31 @@ export function getColorForNewSection(newSection: ScheduleCourse, sectionsInSche
...
+        scheduleName: getDefaultScheduleName(),
+        courses: [],
+        customEvents: [],
+        scheduleNoteId: scheduleNoteId,
+        scheduleId: createId(),
+        id: undefined,
+    };
+}
+
+export function createEmptyShortCourseSchedule(): ShortCourseSchedule {
```
KevinWu098, line 137: question: this is something i could ask claude, but im interested in understanding the use case in this pr.

why do we need to handle empty schedules AND empty short course schedules? whats the difference? why is scheduleNote in one, but the other is just schedule note id?

apps/antalmanac/src/hooks/useIsReadonlyView.tsx
```diff
@@ -0,0 +1,9 @@
+import { useLocation } from 'react-router-dom';
+
+/**
+ * @returns true if the user should not be able to edit/delete anything, false otherwise.
+ */
+export function useIsReadonlyView() {
```
KevinWu098, line 6: thought: this... feels chudded... but i guess I can't think of a reason not to do this... 

🤷‍♂️

apps/antalmanac/src/components/ScheduleManagement/ScheduleManagementTabs.tsx
```diff
@@ -64,6 +64,7 @@ const scheduleManagementTabs: Array<ScheduleManagementTabInfo> = [
 
 export function ScheduleManagementTabs() {
     const { activeTab } = useTabStore();
+    const visibleActiveTab = useIsMobile() || activeTab !== 0 ? activeTab : false;
```
KevinWu098: what? false? what's going on here

apps/antalmanac/src/components/ColorPicker.tsx
```diff
@@ -82,10 +85,11 @@ const ColorPicker = memo(function ColorPicker({
         <>
             <Tooltip title="Change Color">
                 <IconButton
-                    sx={{ color: currColor, padding: 0.5 }}
+                    sx={{ color: currColor, padding: 1 }}
```
KevinWu098: issue: nope don't mess with my color picker styling

apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
```diff
@@ -51,6 +52,8 @@ function SectionTable(props: SectionTableProps) {
 
     const [activeColumns] = useColumnStore((store) => [store.activeColumns]);
     const [activeTab] = useTabStore((store) => [store.activeTab]);
+    const location = useLocation();
+    const sharedSchedulePage = location.pathname.startsWith('/share/');
```
KevinWu098: issue: shouldn't this use that hook or whatever?

KevinWu098 (changes requested):
various additional things

apps/antalmanac/src/components/dialogs/AddSchedule.tsx
```diff
@@ -4,13 +4,14 @@ import { useState, useEffect, useCallback } from 'react';
...
+import { getDefaultScheduleName } from '$stores/scheduleHelpers';
 
 /**
  * Dialog with a text field to add a schedule.
  */
 function AddScheduleDialog({ onClose, onKeyDown, ...props }: DialogProps) {
     const [name, setName] = useState(
-        AppStore.getNextScheduleName(AppStore.getScheduleNames().length, AppStore.getDefaultScheduleName())
+        AppStore.getNextScheduleName(AppStore.getScheduleNames().length, getDefaultScheduleName())
     );
```
KevinWu098: nit: lazy initializer

```suggestion
    const [name, setName] = useState(() =>
        AppStore.getNextScheduleName(AppStore.getScheduleNames().length, getDefaultScheduleName())
    );
```

apps/antalmanac/src/components/dialogs/SignInDialog.tsx
```diff
@@ -1,26 +1,36 @@
...
-    
+
+    const handleSignIn = async () => {
+        setIsLoggingIn(true);
+        try {
+            await loginUser(postHog);
+        } catch {
+            setIsLoggingIn(false);
+        }
+    };
```
KevinWu098, line 32: loggingIn never gets set to false if it throws? i presume this isn't an issue since we redirect for auth, but for correctness this should be fixed

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody/SectionTableBodyCells/action-cell/DeleteAndNotifications.tsx
```diff
@@ -0,0 +1,68 @@
...
+    term: Term['shortName'];
+    lastUpdated: string;
+    lastCodes: string;
+    courseDetails: CourseDetails;
+}
+
+/**
+ * Sections added to a schedule, can be recolored or deleted.
+ */
+export const DeleteAndNotifications = memo(({ ...props }: DeleteAndNotificationsProps) => {
```
KevinWu098: I believe this component / file is no longer used (and should be deleted)

apps/antalmanac/src/components/HomeLayout.tsx
```diff
@@ -0,0 +1,38 @@
...
+interface Props {
+    children: ReactNode;
+    /** When true, the header is hidden (e.g. shared schedule error state). */
+    hideHeader?: boolean;
+}
+
+/**
+ * Wraps components common to all pages that should look and feel like the home page.
+ *
+ * @param chilren The body of the page, usually the desktop/home components
```
KevinWu098: nit: `children`

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -90,27 +91,35 @@ function getCourses() {
 }
 
 function CustomEventsBox() {
```
KevinWu098, line 93: thought: not your fault, but this file is so chudded and should be at least 4 or 5 files

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -231,25 +256,51 @@ function ScheduleNoteBox() {
 }
 
 function SkeletonSchedule() {
-    const [skeletonSchedule, setSkeletonSchedule] = useState(AppStore.getCurrentSkeletonSchedule());
+    const getScheduleData = (): ShortCourseSchedule => {
+        const skeletonSchedule = AppStore.getCurrentSkeletonSchedule();
+        if (!skeletonSchedule.courses || skeletonSchedule.courses.length === 0) {
```
KevinWu098, line 262: nit: i won't ask you to change it, but I will speak aloud that a lot of the functions in the PR would be eaiser to read if the condition was inverted

if (!regularCourses.length) { return skeletonSchedule }

why this is helpful, even though it's the same logic, is that when a reader looks at the code, they no longer have to maintain in their head that there is another codepath

in the current code, i have to think "okay, so when we have courses, we do X Y and Z AND also i need to remember that there is stuff to be done when we do not have courses, which I do not know yet, so I must keep in memory more things in case the other code path is significant".

now of course, this is just my preference and experience, but I think it's worth transmitting!

KevinWu098:
Screenshot 2026-04-29 at 7 26 36 PM

The empty state on friends schedules should probably be amended to not be in the 1st person. Also, i have access to the import dialog panel here

Screenshot 2026-04-29 at 7 26 48 PM

apps/antalmanac/src/components/ScheduleManagement/ScheduleManagementTab.tsx
```diff
@@ -12,34 +13,88 @@ interface ScheduleManagementTabProps {
...
+                onClick={handleClick}
+                value={value}
+                disabled={isSearchTab}
+            />
+        );
+
+        if (isSearchTab) {
+            return (
+                <Tooltip title="Add this schedule to your account before adding classes">
+                    <span
```
KevinWu098, line 65: issue: i know MUI tooltips are tricky, but this is kinda ugly :(

not blocking tho

apps/antalmanac/src/stores/AppStore.ts
brelieu05: I realized this was my code before the introduction of persistent schedules. It was originally used to get the schedule-ids from the backend, so the user can make the schedule link. pretty much leftover dead code from beforehand. Going to remove it.

apps/antalmanac/src/components/ScheduleManagement/ScheduleManagementTab.tsx
brelieu05: will refer this to the design team next meeting, because i got no idea how to make this tooltip look better

apps/antalmanac/src/stores/scheduleHelpers.ts
brelieu05: @Choollol I think this is during your implementation of shared schedules, so chime in if I'm (Claude) is wrong/incomplete.

Claude explanation
> Schedule is the full runtime representation with complete course data, while ShortCourseSchedule is the compact serialized format saved to localStorage/backend with only color, term, and sectionCode per course. They sit on either side of the serialization boundary, which is why scheduleNoteId is a reference into a separate notes array on Schedule, but ShortCourseSchedule just embeds the note string directly.

packages/db/src/schema/friendship.ts
brelieu05: image
image


I'm getting some unhappy typescript, errors when making this into an absolute import. And all our import for users are relative. And I lokes feel unsure about the changes claude is doing for the tsconfig stuff to make this absolute import happen. I can make a new issue for this because it spans a couple of files

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
brelieu05: will make a new issue for this

brelieu05:
> Screenshot 2026-04-29 at 7 26 36 PM
> The empty state on friends schedules should probably be amended to not be in the 1st person. Also, i have access to the import dialog panel here

Change I made:
image

brelieu05:
todo: i will fix the deployment db. db migrate command wasn't working a couple weeks ago, so I ran the sql commands manually into the scheduler, but now i gotta undo said commands and redeploy

KevinWu098:
> todo: i will fix the deployment db. db migrate command wasn't working a couple weeks ago, so I ran the sql commands manually into the scheduler, but now i gotta undo said commands and redeploy

yeah the problem is annoying, but its actually very easy to fix — drizzle doesn't generate idiomatic migrations by default, but cursor will clean it up for you pretty easily

```
DO $$ BEGIN
 CREATE TYPE "public"."friendship_status" AS ENUM('PENDING', 'ACCEPTED', 'DECLINED', 'BLOCKED');
EXCEPTION
 WHEN duplicate_object THEN null;
END $$;--> statement-breakpoint
CREATE TABLE IF NOT EXISTS "friendships" (
	"requester_id" text NOT NULL,
	"addressee_id" text NOT NULL,
	"status" "friendship_status" DEFAULT 'PENDING' NOT NULL,
	"created_at" timestamp with time zone DEFAULT now() NOT NULL,
	"updated_at" timestamp with time zone DEFAULT now() NOT NULL,
	CONSTRAINT "friendships_requester_id_addressee_id_pk" PRIMARY KEY("requester_id","addressee_id"),
	CONSTRAINT "friendships_no_self_friend" CHECK ("friendships"."requester_id" <> "friendships"."addressee_id")
);
--> statement-breakpoint
ALTER TABLE "schedules" ADD COLUMN IF NOT EXISTS "shared_with_friends" boolean DEFAULT true NOT NULL;--> statement-breakpoint
```

KevinWu098:
> todo: i will fix the deployment db. db migrate command wasn't working a couple weeks ago, so I ran the sql commands manually into the scheduler, but now i gotta undo said commands and redeploy

i would very very strongly recommend against executing commands against the DB directly, as that can really fuck with drizzle's state and the rest of the DB (in this case its staging so its whatever, but its a good rule of thumb to be really judicious in your usage)

apps/antalmanac/src/stores/scheduleHelpers.ts
Choollol: That sounds about right

From what I remember and looking at the code, handling empty schedules is for the schedule store and empty short course schedules is for `AppStore`'s state. I think I extracted them both to a separate file to have them live near each other, so when/if people change one, they will know to change the other.

packages/db/src/schema/friendship.ts
KevinWu098: okay yeah that works for me

apps/antalmanac/src/stores/scheduleHelpers.ts
KevinWu098: mm... okay... seems suspect but i guess i can't identify anything incorrect (?)

apps/antalmanac/src/backend/lib/rds.ts
```diff
@@ -409,6 +482,73 @@ export class RDS {
...
+                avatar: user.avatar ?? undefined,
+                userData: {
+                    schedules: userSchedules,
+                    scheduleIndex: 0,
+                },
+            };
+        });
+    }
+
+    private static async getUserAndAccount(
```
KevinWu098: issue: this method is not used anywhere

apps/antalmanac/src/backend/routers/userData.ts
```diff
@@ -275,6 +288,41 @@ const userDataRouter = router({
...
+        }
+        return schedule;
+    }),
+
+    /**
+     * Gets the schedule ID for a schedule by userId and schedule name.
+     * @param input - An object containing the userId and scheduleName.
+     * @returns The schedule ID, or throws NOT_FOUND if not found.
+     */
+    getScheduleIdByName: procedure
```
KevinWu098: issue: this method is unused. im also suspect of fetching a schedule by anything other than its id (although we semi-enforce unique schedule names)

apps/antalmanac/src/backend/lib/rds.ts
```diff
@@ -360,6 +367,72 @@ export class RDS {
...
 
+    /**
+     * Gets the schedule ID for a schedule by userId and schedule name.
+     *
+     * @param db - The database or transaction object to use for the query.
+     * @param userId - The ID of the user who owns the schedule.
+     * @param scheduleName - The name of the schedule.
+     * @returns A promise that resolves to the schedule ID, or null if not found.
+     */
+    static async getScheduleIdByName(
```
KevinWu098: issue: this method's only caller is unused. we should get rid of this

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -256,6 +259,54 @@ export const importScheduleWithUsername = async (username: string, postHog?: Pos
...
+
+    if (!isScheduleLoaded) {
+        scheduleComponentsToggleStore.setState({ openLoadingSchedule: false });
+        return { imported: false, error: new Error('Failed to load shared schedule') };
+    }
+
+    const session = useSessionStore.getState();
+    if (session.sessionIsValid) {
+        try {
+            await autoSaveSchedule({});
```
KevinWu098, line 291: issue: why are we calling auto save schedule

apps/antalmanac/src/backend/routers/friends.ts
```diff
@@ -0,0 +1,182 @@
...
+        });
+    }
+
+    const pendingRow = rows.find((r) => r.status === 'PENDING');
+    if (pendingRow) {
+        const theyRequestedYou = pendingRow.requesterId === addresseeId && pendingRow.addresseeId === requesterId;
+        throw new TRPCError({
+            code: 'BAD_REQUEST',
+            message: theyRequestedYou
+                ? 'This user has already sent you a friend request. Check your Requests tab to accept.'
```
KevinWu098, line 32: thought (non-blocking): we could probably just "accept" the friend req at this point and success it

apps/antalmanac/src/backend/routers/userData.ts
```diff
@@ -275,6 +288,41 @@ const userDataRouter = router({
         return await RDS.flagImportedUser(db, input.username);
     }),
 
+    /**
+     * Retrieves a shared schedule by schedule ID.
+     * All schedules are publicly accessible via their ID.
+     * @param input - An object containing the schedule ID.
+     * @returns The schedule data associated with the schedule ID, or throws NOT_FOUND if not found.
+     */
+    getSharedSchedule: procedure.input(z.object({ scheduleId: z.string() })).query(async ({ input }) => {
```
KevinWu098, line 290: issue: is this method not redundant? is there really no method for getting schedule by id on another router?

if it is not redundant, seems like this should... probably live on another router (?) with validation that the user is actually allowed to fetch this schedule?

KevinWu098 (changes requested):
Some more minor stuff. given that is past midnight already, im opting not to review the Friends- components. if we'd like to merge, im happy to approve in the morning. but otherwise, we can take an extra day to resolve this and maybe take a pass on styling

KevinWu098:
@cubic

PR #1512, opened 2026-02-19
refactor: migrate new quarter check logic from bash to a TypeScript script

sicn4rf:
## Summary
Refactors the CI/CD quarter check automation by implementing an  `update-terms.ts` script that gets called by the `check_new_quarter.yml` workflow. This is an upgrade over the brittle bash and jq scripts written in a string-only, non-typed environment previously implemented in the workflow file itself.

Fixed the reason field to capture both change reasons when the term and count change simultaneously.

Documented permissions given to the workflow that allow it to push directly to the protected main branch.

## Test Plan
- Verified TypeScript compiles with no type errors
- Script can be run locally with pnpm --filter antalmanac update-terms
- Workflow validated against GitHub Actions schema

## Issues
No issues. This is a refactoring of existing functionality.

Closes #
N/A


.github/workflows/check_new_quarter.yml
```diff
@@ -18,74 +18,43 @@ jobs:
...
+            - name: Setup Node and pnpm
+              uses: ./.github/actions/setup-node-and-pnpm
 
-                  LATEST_DEPLOYED=$(jq -r '.data[0].longName // empty' .github/deployed_terms.json)
+            - name: Install dependencies
+              run: pnpm install --frozen-lockfile
 
-                  echo "Latest term from API: $LATEST_TERM"
-                  echo "Latest deployed term: $LATEST_DEPLOYED"
+            - name: Run Update Script
```
dsnsgithub: nit: consistent capitalization

apps/antalmanac/.gitignore
```diff
@@ -1,4 +1,5 @@
 src/generated/
+!src/generated/deployed_terms.json
```
dsnsgithub: Counterintuitively, this won't work, at least on the version of Git I am using, since `src/generated/` will end up skipping it. Replacing the first line with `src/generated/*` should fix that.

apps/antalmanac/scripts/update-terms.ts
```diff
@@ -0,0 +1,112 @@
+import { writeFileSync, readFileSync, existsSync, mkdirSync } from 'node:fs';
+import { dirname, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+import { WebsocTerm } from '@packages/antalmanac-types';
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+
```
dsnsgithub, line 9: Running `pnpm --filter antalmanac update-terms` locally will fail because this file is missing dotenv.

```ts
import 'dotenv/config';
```

apps/antalmanac/scripts/update-terms.ts
```diff
@@ -0,0 +1,112 @@
...
+const COURSE_COUNT_CHANGE_THRESHOLD = 10;
+
+interface DeployedTermsData {
+    latestTerm: string;
+    courseCount: number;
+    updatedAt?: string;
+    reason?: string;
+}
+
+async function getCourseCount(headers: Record<string, string>): Promise<number> {
```
dsnsgithub: nit: type is already implied

```suggestion
async function getCourseCount(headers: Record<string, string>) {
```

.github/workflows/check_new_quarter.yml
dsnsgithub:
```suggestion
            - name: Update deployed terms
```

nit: consistent capitalization, more descriptive name

dsnsgithub (changes requested):

sicn4rf:
https://github.com/sicn4rf/ci-ts/actions
tested it here ^

addressed all the other comments too + cubic. i think it should be good to go 

@dsnsgithub

dsnsgithub:
Looks good to me, just need to fix the merge conflicts and update your test repo with the current workflow that is in this PR.

sicn4rf:
used `git status --porcelain` in test repo and it worked. should be good now!

sicn4rf:
@cubic

sicn4rf:
@dsnsgithub CASH ME OUT

dsnsgithub (approved):
LGTM!

PR #1522, opened 2026-02-23
fix: switch to Oxlint and Oxfmt

dsnsgithub:
## Summary

>[!NOTE]
> This PR is part of a [stack](https://www.stacking.dev/).
> 
> (3) [chore: format](https://github.com/icssc/AntAlmanac/pull/1524)
> (2) [fix: switch to Oxlint and Oxfmt](https://github.com/icssc/AntAlmanac/pull/1522)
> (1) `main`

Switch from ESLint/Prettier to [Oxlint](https://oxc.rs/docs/guide/usage/linter.html)/[Oxfmt](https://oxc.rs/docs/guide/usage/formatter.html). Both are written in [Rust](https://oxc.rs/) and lead to **significant** performance improvements.

As a side effect, the outdated ESLint configuration warnings are now fixed.
Linting and formatting are now global and now work across all packages within the AntAlmanac repo.

### Performance Comparison:

Keep in mind, Oxlint/Oxfmt runs linting and formatting across all packages, while ESLint/Prettier is only targeting `apps/antalmanac`.
| Task | Tooling | Execution Time (Total) | Speedup |
| :--- | :--- | :--- | :--- |
| **Linting** | ESLint | ~7.78s | — |
| **Linting** | **Oxlint** | **~0.61s** | **~12.7x Faster** |
| **Formatting** | Prettier | ~23.30s | — |
| **Formatting** | **Oxfmt** | **~0.96s** | **~24.3x Faster** |

## Test Plan
There should be no functional changes.

A separate PR can be used to run `pnpm format && pnpm lint --fix` and format/lint the entire repository (303 files changed).  

## Issues

Closes https://github.com/icssc/AntAlmanac/issues/1515 
Closes https://github.com/icssc/AntAlmanac/issues/1030

PR #1524, opened 2026-02-24
chore: format

dsnsgithub:
## Summary

>[!NOTE]
> This PR is part of a [stack](https://www.stacking.dev/).
> 
> (3) [chore: format](https://github.com/icssc/AntAlmanac/pull/1524)
> ~~(2) [fix: switch to Oxlint and Oxfmt](https://github.com/icssc/AntAlmanac/pull/1522)~~ (merged)
> (1) `main`

Formats all code in regard to new linting and formatting settings.

## Test Plan
Given the new configuration settings, the other files changes can be replicated by running `pnpm format && pnpm lint --fix --fix-suggestions`.


alexespejo:
TODO: complete regression testing

AureliaSindhu (approved):
LGTM!

AureliaSindhu (dismissed):

alexespejo:
:p

AureliaSindhu (dismissed):
lgtm

dsnsgithub:
Holding off on merging this PR until https://github.com/icssc/AntAlmanac/pull/1511 is merged.

AureliaSindhu (dismissed):
lmk once u fixed the merge conflict! have fun :)

.oxlintrc.json
```diff
@@ -1,8 +1,25 @@
 {
     "$schema": "./node_modules/oxlint/configuration_schema.json",
-    "plugins": null,
-    "categories": {},
-    "rules": {},
+    "plugins": ["eslint", "react", "unicorn", "oxc", "typescript"],
```
KevinWu098, line 3: What is Unicorn?

KevinWu098 (dismissed):

.oxlintrc.json
KevinWu098: And also, do we have that nice warning about type only imports that biome has? And for deps (this one I assume yes)?

dsnsgithub: > What is Unicorn?

Some rules regarding best practices, it is already enabled by default if you don't specify any plugins.

> And also, do we have that nice warning about type only imports that biome has? And for deps (this one I assume yes)?

I don't believe (?) so, at least not by default. I haven't tinkered with Biome but is [typescript/consistent-type-imports ](https://oxc.rs/docs/guide/usage/linter/rules/typescript/consistent-type-imports.html) what you are looking for?

This rule is considered a style rule. Right now `correctness` rules are considered errors and the other rules `suspicious/style/pedantic` are not being considered.

Enabling all of these minor rules would likely create a lot of work for little gain, but we could definitely cherry-pick a few good rules that we deem important. Linter warnings (rather than errors) are pretty much ignored during development, so we also have to consider whether we want to fail CI due to small style choices and whether Oxlint can automatically fix them.

KevinWu098: 1. consistent type imports, yes, that's the one. why would this one be more work? should just be an autofix.
2. exhaustive deps on useEffects should be enforced as an error. effect is already a nasty footgun, and being more explicit is good! i wouldn't expect this to create any significant amount of work, given that we previously had the rule under eslint

KevinWu098:
try again :)

PR #1529, opened 2026-02-25
Clean up ActionCell

### LINKED ISSUE #1514: Clean up ActionCell components
There are two `ActionCell` components and one of them isn't used at all currently. We should clean this component and any related/children components up, deleting any unused components and deduplicating large amounts of copied code.

Choollol:
## Summary

Delete old unused `ActionCell` component and clean up new `ActionCell`'s file organization.

## Test Plan

Ensure anything that uses `ActionCell`, like search results and the added courses pane, work the same as before/on prod.

## Issues

Closes #1514


KevinWu098:
Where are we on this?

sicn4rf:
> Where are we on this?

gonna try to have this reviewed by eod, my bad

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody/SectionTableBodyCells/action-cell/ActionCell.tsx
```diff
@@ -1,22 +1,15 @@
...
 import { memo } from 'react';
 
-import { addCourse, openSnackbar } from '$actions/AppStoreActions';
 import { TableBodyCellContainer } from '$components/RightPane/SectionTable/SectionTableBody/SectionTableBodyCells/TableBodyCellContainer';
 import { DeleteAndNotifications } from '$components/RightPane/SectionTable/SectionTableBody/SectionTableBodyCells/action-cell/DeleteAndNotifications';
-import { NotificationsMenu } from '$components/RightPane/SectionTable/SectionTableBody/SectionTableBodyCells/action-cell/NotificationsMenu';
-import { useIsMobile } from '$hooks/useIsMobile';
-import analyticsEnum, { logAnalytics } from '$lib/analytics/analytics';
-import AppStore from '$stores/AppStore';
+import { ScheduleAddCell } from '$components/RightPane/SectionTable/SectionTableBody/SectionTableBodyCells/action-cell/ScheduleAddCell';
```
sicn4rf, line 6: suggestion: i noticed a circular import between `ActionCell.tsx` and `ScheduleAddCell.tsx`. the code imports `ScheduleAddCell` from `ScheduleAddCell.tsx` into `ActionCell.tsx` but then `ScheduleAddCell.tsx` imports `ActionCellProps` from `ActionCell.tsx`. i'm honestly not entirely sure what this entails. it doesnt break at runtime but i'd assume it could make them harder to reuse independently because of the tight coupling. 

is there a way we could maybe define both the interface and component in `ScheduleAddCell.tsx` and then just import both into `ActionCell.tsx`, that way theres no bidirectional import. (feel free to not take this suggestion at all or correct me if im wrong!!!!!!!!!!!)

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody/SectionTableBodyCells/action-cell/ActionCell.tsx
sicn4rf: nit: potential ghost prop in `scheduleNames`

as far as i can tell, `scheduleNames` prop in the `ActionCellProps` interface appears to be a ghost prop.  it's accepted by the `ScheduleAddCell` component but then immediately aliased to `_scheduleNames` and never used. previously, the add to schedule button used to have the option to add one class to multiple schedules, hence the need for this list of schedule names. however, that feature in the ui has since been removed in this new version of this action cell, so this prop is no longer used. if that feature is gone for good, this may be worth the clean up since its potentially just code smell. 

or like maybe we could just keep it future use and i dont know what im talking about.

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody/SectionTableBodyCells/action-cell/ActionCell.tsx
```diff
@@ -42,81 +35,14 @@ interface ActionProps {
...
-
-export interface ActionCellProps extends Omit<ActionProps, 'classes'> {
     /**
      * Whether the section has been added.
      */
     addedCourse: boolean;
+
+    notifyOn: NotifyOn | undefined;
+    lastUpdated: string;
+    lastCodes: string;
```
sicn4rf, line 21: nit: missing comments / or extraneous comments

i would honestly argue the comments of these props are extraneous. a lot of these are self-explanatory and the comments arent adding any information that you can't already infer. i understand you just copied over the code from the old `ActionCell.tsx` tho so its like chill. 

ironically tho, these last three fields `notifyOn`, `lastUpdated`, and `lastCodes` are actually some that i personally think would benefit from comments. 

anyway, my suggestion would be to clean up the mild inconsistency and either remove all extraneous comments or  add them to the rest of the fields otherwise. 

i'll CC @alexespejo and @KevinWu098 here for a second opinion on the comments on these files.

sicn4rf (dismissed):
overall, this looks like a clean PR to me. i left a few nits and a suggestion inline but none of them are blocking. main thing worth discussing is prob whether `scheduleNames` is worth cleaning up since the `add to [schedule name]` dropdown feature is gone now. but that might be a follow-up conversation thats out of scope for this pr

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody/SectionTableBodyCells/action-cell/ActionCell.tsx
Choollol: Good point about the coupling! I think `ScheduleAddCell.tsx` can type its own props separately then, so it can be more reusable.

sicn4rf (approved):
awesome, this lgtm! thanks for addressing the nits and merge conflicts :) 👍 🚀

PR #1532, opened 2026-02-27
Clean up RDS 

### LINKED ISSUE #1158: Clean up RDS post auth

alexespejo:
## Summary
Containerizes various RDS functions by the tables they primarily interface with. Currently, this PR enforces the separation of concerns but retains the existing RDS calls (reference `rds.ts`). In the future we can deprecate the RDS class and call the functions independently. 

> [!WARNING]
> Large blast radius. Advise merging after registration 

## Test Plan
All changes should be benign. Everything should work the same as on prod 
## Issues

Closes #1158


alexespejo:
ts chopped will come back to this at some point

Issue #1530, opened 2026-02-26
Make AANTS Respect Environment


IsaacNguyen:
On staging, only show notifications created on that instance

Have to update primary key

PR #1536, opened 2026-02-27
fix: isolate AANTS subscriptions by environment

rob-9:
## Summary

Addresses #1530

Staging and prod AANTS instances share the same DB. Previously, subs created on staging would be visible in prod (and vice versa) and the tRPC backend would return/modify subs across environments.

This PR makes `environment` a required part of each subscription so each env is isolated.

## Changes

migration (`0009_handy_prima.sql`)
- backfill existing rows: sets `environment` to `'production'` where it was `NULL` or `''`
- make `environment` `NOT NULL` with a default of `'production'`
- expand primary key from `(userId, sectionCode, year, quarter)` → `(userId, sectionCode, year, quarter, environment)`

schema (`subscription.ts`)
- `environment` column: nullable → `notNull().default('production')`
- primary key now includes `environment`

backend (`rds.ts` + `notifications.ts`)
- all notification queries (`retrieve`, `upsert`, `updateAll`, `delete`, `deleteAll`) now filter by `environment`
- `getStage()` helper reads `process.env.STAGE`, defaulting to `'production'` (the old `set` handler was the only one that read `STAGE`, and it fell back to `''`)
- `upsert` conflict target updated to include `environment`, and `environment` removed from the `onConflictDoUpdate` set (since it's now part of the PK)

## Test plan
- [ ] run migration against a local db with existing subs & verify `NULL`/empty environments backfill to `'production'`
- [ ] create a sub on staging instance & verify it doesn't appear when querying with `environment = 'production'`
- [ ] upsert the same section on two different environments & verify both rows coexist (separate PK)
- [ ] delete a subscription & verify it only deletes the row matching the current environment


IsaacNguyen (dismissed):
So sorry for taking so long to get around to this, but great job on the PR! Everything looks good to me!

I'm not sure I'm the biggest fan of defaulting the stage to "production", but we kind of have to in this situation (I would've done the same).

I think we should hold off from merging this one until after Spring enrollment is completely done (maybe after Week 2).

IsaacNguyen:
Hello! Sorry for the delay in getting this merged (there are some merge conflicts now). But once you squash these conflicts, we'll get this merged in!

rob-9:
migration looks clean

IsaacNguyen (approved):
LGTM!

PR #1542, opened 2026-03-04
AANTS React Email

IsaacNguyen:
## Summary
The old AANTS template was chopped, so we're switching to React Email to make it look nicer.

image

image

image

image

- Able to view email template in browser via `pnpm run preview:email`
- Links directly to AntAlmanac through the banner and link at bottom
- Gmail no longer collapses notifications of same type (fixed by adding a random ID to each email to prevent "duplicate" text)
- Now specifies what was the previous status or restriction codes

Also noticed annoying text wrapping on mobile when it's WAITLISTED, so abbreviated on mobile

image

## Test Plan
- [x] Ensure React Email works with the lambda functions and SQS
- [x] Works for every status change (FULL -> OPEN, OPEN -> WAITLISTED, etc.)
- [x] Works for restriction code changes
- [x] Properly displays one of another
- [x] Handles NULL or empty restriction codes


KevinWu098:
Screenshot 2026-03-03 at 9 52 46 PM

---

image

apps/aants/src/emails/CourseNotificationEmail.styles.ts
KevinWu098, line 1: Not a super big fan of declaring a styles folder — we should continue to err on the side of using a style prop, if possible (is it possible?)

KevinWu098: if style reuse / shared components gets annoying, i'd rather us have a src/components folder in `aants/` rather than using a styles.ts

apps/aants/src/helpers/emailQueue.ts
```diff
@@ -7,11 +7,22 @@ const sqsClient = new SQSClient({});
...
     Destination: {
         ToAddresses: string[];
     };
+    /** Email content from React Email */
+    Content: {
+        Subject: string;
+        Html: string;
+        Text: string;
+    };
+    /** Context for logging */
```
KevinWu098, line 19: question: logging to where?

KevinWu098 (changes requested):
Looks dope, just some small questions

apps/aants/src/helpers/emailQueue.ts
IsaacNguyen: Image

There are logs that get uploaded to Cloudwatch about emails and classes are being updated as AANTS processes. We have to create a new LogContext body because we were initially passing in all our data to fill out the SES template, so it was easy to access things like deptCode and courseNumber. But with React Email, it restructures the way we access/pass in data so we need to explicitly pass in this data to log it.

apps/aants/src/emails/CourseNotificationEmail.styles.ts
IsaacNguyen: OKidoke

apps/aants/src/helpers/emailQueue.ts
IsaacNguyen: image

There are some logs in AANTS that get uploaded to Cloudwatch logs that just help keep track of which notifications are being sent and where. We just need to explicitly define `LogContext` now because React Email changed the way we pass around data.

apps/aants/src/components/WhatChangedBox.tsx
```diff
@@ -0,0 +1,102 @@
+import { Section, Text } from '@react-email/components';
+
+import { StatusPill } from './StatusPill';
+
+export interface StatusChange {
```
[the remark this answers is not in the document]
IsaacNguyen: I fixed this already

KevinWu098 (approved):
LGTM — maybe worth getting a UI/UX member's eyes on it, but no need to be blocked on this by feedback. we can improve incrementally

PR #1544, opened 2026-03-04
fix: Improve AAPI type safety

### LINKED ISSUE #1259: Improve type safety when interfacing with AAPI
As highlighted by the downtime over this past weekend, our routes which interact with an external API, namely the Anteater API are not particularly type safe. 

Currently, we're asserting the response from the API to be the "happy path" and not accounting for potential errors. When these errors do occur, we either fail to handle them or we do not propagate the appropriate error for both observability and user trust.

https://github.com/icssc/AntAlmanac/blob/008286dc79be7244994b9748b66d262b2ec78b91/apps/backend/src/routers/websoc.ts#L62-L74

e.g. `    return sortWebsocResponse(data.data as WebsocAPIResponse);`

A PR resolving this issue should:
1. Widen the type returned from the API to include error responses, handling these as appropriate
2. Properly propagate an error and display an appropriate error to the end user

calebwongg:
## Summary
Beforehand we were wrapping the api response into the `WebsocAPIResponse `which assumes that the API returns a `GET 200`.  This was causing crashes inside `sortWebsocResponse` whenever the API returned a non-200 response.

This PR adds error handling by wrapping the fetch call in a try/catch and checking the response for any HTTP errors. It also widens the response from the API by introducing WebsocAPIResult so that tpescript catches the errors at compile time instead of runtime
## Test Plan
- Tested network error by trying to query with a nonexistent domain instead of `anteaterapi.com`
- Tested bad data shape by setting `data = undefined` and confirming `TRPCError`
- Confirmed normal queries still work
## Issues

Closes #1259


calebwongg:
thanks cubic

packages/anteater-api-types/src/websoc.ts
```diff
@@ -3,9 +3,28 @@ import { paths } from './generated/anteater-api-types';
...
 
+type WebsocAPISuccess = {
+    ok: true;
+    data: WebsocAPIResponse;
+};
+
+type WebsocAPIError = {
+    ok: false;
+    message: string;
+};
```
sicn4rf: suggestion: `WebsocAPIError` and `WebsocAPISuccess` are duplicate types. you can derive these types from the generated spec instead of maintaining it manually. this also lets them stay in sync with AAPI if this endpoint ever changes what it returns. 

this is not blocking since the shapes are identical right now but its definitely worth considering.

`/packages/anteater-api-types/src/generated/anteater-api-types.ts` check lines 996 and 927 for the generated types, you already import `paths` so this shouldnt be difficult.

something like this:
```
export type WebsocAPIResponse =
    paths['/v2/rest/websoc']['get']['responses'][200]['content']['application/json']['data'];

type WebsocAPISuccess = paths['/v2/rest/websoc']['get']['responses'][200]['content']['application/json'];
type WebsocAPIError = paths['/v2/rest/websoc']['get']['responses'][500]['content']['application/json'];
export type WebsocAPIResult = WebsocAPISuccess | WebsocAPIError;
```

apps/antalmanac/src/backend/routers/websoc.ts
```diff
@@ -69,37 +70,79 @@ function sortWebsocResponse(response: WebsocAPIResponse) {
...
+    } catch (err) {
+        throw new TRPCError({
+            code: 'INTERNAL_SERVER_ERROR',
+            message: `Failed to reach the Anteater API: ${err}`,
+        });
+    }
+
+    if (!response.ok) {
+        throw new TRPCError({
+            code: response.status === 401 ? 'UNAUTHORIZED' : 'INTERNAL_SERVER_ERROR',
```
sicn4rf: nit: from my tests, 401 returns when the API key is missing or invalid. 

after further research, tRPC `UNAUTHORIZED` returns 401 to the BROWSER, when a client (user) makes a request to our tRPC endpoint without a valid session.

that is very different from AnteaterAPI returning a 401 to OUR SERVER because the server API key is wrong.

if anything, if the AnteaterAPI response is anything but 200, you should always just note it as `INTERNAL_SERVER_ERROR`, since `UNAUTHORIZED` can be misleading for callers. they may think that the user needs to log in.

sicn4rf (dismissed):
overall, this pr looks clean to me and resolves the issue as needed. the `fetchAnteaterAPI` abstraction noted by cubic looks clean. i also confirmed that the frontend already surfaces displaying these errors to the end user via the snackbar message.  

none of the nits and suggestions i noted are blocking so feel free to dismiss them if you think they're just bogging you down.

good job caleb! :) 👍

packages/anteater-api-types/src/websoc.ts
calebwongg: This type would probably be better for consistency but lowkey this only protects against an http 500. I could add another pathprotecting against a 422 and the anteater api only has responses for 500 and 422 but I'm not sure if there's a possibility for it to throw e.g. a 429 or something unexpected. 

So I think I'm going to keep this shape lowkey

calebwongg:
@cubic-dev-ai cash me out

calebwongg:
cubic

packages/anteater-api-types/src/websoc.ts
sicn4rf: thoughts?
```
  type WebsocAPISuccess = paths['/v2/rest/websoc']['get']['responses'][200]['content']['application/json'];
  type WebsocAPIError =
      | paths['/v2/rest/websoc']['get']['responses'][422]['content']['application/json']
      | paths['/v2/rest/websoc']['get']['responses'][500]['content']['application/json'];
  export type WebsocAPIResult = WebsocAPISuccess | WebsocAPIError;
```

apps/antalmanac/src/backend/routers/websoc.ts
```diff
@@ -69,37 +70,79 @@ function sortWebsocResponse(response: WebsocAPIResponse) {
...
+    } catch (err) {
+        throw new TRPCError({
+            code: 'INTERNAL_SERVER_ERROR',
+            message: `Failed to parse Anteater API response: ${err}`,
+        });
+    }
     console.log('queryWebSoc', data);
-    return sortWebsocResponse(data.data as WebsocAPIResponse);
+
+    if (!data?.ok || !data?.data) {
```
sicn4rf, line 114: nit: now that you added the try/catch, it always throws on failure. so im pretty sure this guarantees that data is always assigned and can't be undefined. `if (!data.ok || !data.data)` is fine. same at line 139

sicn4rf (dismissed):
im gonna approve this because it lgtm. i would consider taking a look at the nits i noted theyre 1 line fixes so I would say it's worth a look. 

now that the `fetchAnteaterAPI` abstraction has been implemented, we should reuse this in other files that face the exact same problem that websoc did. `course.ts`, `enrollHist.ts`, and `grades.ts` all follow this exact same pattern. a good solution to this issue is to move `fetchAnteaterAPI` to a shared location and these three files (or more if this logic is seen anywhere else) should be updated to use it. all of these files fetch from aapi. however, they do not check `response.ok`, the `.json()` is not in a try/catch, and the data is cast statically.

good job!

packages/anteater-api-types/src/websoc.ts
calebwongg: but like if it throws a 429 or something unexpected (im not too certain on the possibility of these but it could be a possibility), we won't catch it

sicn4rf: i may be misunderstanding here, but from what i infer from the implementation, the `fetchAnteaterAPI` abstraction you implemented already throws our any non 200 response. so, if i understand correctly,`WebsocAPIResult` is really only ever a 200 response.

the concern about my suggestion is that as of today, the handwritten type and the spec-generated type are identically the same and willl behave the same today. but the benefit of deriving from the generated openAPI spec is that if the API changes, they stay in sync automatically. that said, this is non-blocking just something to consider.

in `fetchAnteaterAPI`:
```
    if (!response.ok) {
        throw new TRPCError({
            code: 'INTERNAL_SERVER_ERROR',
            message: `Anteater API returned an error: ${response.status} ${response.statusText}`,
        });
    }
```
response is only ok for codes 200-299. so this function will throw an error if its not ok. and this gets called before ever typing data as a `WebsocAPIResult`, that means this is always going to be a 200 code is it not?

apps/antalmanac/src/backend/routers/websoc.ts
```diff
@@ -69,37 +70,79 @@ function sortWebsocResponse(response: WebsocAPIResponse) {
...
+    } catch (err) {
+        throw new TRPCError({
+            code: 'INTERNAL_SERVER_ERROR',
+            message: `Failed to parse Anteater API response: ${err}`,
+        });
+    }
     console.log('queryWebSoc', data);
-    return sortWebsocResponse(data.data as WebsocAPIResponse);
+
+    if (!data.ok || !data.data) {
```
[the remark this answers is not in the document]
sicn4rf: the try/catch guarantees that `data` is _something_. this is unecessary.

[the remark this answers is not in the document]
sicn4rf: i lied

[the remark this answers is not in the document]
sicn4rf: yea

sicn4rf (dismissed):
lgtm

packages/anteater-api-types/src/websoc.ts
calebwongg: yeah i think widening the api type might just be redundant ...

calebwongg:
lgtm

sicn4rf (changes requested):
blocked by james

try using trpc's `fetchRequestHandler` instead.

sicn4rf (approved):
nevermind looks lgtm to me bro

calebwongg:
> nevermind looks lgtm to me bro

yippee

PR #1548, opened 2026-03-05
Successful navigation after save

### LINKED ISSUE #1503: Planner Button Bug
## Description
After adding classes onto my AntAlmanc page and then clicking the planner button, I'm prompted that I have unsaved changes. Once I go ahead and save it and I get the confirmation toast that it's saved and I click on the planner button it won't let me switch to the planner page.  

## Steps to Recreate
1. Add a class to your schedule 
2. Click on the planner button
3. Once prompted that you have unsaved changes, go ahead and press the save button 
4. Click the planner button 

The behavior that's observed is that I'm not able to switch onto the page.

victorialee13:
## Summary
Adding a class and clicking the Planner button, the dialog for "Leave site? Changes you made may not be saved." would be prompted. Pressing cancel, then the save button for your schedule would successfully save your schedule. On the second attempt to switch to planner, we couldn't since the page reload never happens. 

Instead of using navigation, it completely leaves this page and load the new page (full-page nav). beforeunload would still run and unsavedChanges is set to false after save, then we direct to the planner.
## Test Plan
1. Add a class to your schedule
2. Click on the planner button
3. Once prompted that you have unsaved changes, go ahead and press the save button
4. Click the planner button
## Issues

Closes #1503

### DIFF
--- a/apps/antalmanac/src/components/Header/Header.tsx
+++ b/apps/antalmanac/src/components/Header/Header.tsx
@@ -28,7 +28,7 @@ import {
     removeLocalStorageDataCache,
     removeLocalStorageImportedUser,
 } from '$lib/localStorage';
-import { BLUE } from '$src/globals';
+import { BLUE, PLANNER_LINK } from '$src/globals';
 import { useIsMobile } from '$src/hooks/useIsMobile';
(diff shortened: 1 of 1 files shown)


KevinWu098:
Closing as succeeded by #1551

PR #1554, opened 2026-03-07
Reorderable added courses

### LINKED ISSUE #903: Reorder classes in added courses pane
Allow the user to reorder classes in added courses pane. I think this would be a nice feature, but its value-for-time can be debated.

Choollol:
## Summary

Allows reordering of classes within the added courses pane.

## Test Plan

1. Ensure styling is the same for added courses pane and other places that use affected components, like search results
2. Ensure drag and drop works
3. Ensure order of courses is saved after reordering

## Issues

Closes #903


apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
```diff
@@ -110,6 +118,12 @@ function SectionTable(props: SectionTableProps) {
                     marginTop: '4px',
                 }}
             >
+                {sortable ? (
```
alexespejo, line 136: I think it would make sense to have the "grabbing hand" cursor on hover

alexespejo (changes requested):
Sorry for the delay I'm surveying some feedback – and this PR lwky keeps getting lost in my mind when I scroll up and down the PR board. I'm thinking that after we do #1643 and #1630 there will be new space to reorder some of the buttons and change up the design. Lmk what you think. Otherwise feature works great

Choollol:
> Sorry for the delay I'm surveying some feedback – and this PR lwky keeps getting lost in my mind when I scroll up and down the PR board. I'm thinking that after we do #1643 and #1630 there will be new space to reorder some of the buttons and change up the design. Lmk what you think. Otherwise feature works great

I think holding off until after the redesigns makes sense!

apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
Choollol: Getting the "grab" cursor is easy, but making it "grabbing" when dragging is harder. Do you think it would be worth it to get the "grabbing" cursor too, or leave it at just "grab"?

Choollol:
@cubic-dev-ai review

apps/antalmanac/src/backend/lib/rds.ts
```diff
@@ -420,16 +423,35 @@ export class RDS {
...
             }
 
             schedulesMapping[scheduleId] = scheduleAggregate;
         });
 
+        for (const [scheduleId, indexes] of Object.entries(courseIndexes)) {
+            schedulesMapping[scheduleId].courses.sort((a, b) => {
+                const aIndex = indexes[a.sectionCode];
+                const bIndex = indexes[b.sectionCode];
+                if (aIndex === null || bIndex === null) {
```
sicn4rf: blocking: this still looks fragile for schedules with missing/legacy course indexes. `indexes[a.sectionCode]` can be `undefined`, not just `null`, and the comparator also never returns `0` for equal positions. Could we handle `== null` on both sides and use `aIndex - bIndex` once both exist?

(+1 on cubics comment)

apps/antalmanac/src/stores/AppStore.ts
```diff
@@ -324,6 +329,18 @@ class AppStore extends EventEmitter {
...
+    reorderAddedCourses(scheduleIndex: number, movedCourseId: string, nextCourseId: string | null) {
+        this.schedule.reorderAddedCourses(scheduleIndex, movedCourseId, nextCourseId);
+        this.unsavedChanges = true;
+        const action: ReorderAddedCoursesAction = {
+            type: 'reorderAddedCourses',
+            scheduleIndex: scheduleIndex,
+            movedCourseId: movedCourseId,
+            nextCourseId: nextCourseId,
+        };
+        actionTypesStore.autoSaveSchedule(action);
```
sicn4rf, line 341: issue/question: after reordering, should this emit `addedCoursesChange` like add/delete/clear/load do? the added courses pane updates optimistically, but other listeners that depend on schedule course order won’t hear about this mutation.

apps/antalmanac/src/components/drag-and-drop/SortableList.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { List, SxProps } from '@mui/material';
+import { mergeSx } from '@mui/x-date-pickers/internals';
+import type { ContextType, ReactNode } from 'react';
+import { createContext, Fragment, useMemo, useState } from 'react';
+
+interface BaseItem {
+    id: UniqueIdentifier;
+}
+
+interface Props<T extends BaseItem> {
```
sicn4rf: suggestion: convention: props type naming: `{ComponentName}Props`. can we rename `Props` to `SortableListProps` here? we’ve been using `{ComponentName}Props` for component props types so grepping and references stay predictable

apps/antalmanac/src/components/drag-and-drop/DragHandle.tsx
```diff
@@ -0,0 +1,55 @@
...
+            ref={ref}
+            sx={mergeSx(
+                {
+                    display: 'flex',
+                    alignItems: 'center',
+                    justifyContent: 'center',
+                    cursor: draggingState !== null ? 'grabbing' : disabled ? 'auto' : 'grab',
+                    borderRadius: 1,
+                    touchAction: 'none',
+                    '&:hover': {
```
sicn4rf, line 33: suggestion: minor UX polish from smoke testing: the drag handle hover state appears darker than the adjacent course info button hover, even though they sit next to each other as peer buttons/controls. could we align the hover styling with the existing button treatment so the controls feel consistent?

sicn4rf (changes requested):

apps/antalmanac/src/stores/AppStore.ts
Choollol: I dont think it needs to rn, but it might matter in the future so I think it makes sense to add now

apps/antalmanac/src/components/drag-and-drop/DragHandle.tsx
Choollol: could you send a screenshot? i'm looking at them with my eyes and with the inspector and i can't see the difference

sicn4rf: holy i might be blind i think it actually is the same my fault. color theory

apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -182,6 +186,29 @@ export class Schedules {
...
+        const courses = this.schedules[scheduleIndex].courses;
+
+        const fromIndex = courses.findIndex((course) => getCourseId(course) === movedCourseId);
+        const toIndex =
+            nextCourseId !== null
+                ? courses.findIndex((course) => getCourseId(course) === nextCourseId)
+                : courses.length;
+        const sectionCount = courses.findLastIndex((course) => getCourseId(course) === movedCourseId) - fromIndex + 1;
+
+        moveArrayElements(courses, fromIndex, toIndex, { elementMoveCount: sectionCount });
```
sicn4rf, line 222: question: following up on cubic’s `findIndex === -1` concern: are we relying on `movedCourseId` / `nextCourseId` always coming from the current rendered course list, or should this guard before `moveArrayElements`? If either id is missing, `-1` could move the wrong course.

apps/antalmanac/src/components/drag-and-drop/SortableList.tsx
```diff
@@ -0,0 +1,106 @@
+import { DragHandle } from '$components/drag-and-drop/DragHandle';
+import { SortableItem } from '$components/drag-and-drop/SortableItem';
+import { SortableOverlay } from '$components/drag-and-drop/SortableOverlay';
+import { DndContext, KeyboardSensor, PointerSensor, useSensor, useSensors } from '@dnd-kit/core';
+import type { Active, UniqueIdentifier } from '@dnd-kit/core';
+import { restrictToVerticalAxis } from '@dnd-kit/modifiers';
+import { SortableContext, SortingStrategy, arrayMove, sortableKeyboardCoordinates } from '@dnd-kit/sortable';
+import { List, SxProps } from '@mui/material';
+import { mergeSx } from '@mui/x-date-pickers/internals';
```
sicn4rf: question: again here cubic flagged this private MUI import. i see existing `mergeSx` usage elsewhere, but do we want to keep expanding imports from `@mui/x-date-pickers/internals`, or should we use a small local helper here?

apps/antalmanac/src/components/drag-and-drop/SortableList.tsx
```diff
@@ -0,0 +1,106 @@
...
+            sensors={sensors}
+            modifiers={[restrictToVerticalAxis]}
+            autoScroll={{ threshold: { x: disableHorizontalScroll ? 0 : 0.2, y: 0.2 } }}
+            onDragStart={({ active }) => {
+                setActive(active);
+                document.body.style.cursor = 'grabbing';
+            }}
+            onDragEnd={({ active, over }) => {
+                if (over && active.id !== over?.id) {
+                    const activeIndex = items.findIndex(({ id }) => id === active.id);
```
sicn4rf, line 71: question: also  following up on cubic’s guard comment: should we return early if either `activeIndex` or `overIndex` is `-1` before calling `arrayMove`? it may be impossible in normal dnd-kit flow, but confirming that invariant would make this easier to review.

sicn4rf (changes requested):
all my other comments have been resolved but just wanted to ask about cubic comments. otherwise lgtm

apps/antalmanac/src/components/drag-and-drop/SortableList.tsx
Choollol: True, that might be a separate PR thing since I used it in other places too. I'll do the copy-paste in this one tho, if there are no objections

apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -182,6 +187,41 @@ export class Schedules {
...
+            console.error(`Course id ${movedCourseId} was not found in schedule courses`);
+            openSnackbar('error', 'Could not reorder added courses');
+            return;
+        }
+
+        const toIndex =
+            nextCourseId !== null
+                ? courses.findIndex((course) => getCourseId(course) === nextCourseId)
+                : courses.length;
+        if (fromIndex === -1) {
```
sicn4rf: issue: the new missing-course guard looks like it has a small typo: the second check still uses `fromIndex === -1`, but this block seems intended to guard `toIndex` when `nextCourseId !== null`. as written, a missing `nextCourseId` can still pass through as `-1` and reorder incorrectly. could this be changed to something like `if (nextCourseId !== null && toIndex === -1)` and log `nextCourseId` instead?

thanks ai

sicn4rf (changes requested):

apps/antalmanac/src/stores/Schedules.ts
Choollol: oops again

sicn4rf:
thanks for fixing the typo, this LGTM to now just resolve the merge conflicts and ill approve

really nice feature! @Choollol

sicn4rf (approved):
lgtm fire feature 🔥

PR #1555, opened 2026-03-08
1543 persistent schedule ids

### LINKED ISSUE #1543: Refactor: Retain Schedule Ids on Save
There is existing legacy behavior when a user saves a schedule we destroy and re-insert the schedule as a new row(s). This is both excessive and limiting to planned features. 

Blocks: #1357 

## Potential Test Plan

1) Edit and save an **existing** schedule
2) Save a **new** schedule
3) Edit multiple schedules and create multiple schedules
4) Repeat 1 and 2 with custom events 
5) Repeat 1, 2, 4 with schedule notes 
6) Repeat the above with renaming schedules

brelieu05:
## Summary
Previously, every save deleted all of a user's schedules and re-inserted them, causing schedule IDs to change on every save. To maintain persistent schedule IDs, schedules with a known DB ID are updated in place, new schedules are inserted, and removed schedules are deleted. Schedule IDs now round-trip through the frontend store so subsequent saves always take the update path.
                                                                                                                   
The schedules table has unique constraints on (user_id, name) and (user_id, index) which caused mid-transaction constraint violations when updating indices. These are converted to DEFERRABLE INITIALLY DEFERRED so Postgres only validates uniqueness at commit time, allowing the transaction to freely update all schedules before checking the final state. After each successful save, the server returns the DB-assigned schedule IDs which are written back into the frontend store, ensuring subsequent saves always have the correct IDs to take the update path.

## Test Plan
- Saved a schedule, reload the page, verify the schedule ID is the same in the DB
- Added a class, save —verified no constraint errors and the existing schedule is updated in place
- Added a second schedule, save — verified both schedules persist with stable IDs on subsequent saves
- Renamed a schedule, save — verified the update succeeds without constraint violation
- Deleted a schedule, save — verified it is removed from the DB
- Used manual save to remove a schedule and add back in. (Testing constraints)
    1. Added a schedule to the end of the schedule list
    2. Saved
    3. Removed said schedule
    4. Added a new schedule at the end + with the same name 
    5. Saved
    6. When refreshed, all saves will be properly saved with no rollback. When schedule ids were console logged, new schedule had a new unique schedule id
- Sent simultaneous post request of two schedules with the same name -> schedules were automatically renamed and appended (1) before sent to DB -> never triggers duplicate name error
    -image

- Tested deadlock scenario by sending two simultaneous "saves" or mutates of the same schedule
    -image
    -image
    - image
    - One succeeded (B) while the other returned null (A). And the error message is intended since DB aborted one transaction, caused the error and logged it, and allowed the other transaction to complete
- Race condition unit tests (apps/antalmanac/tests/schedule-id-race-condition.test.ts): reproduced the ID-corruption bug by throttling to Slow 4G and inserting a schedule 
  mid-flight, then verified the CUID-keyed fix resolves it                                                                                                                   
    - To run: pnpm --filter antalmanac exec vitest run tests/schedule-id-race-condition.test.ts                                                                              
    - To observe the bug without the fix: revert updateScheduleIds to the positional ids: string[] implementation — the regression test passes (proves the bug exists) while the four fix-verification tests fail
    - Reproduce bug by hand as well
    > Setup — user has two schedules:
    > this.schedules[0] = { scheduleName: "Fall", scheduleId: "cuid-A" }
    > this.schedules[1] = { scheduleName: "Winter", scheduleId: "cuid-B" }
    > 
    > Step 1 — auto-save fires. getScheduleAsSaveState() serializes [Fall, Winter] and the request goes out.
    > 
    > Step 2 — while the request is in-flight, the user adds a new schedule at the front (or reorders):
    > this.schedules[0] = { scheduleName: "NEW", scheduleId: "cuid-NEW" } ← inserted at 0
    > this.schedules[1] = { scheduleName: "Fall", scheduleId: "cuid-A" }
    > this.schedules[2] = { scheduleName: "Winter", scheduleId: "cuid-B" }
    > 
    > Step 3 — save response arrives with scheduleIds = ["db-id-for-Fall", "db-id-for-Winter"]. updateScheduleIds runs:
    > this.schedules[0].scheduleId = "db-id-for-Fall" // ← NEW gets Fall's DB id
    > this.schedules[1].scheduleId = "db-id-for-Winter" // ← Fall gets Winter's DB id
    > // schedules[2] (Winter) is untouched, still has "cuid-B"
    > 
    > Step 4 — next save fires with these corrupted IDs:
    > 
    > NEW tries to upsert with db-id-for-Fall → existingIds contains it → updates Fall's DB row with NEW's data (wrong name, wrong courses)
    > Fall tries to upsert with db-id-for-Winter → updates Winter's DB row with Fall's data
    > Winter tries to upsert with cuid-B → existingIds doesn't contain it → inserts a fresh row
    > End result in the DB:
    > db-id-for-Fall → "NEW" (was "Fall")
    > db-id-for-Winter → "Fall" (was "Winter")
    > new-db-id → "Winter" (duplicate, correct data but new ID)

## Issues

Closes #1543

### DIFF
--- a/apps/antalmanac/src/generated/deployed_terms.json
+++ b/apps/antalmanac/src/generated/deployed_terms.json
@@ -1,6 +1,6 @@
 {
   "latestTerm": "2026 Summer Session 2",
-  "sectionCount": 486,
-  "updatedAt": "2026-03-27T04:28:14.258Z",
+  "sectionCount": 487,
+  "updatedAt": "2026-03-27T18:45:14.059Z",
   "reason": "Section count changed"
 }


packages/types/src/schedule.ts
```diff
@@ -19,6 +19,7 @@ export type Schedule = {
     courses: ScheduleCourse[];
     customEvents: RepeatingCustomEvent[];
     scheduleNoteId: number;
+    scheduleId?: string;
```
KevinWu098: question: why would ids be optional?

packages/types/src/schedule.ts
KevinWu098, line 17: question: i know this isn't part of the diff, but i feel like we should move towards deriving these types from our DB schema rather than repeating ourselves (and allowing for drift)

apps/antalmanac/src/stores/Schedules.ts
```diff
@@ -546,6 +546,7 @@ export class Schedules {
     getScheduleAsSaveState(): ScheduleSaveState {
         const shortSchedules: ShortCourseSchedule[] = this.schedules.map((schedule) => {
             return {
+                ...(schedule.scheduleId ? { id: schedule.scheduleId } : {}),
```
KevinWu098: question: again, when would schedules not have an id? this seems bad / wrong

apps/antalmanac/src/backend/lib/rds.ts
```diff
@@ -199,57 +199,17 @@ export class RDS {
...
-
     /**
      * Does the same thing as `insertGuestUserData`, but also updates the user's schedules and courses if they exist.
      *
      * @param db The Drizzle client or transaction object
      * @param userData The object of data containing the user's schedules and courses
      * @returns The user's ID
      */
-    static async upsertUserData(db: DatabaseOrTransaction, userData: User): Promise<string> {
+    static async upsertUserData(
```
KevinWu098: question/issue: i haven' read this very deeply, but doesn't drizzle support upserts https://orm.drizzle.team/docs/guides/upsert via `onConflictDoUpdate`?

KevinWu098 (changes requested):
Some initial comments. Mostly, I'm concerned about type safety and getting us to use more of the built-in APIs rather than doing more (possibly bug-prone) custom logic

packages/types/src/schedule.ts
brelieu05: scheduleId was optional because IDs were only assigned after it was saved to the DB. Thus, when a user added a schedule, the scheduleId would be undefined until they saved.

To make it strictly string, I could try moving the id generation to the frontend using cuid's createId(), so it will have an id even when just added

packages/types/src/schedule.ts
brelieu05: https://github.com/icssc/AntAlmanac/issues/1566

Made it a separate issue after discussion with Alex because it'll need a more comprehensive refactor

packages/db/src/schema/schedule/schedule.ts
```diff
@@ -1,37 +1,33 @@
 import { createId } from '@paralleldrive/cuid2';
-import { pgTable, unique, text, timestamp, integer } from 'drizzle-orm/pg-core';
+import { pgTable, text, timestamp, integer } from 'drizzle-orm/pg-core';
```
alexespejo: The deferred unique constraints on (user_id, name) and (user_id, index) are not declared here — they only exist via the raw SQL in migration  
  0009. If pnpm db:generate is run in the future after an unrelated schema change, drizzle-kit may regenerate those constraints as non-deferred or
   drop them entirely. Add a comment here so future devs don't unknowingly clobber them:                                                          
```
// NOTE: unique constraints on (userId, name) and (userId, index) are intentionally                                                           
// not declared here. They are managed manually in migration 0009 as DEFERRABLE INITIALLY DEFERRED                                            
// to allow index/name swaps within a transaction. Do not let drizzle-kit regenerate them.    
```

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -137,6 +137,10 @@ export const saveSchedule = async (
                     },
                 });
 
+                if (result?.scheduleIds) {
+                    AppStore.schedule.updateScheduleIds(result.scheduleIds);
```
alexespejo: Niche bug: 

`updateScheduleIds` maps returned IDs by position (ids[0] → this.schedules[0]), which assumes the schedule array hasn't changed between when                 
  getScheduleAsSaveState() was called and when the response comes back. If the user reorders or adds a schedule while a save is in-flight, the IDs
   could be written to the wrong slots. On the next save, the mismatched ID would be caught by existingIds.has(schedule.id) and fall through to a 
  fresh insert rather than an update — so no data is corrupted — but it's a silent failure worth documenting. At minimum, add a comment noting the
   assumption. A more robust fix would be to key the returned IDs by the frontend CUID that was sent up rather than by array position.

alexespejo (changes requested):
Honestly, the race condition doesn't seem like a practical issue since by default schedules are created at the end of the options 

I was able to reproduce by setting my throttling to Slow 4G: 
> Setup — user has two schedules:                                                                                                                 
>   this.schedules[0] = { scheduleName: "Fall",   scheduleId: "cuid-A" }                                                                            
>   this.schedules[1] = { scheduleName: "Winter", scheduleId: "cuid-B" }                                                                            
>                                                                                                                                                   
>   Step 1 — auto-save fires. getScheduleAsSaveState() serializes [Fall, Winter] and the request goes out.                                          
>                                                                                                                                                   
>   Step 2 — while the request is in-flight, the user adds a new schedule at the front (or reorders):                                               
>   this.schedules[0] = { scheduleName: "NEW",    scheduleId: "cuid-NEW" }  ← inserted at 0                                                         
>   this.schedules[1] = { scheduleName: "Fall",   scheduleId: "cuid-A" }                                                                            
>   this.schedules[2] = { scheduleName: "Winter", scheduleId: "cuid-B" }                                                                            
>                                                                                                                                                   
>   Step 3 — save response arrives with scheduleIds = ["db-id-for-Fall", "db-id-for-Winter"]. updateScheduleIds runs:                               
>   this.schedules[0].scheduleId = "db-id-for-Fall"   // ← NEW gets Fall's DB id                                                                    
>   this.schedules[1].scheduleId = "db-id-for-Winter" // ← Fall gets Winter's DB id                                                                 
>   // schedules[2] (Winter) is untouched, still has "cuid-B"                                                                                       
>                                                                                                                                                   
>   Step 4 — next save fires with these corrupted IDs:                                                                                              
>   - NEW tries to upsert with db-id-for-Fall → existingIds contains it → updates Fall's DB row with NEW's data (wrong name, wrong courses)         
>   - Fall tries to upsert with db-id-for-Winter → updates Winter's DB row with Fall's data                                                         
>   - Winter tries to upsert with cuid-B → existingIds doesn't contain it → inserts a fresh row                                                     
>                                                                                                                                                   
>   End result in the DB:                                                                                                                           
>   db-id-for-Fall  → "NEW"    (was "Fall")                                                                                                         
>   db-id-for-Winter → "Fall"  (was "Winter")                                                                                                       
>   new-db-id       → "Winter" (duplicate, correct data but new ID)

apps/antalmanac/src/backend/lib/rds.ts
```diff
@@ -265,34 +225,114 @@ export class RDS {
             }
 
             // Add schedules and courses
-            const scheduleIds = await this.upsertSchedulesAndContents(tx, userId, userData.userData.schedules);
+            const scheduleIdMap = await this.upsertSchedulesAndContents(tx, userId, userData.userData.schedules);
 
             // Update user's current schedule index
             const scheduleIndex = userData.userData.scheduleIndex;
+            const scheduleDbIds = Object.values(scheduleIdMap);
```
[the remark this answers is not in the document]
brelieu05: Invalid because I made it so schedules created will be constructed with an id with createId() in Schedule.ts. Thus I'm resolving this conversation

PR #1562, opened 2026-03-09
add: add back colour picker

### LINKED ISSUE #1558: Bring back color picker
In implementing AANTS, we removed the color picker in the section table component. Having spent some time with it, I'm noticing that because 1. the calendar events are not super readable and 2. there is no longer a color picker, it's harder for me to actually read the calendar because i can't built the mental association via color.

We should fix this somehow without regressing UI (which is particularly noticeable on mobile)

---
*Created by <@227113377145552896> (Kevin 💖) from #antalmanac-internal in Discord*
*[View in Discord](https://discord.com/channels/772739905981644850/1320360830251827272)*

vicksey:
## Summary
colour associativity is BACK 
image

## Issues

Closes #1558


KevinWu098:
Can we reduce layout shift? it's not a great UX if the trash button shifts when added/not-added. what if we brought back the original dropdown menu button which let people copy link and add to schedule?

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody/SectionTableBodyCells/action-cell/DeleteAndNotifications.tsx
```diff
@@ -44,22 +44,40 @@ export const DeleteAndNotifications = memo(({ ...props }: DeleteAndNotifications
...
+            >
+                {initialized ? (
+                    <NotificationsMenu {...props} />
+                ) : (
+                    <IconButton disabled>
+                        <CircularProgress size={15} />
+                    </IconButton>
+                )}
+
+                <ColorPicker
```
alexespejo: I'm sure will deliberate with UI/UX but at least looking at the components: the padding is uneven across the three buttons I'm sure there's a way to orient the buttons cleanly and also reduce some of the space the column will take up

Image

Image

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody/SectionTableBodyCells/action-cell/DeleteAndNotifications.tsx
```diff
@@ -44,22 +44,40 @@ export const DeleteAndNotifications = memo(({ ...props }: DeleteAndNotifications
...
-                <NotificationsMenu {...props} />
-            ) : (
-                <IconButton disabled>
-                    <CircularProgress size={15} />
-                </IconButton>
-            )}
+            <Box
+                sx={{
+                    display: 'flex',
+                    flexDirection: isMobile ? 'column' : 'row',
```
alexespejo: I would prob still fat finger the delete button 
Image

alexespejo (commented):
I like the idea. Kevin's #1592 PR adds back the color picker itself. But I can see this being a mobile related feature? But I know there's an argument about how knowing the color of a course is unnecessary to mobile users.

alexespejo (changes requested):
Based on recent conversations Kevin suggests making the color picker a clickable element that can still change the color. So all behavior should be the same. Maybe on hover the color picker will expand in width and you can click it. On mobile I think we've considered that users won't likely be interfacing with the color picker so it wouldn't be the biggest deal 

image

vicksey:
merged into #1585

PR #1565, opened 2026-03-10
feat: dev mode

### LINKED ISSUE #1545: RFC: DEV Mode in AntAlmanac
> thought: What if we expand this PR and create a "DEV mode" where we can basically hide special features like this behind a toggle. Power users can also use this feature if they really wanted to. 

_Originally posted by @alexespejo in https://github.com/icssc/AntAlmanac/pull/1460#pullrequestreview-3810825316_

However we decide to approach this, we could use a mechanism like this to create a suite of dev tools we can use in prod and development that we'd hide from users

xgraceyan:
## Summary
Based off #1460, adds a dev mode toggle (on for dev instance of AA by default). Moved import export schedules to be a feature of dev mode.
image

## Test Plan
- Ensure dev mode state persists in local storage
- Ensure import export schedules doesn't show when dev mode is disabled.

**For import/export**
1) Test with single schedule (signed in)
2) 1 but signed out 
3) Repeat 1 and 2 with combination of classes and custom events 
4) Repeat 1, 2, and 3 with schedule note 

## Issues

Closes #1545


alexespejo:
image

There seems to be some false validation for custom events and building

xgraceyan:
Custom events only need schedule and day as required fields right. CC @brelieu05 but I'm also happy to resolve

alexespejo:
@xgraceyan hows the fixes looking regarding the import/export feature

xgraceyan:
Lwk clocked out during spring break. Working on them now 🫡

alexespejo (changes requested):
Works. Change the primary blue color to match the new ones we have and I'll approve

alexespejo:
bump

alexespejo (approved):

PR #1571, opened 2026-03-16
Ignore Starting and Ending Whitespaces in Course Search

galacticaledge:
## Summary
Small Quality of Life (QoL) improvement that makes the course search ignore starting and ending whitespaces. This is to avoid situations where ` 36006`, for example, result in `No results found! Please try broadening your search.`


KevinWu098 (approved):
Nice, thanks for the contribution!

PR #1574, opened 2026-03-20
feat: remove unsaved actions

KevinWu098:
## Summary
Removes unsaved actions code path

## Issues

Closes #

### DIFF
--- a/apps/antalmanac/src/actions/ActionTypesStore.ts
+++ b/apps/antalmanac/src/actions/ActionTypesStore.ts
@@ -4,17 +4,10 @@ import { CustomEventId, RepeatingCustomEvent, ScheduleCourse } from '@packages/a
 
 import { autoSaveSchedule } from '$actions/AppStoreActions';
 import trpc from '$lib/api/trpc';
-import {
-    getLocalStorageAutoSave,
-    getLocalStorageUnsavedActions,
-    removeLocalStorageUnsavedActions,
-    setLocalStorageUnsavedActions,
-} from '$lib/localStorage';
+import { getLocalStorageAutoSave, removeLocalStorageUnsavedActions } from '$lib/localStorage';
 import AppStore from '$stores/AppStore';
 import { scheduleComponentsToggleStore } from '$stores/ScheduleComponentsToggleStore';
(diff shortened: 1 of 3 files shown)


apps/antalmanac/src/actions/ActionTypesStore.ts
```diff
@@ -113,20 +106,12 @@ export type ActionType =
...
-    }
-}
-
 class ActionTypesStore extends EventEmitter {
     constructor() {
         super();
     }
 
-    async autoSaveSchedule(action: ActionType) {
+    async autoSaveSchedule(_action: ActionType) {
```
alexespejo, line 114: _?

KevinWu098: field wasn't used anymore, but I didn't want to change the signature. `_` satisfies the linter

apps/antalmanac/src/actions/ActionTypesStore.ts
```diff
@@ -148,81 +133,11 @@ class ActionTypesStore extends EventEmitter {
...
-                if (unsavedActions.length > MAX_UNSAVED_ACTIONS) {
-                    unsavedActions = unsavedActions.slice(100);
-                }
-                unsavedActions.push(action);
-                setLocalStorageUnsavedActions(JSON.stringify(unsavedActions));
-            }
         }
     }
 
     async loadScheduleFromUnsavedActions() {
```
alexespejo: if we're removing unsaved changes we might as well get rid of any prior remnants

alexespejo (changes requested):
thought: maybe we could have our own dialog to warn users that they haven't saved their changes rather than using the built in `BeforeUnload` prompt? Not the biggest deal but I feel like people are pretty dismissive of these built in browser dialogs.

KevinWu098:
> thought: maybe we could have our own dialog to warn users that they haven't saved their changes rather than using the built in `BeforeUnload` prompt? Not the biggest deal but I feel like people are pretty dismissive of these built in browser dialogs. 

I'm amenable to this, but I'd rather not do this on this PR if possible 🥺

alexespejo (approved):
LGTM

PR #1580, opened 2026-04-03
Fix aants package dependency on db package

### LINKED ISSUE #1579: Fix AANTS dependency on @packages/db
Files in the aants package currently need to do some pretty hacky stuff that also causes errors:

Image

This can be fixed by adding the correct local workspace package as a dependency.

Choollol:
## Summary

Add @packages/db as a dependency to the AANTS package to fix errors.

## Test Plan

Confirm that there are no import errors in `subscriptionData.ts`.

## Issues

Closes #1579

### DIFF
--- a/apps/aants/package.json
+++ b/apps/aants/package.json
@@ -15,6 +15,7 @@
         "@aws-sdk/client-sesv2": "^3.750.0",
         "@aws-sdk/client-sqs": "^3.750.0",
         "@packages/anteater-api-types": "workspace:*",
+        "@packages/db": "workspace:*",
         "@react-email/components": "^1.0.8",
         "@react-email/render": "^2.0.4",
         "@trpc/server": "^10.30.0",
--- a/apps/aants/src/helpers/subscriptionData.ts
+++ b/apps/aants/src/helpers/subscriptionData.ts
@@ -1,9 +1,8 @@
 import type { WebsocAPIResponse, WebsocSection } from '@packages/anteater-api-types';
+import { db } from '@packages/db/src/index';
+import { type User as DbUser, users } from '@packages/db/src/schema/auth/user';
+import { type Subscription, subscriptions } from '@packages/db/src/schema/subscription';
 import { and, eq, inArray } from 'drizzle-orm';
-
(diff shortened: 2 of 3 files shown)


sicn4rf (approved):
lgtm! good catch and clean fix :) 👍 

a good follow-up clean-up would be to make the import from `@packages/db/src` consistent across the codebase. node modules automatically look for an `index.js`/`index.ts` file in that directory so `@packages/db/src/index` is equivalent to just `@packages/db/src`. i dont think theres a right or wrong here since they both work but i think its good to note for consistency's sake. we do this import inconsistently across files in antalmanac at the moment

sicn4rf:
Screenshot 2026-04-06 at 12 38 31 AM

IsaacNguyen (approved):

Choollol:
> lgtm! good catch and clean fix :) 👍
> 
> a good follow-up clean-up would be to make the import from `@packages/db/src` consistent across the codebase. node modules automatically look for an `index.js`/`index.ts` file in that directory so `@packages/db/src/index` is equivalent to just `@packages/db/src`. i dont think theres a right or wrong here since they both work but i think its good to note for consistency's sake. we do this import inconsistently across files in antalmanac at the moment

Good point! What do you think about just `import { db } from '@packages/db';`? We could do this by adding `"main": "./src/index.ts",` to the db package's `package.json`. I think it makes more sense than importing from a `src` directory, but I also don't have much experience in these parts so lmk!

sicn4rf:
> Good point! What do you think about just `import { db } from '@packages/db';`? We could do this by adding `"main": "./src/index.ts",` to the db package's `package.json`. I think it makes more sense than importing from a `src` directory, but I also don't have much experience in these parts so lmk!

yeah that's a good idea! that's pretty much the pattern we see with `@packages/anteater-api-types` in `packages/anteater-api-types/package.json`. so, we know its already an established pattern. we just need to add `"main":` and `"types":` to `package.json` (like how `anteater-api-types` does it) and update the imports to use just `@packages/db`. that'd be a good follow-up, it could be its own small pr :)

PR #1581, opened 2026-04-05
Use fetchAnteaterAPI for all AAPI calls

### LINKED ISSUE #1570: Refactor: use fetchAnteaterAPI for all AAPI fetch calls
The tRPC routers for `course.ts`, `enrollHist.ts`, and `grades.ts` all make raw fetch calls to the Anteater API without checking response.ok, without try/catch on .json(), and with unsafe static type casts. The `fetchAnteaterAPI `helper in `websoc.ts` already solves these issues — it should be moved to a shared location and adopted by all routers that call the Anteater API.
 

Comment from francis: 
now that the `fetchAnteaterAPI` abstraction has been implemented, we should reuse this in other files that face the exact same problem that websoc did. `course.ts`, `enrollHist.ts`, and `grades.ts` all follow this exact same pattern. a good solution to this issue is to move `fetchAnteaterAPI` to a shared location and these three files (or more if this logic is seen anywhere else) should be updated to use it. all of these files fetch from aapi. however, they do not check `response.ok`, the `.json()` is not in a try/catch, and the data is cast statically.

_Originally posted by @sicn4rf in https://github.com/icssc/AntAlmanac/pull/1544#pullrequestreview-3931370389_

Choollol:
## Summary

Wrap all Anteater API calls in `fetchAnteaterAPI` to catch errors.

## Test Plan

1. All AAPI features should work as before
2. Any errors when calling AAPI should be caught and a TRPC error should be thrown

## Issues

Closes #1570

### DIFF
--- a/apps/antalmanac/scripts/fetch-term-data.ts
+++ b/apps/antalmanac/scripts/fetch-term-data.ts
@@ -8,7 +8,9 @@ import { mkdir, writeFile } from 'node:fs/promises';
 import { dirname, join } from 'node:path';
 import { fileURLToPath } from 'node:url';
 
-import type { CalendarTerm } from '@packages/antalmanac-types';
+import type { CalendarTerm, CalendarAllAPIResult } from '@packages/antalmanac-types';
+
+import { fetchAnteaterAPI } from '$src/backend/lib/helpers';
 
 const PUBLIC_ANTEATER_API_KEY = 'INSqn9qP1pXlEwihpQa_GtrJhGOxQyjE5zcAKYLptLg.pk.prj9hlf3sf7q638jkq61u282';
 const __dirname = dirname(fileURLToPath(import.meta.url));
@@ -33,18 +35,16 @@ function sanitizeTermName(year: string, quarter: keyof typeof QUARTER_MAP): `${s
 }
 
 async function fetchCalendarTerms(): Promise<CalendarTerm[]> {
-    const res = await fetch(API_URL, {
+    const { data } = await fetchAnteaterAPI<CalendarAllAPIResult>(API_URL, {
         headers: {
             Authorization: `Bearer ${PUBLIC_ANTEATER_API_KEY}`,
             Origin: ORIGIN,
         },
+        invalidResponseCallback: (res) => {
+            throw new Error(`Failed to fetch terms: ${res.statusText}`);
+        },
     });
 
-    if (!res.ok) {
-        throw new Error(`Failed to fetch terms: ${res.statusText}`);
-    }
-
-    const { data } = await res.json();
     return data;
 }
 
--- a/apps/antalmanac/src/backend/routers/course.ts
+++ b/apps/antalmanac/src/backend/routers/course.ts
@@ -3,11 +3,11 @@ import { z } from 'zod';
 
 import { procedure, router } from '../trpc';
 
-import { fetchAnteaterAPIData } from '$src/backend/lib/helpers';
+import { fetchAnteaterAPI } from '$src/backend/lib/helpers';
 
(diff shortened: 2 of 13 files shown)


apps/antalmanac/scripts/fetch-term-data.ts
```diff
@@ -33,18 +35,16 @@ function sanitizeTermName(year: string, quarter: keyof typeof QUARTER_MAP): `${s
 }
 
 async function fetchCalendarTerms(): Promise<CalendarTerm[]> {
-    const res = await fetch(API_URL, {
+    const { data } = await fetchAnteaterAPI<CalendarAllAPIResult>(API_URL, {
```
sicn4rf, line 38: thought: the scripts in the `scripts/` directory are used outside the context of tRPC, but the fetchAnteaterAPI call is used and will throw tRPC errors on failure which is misleading.

however, i undrestand you didn't implement `fetchAnteaterAPI` yourself so this isn't really on you since that wasnt your design decision.

this is mainly just a design concern since it is misleading to get tRPC errors from code that has nothing to do with tRPC.  i think a follow up worth having would be to have `fetchAnteaterAPI` throw plain `Error` and let tRPC routers wrap it instead

apps/antalmanac/src/backend/routers/course.ts
```diff
@@ -1,17 +1,16 @@
...
-            headers: {
-                ...(process.env.ANTEATER_API_KEY && { Authorization: `Bearer ${process.env.ANTEATER_API_KEY}` }),
-            },
-        })
-            .then((data) => data.json())
-            .then((data) => (data.ok ? (data.data as Course) : null));
+        const data = await fetchAnteaterAPI<CourseByIdAPIResult>(
+            `https://anteaterapi.com/v2/rest/courses/${encodeURIComponent(input.id)}`
+        );
+        return data.ok ? data.data : null;
```
sicn4rf: nit: per AAPI's OpenAPI spec, a 200 response always has `ok: true`. so this check is pretty much dead code since `fetchAnteaterAPI` already throws on non-200 HTTP responses.

apps/antalmanac/src/backend/routers/websoc.ts
sicn4rf: nit: same thing here, `!data?.ok` can't be true after `fetchAnteaterAPI` passes. AAPI types enforce `ok: true` on 200 responses. its prob fine to keep as defensive code if you prefer, just noting it. however its not used in other calls so it might be wortht he clean up to avoid confusion

sicn4rf (changes requested):
take the nits n thoughts into consideration (and validate), and feel free to dismiss if you think theyre not valid and just request my review again :)

apps/antalmanac/scripts/fetch-term-data.ts
Choollol: Oh yeah I didn't notice that. Do you think we could create a "core" package? It would just be somewhere to put shared utilities that any other package can use, like `fetchAnteaterAPI`.

sicn4rf: i'm gonna CC @alexespejo here since this seems like a bigger scope change. i think the idea is reasonable, especially if we accumulate a lot more shared utils in the future.

in this case, however, i think creating a core package may be over-engineering a solution to fix an error type. 

the simpler fix might just be to throw a plain `Error` (`TRPCError` extends `Error`), and let the tRPC routers handle the wrapping instead. this way, it can stay in `helpers.ts` without needing to change the package infrastructure.

sicn4rf (approved):
nice, this LGTM !

PR #1582, opened 2026-04-06
1543 persistent schedule ids

### LINKED ISSUE #1543: Refactor: Retain Schedule Ids on Save

brelieu05:
## Summary
Previously, every save deleted all of a user's schedules and re-inserted them, causing schedule IDs to change on every save. To maintain persistent schedule IDs, schedules with a known DB ID are updated in place, new schedules are inserted, and removed schedules are deleted. Schedule IDs now round-trip through the frontend store so subsequent saves always take the update path.
                                                                                                                   
The schedules table has unique constraints on (user_id, name) and (user_id, index) which caused mid-transaction constraint violations when updating indices. These are converted to DEFERRABLE INITIALLY DEFERRED so Postgres only validates uniqueness at commit time, allowing the transaction to freely update all schedules before checking the final state. After each successful save, the server returns the DB-assigned schedule IDs which are written back into the frontend store, ensuring subsequent saves always have the correct IDs to take the update path.

## Test Plan
- Saved a schedule, reload the page, verify the schedule ID is the same in the DB
- Added a class, save —verified no constraint errors and the existing schedule is updated in place
- Added a second schedule, save — verified both schedules persist with stable IDs on subsequent saves
- Renamed a schedule, save — verified the update succeeds without constraint violation
- Deleted a schedule, save — verified it is removed from the DB
- Used manual save to remove a schedule and add back in. (Testing constraints)
    1. Added a schedule to the end of the schedule list
    2. Saved
    3. Removed said schedule
    4. Added a new schedule at the end + with the same name 
    5. Saved
    6. When refreshed, all saves will be properly saved with no rollback. When schedule ids were console logged, new schedule had a new unique schedule id
- Sent simultaneous post request of two schedules with the same name -> schedules were automatically renamed and appended (1) before sent to DB -> never triggers duplicate name error
    -image

- Tested deadlock scenario by sending two simultaneous "saves" or mutates of the same schedule
    -image
    -image
    - image
    - One succeeded (B) while the other returned null (A). And the error message is intended since DB aborted one transaction, caused the error and logged it, and allowed the other transaction to complete
- Race condition unit tests (apps/antalmanac/tests/schedule-id-race-condition.test.ts): reproduced the ID-corruption bug by throttling to Slow 4G and inserting a schedule 
  mid-flight, then verified the CUID-keyed fix resolves it                                                                                                                   
    - To run: pnpm --filter antalmanac exec vitest run tests/schedule-id-race-condition.test.ts                                                                              
    - To observe the bug without the fix: revert updateScheduleIds to the positional ids: string[] implementation — the regression test passes (proves the bug exists) while the four fix-verification tests fail
    - Reproduce bug by hand as well
    > Setup — user has two schedules:
    > this.schedules[0] = { scheduleName: "Fall", scheduleId: "cuid-A" }
    > this.schedules[1] = { scheduleName: "Winter", scheduleId: "cuid-B" }
    > 
    > Step 1 — auto-save fires. getScheduleAsSaveState() serializes [Fall, Winter] and the request goes out.
    > 
    > Step 2 — while the request is in-flight, the user adds a new schedule at the front (or reorders):
    > this.schedules[0] = { scheduleName: "NEW", scheduleId: "cuid-NEW" } ← inserted at 0
    > this.schedules[1] = { scheduleName: "Fall", scheduleId: "cuid-A" }
    > this.schedules[2] = { scheduleName: "Winter", scheduleId: "cuid-B" }
    > 
    > Step 3 — save response arrives with scheduleIds = ["db-id-for-Fall", "db-id-for-Winter"]. updateScheduleIds runs:
    > this.schedules[0].scheduleId = "db-id-for-Fall" // ← NEW gets Fall's DB id
    > this.schedules[1].scheduleId = "db-id-for-Winter" // ← Fall gets Winter's DB id
    > // schedules[2] (Winter) is untouched, still has "cuid-B"
    > 
    > Step 4 — next save fires with these corrupted IDs:
    > 
    > NEW tries to upsert with db-id-for-Fall → existingIds contains it → updates Fall's DB row with NEW's data (wrong name, wrong courses)
    > Fall tries to upsert with db-id-for-Winter → updates Winter's DB row with Fall's data
    > Winter tries to upsert with cuid-B → existingIds doesn't contain it → inserts a fresh row
    > End result in the DB:
    > db-id-for-Fall → "NEW" (was "Fall")
    > db-id-for-Winter → "Fall" (was "Winter")
    > new-db-id → "Winter" (duplicate, correct data but new ID)

## Issues

Closes #1543

Github shenanigans where the last pr just closed and I can't reopen :(


apps/antalmanac/tests/schedule-id-race-condition.test.ts
alexespejo, line 1: bruh wat

alexespejo:
Test plan for my own intents and porpoises 

---

## Test 1 — Verify IDs are stable across saves (Core Feature)

### Before Save

```sql
SELECT id, name, index
FROM schedules
WHERE user_id = '<your_user_id>'
ORDER BY index;
```

### Steps

1. In the app, add a course to schedule "Fall".
2. Click Save.

### After Save

```sql
SELECT id, name, index, last_updated
FROM schedules
WHERE user_id = '<your_user_id>'
ORDER BY index;
```

### Expected Result

* **Pass:** Same `id` values, only `last_updated` changes.
* **Fail:** New IDs are generated (indicates delete-reinsert behavior still happening).

---

## Test 2 — Verify `current_schedule_id` survives a save

### Before

```sql
SELECT current_schedule_id
FROM users
WHERE id = '<your_user_id>';
```

### Steps

1. Switch to your second schedule in the app.
2. Click Save.

### After

```sql
SELECT u.current_schedule_id, s.name
FROM users u
JOIN schedules s ON s.id = u.current_schedule_id
WHERE u.id = '<your_user_id>';
```

### Expected Result

* **Pass:** Points to the correct schedule name.
* **Fail:** `NULL` or incorrect schedule name.

---

## Test 3 — Stress the positional `currentScheduleId` edge case

### Setup

* Have 3 schedules: `Fall`, `Winter`, `Spring`.
* Set `Spring` (index 2) as active.

### Steps

1. Open a second browser tab.
2. In DevTools → Network, throttle to **Slow 3G**.
3. Click Save.
4. Before the response returns:

   * Add a new schedule `Summer` at the front OR reorder to index 0.
5. Let the original save response complete.

### Verification

```sql
SELECT u.current_schedule_id, s.name, s.index
FROM users u
JOIN schedules s ON s.id = u.current_schedule_id
WHERE u.id = '<your_user_id>';
```

### Expected Result

* **Pass:** `name = 'Spring'`
* **Fail:** Points to `Summer` or another incorrect schedule (indicates stale positional lookup).

---

## Test 4 — Verify delete-then-reinsert reuses IDs

### Before

```sql
SELECT id, name
FROM schedules
WHERE user_id = '<your_user_id>'
ORDER BY index;
```

### Steps

1. Rename a schedule in the app.
2. Click Save.

### After

```sql
SELECT id, name
FROM schedules
WHERE user_id = '<your_user_id>'
ORDER BY index;
```

### Expected Result

* **Pass:** Same IDs, updated name.
* **Fail:** New IDs (safeId reuse path not working).

---

## Test 5 — Simulate a foreign schedule ID rejection

### Step 1 — Get a foreign ID

```sql
SELECT id
FROM schedules
WHERE user_id != '<your_user_id>'
LIMIT 1;
```

Example:

```
foreign-sched-id-abc123
```

### Step 2 — Inject via request

* Use tRPC or replay a network request.
* Replace one of your schedule IDs with the foreign ID.
* Submit Save.

### Verification

```sql
SELECT id, name
FROM schedules
WHERE user_id = '<your_user_id>'
ORDER BY index;
```

### Expected Result

* **Pass:** No schedule has the foreign ID.
* **Fail:** Foreign ID appears (security issue — validation failed).

---

## Test 6 — Verify cascade integrity after delete

### Before

```sql
SELECT COUNT(*)
FROM "coursesInSchedule"
WHERE "scheduleId" = '<sched_id>';

SELECT COUNT(*)
FROM "customEvents"
WHERE "scheduleId" = '<sched_id>';
```

### Steps

1. Delete a schedule containing courses.
2. Click Save.

### After

```sql
SELECT COUNT(*)
FROM "coursesInSchedule"
WHERE "scheduleId" = '<old_sched_id>';

SELECT COUNT(*)
FROM "customEvents"
WHERE "scheduleId" = '<old_sched_id>';

SELECT id
FROM schedules
WHERE id = '<old_sched_id>';
```

### Expected Result

* **Pass:**

  * Counts return `0`
  * No schedule row exists
* **Fail:** Orphaned course or event rows remain

---

## Notes

* Replace `<your_user_id>` and `<sched_id>` with actual values.
* Run queries directly against your development database.
* Use browser DevTools for request manipulation when needed.

alexespejo:
image

alexespejo (approved):
LGTM 🐬

PR #1588, opened 2026-04-07
feat: add filters to search results page

### LINKED ISSUE #937: Add way to add filters from search results page
We already support filtering by start time or class fill-level but people might not know because it's hidden behind the advanced search options (I know because a friend messaged me asking to add filtering since he didn't know it was a feature). I haven't thought at all about how to implement a solution but the problem is worth noting down.

calebwongg:
## Summary
Add filters for each displayed course on the search results page. Filtering option persists across searches, but not across sessions

image
image 

## Test Plan
- Check each filtering option and ensure that classes are sorted accordingly 
- Check filtering methods across multiple searches 
## Issues
Closes #937


alexespejo:
I know this is still a draft but I don't think I see the purpose of sorting the individual class sections (granted I've only had to enroll in upper div cs classes). I can see another button being added to that top row where we can filter the classes against each other. Mb if this is redundant from a design discussion 

image

calebwongg:
@cubic hi

calebwongg:
@cubic

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableBody/SectionTableBody.tsx
```diff
@@ -1,12 +1,55 @@
...
+
+function getMeetingDays(section: AASection): string {
+    const meeting = section.meetings[0];
+    if (!meeting || meeting.timeIsTBA) return '';
+    return meeting.days;
+}
+
+const STATUS_ORDER: Record<string, number> = { OPEN: 0, NewOnly: 1, Waitl: 2, FULL: 3, '': 4 };
+
+function sortSections(sections: AASection[], sortBy: SortOption): AASection[] {
```
alexespejo: issue: the classes still seem to filter by their individual sections rather than across all classes

apps/antalmanac/src/stores/SectionFilterStore.ts
```diff
@@ -0,0 +1,21 @@
+import { create } from 'zustand';
+
+export const SORT_OPTIONS = [
+    { value: 'default', label: 'Default' },
```
alexespejo, line 4: suggestion: I think GPA would be a valuable feature

apps/antalmanac/src/stores/SectionFilterStore.ts
```diff
@@ -0,0 +1,21 @@
...
+    { value: 'status', label: 'Status' },
+    { value: 'time_asc', label: 'Time' },
+    { value: 'days_mwf', label: 'Date: MWF' },
```
alexespejo: thought: not sure how valuable it would be to filter by weekdays. Since some people might use the filter for different reasons. For example, they want a lec on MWF and a disc on TT or somn like that

alexespejo (changes requested):
I

calebwongg:
@cubic

alexespejo:
how's progress here

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
```diff
@@ -394,6 +484,50 @@ export default function CourseRenderPane(props: { id?: number }) {
...
+            if (!('sections' in course)) continue;
+            for (const section of getRepresentativeSections(course)) {
+                //get the lecs / everything else if no lecs
+                for (const instructor of section.instructors) {
+                    if (instructor === 'STAFF') continue; //teachers displayed as 'STAFF' have no data
+                    const key = courseGpaKey(course.deptCode, course.courseNumber, instructor);
+                    if (requested.has(key)) continue;
+                    requested.add(key);
+                    tasks.push(
+                        Grades.queryGrades(course.deptCode, course.courseNumber, instructor, true)
```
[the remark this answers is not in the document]
calebwongg: everything displayed in the search pane should already be in the cache? because theres no way that classes can get added without requerying? 

@alexespejo not sure if im misinterpreting me know your thoughts on this please :D

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
alexespejo, line 1: Uhhh something seems busted. Filter -> rows disappear -> scroll down the reappear

apps/antalmanac/src/stores/SectionFilterStore.ts
```diff
@@ -0,0 +1,20 @@
+import { create } from 'zustand';
+
+export const SORT_OPTIONS = [
+    { value: 'default', label: 'Default' },
+    { value: 'status', label: 'Status' },
+    { value: 'time_asc', label: 'Time' },
```
alexespejo, line 6: it probably isn't the most useful when the top result is TBA 

Image

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
```diff
@@ -80,6 +81,93 @@ const flattenSOCObject = (SOCObject: WebsocAPIResponse): (WebsocSchool | WebsocD
...
+    gpaMap: CourseGpaMap
+): (WebsocSchool | WebsocDepartment | AACourse)[] {
+    if (sortBy === 'default') return items;
+
+    const result: (WebsocSchool | WebsocDepartment | AACourse)[] = [];
+    let buffer: AACourse[] = [];
+
+    const flush = () => {
+        if (buffer.length === 0) return;
+        const ascending = sortBy === 'time_asc';
```
alexespejo, line 141: use enum

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
```diff
@@ -394,6 +484,50 @@ export default function CourseRenderPane(props: { id?: number }) {
         };
     }, [loadCourses, props.id]);
 
+    useEffect(() => {
+        if (sortBy !== 'gpa_descending' || courseData.length === 0) return;
```
alexespejo, line 488: there is no gpa_descending right?

alexespejo: When the user selects GPA sort, the useEffect fires and fetches grades for every course, then populates gpaMap. That map lives in component state and is never 
  explicitly cleared.

  The issue is: the map only gets rebuilt when both sortBy === 'gpa_descending' AND courseData changes. So this sequence breaks it:

  1. User searches for "CS" → selects GPA sort → map is populated with CS GPA data
  2. User switches to Time sort → map still holds the CS data in state (unused but present)
  3. User searches for "MATH" → courseData changes, but sortBy !== 'gpa_descending', so the effect early-returns and the map is not rebuilt
  4. User switches back to GPA sort → sortBy changes but courseData did NOT change, so the effect doesn't re-run → the map still has the old CS data → MATH courses
  show GPA data from CS instructors (or no data at all if the keys don't match)

  The key condition in the effect is:

  // CourseRenderPane.tsx
  if (sortBy !== 'gpa_descending' || courseData.length === 0) return;

  This means a fresh search while on a different sort option leaves the map stale. The fix would be to either clear gpaMap whenever courseData changes (regardless of
   sortBy), or include a separate effect that clears the map on courseData change.

-- Claude explained it better than me

alexespejo (changes requested):
Let's discuss at meeting I want a team discussion over how this features behaves and maybe have designers take a look at how it behaves and things we can consider further

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
calebwongg: shoot

calebwongg:
shelving pr for now. this feature doesn't provide enough value given the complexity and maintenance cost.

PR #1600, opened 2026-04-08
feat: add server-rendered SEO content for search engine indexing

KevinWu098:
## Problem

Right now, Googlebot sees an essentially **empty `<body>`** when it crawls antalmanac.com:

```html
<div hidden=""><!--$--><!--/$--></div>
<noscript>You need to enable JavaScript to run this app.</noscript>
<template data-dgst="BAILOUT_TO_CLIENT_SIDE_RENDERING"></template>
```

The `<head>` metadata (title, description, OG tags) is fine, but Google ranks pages based on **body content**, not just metadata. This is a major reason we underperform competitors like ZotCourse in search results despite having a better domain — ZotCourse serves ~200 lines of keyword-rich, server-rendered HTML (department names, headings, instructions) that Google can immediately index.

Additionally, `/robots.txt` and `/sitemap.xml` were caught by the `[[...slug]]` catch-all route and returned the app shell instead of valid responses.

## Solution

**Server-rendered SEO content** — Add a `<SeoContent />` Server Component to the page that provides real, indexable HTML in the initial response. The content is visually hidden using the standard `sr-only` accessibility pattern (same as Tailwind/Bootstrap) so it does not affect the UI, but it gives crawlers and screen readers actual text to work with:

- **`<h1>`** with "UCI Schedule Planner" (Google heavily weights heading tags)
- **Keyword-rich paragraphs** describing features: course search, schedule builder, campus map, grade distributions
- **Department and GE category mentions** woven into natural prose
- **Internal `<nav>` links** to `/added`, `/map`, `/feedback` (Google can now discover these routes without executing JS)
- **Footer** with ICSSC attribution and link

**`robots.txt` and `sitemap.xml`** — Add Next.js metadata file conventions so these URLs return proper responses instead of the app shell.

## Changes

- **`seo-content.tsx`** (new): Server Component with sr-only content for crawlers and screen readers
- **`robots.ts`** (new): Generates `/robots.txt` allowing all crawlers, pointing to sitemap
- **`sitemap.ts`** (new): Generates `/sitemap.xml` listing the homepage
- **`page.tsx`**: Renders `<SeoContent />` above `<ClientOnly />`; switched to absolute imports
- **`globals.css`**: Added `.sr-only` utility class

### DIFF
--- a/apps/antalmanac/src/app/[[...slug]]/page.tsx
+++ b/apps/antalmanac/src/app/[[...slug]]/page.tsx
@@ -1,9 +1,15 @@
-import { ClientOnly } from './client';
+import { ClientOnly } from '$src/app/[[...slug]]/client';
+import { SeoContent } from '$src/app/[[...slug]]/seo-content';
 
 export function generateStaticParams() {
     return [{ slug: [] }, { slug: ['added'] }, { slug: ['map'] }];
 }
 
 export default function Page() {
-    return <ClientOnly />;
+    return (
+        <>
+            <SeoContent />
+            <ClientOnly />
+        </>
+    );
 }
--- a/apps/antalmanac/src/app/[[...slug]]/seo-content.tsx
+++ b/apps/antalmanac/src/app/[[...slug]]/seo-content.tsx
@@ -1,0 +1,70 @@
+/**
+ * Server-rendered content for search engine indexing and screen readers.
+ *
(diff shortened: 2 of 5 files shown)


sicn4rf (approved):
lgtm! great pr, well-scoped, and well-explained :)

PR #1601, opened 2026-04-10
Change waitlist color on darkmode

### LINKED ISSUE #1560: Change color of Waitlist status to orange/yellow
The blue is a bit unintuitive and has poor contrast in dark mode.

---
*Created by <@720459930561150976> (anthony) from #github in Discord*
*[View in Discord](https://discord.com/channels/772739905981644850/893360601508741171)*

alexespejo:
## Summary
Dark mode 
image

Light mode
image

image

## Issues

Closes #1560


sicn4rf (dismissed):
i was thinking why don't we just use the `useThemeStore` hook + `isDark` so that the code can just be:

```
Waitl: {
    color: isDark ? '#f5c518' : '#1c44b2'
}
```
this way, we don't need to change the type of SxProps

but we would have to move the constant inside the component to use the hook which  isn't really a win if its just to save the type change. this approach works and the color change is what matters

LGTM 👍

sicn4rf:
wait i just realized, why don't we use the gold/yellow color for both light and dark mode? the same way the red FULL and green OPEN are readable on both light and dark mode, can't we set it to a yellow that is readable on both light and dark. that way we don't even need the `Theme` callback at all. thoughts? idk if its possible to find a readable hex for a color like gold with high luminance tho

i would just like to see consistency between light and dark mode. honestly a simple fix without bogging you down might even be to just change this one line to:
```
color: (theme) => (theme.palette.mode === 'dark' ? '#f5c518' : <some darker gold for light mode>),
 # like #B8860B or smth
```

@alexespejo

KevinWu098 (commented):
Are we updating AANTS as well?

alexespejo:
> Are we updating AANTS as well?

AANTs doesn't have any color indicators. Pretty sure :3 

CC: @IsaacNguyen

alexespejo:
@sicn4rf how does this look papi

image
image

KevinWu098:
> > Are we updating AANTS as well?
> 
> 
> 
> AANTs doesn't have any color indicators. Pretty sure :3 
> 
> 
> 
> CC: @IsaacNguyen 

It does and it's blue IIRCimage

IsaacNguyen:
Yeah you should update the color in React Email template file

alexespejo:
> > > Are we updating AANTS as well?
> > 
> > 
> > 
> > AANTs doesn't have any color indicators. Pretty sure :3 
> > 
> > 
> > 
> > CC: @IsaacNguyen 
> 
> It does and it's blue IIRCimage

Oh yeah 👍

KevinWu098:
small nit is that it would be nice if we could derive the yellow/orange from `Theme.tsx` rather than magic string it

apps/aants/src/theme.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Enrollment status text colors — matches lightTheme.enrollmentStatus in Theme.tsx.
+ * Used as the semantic color anchor; pill variants below are derived from these hues.
+ */
+export const ENROLLMENT_STATUS_COLORS = {
+    open: '#00c853',
+    waitlist: '#ff9800',
+    full: '#e53935',
+} as const;
```
sicn4rf, line 17: nit: this might be dead code, i don't see it being imported anywhere, since the actual pill styles are derived from `STATUS_PILL_STYLES` below

sicn4rf (approved):
this LGTM, maybe clean up the unused code unless it's for future use. nice clean up by stepping away from the magic strings !

PR #1608, opened 2026-04-13
feat: added courses loading skeleton

### LINKED ISSUE #1718: Add Courses Loading Skeleton

dsnsgithub:
## Summary
- Distinguish loading state from empty state through a loading skeleton for the added pane.
- Modifies the course skeleton to be closer to the new, lighter MUI skeleton components.

https://github.com/user-attachments/assets/ea60ad39-2784-4b6f-8e63-147461628912

## Issues

Fixes inability to distinguish loading state from empty state.
chrome-capture-2026-4-12
(from Kevin)

Closes #1718


KevinWu098:
How's progress here?

dsnsgithub:
> How's progress here?

I modified the colors of the calendar and course loading skeleton to be slightly lighter, so it should be ready for review.

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
```diff
@@ -134,7 +134,7 @@ export const ScheduleCalendar = memo(() => {
             end.setHours(blueprint.endHour, blueprint.endMinute, 0, 0);
 
             return {
-                color: '#6d6d6d',
+                color: '#8d8d8d',
```
KevinWu098: nit: sorry to nit again. is there any chance this could come from the MUI theme? is the default skeleton style not valid here?

apps/antalmanac/src/lib/localStorage.ts
KevinWu098, line 1: thought: man this file sucks and i wrote this file

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableLazyWrapper.tsx
```diff
@@ -1,13 +1,14 @@
...
 
+import { SectionTableSkeleton } from '$components/RightPane/LoadingSkeleton';
 import { SectionTableProps } from '$components/RightPane/SectionTable/SectionTable.types';
 
 // This should be in SectionTable.tsx, IMO
 const SectionTable = lazy(() => import('./SectionTable'));
 
 export default function SectionTableLazyWrapper(props: SectionTableProps) {
     return (
-        <Suspense fallback={<div></div>}>
```
KevinWu098: question: when does this codepath get hit? trying to understand where `SectionTableSkeleton` and `AddedCoursesLoadingSkeleton` differ

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursePane.tsx
```diff
@@ -334,6 +341,17 @@ function AddedSectionsGrid() {
         };
     }, []);
 
+    useEffect(() => {
```
KevinWu098: issue: do we need a useEffect here? can't updating localStorage be handled in the callback to `setCourses`?

apps/antalmanac/src/components/RightPane/LoadingSkeleton.tsx
```diff
@@ -0,0 +1,128 @@
+import { Box, Button, SxProps } from '@mui/material';
+
+import { getLocalStorageAddedCoursesSkeletonBlueprint } from '$lib/localStorage';
+
+export interface AddedCourseSkeletonEntry {
+    sectionCount: number;
+}
+
+// shade is #9d9d9d instead of #8d8d8d to account for gray overlay on calendar
+const shimmerSx: SxProps = {
```
KevinWu098: question: I... assume that this mirrors code from the calendar, but I just want to confirm that it is?

apps/antalmanac/src/components/RightPane/LoadingSkeleton.tsx
```diff
@@ -0,0 +1,128 @@
...
+            return parsed;
+        }
+    } catch {
+        // ignore malformed data
+    }
+
+    return null;
+}
+
+export function AddedCoursesLoadingSkeleton() {
```
KevinWu098: issue + question (repeat): still trying to figure out when each is rendered, but I'm seeing layout shift where it looks like the non-blueprint skeleton loads, then the blueprint skeleton loads... but if we have a copy in localStorage... why do we ever hit the non-blueprint?

similarly, if we have this in localStorage, why is there a separate default skeleton? if we have no blueprint, wouldn't we just render nothing / the empty state, like the calendar?

chrome-capture-2026-4-20

KevinWu098 (changes requested):
A handful of comments — feature looks good, but just hoping to flesh out the implementation a bit more

apps/antalmanac/src/components/RightPane/LoadingSkeleton.tsx
dsnsgithub: I tried to mirror the loading animation as close as I could to the calendar skeleton.

Since the calendar skeleton piggybacks off of the existing calendar system instead of using the MUI skeleton component it makes it harder.

https://github.com/icssc/AntAlmanac/blob/585d5c2abb0f08429b364c3f544c248b2220901a/apps/antalmanac/src/components/Calendar/calendar.css#L102-L120

apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
dsnsgithub: I would likely have to modify the existing calendar skeleton loading animation to match MUI skeleton, as the existing calendar skeleton doesn't actually use the [MUI skeleton component](https://mui.com/material-ui/react-skeleton/). Instead, it uses the existing calendar system with some special CSS.

I could attempt to replicate the MUI skeleton component, but if neither the calender skeleton nor the added course skeleton use it, it would mean trying to compare and replicate it with vanilla CSS and rewriting both of these skeletons (added scope).

apps/antalmanac/src/components/RightPane/SectionTable/SectionTableLazyWrapper.tsx
dsnsgithub: This is hit right when initial section table data is loaded. I believe the section table is lazy loaded because some data takes extra time to load (haven't really dug into if that is necessary or the true reason). I know that before I made this change, the loading skeleton would have some flicker as it initially uses my skeleton, then switches over to the section table lazy loading fallback, then finally renders the full table.

I modified the fallback so that I could use the section table lazy loading as part of my skeleton.

apps/antalmanac/src/components/RightPane/LoadingSkeleton.tsx
dsnsgithub: Since my skeleton doesn't store the full names of the classes, there will still be some layout shift as I assume the length of the course.

I will try to see if I can use localStorage with the entire schedule similar to the calendar skeleton loader.

> similarly, if we have this in localStorage, why is there a separate default skeleton? if we have no blueprint, wouldn't we just render nothing / the empty state, like the calendar?

I think that in attempting to "distinguish loading state from empty state" we need to signify that there might be a schedule that is not in localStorage (because it is being pulled from AntAlmanac), so a default skeleton would still make sense here.

apps/antalmanac/src/lib/localStorage.ts
dsnsgithub: localStorage is a necessary evil for this skeleton 🥀

KevinWu098:
i promise ill take a look at this soon, in the meanwhile do you mind handling the conflicts

apps/antalmanac/src/components/RightPane/AddedCourses/AddedCoursesLoadingSkeleton.tsx
```diff
@@ -0,0 +1,107 @@
...
+            const entries = parsed.flatMap((entry): AddedCourseSkeletonEntry[] => {
+                if (typeof entry?.height !== 'number' || entry.height <= 0) return [];
+                return [
+                    {
+                        deptCode: typeof entry.deptCode === 'string' ? entry.deptCode : '',
+                        courseNumber: typeof entry.courseNumber === 'string' ? entry.courseNumber : '',
+                        courseTitle: typeof entry.courseTitle === 'string' ? entry.courseTitle : '',
+                        height: entry.height,
+                    },
+                ];
```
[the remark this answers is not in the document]
dsnsgithub: There are no existing users for the added courses loading skeleton, it really doesn't matter.

apps/antalmanac/src/components/RightPane/AddedCourses/AddedSectionsGrid.tsx
```diff
@@ -159,7 +189,9 @@ export function AddedSectionsGrid() {
             <Box sx={{ marginTop: 7 }}>
                 <Typography variant="h6">{`${scheduleName} (${scheduleUnits} Units)`}</Typography>
                 {isMobile && <SelectSchedulePopover />}
-                {courses.length === 0 && (
+                {loadingSchedule ? (
+                    <AddedCoursesLoadingSkeleton />
```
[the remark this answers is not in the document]
dsnsgithub: should be fixed with https://github.com/icssc/AntAlmanac/commit/e1fc1f1c00f42274ad475c5c003a1efb580945b8?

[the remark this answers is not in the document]
dsnsgithub: Reverting these changes, as creating this skeleton on first load creates flickering as the skeleton is likely to be replaced with the "No Courses Added Yet" message.

dsnsgithub:
@cubic review this PR

dsnsgithub:
@cubic-dev-ai review

vicksey (commented):
LGTM. Tested on staging the skeleton loads and smoothly transitions to the sections when the data loads. 

I just noticed 1 small thing for consistency sake: custom events don't have a skeleton on the added pane. I think it's worth adding since on the calendar pane custom events load with a skeleton.

Besides that nice pr!

dsnsgithub:
Custom events should now also have a skeleton (only when the user has any in their schedule).

vicksey (approved):
LGTM 🚀 🔥 ✅

PR #1611, opened 2026-04-14
Migrate to better auth

### LINKED ISSUE #1578: Migrate to better-auth
Use better-auth instead of implementing Google authentication manually to simplify code and hopefully clear any bugs that have to do with auth, like some auto-save stuff.

Choollol:
## Summary

Migrate auth from manual implementation to using better-auth.

Notes:
- The ~900 line change in `pnpm-lock.yaml` comes from better auth being installed and drizzle-kit being updated
- ~1000 added lines are from drizzle-kit's snapshot.json
- There are a lot of small formatting changes that were caused by the move from eslint to oxlint

## Test Plan

- Signing in works with both Google and Apple and schedules are loaded
- Schedules can be saved as before
- Signing out works and tells auth.icssc that the user signed out
- Return URL works (things like url search params are kept after signing in)
- Planner features like search with planner still work
- Analytics are sent correctly
- iOS testflight auth works

## Issues

Closes #1578


KevinWu098:
> AutoSignIn no longer passes prompt: none because https://github.com/better-auth/better-auth/issues/2351

Isn't... this a problem? this would defeat the purpose of auto sign in, if... we can't... automatically signin without a prompt

Choollol:
> > AutoSignIn no longer passes prompt: none because [better-auth/better-auth#2351](https://github.com/better-auth/better-auth/issues/2351)
> 
> Isn't... this a problem? this would defeat the purpose of auto sign in, if... we can't... automatically signin without a prompt

Yeahh, do you think it'd be better to get rid of `AutoSignIn` until better auth lets us pass different prompts?

KevinWu098:
> > > AutoSignIn no longer passes prompt: none because [better-auth/better-auth#2351](https://github.com/better-auth/better-auth/issues/2351)
> > 
> > 
> > Isn't... this a problem? this would defeat the purpose of auto sign in, if... we can't... automatically signin without a prompt
> 
> Yeahh, do you think it'd be better to get rid of `AutoSignIn` until better auth lets us pass different prompts?

No, i think the opposite / we should have a workaround. breaks the scheduler <> planner flow quite a bit having to sign in twice

apps/antalmanac/src/lib/auth/authClient.ts
```diff
@@ -0,0 +1,52 @@
...
+        const { logoutUrl } = await trpc.userData.getLogoutUrl.query({
+            redirectUrl: window.location.origin,
+        });
+
+        window.location.href = logoutUrl;
+    } catch (error) {
+        console.error('Error during logout', error);
+    }
+}
+
```
sicn4rf, line 45: blocking: federated sign-out is broken here. `clearSession()` calls `window.location.reload()`, which schedules a page reload. the subsequent `trpc.userData.getLogoutUrl.query(...)` fires an XHR, but the execution yields at await, the browser processes the pending reload, and the `window.location.href` is never reached.

i verified this against prod, prod hits `auth.icssc.club/logout?` and returns a 302 before redirecting back, but staging hits better-auth's local `sign-out` endpoint and reloads. the `auth.icssc.club` IdP session stays alive and google OAuth consent is still granted. 

this is crucial for shared devices, if a user "signs out", the next person will silently sign in as them when they click sign in.

a potential fix might be to fetch `logoutUrl` first, and then clear the local session, then redirect.

(+1 to Cubic's P1, confirmed liva via prod and staging comparison)

apps/antalmanac/src/components/AuthInitializer.tsx
```diff
@@ -0,0 +1,151 @@
...
+                    id: googleId,
+                    userData: scheduleSaveState,
+                },
+            });
+
+            removeLocalStorageDataCache();
+
+            openSnackbar('success', `Unsaved changes have been saved to your account!`);
+        }
+    }, [sessionData, googleId]);
```
sicn4rf: issue: this `handleUnsavedChanges` `useCallback` has `googleId` in the deps. And the `useEffect` below this depends on `handleUnsavedChanges`, so when `updateSession(sessionData)` gets `googleId` from zustand, the component re-renders and `useCallback` recomputes and the `useEffect` fires again.

this means every `authClient.useSession()` revalidation that happens on every tab focus by default in better-auth triggers:
- full auth flow re-run
- `handleUnsavedChanges` re-execution
- `loadScheduleAndSetLoadingAuth` re-execution, loading skeleton + "Schedule loaded." snackbar
-  `loadNotifications` re-execution

users will see the loading skeleton and snackbar on every tab switch

potentially fixable by reading `googleId` imperatively in the callback itself instead of declaring it as a dep (?)

apps/antalmanac/next.config.mjs
```diff
@@ -12,6 +12,15 @@ const nextConfig = {
     },
     serverExternalPackages: ['@node-rs/argon2'],
     turbopack: {},
+    async redirects() {
+        return [
+            {
+                source: '/auth',
+                destination: '/api/auth/oauth2/callback/icssc',
+                permanent: true,
```
sicn4rf: suggestion: `permanent: true` produces a 301 that the browser caches. If this endpoint needs to change or roll back later, users with the cached 301s will keep following this stale path even though the source code destination has been changed. consider `permanent: false`

sicn4rf (changes requested):
thanks for taking this PR on, migrating to better-auth is the right move and the happy path seems to work correctly for both new and existing users. schema migration applies cleanly for my real account.

however, after smoke testing against both prod and this PRs staging, i found a couple issues.
summary:

**blocking**
1. sign-out regression - the IdP session at `auth.icssc.club` is never invalidated because `clearSession()` fires `window.location.reload()` before the `getLogoutUrl` redirect runs. Prod visits `auth.icssc/logout?post_logout..` on sign-out but staging doesn't when i compared network logs of both flows.

2. `AuthInitializer` useEffect cascades on every single tab focus which causes redundant API calls and a loading skeleton flash on every tab switch

Choollol:
@sicn4rf Thanks for catching these problems! I've (hopefully) fixed them, along with some extra changes to optimize and protect against stuff like auth race conditions.

Also btw, I still need to test this with PWA and probably make some changes too.

Choollol:
@sicn4rf I think this PR is pretty much ready except for testing PWA auth. Would it be possible to deploy PWA on this PR's staging instance so we could test it?

sicn4rf:
> @sicn4rf I think this PR is pretty much ready except for testing PWA auth. Would it be possible to deploy PWA on this PR's staging instance so we could test it?

CC @KevinWu098 is this possible? the PWA auth can probably only be tested against `antalmanac.com` and not this staging deployment right? so like we can prob test it after this merges using TestFlight pointed at prod. wouldn't we need to modify the xcode project to point at `schedluer-1611.antalmanac.com`to test PWA auth against this PRs staging? I'm not sure how this works

KevinWu098:
> > @sicn4rf I think this PR is pretty much ready except for testing PWA auth. Would it be possible to deploy PWA on this PR's staging instance so we could test it?
> 
> CC @KevinWu098 is this possible? the PWA auth can probably only be tested against `antalmanac.com` and not this staging deployment right? so like we can prob test it after this merges using TestFlight pointed at prod. wouldn't we need to modify the xcode project to point at `schedluer-1611.antalmanac.com`to test PWA auth against this PRs staging? I'm not sure how this works

uhhh, i... think it should be okay... but someone should ask an AI...

sorry i forgot about this one. someone ping me when they think its in a stable state to be tested on PWA

Choollol:
@cubic-dev-ai review

Choollol:
@sicn4rf Ready for review I think

Choollol:
@cubic-dev-ai review

packages/db/src/schema/auth/account.ts
```diff
@@ -19,22 +20,26 @@ export type AccountType = (typeof accountTypes)[number];
 export const accounts = pgTable(
     'accounts',
     {
+        id: text('id').$defaultFn(createId),
```
sicn4rf: blocking: +1 on cubic here, i think `accounts.id` needs to be treated as a real better-auth account identifier. `accounts.id` is nullable, unbackfilled, and not unique/primary even though better-auth exptects `accounts.id` to be the row identifier. this means it has to, at the bare minimum, be unique. 

`$defaultFn(createId` will only run through drizzle inserts and does not backfill on existing rows, its not a postgres DB level default. existing accounts will end up with NULL `id`s, and new ids will not be enforced to be unique at the database level. 

better-auths drizzle schema expects `account.id` to be the primary key

better-auth’s database docs describe the account model as having its own unique `id`, separate from provider/account identifiers, and the drizzle adapter docs recommend generating/migrating the schema from the better-auth config:
https://www.better-auth.com/docs/concepts/database
https://www.better-auth.com/docs/adapters/drizzle#schema-generation--migration

apps/antalmanac/src/stores/SessionStore.ts
```diff
@@ -1,85 +1,59 @@
...
-                isGoogleUser: false,
-                email: null,
-                name: null,
-                avatar: null,
+                session: sessionData.session,
+                sessionId: sessionData.session.id,
+                sessionIsValid: true,
+                user: sessionData.user,
+                userId: sessionData.user.id,
+                isGoogleUser: true,
```
sicn4rf: suggestion: `isGoogleUser` now appears to mean “authenticated user” rather than literally Google user.

this PR adds Apple sign-in too, better-auth’s account model distinguishes provider/account identity separately:
https://www.better-auth.com/docs/concepts/database

since this flag gates notifications and other signed-in behavior, could we rename it to something provider-neutral like `isAuthenticatedUser` / `canUseNotifications`, or derive it from the actual capability we need? maybe?

apps/antalmanac/src/components/SignInAlertDialog.tsx
```diff
@@ -0,0 +1,22 @@
...
+    open: boolean;
+    title: string;
+    severity?: ComponentProps<typeof AlertDialog>['severity'];
+    onClose: () => void;
+}
+
+export const SignInAlertDialog = ({ open, title, severity = 'info', onClose }: Props) => {
+    return (
+        <AlertDialog open={open} onClose={onClose} title={title} severity={severity}>
+            <DialogContentText>To load your schedule sign in with your Google account</DialogContentText>
```
sicn4rf: nit: this copy still says “Google account,” but the dialog now renders both Google and Apple sign-in buttons.

better-auth now supports multiple providers in this flow, so maybe use provider-neutral copy like “To load your schedule, sign in to your account.”?

apps/antalmanac/next.config.mjs
```diff
@@ -12,6 +12,15 @@ const nextConfig = {
     },
     serverExternalPackages: ['@node-rs/argon2'],
     turbopack: {},
+    async redirects() {
+        return [
+            {
+                source: '/auth',
```
sicn4rf, line 18: question: can we confirm  somehow the native iOS/PWA callback still works with the better-auth callback route?

`NATIVE_IOS_REDIRECT_URI` is still `https://antalmanac.com/auth/native` at `apps/antalmanac/src/lib/platform.ts:25`, while `next.config.mjs` only redirects `/auth` to better-auth’s callback. i see the Swift wrapper strips `/native` before loading the web view, so this may be fine, but better-auth's OAuth callback flow is now different from the removed `AuthPage.tsx` flow:
https://www.better-auth.com/docs/concepts/oauth

since we can’t test PWA auth on this staging deployment, can we confirm the `/auth/native -> /auth -> /api/auth/oauth2/callback/icssc` path still works in the native wrapper?

CC @KevinWu098

sicn4rf (changes requested):
this is looking good, just one blocking issue to address and a few suggestions that incorporate the new Apple/PWA auth paths.

sicn4rf (dismissed):
core web auth looks stable to me now. i re-smoke-tested Google/Apple sign-in, schedule save/load, sign-out, return URL preservation, planner search, and analytics, and the follow-up fixes addressed my blocking schema concern plus the provider-neutral cleanup.

lgtm, i'm approving from the web app side.

CC @KevinWu098 this PR looks stable enough for PWA auth testing now. the remaining validation d’d like before merge is confirming the native `/auth/native -> /auth -> /api/auth/oauth2/callback/icssc` flow still works in the iOS wrapper.

Choollol:
@sicn4rf Thanks for the detailed comments and feedback! Should I hold off on merging until we can fully test PWA auth and make sure everything works?

sicn4rf:
> @sicn4rf Thanks for the detailed comments and feedback! Should I hold off on merging until we can fully test PWA auth and make sure everything works?

yeah, id hold off until we can do the PWA auth smoke test.

i would wait until kevin responds on this

KevinWu098:
I'll take a look at this tomorrow and deploy a testflight. please ping me if i dont get around to it

Choollol:
@sicn4rf Auth seems to work fine on PWA (https://testflight.apple.com/join/73RdQPKy). Sessions aren't invalidated but that bug is present in earlier builds too.

apps/antalmanac/src/lib/auth/authActions.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export async function getSignInUrl(
+    provider: Provider,
+    { authorizationUrlParams, returnUrl }: GetSignInUrlOptions = {}
+) {
+    const { url } = await auth.api.signInWithOAuth2({
+        body: {
+            providerId: AUTH_PROVIDER_ID,
+            newUserCallbackURL: '/welcome',
+            callbackURL: isNativeIosApp() ? getNativeIosRedirectUri(BETTER_AUTH_URL) : undefined,
```
sicn4rf: question: does this native iOS check still work now that `getSignInUrl` is a server action?

since this file is marked `'use server'`, `isNativeIosApp()` can't actually read `document.cookie` here and should always return false. so this means the `/auth/native` callback branch is probably dead code now.

i know TestFlight auth passed, so maybe we no longer need the native callback override with better-auth, but if that's the case should we remove this branch? otherwise, if `/auth/native` is still required, should this check use server-readable cookies/headers instead?

apps/antalmanac/src/components/AuthInitializer.tsx
```diff
@@ -0,0 +1,170 @@
...
+        // Clean up stale localStorage token from before the cookie migration
+        window.localStorage.removeItem('sessionId');
+
+        if (sessionData) {
+            (async () => {
+                if (sessionData.session.expiresAt < new Date()) {
+                    setOpenAlert(true);
+                    setHasCheckedAuth(true);
+                    return;
+                }
```
sicn4rf, line 123: issue: i think this expired-session branch can leave the schedule loading skeleton open.

`openLoadingSchedule` now defaults to true, and the normal auth paths call `handleAuthChecked()` to set it back to false. but this branch sets the alert and `hasCheckedAuth`, then returns before calling `handleAuthChecked()`.

could we call `handleAuthChecked()` before this return so users with an expired session don't get stuck in the loading state?

sicn4rf (changes requested):
few questions about the new code changes. one more thing, did u test pwa before or after making these changes that brought up a couple issues?

Choollol:
@sicn4rf Thanks for the catches! I tested the iOS testflight and website before and after and it seemed to work fine. I think the bug was hidden because /auth/native redirects to the better auth callback url anyway.

sicn4rf (approved):
fire lgtm, im approving now.

the latest changes addressed my concerns: the native iOS check now runs client-side before calling the server action, and the expired-session path now clears the loading state before returning. PWA/TestFlight auth has also been smoke tested, with the remaining session invalidation behavior noted as pre-existing.

PR #1619, opened 2026-04-15
revert: SST 3.19.3 upgrade to fix I&C SCI search

sicn4rf:
## Summary
- Reverts the SST upgrade from 3.17.25 to 3.19.3 introduced in #1598
- SST 3.19.3 silently bumps OpenNext from 3.6.6 to 3.9.14 (due to Next.js 16 version detection), which introduces a bug where URL-encoded query parameter values are decoded but not re-encoded when reconstructing `req.url`, breaking any request with `&` in query values
- This broke all searches for the `I&C SCI` department (fuzzy search and manual dropdown)

## Context
See the full root cause analysis in the [comment on opennextjs/opennextjs-aws#1133](https://github.com/opennextjs/opennextjs-aws/issues/1133#issuecomment-4248434215).

Once OpenNext fixes the underlying issue, we can re-upgrade SST.


KevinWu098 (approved):
Really nice investigation, thanks for tackling this p0

PR #1632, opened 2026-04-17
feat: protect remaining authenticated tRPC procedures

KevinWu098:
## Summary

Second-pass audit of the tRPC routers after the `kwu/protected-auth` migration. Locks down the procedures that were still unprotected, removes now-redundant caller-supplied identifiers, and consolidates the guest-schedule read surface so it's safe by construction rather than by application-layer guard.

## What changed — at a glance

| Area | Before | After |
| --- | --- | --- |
| `userData.saveUserData` | `procedure`, takes `{ id, data: User }` | `protectedProcedure`, takes `{ userData: ScheduleSaveState }`; owner derived from `ctx.userId` |
| `userData.flagImportedSchedule` | `procedure` | `protectedProcedure`; input renamed `providerId` → `username` |
| `userData.getUserByUid` | `procedure`, input `{ userId }` | Renamed to `getCurrentUser`; `protectedProcedure`, no input |
| Public guest-read surface (3 endpoints) | `getUserData` + `getAccountByProviderId` + `getGuestAccountAndUserByName` | Collapsed into one: `getGuestScheduleByUsername` |
| `loadSchedule` / `loadScheduleWithSessionToken` | username-based was primary, session-based was "WithSessionToken" | Swapped: session-based is `loadSchedule`, legacy is `loadGuestSchedule` |

> `notifications.updateNotifications` is intentionally **not** included in this PR. Protection alone isn't a real fix — the endpoint performs a cross-user write that has pre-existing behavioral coupling with the notification cron. Tracked separately.

## 1. Additional protected procedures

Three endpoints that accepted or mutated per-user state without authentication:

- **`userData.saveUserData`** — accepted `{ id, data }` from any client. The server trusted `id` to decide whose schedules to overwrite. Now:
  ```ts
  saveUserData: protectedProcedure.input(saveInputSchema).mutation(async ({ input, ctx }) => {
      const userData = input.userData;
      userData.schedules = mangleDuplicateScheduleNames(userData.schedules);
      return await RDS.upsertUserData(db, ctx.userId, userData);
  })
  ```
  The `RDS` layer matched: `upsertUserData(db, userId, saveState)` — writes schedules by internal id, no account re-registration (that happens exclusively in `handleGoogleCallback` now).

- **`userData.flagImportedSchedule`** — was flippable by anyone for any guest account. Now `protectedProcedure`.

- **`userData.getUserByUid`** — took `{ userId }` and returned the full `users` row (`email`/`name`/`avatar`/…). Renamed to `getCurrentUser`, no input, derives from `ctx.userId`.

## 2. Guest-schedule read surface, consolidated

Three overlapping public endpoints replaced with one:

```ts
// Before: three hops, one of which took a client-supplied CUID that
//         needed an application-layer "is this actually GUEST?" guard.
getUserData:                 procedure.input({ userId }).query(...)
getAccountByProviderId:      procedure.input({ accountType, providerId }).query(...)
getGuestAccountAndUserByName: procedure.input({ name }).query(...)

// After: one hop, sole gate is a SQL join filter.
getGuestScheduleByUsername: procedure
    .input(z.object({ username: z.string() }))
    .query(async ({ input }) => {
        const result = await RDS.getGuestScheduleByUsername(db, input.username);
        if (!result) throw new TRPCError({ code: 'NOT_FOUND', ... });
        return result;
    })
```

The RDS implementation has one query:

```ts
const row = await tx
    .select()
    .from(accounts)
    .innerJoin(users, eq(accounts.userId, users.id))
    .where(and(eq(accounts.accountType, 'GUEST'), eq(accounts.providerAccountId, username)))
    .limit(1);
```

**Why this is safe by construction:**
- The endpoint no longer accepts a user id of any kind. The only input is a username interpreted as a GUEST `providerAccountId`.
- The `WHERE accountType = 'GUEST'` clause is enforced by Postgres, not JavaScript. There is no input the client can supply that makes the query return a row for a user who doesn't have a GUEST account.
- OIDC users are structurally unreachable: their `accounts` row has `accountType = 'OIDC'`, which fails the filter.
- Return shape is trimmed: `{ user: { imported }, userData: { schedules, scheduleIndex } }`. `email`/`name`/`avatar`/`phone` from the `users` row are not exposed to public viewers.

**What about dual-account users (former guest, linked via OIDC)?** The guest username still resolves — this is intentional. Guest share URLs are a product promise and must keep working after the owner signs in with Google. The `users.imported` flag lets the UI warn about duplicate imports without changing the public data's availability.

## 3. Cleanup enabled by auth-derived identity

Once `ctx.userId` is authoritative, several pieces of caller plumbing become dead weight:

- `saveSchedule(rememberMe, postHog?)` and `autoSaveSchedule(options)` no longer take a `providerId`. It only ever served as an analytics label and as a gate on an unreachable "Schedule saved under username …" snackbar branch.
- Dropped the `getUserAndAccountBySessionToken` round-trips in `Save.tsx`, `ActionTypesStore`, and `handleScheduleImport` that existed only to fetch `providerAccountId`.
- `ExperimentalMenu` no longer consults `localStorage` userId before auto-saving; the session answers that directly.
- `AuthPage` dropped its separate `getUserByUid` profile fetch before `saveUserData`, and switched its post-callback self-read from `getUserData` to `getUserDataWithSession`.
- `loadScheduleWithSessionToken` renamed to `loadSchedule` (now the primary path); legacy guest-username loader renamed to `loadGuestSchedule`.
- `RDS.upsertUserData` no longer re-runs `registerUserAccount` inside a save. That function is exclusively called from `handleGoogleCallback`, which is where sign-up / email-link-to-existing-guest actually belongs.

## Endpoints deliberately left public

- `auth.validateSession`, `auth.getSessionUserId` — return `false` / `''` when logged out; callers depend on this.
- `userData.getGoogleAuthUrl`, `handleGoogleCallback`, `logout` — the auth/logout surface itself, must run without a session.
- `userData.getGuestScheduleByUsername` — the one remaining public guest-read endpoint; GUEST-only by SQL filter.
- `notifications.deleteNotification` / `deleteAllNotifications` — email unsubscribe links (comment retained inline).
- `course`, `enrollHist`, `grades`, `search`, `websoc`, `zotcourse` — public catalog data, no per-user state.

## Test Plan

- `pnpm lint` passes.
- `tsc --noEmit` on `apps/antalmanac` produces no new errors (only pre-existing unrelated ones from missing generated files / image asset modules).
- Every frontend caller of every touched endpoint was traced manually to confirm the call path is either inside an authenticated session or operating on a legacy guest username.
- Guest-schedule-by-username flow exercised end-to-end mentally across the three shapes (guest-only / OIDC-only / dual-account).

## Issues

Closes #

<div><a href="https://cursor.com/agents/bc-579ec9cb-bb4c-44e6-bb9c-01ac246b0863"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-579ec9cb-bb4c-44e6-bb9c-01ac246b0863"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


apps/antalmanac/src/backend/routers/userData.ts
```diff
@@ -295,34 +250,23 @@ const userDataRouter = router({
...
+                code: 'INTERNAL_SERVER_ERROR',
+                message: 'Failed to save user data',
+            });
+        }
     }),
 
-    flagImportedSchedule: procedure.input(z.object({ providerId: z.string() })).mutation(async ({ input }) => {
-        return await RDS.flagImportedUser(db, input.providerId);
+    flagImportedSchedule: protectedProcedure.input(z.object({ username: z.string() })).mutation(async ({ input }) => {
+        return await RDS.flagImportedUser(db, input.username);
```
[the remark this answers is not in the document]
KevinWu098: That's fine, this is just something that we allow on a guest user

apps/antalmanac/src/backend/routers/userData.ts
```diff
@@ -14,21 +14,11 @@ import { z } from 'zod';
...
-    /**
-     * Schedule data being saved.
-     *
-     * The ID of the requester and user ID in the schedule data may differ,
-     * i.e. if the user is editing and saving another user's schedule.
+     * Schedule data being saved
      */
-    data: z.custom<User>(),
+    userData: z.custom<ScheduleSaveState>(),
 });
```
Choollol: Small nitpick: This can probably just be inlined

apps/antalmanac/src/backend/routers/userData.ts
Choollol: Could consider renaming to `getUserAndAccount` since how it gets the data is hidden now

Choollol: Could consider renaming to `getGoogleId` since how it gets the data is hidden now

apps/antalmanac/src/backend/routers/userData.ts
```diff
@@ -66,48 +49,20 @@ const userDataRouter = router({
...
-            const account = await RDS.getAccountByProviderId(db, input.accountType, input.providerId);
-            if (!account) {
-                throw new TRPCError({
-                    code: 'NOT_FOUND',
-                    message: `Couldn't find schedules for username "${input.providerId}".`,
-                });
-            }
-            return account;
-        }),
+    getUserDataWithSession: protectedProcedure.query(async ({ ctx }) => {
```
Choollol: Could consider renaming to `getUserData` since how it gets the data is hidden now

apps/antalmanac/src/backend/routers/userData.ts
```diff
@@ -314,57 +269,42 @@ const userDataRouter = router({
...
      * Logs out a user by invalidating their session and redirecting to OIDC logout
      */
     logout: procedure.input(z.object({ redirectUrl: z.string().optional() })).mutation(async ({ input, ctx }) => {
+        if (ctx.sessionToken) {
+            const session = await RDS.getCurrentSession(db, ctx.sessionToken);
+            if (session) {
+                await RDS.removeSession(db, session.userId, session.refreshToken);
+            }
+        }
+
```
Choollol: What's the reasoning behind not catching errors here anymore?

apps/antalmanac/src/actions/ActionTypesStore.ts
```diff
@@ -118,12 +117,14 @@ class ActionTypesStore extends EventEmitter {
         }
 
         if (autoSave) {
-            const { users, accounts } = await trpc.userData.getUserAndAccountBySessionToken.query();
+            this.emit('autoSaveStart');
 
-            if (accounts.providerAccountId) {
-                this.emit('autoSaveStart');
-                await autoSaveSchedule(accounts.providerAccountId, { userInfo: users });
+            try {
```
Choollol: Can `autoSaveSchedule` throw any errors? If it does, it should probably be handled in the function itself because other errors are already being handled in it. Either way, I don't think we need to wrap the call in a try-catch.

apps/antalmanac/src/backend/lib/rds.ts
```diff
@@ -1,27 +1,34 @@
...
 } from '@packages/db/src/schema';
-import { and, eq, ExtractTablesWithRelations, gt } from 'drizzle-orm';
-import { PgTransaction, PgQueryResultHKT } from 'drizzle-orm/pg-core';
+import { and, eq, type ExtractTablesWithRelations, gt } from 'drizzle-orm';
+import type { PgTransaction, PgQueryResultHKT } from 'drizzle-orm/pg-core';
 
 type Transaction = PgTransaction<PgQueryResultHKT, typeof schema, ExtractTablesWithRelations<typeof schema>>;
 type DatabaseOrTransaction = Omit<typeof db, '$client'> | Transaction;
 
+// biome-ignore lint/complexity/noStaticOnlyClass: todo
```
Choollol: Do we use biome? `settings.json` says it's disabled.

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -180,8 +137,8 @@ export async function autoSaveSchedule(providerID: string, options: AutoSaveSche
         deleteTempSaveData();
         AppStore.saveSchedule();
     } catch (e) {
-        if (e instanceof TRPCError) {
-            openSnackbar('error', `Schedule could not be auto-saved under username "${providerID}`);
+        if (e instanceof TRPCClientError) {
+            openSnackbar('error', 'Schedule could not be auto-saved');
         } else {
             openSnackbar('error', 'Network error or server is down.');
```
Choollol, line 143: We should say something about auto-save failing in the snackbar:
```suggestion
            openSnackbar('error', 'Schedule auto-save failed: Network error or server is down.');
```
Feel free to tweak the message or do something different.

apps/antalmanac/src/actions/AppStoreActions.ts
```diff
@@ -80,97 +78,56 @@ export function isEmptySchedule(schedules: ShortCourseSchedule[]) {
...
-                }
-            }
+        openSnackbar('success', `Schedule saved. Don't forget to sign up for classes on WebReg!`);
+        deleteTempSaveData();
+        AppStore.saveSchedule();
+    } catch (e) {
+        if (e instanceof TRPCClientError) {
+            openSnackbar('error', `Schedule could not be saved`);
+        } else {
+            openSnackbar('error', 'Network error or server is down.');
```
Choollol, line 115: We should say something about save failing in the snackbar:
```suggestion
            openSnackbar('error', 'Schedule save failed: Network error or server is down.');
```
Feel free to tweak the message or do something different.

Choollol (changes requested):
Generally looks good, some small questions and suggestions

apps/antalmanac/src/backend/routers/userData.ts
KevinWu098: not sure what you're referring to? i don't think so since we're mixing arktype and zod (which, i realize, we should write an issue to migrate off of arktype)

apps/antalmanac/src/backend/lib/rds.ts
KevinWu098: removed. wasn't the case when i first started this pr

apps/antalmanac/src/actions/AppStoreActions.ts
KevinWu098: some other PR, perhaps? 🥺

apps/antalmanac/src/actions/AppStoreActions.ts
KevinWu098: some other PR, perhaps? 🥺

apps/antalmanac/src/backend/routers/userData.ts
KevinWu098: eh, i (or cursor) must've had a reason, but can't figure it out now. reverted

apps/antalmanac/src/backend/routers/userData.ts
Choollol: Sorry, I meant that we could get rid of `saveInputSchema` and just put the zod schema in the same place as where it's used.

KevinWu098: ah yeah, its an arktype schema, so doesn't mix proper

Choollol: Oh huh, is it an issue that only happens at runtime? TypeScript doesn't complain about this:
```ts
    saveUserData: protectedProcedure
        .input(z.object({ userData: z.custom<ScheduleSaveState>() }))
        .mutation(async ({ input, ctx }) => {
```

Choollol (dismissed):

apps/antalmanac/src/backend/routers/userData.ts
KevinWu098: ohhh, fixed

Choollol (approved):
github could use an improvement in their whitespace diffing haha

PR #1637, opened 2026-04-19
Calendar skeleton dayOffset calculation breaks for Sunday courses

### LINKED ISSUE #1631: Calendar skeleton doesn't handle sunday properly
https://cdn.discordapp.com/attachments/813958408495300609/1494384007809470534/Screen_Recording_2026-04-17_at_2.08.20_AM.mov?ex=69e2690f&is=69e1178f&hm=60ae256c21d96098c3450d443b37df44c535f5436f5324742bbfb5255983b5bd&

---
*Created by <@227113377145552896> (Kevin 💖) from #antalmanac-uiux in Discord*
*[View in Discord](https://discord.com/channels/772739905981644850/813958408495300609)*

brelieu05:
## Summary
- Sunday courses are stored on new Date(2018, 0, 0) = Dec 31, 2017 in calendarizeHelpers.ts                                
- The old skeleton blueprint calculation used event.start.getDate() - BASE_DATE.getDate() which returned 31 - 1 = 30 for Sunday events, failing the >= 0 && <= 6 filter and dropping Sunday entirely from the saved skeleton                        
- Fixed by using event.start.getDay() - 1 instead (Sun→-1, Mon→0, …, Sat→5) and updating the filter bounds to >= -1 && <= 5

image

## Test Plan
- Add a Sunday course to your schedule                                                                                     
- Reload the page — the loading skeleton should show an event in the Sunday column
- Verify Mon–Sat courses still render correctly in the skeleton

## Issues

Closes #1631

### DIFF
--- a/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
+++ b/apps/antalmanac/src/components/Calendar/CalendarRoot.tsx
@@ -23,7 +23,7 @@ import { useThemeStore, useTimeFormatStore } from '$stores/SettingsStore';
 import { useTabStore } from '$stores/TabStore';
 import { CalendarMonth } from '@mui/icons-material';
 import { Box, Backdrop, useTheme } from '@mui/material';
-import { format, getDay, startOfWeek, type Locale } from 'date-fns';
+import { differenceInCalendarDays, format, getDay, startOfWeek, type Locale } from 'date-fns';
 import { enUS } from 'date-fns/locale';
 import { memo, useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { Calendar, Components, DateLocalizer, dateFnsLocalizer, Views, ViewsProps } from 'react-big-calendar';
@@ -97,7 +97,7 @@ export const ScheduleCalendar = memo(() => {
                 hasHadEventsRef.current = true;
                 const skeletonBlueprint = eventsInCalendar
                     .map((event) => {
-                        const dayOffset = event.start.getDay() - 1;
+                        const dayOffset = differenceInCalendarDays(event.start, BASE_DATE);
                         return {
                             dayOffset,
                             startHour: event.start.getHours(),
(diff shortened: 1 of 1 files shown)


KevinWu098 (changes requested):
I smelled something fishy with this solution (why do we have -1?), so I asked cursor to dive into it a bit:

---------

The fix works, but the - 1 is load-bearing in a non-obvious way. BASE_DATE = new Date(2018, 0, 1) is a Monday, and reconstruction does BASE_DATE + dayOffset, so dayOffset here means "days relative to Monday". Subtracting 1 from getDay() (which is 0=Sun..6=Sat) re-centers it on Monday, giving Sun=-1..Sat=5. This convention is preserved to stay compatible with skeletonBlueprintVariations.ts (which uses 0=Mon) and any existing localStorage blueprints.

The underlying bug isn't really Sunday-specific — it's that getDate() - BASE_DATE.getDate() does day-of-month arithmetic, which breaks across the Dec/Jan boundary that Sunday events roll back over.

Two cleaner alternatives worth considering:

A. Use differenceInCalendarDays (minimal change). date-fns is already imported. Replace the expression with:


const dayOffset = differenceInCalendarDays(event.start, BASE_DATE);
Same values (-1..5), same filter, but it directly expresses what the reconstruction code consumes ("days from BASE_DATE") instead of relying on the coincidence that BASE_DATE.getDay() === 1. If BASE_DATE ever changes, getDay() - 1 silently breaks; this doesn't.

B. Normalize on 0=Sun..6=Sat to match calendarizeHelpers.ts. Use event.start.getDay() directly (filter >= 0 && <= 6) and reconstruct with new Date(2018, 0, blueprint.dayOffset, ...), mirroring how events are originally built in calendarizeCourseEvents. Forward and reverse transforms become symmetric and no negative offsets. Costs: shift skeletonBlueprintVariations.ts by +1, and existing users' stored blueprints render off-by-one for one session before being overwritten (mitigable with a versioned localStorage key).

--------

I think A is probably fine here

KevinWu098 (approved):

PR #1615, opened 2026-04-14
refactor: replace moment with date-fns for calendar localizer

KevinWu098:
## Summary

Replace `moment` (~300kB minified) with `date-fns` (already in the project) across all three files that imported it:

**CalendarRoot.tsx** — the primary change:
- Switch from `momentLocalizer` to `dateFnsLocalizer` from `react-big-calendar`
- Replace `moment.defineLocale()` calls with date-fns `Locale` objects that set `weekStartsOn` (0 for normal schedules, 6 for finals)
- Convert moment format tokens to date-fns equivalents (`'h:mm A'` → `'h:mm a'`, `'ddd'` → `'eee'`, `'M/DD'` → `'M/dd'`)

**termData.ts** — enrollment deadline calculation:
- Replace `moment().diff(..., 'week')` with `differenceInWeeks()`
- Replace `moment().add().day()` chain with `addWeeks()` + `setDay()`
- Replace `moment() <=` with native `new Date() <=` to preserve the original inclusive semantics

**CustomEventDetailView.tsx** — time display formatting:
- Replace `moment({ hours, minutes }).format()` with `set()` + `format()` from date-fns

**package.json** — remove `moment` from dependencies.

## Test Plan

- Verified TypeScript compilation produces no new errors (all errors are pre-existing)
- Verified linter produces no new errors or warnings
- Calendar week start behavior preserved: Sunday (dow=0) for normal view, Saturday (dow=6) for finals
- Time format strings produce equivalent output (12h/24h toggle)
- The `culture` prop on `<Calendar>` selects the correct locale from the `locales` map at render time

## Issues

<div><a href="https://cursor.com/agents/bc-2067d2b4-a7e5-44b1-91a0-a00b7d41e253"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-2067d2b4-a7e5-44b1-91a0-a00b7d41e253"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #1638, opened 2026-04-19
fix: prevent scrolling to the bottom when loading calendar

dsnsgithub:
## Summary
Before replacing `moment` with `date-fns`, loading AntAlmanac would position the calendar at the top (7am), but after it would by default lock the calendar to the bottom (10pm).

While minor, I think it makes more sense (from a web perspective) to scroll down to see more instead of scrolling up.

## Test Plan
- Load AntAlmanac on any resolution that requires the calendar to scroll to view all events.

## Issues
regression from https://github.com/icssc/AntAlmanac/pull/1615

https://github.com/user-attachments/assets/8af69f0d-0052-43a4-aa25-dacf952ad6e3


KevinWu098 (approved):

KevinWu098:
I was wondering why that was happening!

PR #1640, opened 2026-04-21
feat: createdAt and updatedAt

KevinWu098:
## Summary

1. Adds `createdAt` / `updatedAt` timestamps across the schema.
2. Refactors backend code paths to correctly use `onConflictDoUpdate`, following up from #1582 (which stabilized ids, but continued to delete then reinsert)
3. Refactors some frontend custom event code (because it was bothering me)

### Schema (`packages/db`)
- Added `createdAt`/`updatedAt` (or kept existing `lastUpdated`) with `defaultNow().notNull().$onUpdate(() => new Date())` on `users`, `accounts`, `sessions`, `schedules`, `coursesInSchedule`, `customEvents`, `subscriptions`.
- Migration `0011_public_adam_destine.sql` (additive only).
- `customEvents.id` is now the primary key directly; dropped `$defaultFn(createId)` — clients send their own CUID.

### Upsert helper (`packages/db/src/utils.ts`)
- New `buildConflictUpdateSet(table, policy)` + `ConflictUpdatePolicy` type. Exhaustive, per-column `'keep' | 'update'` policy generates the `set` clause with `excluded."columnName"` references (quoted to survive Postgres case-folding on camelCase columns — this was the `subscriptions` bug).

### Upsert refactors (`apps/antalmanac/src/backend/lib/rds.ts`)
- `upsertCustomEvents`: delete-all → diff-based `onConflictDoUpdate` on `customEvents.id`. Preserves `createdAt`.
- `upsertCourses`: diff-based `onConflictDoUpdate` on composite `(scheduleId, sectionCode, term)` with client-side dedup to avoid "cannot affect row a second time" errors.
- `upsertSchedulesAndContents`: inlined the old `insertScheduleAndContents` helper; diff-based on `schedules.id`, reusing trusted IDs and minting CUIDs for unknown/foreign ones.
- `upsertNotification`: `set` clause now goes through `buildConflictUpdateSet`.
- `registerUserAccount`: single transaction, `onConflictDoUpdate` on `(userId, accountType)` so re-auth with a rotated providerId no longer crashes.

### Custom event ID standardization
- `customEventID` is a client-generated CUID string throughout (`@paralleldrive/cuid2`). Replaces the old `Date.now()` numeric ID.
- Type tightened to `'string'` in `packages/types/src/customEvent.ts`.
- Generator switched in `CustomEventDialog.tsx` and `zotcourse.ts`; fixtures updated in `tests/`.

### Frontend
- `CustomEventDialog`: re-seed form state inside `handleOpen` instead of only at mount — fixes a stale-values bug when editing the same event twice. `resetForm` deleted (no longer needed).
- `CustomEventDetailView`: switched to `CardContent` layout, named export, icon padding aligned to `ColorPicker` (`sx={{ padding: 0.5 }}`); Delete no longer uses `size="large"`.
- `CourseCalendarEvent`: Delete icon padding aligned to match.

### Cleanup
- Removed unused `apps/antalmanac/src/stores/legacyScheduleHelpers.ts`.
- `docker-compose.yml` port fix: `5432:5432` → `5433:5432` (matches `DB_URL` convention used locally).
- Tests: restored imports stripped by the formatter in `calendarize-helpers.test.ts`; added `afterEach(cleanup)` to shared setup to stop React DOM leakage between tests.

## Test Plan
- [ ] Sign in via Google — new user path creates `users` + `accounts` rows with matching `createdAt`/`updatedAt`.
- [ ] Sign in again — `users.updatedAt` bumps, `createdAt` unchanged; no duplicate account row.
- [ ] Save a schedule, edit a custom event, save again — `customEvents.createdAt` stays stable, `lastUpdated` advances.
- [ ] Add/remove/reorder courses in a schedule — `coursesInSchedule.createdAt` stable for kept rows.
- [ ] Rename a schedule — `schedules.createdAt` stable, `lastUpdated` advances.
- [ ] Open the edit dialog on a custom event twice in a row — form shows that event's values both times.
- [ ] Toggle a notification preference — `subscriptions.updatedAt` advances, no case-fold SQL error.
- [ ] Run `pnpm test` in `apps/antalmanac` — green.

## Issues
Closes #


packages/db/migrations/0011_public_adam_destine.sql
```diff
@@ -0,0 +1,14 @@
+ALTER TABLE "users" ALTER COLUMN "last_updated" SET NOT NULL;--> statement-breakpoint
```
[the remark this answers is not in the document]
KevinWu098: We'll see if this is a problem - if so, then we'll backfill

apps/antalmanac/src/backend/lib/rds.ts
```diff
@@ -153,61 +148,59 @@ export class RDS {
...
-        // First check if an account with OIDC providerId already exists
-        const existingAccount = await this.getAccountByProviderId(db, accountType, oidcProviderId);
-        if (existingAccount && accountType === 'OIDC') {
-            return { ...existingAccount, newUser: false };
-        }
+        const oidcProviderId = providerId.startsWith('google_') ? providerId : `google_${providerId}`;
 
-        const existingUser = email ? await this.getUserByEmail(db, email) : null;
+        return db.transaction(async (tx) => {
+            const existingAccount = await this.getAccountByProviderId(tx, accountType, oidcProviderId);
```
sicn4rf, line 113: question: `registerUserAccount` opens its own `db.transaction(...)` wrapper, but it's also called from inside `upsertUserData`'s transaction. i think drizzle turns that into a savepoint so it works, but i noticed none of the other private static helpers here do that. is  there an intention behind using nested transactions instead of just taking `tx` from the caller ?

apps/antalmanac/src/backend/lib/rds.ts
```diff
@@ -191,153 +184,167 @@ export class RDS {
...
+                .onConflictDoUpdate({
+                    target: schedules.id,
+                    set: buildConflictUpdateSet(schedules, {
+                        id: 'keep',
+                        userId: 'keep',
+                        name: 'update',
+                        notes: 'update',
+                        index: 'update',
+                        createdAt: 'keep',
+                        lastUpdated: 'update',
```
sicn4rf: thought: while smoke testing, noticed that every save bumps `last_updated` on every schedule in the payload, not just the one that changed. i think this is because the client sends the full schedules array each time and `onConflictDoUpdate` always runs the SET clause on conflicts, so every existing row gets `last_updated = now()` even when its content didn't change. same fanout behavior happens on `coursesInSchedule`.

is this the intended semantic of `last_updated`? if the idea is "last time this row was touched by a save," this is fine. but if it should mean "last time the row's content actually changed," we'd either need a conditional update or only send changed rows from the client, or something along those lines yada yada
image

not blocking, but worth being intentional about so future code that relies on `last_updated`(sync, activity feeds, last-modified headers) doesn't end up misleading.

sicn4rf (dismissed):
LGTM! i verified the bugs from #1597 are fixed. migration 0011 correctly adds `DEFAULT now()` to `schedules.last_updated` and `SET NOT NULL` on the three drift tables. i also confirmed via local smoke test against a freshly-migrated Postgres the following:

- schedule `INSERT` now succeeds (the blocking bug i flagged from #1597 )
- `createdAt` is preserved across saves on schedules, courses, and custom events (the `onConflictDoUpdate` refactor works as intended)
- custom events propagate correctly across schedules that share them
- sign out / sign in produces a single account row (atomic upsert works)

---

note: unrelated to the dialog fix in this PR, but thought id bring it up since i noticed while testing custom events. after a name change, the already-open dialog doesn't update until you close and reopen it. probably separate scope, just flagging.
Image

apps/antalmanac/src/backend/lib/rds.ts
KevinWu098: mm, its because our rds code is actually not structured the best — the `DatabaseOrTransaction` type is pretty gnarly if you think about it. in this case, the extra savepoint is acceptable to me since upsert is called with `db` initially sometimes, or from register (which passes it a `tx`).

would be open to fixing this separately. would you mind opening an issue?

KevinWu098:
> note: unrelated to the dialog fix in this PR, but thought id bring it up since i noticed while testing custom events. after a name change, the already-open dialog doesn't update until you close and reopen it. probably separate scope, just flagging.

open an issue :3 — maybe phrase it as revamp the event popover? it misses a lot of good info like the location, and if this is linked to other schedules, etc

apps/antalmanac/src/backend/lib/rds.ts
sicn4rf: that makes sense to me. the extra savepoint is reasonable to me as well. i'll go ahead and open an issue for this

sicn4rf:
> open an issue :3 — maybe phrase it as revamp the event popover? it misses a lot of good info like the location, and if this is linked to other schedules, etc

i agree, it does miss a lot of good info, and the popover is just sparse overall. everything else LGTM! i'll write up the two issues right now

sicn4rf (dismissed):
this looks good, i verified that it only updates per schedule now. 

Image

after the fanout fix, adding or removing courses no longer bumps `schedules.last_updated`. only direct edits to the schedule row's columns like changing its name, notes, or index bump it. the course changes are captured on `coursesInSchedule.last_updated` instead.

this is honestly more precise, but if `schedules.last_updated` is meant to represent "last time anything about this schedule changed, including contents," we'd either need to bump it explicitly after upserting courses/custom events, or compute it at read time across the children? not sure if thats the intention here

everything else LGTM 👍

KevinWu098:
> this looks good, i verified that it only updates per schedule now.
> 
> Image
> after the fanout fix, adding or removing courses no longer bumps `schedules.last_updated`. only direct edits to the schedule row's columns like changing its name, notes, or index bump it. the course changes are captured on `coursesInSchedule.last_updated` instead.
> 
> this is honestly more precise, but if `schedules.last_updated` is meant to represent "last time anything about this schedule changed, including contents," we'd either need to bump it explicitly after upserting courses/custom events, or compute it at read time across the children? not sure if thats the intention here
> 
> everything else LGTM 👍

Right. Yeah that's an intentional design choice since I didn't want to get deeper in the rabbit hole of `last_updated` semantics that aren't managed directly in the schema of `schedule.ts`. I've added a comment to the schema describing this

PR #1641, opened 2026-04-21
perf(rds): remove redundant db.transaction wrappers and dead guest-user helpers

KevinWu098:
## Summary

Two cleanups in `apps/antalmanac/src/backend/lib/rds.ts`:

### 1. Remove redundant single-statement and read-only `db.transaction` wrappers

Several helpers wrap a single SQL statement in `db.transaction(...)`. Single statements are already atomic in Postgres, so the wrapper is pure overhead — and worse, when these helpers are invoked with an existing `Transaction` (Drizzle's `DatabaseOrTransaction` union allows either), Drizzle's `db.transaction(...)` opens a **savepoint** rather than reusing the outer tx. In either case the wrapper is dead weight.

Two more helpers wrap a small sequence of independent read-only `SELECT`s in `db.transaction(...)`. The justification ("snapshot consistency across the reads") does not hold at Postgres's default `READ COMMITTED` isolation level — each statement takes its own snapshot, so the wrapper provides no meaningful cross-statement consistency and again just costs a `BEGIN`/`COMMIT` round-trip (or a savepoint when a caller passes in an existing `Transaction`). If cross-read consistency is ever actually needed, it should be enforced explicitly with `REPEATABLE READ`.

Unwrapped helpers (13):

- `getAccountByProviderId` — single `SELECT`
- `getUserById` — single `SELECT`
- `getUserByEmail` — single `SELECT`
- `getGoogleIdByUserId` — single `SELECT`
- `getCurrentSession` — single `SELECT`
- `getUserAndAccountBySessionToken` — single `SELECT` with joins
- `retrieveNotifications` — single `SELECT`
- `upsertNotification` — single `INSERT ... ON CONFLICT DO UPDATE`
- `updateAllNotifications` — single `UPDATE`
- `deleteNotification` — single `DELETE`
- `deleteAllNotifications` — single `DELETE`
- `getGuestScheduleByUsername` — 3 independent read-only `SELECT`s
- `fetchUserDataWithSession` — 3 independent read-only `SELECT`s (via the private `getUserDataWithSession` helper, whose parameter was widened from `Transaction` to `DatabaseOrTransaction`)

Behavior is unchanged: each call still executes the same SQL against the provided `DatabaseOrTransaction`, just without opening a new transaction/savepoint. Practical effect today is removing a `BEGIN`/`COMMIT` round-trip per call — e.g. on the OAuth callback flow (`registerUserAccount`), which internally calls `getAccountByProviderId` and `getUserByEmail` with the raw `db`, this drops 2 extra top-level transactions per sign-in; `fetchUserDataWithSession` runs on every authenticated session load.

### 2. Remove dead guest-user helpers

- `createGuestUserOptional` has zero callers anywhere in the repo.
- `guestUserIdWithNameOrNull` was only called by `createGuestUserOptional`, so transitively dead.

Both were orphaned by the guest-schedule endpoint restructuring earlier on this branch stack.

## Transactions left in place (and why)

Three `db.transaction` call sites remain in `rds.ts`. Each was evaluated and kept intentionally:

| # | Function | What it wraps | Why it's kept |
|---|---|---|---|
| 1 | `upsertUserData` | delete schedules → insert schedules → insert courses + custom events → update `users.currentScheduleId` | Multi-statement write; atomicity required. Only caller (`saveUserData` tRPC mutation) passes the raw `db`, so this opens a fresh top-level tx, not a savepoint. |
| 2 | `upsertSession` | `SELECT` refresh token → conditional `INSERT` session | Read-modify-write; reviewer explicitly flagged as `keep`. |
| 3 | `flagImportedUser` | `SELECT userId + imported` → conditional `UPDATE imported=true` | Read-modify-write; tx prevents racing with a concurrent flag update. |

## Test Plan

- `pnpm lint` passes.
- `tsc --noEmit` on `apps/antalmanac` produces the same set of pre-existing errors as the base branch `kwu/backend-perf` (all unrelated — missing `$generated/termData`, `$generated/searchData`, image asset modules). No new type errors.
- Diff is a mechanical unwrap of `db.transaction((tx) => tx.X)` → `db.X` plus deletion of two unreferenced private methods; no logic or query changes to live code paths.

## Issues

Closes #

<div><a href="https://cursor.com/agents/bc-91e5ae55-b416-4e98-ab1f-a7f84027e11e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-91e5ae55-b416-4e98-ab1f-a7f84027e11e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


apps/antalmanac/src/backend/lib/rds.ts
Choollol: Could this transaction also be removed? It looks like this action is readonly which doesn't seem like it needs a transaction. I'm not too well-versed in backend dev though, so feel free to correct me!

Choollol: If you decide to remove the transaction in `fetchUserDataWithSession`, then this parameter should also probably be changed.

Choollol: Same question as https://github.com/icssc/AntAlmanac/pull/1641/changes#r3135302983

Choollol (approved):
LGMT! Just a small question about some possible changes

KevinWu098:
> LGMT! Just a small question about some possible changes

Nice catches, think cursor resolved them

Choollol (dismissed):
LGTM! There are some cubic comments that look valid but that logic is from before so doesn't necessarily need to be changed in this PR.

apps/antalmanac/src/backend/lib/rds.ts
```diff
@@ -591,62 +537,59 @@ export class RDS {
...
+
+        const customEventResults = await db
+            .select()
+            .from(schedules)
+            .where(eq(schedules.userId, user.id))
+            .leftJoin(customEvents, eq(schedules.id, customEvents.scheduleId));
+
+        const userSchedules = RDS.aggregateUserData(sectionResults, customEventResults);
+
+        const scheduleIndex = user.currentScheduleId
```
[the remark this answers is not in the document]
KevinWu098: will deal with it elsewhere

apps/antalmanac/src/backend/lib/rds.ts
```diff
@@ -398,47 +348,45 @@ export class RDS {
...
-        });
+        const customEventResults = await db
+            .select()
+            .from(schedules)
+            .where(eq(schedules.userId, userId))
+            .leftJoin(customEvents, eq(schedules.id, customEvents.scheduleId));
+
+        const userSchedules = RDS.aggregateUserData(sectionResults, customEventResults);
+
+        const scheduleIndex = row.users.currentScheduleId
```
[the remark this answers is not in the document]
KevinWu098: will deal with it elsewhere

PR #1645, opened 2026-04-22
Fix planner response structure validation

### LINKED ISSUE #1644: Planner fetching is broken
Planner changed the structure of their response (to support user-defined units I think), which broke our side since we have to manually change types.

Choollol:
## Summary

Update planner response types to match actual returned object.

## Test Plan

Ensure planner fetch doesn't result in zod error in the server console.

## Issues

Closes #1644


Choollol:
Will just be part of https://github.com/icssc/AntAlmanac/pull/1647

Issue #1646, opened 2026-04-22
Refactor planner hook


Choollol:
Move planner-related stuff out of the session store and into a separate planner store that will also replace `usePlannerRoadmaps`.

Choollol:
Will be done as part of migration to better auth

Choollol:
Nvm, can be done after migration

PR #1647, opened 2026-04-22
Search with planner

### LINKED ISSUE #1644: Planner fetching is broken

Choollol:
## Summary

Populate search results with courses from a planner roadmap's term

## Test Plan

1. Searching with planner works normally
2. Loading states are properly set when loading planner roadmaps and loading search
3. Roadmaps that include/exclude current term are correct
4. Unoffered courses are ignored and displayed in alerts
5. If Planner gives an unavailable term, fallback to default term and display in alert
6. Custom courses are ignored

## Issues

Closes #1644 (side thing)


anthonyj33:
~~oh probably need to update the branch from main lol~~
oh i forgot i dont have permissions in this repo
> Command 'deploy-shared' is not configured for the user's permission level 'none'.

apps/antalmanac/src/components/RightPane/RightPaneStore.ts
```diff
@@ -76,17 +89,27 @@ class RightPaneStore extends EventEmitter {
...
     };
 
+    setMultiSearchData = (data: Partial<(typeof this.multiSearchData)[number]>[]) => {
+        this.multiSearchData = data.map((params) => ({ ...defaultFormValues, ...params }));
+    };
+
+    clearMultiSearchData = () => {
+        this.multiSearchData = [];
+    };
+
```
sicn4rf, line 132: blocking: `setMultiSearchData` is fetching sections from the wrong term

repro 1: i'm on the default term (Spring 2026). I change the form's term picker to Fall 2025, which exists in my roadmap. I click Search with Planner.

what happens: the planner-side lookup correctly pulls my Fall 2025 course IDs from the roadmap, but the resulting section table shows Spring 2026 sections. the courses have open enrollment, current-quarter section codes, etc. Fall 2025 sections would be historical/closed, so the "all still open" status confirms it's actually fetching the default term.

repro 2: i added CS 161 and CS 147 to my Summer 10wk plan. form set to Summer 10wk, then Search with Planner.

what happens: CS 161 shows 6 sections, but a manual Summer 10wk search for CS 161 only returns 2. those 6 are Spring 2026's. CS 147 disappears entirely with no "not offered" banner. it passed the Summer 10wk `filterOfferedCourses` check, but the downstream WebSOC query ran for Spring 2026 (where CS 147 isn't offered) and returned nothing.

potential root cause: `setMultiSearchData` spreads `defaultFormValues` (whose `term` is captured once at module load), so every entry's `term` becomes the static module default rather than what the user picked. `loadCourses` then passes `course.term` into `getQueryParams`, which builds the WebSOC query. so WebSOC gets called for the wrong term. the `filterOfferedCourses` path uses `getTermParts()` (form term) so the two paths disagree.

Fix: spread the live form values instead, e.g.
```
this.multiSearchData = data.map((params) => ({ ...this.formData, ...params }));
```

or accept the term as a parameter into `setMultiSearchData`. fixing this should also resolve the "course silently disappears" case (CS 147) since the offered-check and the section-fetch will agree on the term.

apps/antalmanac/src/hooks/usePlanner.ts
```diff
@@ -71,18 +73,23 @@ export function usePlannerRoadmaps() {
     }, []);
 
     useEffect(() => {
+        if (isPlannerLoading) {
+            return;
+        }
```
sicn4rf: issue: loading-flag race in `usePlannerRoadmaps`

two compounding problems in this effect. (1) the `if (isPlannerLoading) return` guard reads `isPlannerLoading` but the dep array is `[googleId, setPlannerRoadmaps]`, so the guard is checking a stale snapshot. (2) `setIsPlannerLoading(false)` runs unconditionally outside the `if (active)` cleanup guard, so a stale earlier request can flip loading off while a newer fetch is still in flight.

doesn't bite often in practice because `googleId` is stable post-login, but it's an easy cleanup.

possible fix:
```
useEffect(() => {
    let active = true;
    async function loadRoadmaps() {
        if (!googleId) {
            setPlannerRoadmaps([]);
            return;
        }
        setIsPlannerLoading(true);
        try {
            const data = await trpc.roadmap.fetchUserPlannerRoadmaps.query();
            if (active) setPlannerRoadmaps(data ?? []);
        } catch (e) {
            console.error('Failed to fetch Planner roadmaps:', e);
        } finally {
            if (active) setIsPlannerLoading(false);
        }
    }
    loadRoadmaps();
    return () => {
        active = false;
    };
}, [googleId, setPlannerRoadmaps, setIsPlannerLoading]);
```

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/LabeledInputs/LabeledTextField.tsx
```diff
@@ -1,21 +1,52 @@
...
-        <LabeledInput label={label} isAligned={isAligned} id={textFieldId}>
-            <TextField size="small" variant="outlined" id={textFieldId} {...textFieldProps} />
+        <LabeledInput labelProps={{ ...labelProps, id: textFieldId, label, isAligned }}>
+            <TextField
+                size="small"
+                variant="outlined"
+                id={textFieldId}
+                disabled={disabled}
+                {...textFieldProps}
+                {...(loading
```
sicn4rf: nit: `disabled` is overridable by `textFieldProps`
```
disabled={disabled}
{...textFieldProps}
```

no current caller passes `disabled` through `textFieldProps`, so this is latent. but since the prop is being introduced, it should win.

fix:

```
{...textFieldProps}
disabled={disabled ?? textFieldProps?.disabled}
```

sicn4rf (changes requested):
one blocking bug + one race-condition issue. the blocker is reproducible and breaks the core intent of the feature (search using the term i picked). tested against staging-shared:

term clobber: searching with planner uses the form's term to find the right roadmap quarter and to run the offered/not-offered check, but then issues the actual WebSOC query for the static module-default term. confirmed two ways:
- Fall 2025 planner search returns Spring 2026 (currently-open) sections.
- Summer 10wk planner search for CS 161 returns 6 Spring 2026 sections instead of the 2 real Summer 10wk sections; CS 147 silently disappears entirely because it's not in Spring 2026.

apps/antalmanac/src/hooks/usePlanner.ts
Choollol: I miss dependency array linting 😔 

I also plan to convert the usePlanner hook into a zustand store soon after this PR so this will probably be less of a problem after that

sicn4rf (dismissed):
all four items i flagged got addressed in the latest commits:
- term clobber fixed in `setMultiSearchData` by pinning `term: this.formData.term` last in the spread
- usePlanner race fixed by removing the stale `isPlannerLoading` guard and gating `setIsPlannerLoading(false)` behind `active`
- LabeledTextField `disabled` prop now wins over `textFieldProps`
- missing key on the no-courses Tooltip

retested both reproductions:
- Fall 2025 planner search from the Spring 2026 default now correctly returns Fall 2025 sections (CRNs match historical data even though the status shows OPEN)
- Summer 10wk planner search returns both CS 161 and CS 147; CS 161's section count matches a manual Summer 10wk search

---

one thing i noticed during testing that's NOT in scope for this PR but im flagging it here for context, will put up a hotfix shortly:

MATH 1A and 1B show "not offered" for Summer Session 1 even though they are offered. same wrong call appears in the QuickSearch autocomplete "at a glance" tag. both surfaces share the same backend path (`getOfferedCourses` in `search.ts`), which reads from `apps/antalmanac/src/generated/terms/${quarter}_${year}.json`.

root cause is in `scripts/get-search-data.ts`: the cache only refreshes for `termData[0]` (currently `2026 Summer Session 2` per `deployed_terms.json`). every other term, including Summer Session 1, keeps whatever was generated the first time it was cached. so courses UCI added to Summer 1 after that initial snapshot stay invisible.

this is a pre-existing cache-invalidation bug that's been latent since the autocomplete tag feature i worked on was shipped. the new `filterOfferedCourses` endpoint just makes it more visible by surfacing it as a banner instead of a subtle tag. not introduced by this PR, fully separable scope (only touches `get-search-data.ts`), so i'm approving #1647 and will open a small hotfix PR shortly to refresh upcoming terms in addition to the current one.

related but separate: the "Search for X on AntAlmanac Planner!" when you actually click on a course thats on offered is CTA in `ErrorMessage` and reads from `formData` directly, so on a multi-search it shows whatever the user last typed manually instead of the actually-failed course. also pre-existing; might bundle into the same hotfix.

so if you search POLSCI 6B: "Nothing found. Search for POLSCI 6B on AntAlmanac Planner", and then you trigger the "Search with Planner" feature on courses that are all 
"Not offered" it shows the stale error message. see below.

Image

LGTM im approving. good to merge tonight.

apps/antalmanac/src/stores/SessionStore.ts
```diff
@@ -15,16 +15,20 @@ interface SessionState {
     loadSession: () => Promise<boolean>;
     clearSession: () => Promise<string | null>;
 
+    hasCheckedAuth: boolean;
```
KevinWu098, line 18: issue: why have we added this prop? is the presence of a userId not enough?

apps/antalmanac/src/stores/SessionStore.ts
```diff
@@ -15,16 +15,20 @@ interface SessionState {
...
+
     googleId: string | null;
     filterTakenCourses: boolean;
     userTakenCourses: Set<string>;
 
     plannerRoadmaps: Roadmap[];
+    isPlannerLoading: boolean;
```
KevinWu098, line 25: issue: i know we didn't start doing this in this PR, but this information definitely should not live on the session (which I usually understand to be specific to auth/identity)

at the least, i want to see an issue created for this

apps/antalmanac/src/backend/routers/websoc.ts
```diff
@@ -133,7 +133,7 @@ function combineWebsocResponses(responses: WebsocAPIResponse[]) {
 
 const websocRouter = router({
     getOne: procedure.input(z.record(z.string(), z.string())).query(queryWebSoc),
-    getMany: procedure
+    getManyOfField: procedure
```
KevinWu098, line 136: question: just curious, what does this name mean? why is it different to getMultiple?

apps/antalmanac/src/components/RightDivider.tsx
```diff
@@ -0,0 +1,27 @@
...
+                    '&::after': { borderColor: theme.palette.text.primary },
+                }),
+                sx
+            )}
+        >
+            {children}
+        </Divider>
+    );
+};
+export default RightDivider;
```
KevinWu098: nit (non-blocking): we do not do defualt exports in this house

apps/antalmanac/src/components/RightDivider.tsx
```diff
@@ -0,0 +1,27 @@
+import { Divider, SxProps, Theme } from '@mui/material';
+import { mergeSx } from '@mui/x-date-pickers/internals';
+import { ReactNode } from 'react';
+
+interface Props {
```
KevinWu098: nit (non-blocking): we name props {ComponentName}Props in this household

apps/antalmanac/src/components/RightDivider.tsx
```diff
@@ -0,0 +1,27 @@
...
+    children: ReactNode;
+    sx?: SxProps<Theme>;
+}
+
+const RightDivider = ({ children, sx }: Props) => {
```
KevinWu098: nit (non-blocking): it would be nice if this was a little more descriptive to what it was (e.g. SearchResultsRightDivider)

apps/antalmanac/src/components/WarningAlert.tsx
```diff
@@ -0,0 +1,34 @@
+import { Alert } from '@mui/material';
+import { ReactNode, useState } from 'react';
+
+interface Props {
```
KevinWu098: nit (non-blocking) (repeat): props name, default export

KevinWu098 (changes requested):
Some initial comments. haven't gotten into the UI components yet

apps/antalmanac/src/stores/SessionStore.ts
Choollol: userId isn't defined while auth is loading, which causes the dialog to appear for a fraction of a second before auth is fully loaded then disappear.

apps/antalmanac/src/stores/SessionStore.ts
Choollol: Made one some time ago: https://github.com/icssc/AntAlmanac/issues/1646

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/LabeledInputs/LabeledTextField.tsx
```diff
@@ -1,21 +1,52 @@
...
+                    ? {
+                          slotProps: {
+                              ...textFieldProps?.slotProps,
+                              input: {
+                                  ...textFieldProps?.slotProps?.input,
+                                  endAdornment: loading ? (
+                                      <Box sx={{ display: 'flex', justifyContent: 'center' }}>
+                                          <CircularProgress size="1.5em" />
+                                      </Box>
```
KevinWu098: suggestion: seems suspect! we don't really use em in this household!

apps/antalmanac/src/backend/routers/websoc.ts
Choollol: From my understanding of it, `getManyOfField` fetches multiple times for one set of form data with only a single field changed between each. `getMultiple` fetches multiple sets of form data.

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/QuickSearch.tsx
```diff
@@ -0,0 +1,34 @@
+import FuzzySearch from '$components/RightPane/CoursePane/SearchForm/FuzzySearch';
+import SearchWithPlanner from '$components/RightPane/CoursePane/SearchForm/SearchWithPlanner';
+import { useIsMobile } from '$hooks/useIsMobile';
+import { Typography } from '@mui/material';
+import { Box, Stack } from '@mui/system';
+import { usePostHog } from 'posthog-js/react';
+import { ComponentProps } from 'react';
+
+interface Props {
```
KevinWu098: nit: yada yada you know

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/QuickSearch.tsx
```diff
@@ -0,0 +1,34 @@
...
+    const fuzzySearch = <FuzzySearch toggleSearch={toggleSearch} postHog={postHog} labelProps={labelProps} />;
+    const plannerSearch = <SearchWithPlanner labelProps={labelProps} />;
+
+    return isMobile ? (
+        <>
+            {fuzzySearch}
+            {plannerSearch}
+        </>
+    ) : (
+        <Stack direction="row" spacing={2} sx={{ alignItems: 'center' }}>
```
KevinWu098: issue (blocking): guys this is chudded and the smaller sizing is chudded too. why aren't we just flex wrapping this stuff when we need to? i dont understand

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchForm.tsx
```diff
@@ -1,22 +1,25 @@
...
 import { useThemeStore } from '$stores/SettingsStore';
+import { alpha, Box, Stack, SxProps, Theme, ToggleButton, ToggleButtonGroup } from '@mui/material';
+import { usePostHog } from 'posthog-js/react';
+import { useCallback, type FormEvent } from 'react';
 
 interface SearchFormProps {
     toggleSearch: () => void;
 }
 
+const QUICKSEARCH_LABEL_SX: SxProps<Theme> = {
```
KevinWu098: issue: this is definitely chudded

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchWithPlanner.tsx
```diff
@@ -0,0 +1,253 @@
+import { SignInDialog } from '$components/dialogs/SignInDialog';
```
KevinWu098, line 1: thought: i'm pretty suspect of all the useEffect-ing that we're doing here, but I'm not interested in unravelling it now

apps/antalmanac/src/components/RightDivider.tsx
Choollol: Could it be reused in the future for something that isn't related to search results?

I do agree that it is a little vague. What do you think about "HorizontalRightDivider"?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/TermSelector.tsx
```diff
@@ -31,13 +47,27 @@ export function TermSelector() {
...
+            const fallbackTerm = getDefaultTerm().shortName;
+            const message = `${term} is currently unavailable, falling back to ${fallbackTerm}`;
+            openSnackbar('error', message);
+            console.error('Error setting term from URL:', message);
+
+            handleChange(null, fallbackTerm);
+
+            setWarningMessages(CoursePaneWarningType.TermUnavailable, [message]);
+        }
+    }, [term]);
```
KevinWu098: issue: this should not be a useEffect. this should be handled in a synchronous callback

apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
```diff
@@ -129,69 +141,69 @@ function SectionTable(props: SectionTableProps) {
                         />
                     }
                 />
+
+                <IconButton
+                    title={`${openContent ? 'Collapse' : 'Expand'} courses`}
```
KevinWu098, line 146: suggestion: would be nice if this was positioned next to quick search, so as to be more... aligned (?)

anthonyj33 (approved):
Looks good- I don't have anything to add that other reviews haven't covered

apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
Choollol: Could you explain a little more about what you mean by next to quick search? Do you mean the quick search autocomplete on the other display?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/QuickSearch.tsx
Choollol: This might have to something addressed in a redesign cuz I don't think there's a way to conditionally render the "or" with just flex?

apps/antalmanac/src/backend/routers/websoc.ts
KevinWu098: 🌚 im gonna need an example, but ill take your word for it

apps/antalmanac/src/stores/SessionStore.ts
KevinWu098: :| i guess...

apps/antalmanac/src/components/RightDivider.tsx
KevinWu098: works for me

apps/antalmanac/src/components/RightPane/SectionTable/SectionTable.tsx
KevinWu098: looks like the order is 

[quick search icon] [info icon] [planner] [syllabi]

KevinWu098 (approved):

PR #1648, opened 2026-04-22
feat: optimize react render path

KevinWu098:
## Summary

Measured render times via React Profiler heat maps and reduced **~0.5–0.8 s** shaved from schedule load.

### Render

- **`providers/PostHog.tsx`** — hoisted `posthog.init()` to module scope (was re-initializing on every render of `AppPostHogProvider`).
- **`app/Theme.tsx`** — split `useThemeStore((s) => [s.appTheme, s.setAppTheme])` into two scalar selectors; the tuple returned a new reference on every call and forced `AppThemeProvider` to re-render on any store write.
- **`Calendar/CalendarCourseEventWrapper.tsx`** — `memo()` the component and replace `<Box>` with `<div>`. Was the top self-time fiber (243 ms) in the initial profile.
- **`Calendar/CalendarRoot.tsx`** — memoized `events` and `startTime`; `getStartTime()` produced a new `Date` every render, breaking `react-big-calendar`'s internal equality checks.
- **`Calendar/Toolbar/CalendarToolbar.tsx`** — replaced the `handleUndo(postHog)` / `handleRedo(postHog)` factory-invocation pattern with stable `useCallback` handlers. The prior pattern created fresh function identities per render (~87 ms `MuiButtonRoot` in flame graph).
- **`buttons/Clear.tsx`** — same `useCallback` fix; also actually threads `analyticsCategory` through to the handler.
- **`hooks/useQuickSearch.tsx`** — swapped full-store destructures for narrow scalar selectors.

### Network

- **`stores/SessionStore.ts` + `components/Header/Signin.tsx`** — moved `loadNotifications()` out of `loadSession()`; it now fires after `loadSchedule` resolves. `notifications.get` was landing in the same tRPC `httpBatchLink` window as the `websoc.getCourseInfo` calls, so the combined batch blocked the critical path on the slowest arm.

### Measured impact

- Render-side: removed the 243 ms `CalendarCourseEventWrapper` and 87 ms `MuiButtonRoot` hotspots plus redundant `AppThemeProvider` / `PostHog` re-renders. No clean before/after render total, but ~200–300 ms order-of-magnitude based on the flame graphs.

## Test Plan

- [ ] Sign in (Google), schedule loads and courses render.
- [ ] Notifications tab/UI initializes correctly for signed-in Google users.
- [ ] Undo / redo / clear schedule / quick-search from calendar event still work.
- [ ] Theme toggle (dark/light) still works; no extra `AppThemeProvider` churn in Profiler.
- [ ] Silent SSO via `AutoSignIn` (SSO cookie present, session expired) still loads notifications after schedule.
- [ ] Posthog initialized correctly and events sent

## Issues

Closes #


apps/antalmanac/src/providers/PostHog.tsx
```diff
@@ -5,17 +5,25 @@ interface Props {
...
-
-        return <PostHogProvider client={posthog}>{props.children}</PostHogProvider>;
     } else {
         console.warn('PostHog not initialized: Missing API key');
+    }
+}
+
+export default function AppPostHogProvider(props: Props) {
+    if (!POSTHOG_KEY) {
         return <>{props.children}</>;
```
Choollol: I know you weren't the one that did this, but since you're changing this file, could this be simplified into just `return props.children;`?

Choollol (changes requested):
Changes that are here generally look good! Just a few thoughts:

1. In `CalendarRoot.tsx`, it looks like `blueprintToSkeletonEvent` and `createSkeletonEvents` can be moved to outside the component function. If it's not related to the changes you want to make in this PR, feel free to leave this for later.
2. Should we set a convention for consistent shallow zustand store selects? In some places we have them separated, like in this PR, and in other places we use `useShallow`. It's not a big issue, but I think this consistency would be better for the codebase's readability.

Also to double check, is the target branch of this PR (kwu/remove-redundant-transactions-e11e) correct?

KevinWu098:
Yes the target is correct (although this is not necessary, I don't think) -- I just want the deploy to also benefit from the speed ups from the other unmerged PRs

KevinWu098:
> Changes that are here generally look good! Just a few thoughts:
> 
> 
> 
> 1. In `CalendarRoot.tsx`, it looks like `blueprintToSkeletonEvent` and `createSkeletonEvents` can be moved to outside the component function. If it's not related to the changes you want to make in this PR, feel free to leave this for later.
> 
> 2. Should we set a convention for consistent shallow zustand store selects? In some places we have them separated, like in this PR, and in other places we use `useShallow`. It's not a big issue, but I think this consistency would be better for the codebase's readability.
> 
> 
> 
> Also to double check, is the target branch of this PR (kwu/remove-redundant-transactions-e11e) correct? 

1. Sure
2. Eh don't want to bloat this too much -- I can sic cursor on the rest later. Do you want to open an issue?

Choollol:
> 2\. Eh don't want to bloat this too much -- I can sic cursor on the rest later. Do you want to open an issue?

Makes sense to me, I'll open the issue

Choollol (dismissed):
LGTM

PR #1649, opened 2026-04-22
feat: add db indexes

KevinWu098:
## Summary

Adds indexes on `sessions` and `customEvents`, turning sequential scans into index lookups on two hot paths:

- `sessions.refresh_token` — every authenticated request runs `getUserAndAccountBySessionToken`, which filters on `refresh_token`. Today this seq-scans all **170k** session rows. In the latest prod trace this single call is **~497 ms** (and was ~1 s in an earlier trace); a B-tree lookup drops it to the single-digit millisecond range.
- `sessions.user_id` — covers `createSession`'s existence check and the `users ⨝ sessions` joins. Postgres doesn't auto-index foreign keys.
- `customEvents.scheduleId` — covers the `leftJoin(customEvents, eq(schedules.id, customEvents.scheduleId))` inside `fetchUserDataWithSession`. Today seq-scans all **82.7k** custom event rows on every signed-in page load.

Expected net win on the warm authed critical path: **~450–500 ms** off time-to-courses, dominated by the `sessions.refresh_token` lookup.

## Test Plan

- [ ] Sign in (Google), schedule loads; verify `userData.getUserAndAccountBySessionToken` wall time drops from ~500 ms to <50 ms in the network panel.
- [ ] Sign out / session creation still works (exercises `createSession` path, which uses the new `sessions_user_id_idx`).
- [ ] Schedule with custom events renders correctly (exercises `customEvents_scheduleId_idx`).

## Issues

Closes #


KevinWu098:
eh... nah

PR #1658, opened 2026-04-27
feat: review pop-up prompt

### LINKED ISSUE #1628: feat: in-scheduler review prompt for past (potentially) taken courses
# Context
Planner has a professor review/rating system, and we want to increase penetration for the feature. Students typically don't think of leaving reviews when they're planning for future courses on Planner. Since Scheduler users already have course history in their schedules, we can prompt them to leave quick reviews and potentially driving data into the review system with minimal friction

## Feature

A non-intrusive popup (bottom right corner of Scheduler) that:

**1. Detects candidate courses** - pulls 1 course from the user's schedule from a term within the last 2-3 quarters. 
**2. Confirm enrollment first** - since users can have multiple schedules per term (planning alternatives and what-if scenarios), a schedule entry doesn't guarantee they actually took the course. the prompt should open with something like
```
"Did you take [Course] with [Professor] recently?"
Yes / No
```
- No -> dismiss, don't ask for that course+professor combo (probably)
- Yes -> proceed with the review flow

**3. Collect a quick review**
- 1-5 star rating
- Predefined tag selection (e.g. "Clear grading criteria", "Tough grader", "Amazing lectures", "Lots of homework", etc.)

**4. Store the data** - in our own DB that matches Planner's existing `reviews` and we can reconcile later

**5.  Dismissable** - users can close at any point. don't re-prompt for courses that are already reviewed or dismissed

## Existing infrastructure to build on
- **Schedule store** (`Schedules.ts`) - this has course history with `term` per course; note that users can have multiple schedules per term
- **Snackbar system** (`SnackbarStore.ts`) - existing popup pattern. the review prompt will need a custom component though (should be richer than the snackbar)
- **MUI** - Dialogs, Popovers, Cards are all available
- **Drizzle ORM and Postgres** - same stack as Planner, makes schema alignment easier

### References to Planner's current `review` schema
source: [api/src/db/schema.ts](https://github.com/icssc/peterportal-client/blob/main/api/src/db/schema.ts)
full tag list and type definitions: [types/src/review.ts](https://github.com/icssc/peterportal-client/blob/main/types/src/review.ts]

I think our review table should mirror Planner's schema column-for-column. For fields we don't collect in the UI, we can leave them as nullable. Reach out to me or Alex for more discussion about the design decisions with how we can approach this mirrored DB.

AureliaSindhu:
## Summary

- Adds a new `instructorReviews` table to the DB schema (Drizzle + migration `0011`) to store quick in-app course/professor reviews submitted by users.
- Adds a `review` tRPC router with two endpoints: `submitReview` (enforces one review per user/professor/course combo) and `getReviewedCombos` (used to skip already-reviewed candidates on the client).
- Adds `ReviewPromptStore` (Zustand) that selects a random eligible candidate from the user's past-term schedule history, filters out already-dismissed (localStorage) and already-reviewed (DB) combos, and manages the multi-step prompt state.
- Adds the `ReviewPrompt` component — a fixed bottom-right MUI card that appears 15 s after login with two steps: an enrollment-confirmation step and a star-rating + tag-selection step.
- Mounts `ReviewPrompt` in the Home route so it is globally available.

## Test Plan

- [ ] Log in with an account that has courses from past terms in the schedule.
- [ ] Wait ~15 s — the "Quick question" card should appear in the bottom-right corner.
- [ ] Click **Not really** — card closes and does not re-appear for that course/professor combo (check `localStorage` key `aa_review_dismissed`).
- [ ] Reload and wait — a different eligible candidate (if any) should be picked; dismissed combos are excluded.
- [ ] Click **Yes!** — the rating + tag step should appear.
- [ ] Select a star rating and optionally toggle some tags, then click **Submit** — a success snackbar appears and the review is persisted in the DB.
- [ ] Submit again for the same course/professor — the server should return a `CONFLICT` error and the error snackbar should appear.
- [ ] Log in without any past-term courses — the prompt should never appear.

## Issues

Closes #1628


apps/antalmanac/src/components/ReviewPrompt/ReviewPrompt.tsx
```diff
@@ -0,0 +1,259 @@
...
+    Fade,
+    IconButton,
+    Rating,
+    Stack,
+    Typography,
+} from '@mui/material';
+import { useEffect, useRef } from 'react';
+import { useShallow } from 'zustand/react/shallow';
+
+/** Delay (ms) after mount before the prompt is surfaced. */
```
KevinWu098: nit: comment feels redundant

apps/antalmanac/src/components/ReviewPrompt/ReviewPrompt.tsx
```diff
@@ -0,0 +1,259 @@
...
+    if (step === 'hidden' || !candidate) return null;
+
+    const handleSubmit = () => {
+        if (userId) submitReview(userId);
+    };
+
+    return (
+        <Fade in timeout={400}>
+            <Card
+                elevation={6}
```
KevinWu098: issue: it would be nice if we didn't magic string elevation. does MUI not have some sort of component that works for this better? its okay if not, but id prefer it if we could take a look and see

apps/antalmanac/src/components/ReviewPrompt/ReviewPrompt.tsx
KevinWu098, line 1: issue: break this up into multiple components please. best practices should be one component, one file

apps/antalmanac/src/stores/ReviewPromptStore.ts
```diff
@@ -0,0 +1,193 @@
+import trpc from '$lib/api/trpc';
+import { termData } from '$lib/termData';
+import AppStore from '$stores/AppStore';
+import { create } from 'zustand';
+import { combine } from 'zustand/middleware';
+
+import { openSnackbar } from './SnackbarStore';
```
KevinWu098: nit: absolute imports please

apps/antalmanac/src/stores/ReviewPromptStore.ts
```diff
@@ -0,0 +1,193 @@
...
+    'Mandatory Discussion',
+    'Curved',
+    'Extra Credit',
+    'Project Based',
+    'Test Heavy',
+] as const;
+
+export type ReviewTag = (typeof REVIEW_TAGS)[number];
+
+// Review candidate types
```
KevinWu098: nit: you should do a pass and remove redundant comments

apps/antalmanac/src/stores/ReviewPromptStore.ts
```diff
@@ -0,0 +1,193 @@
...
+export type ReviewCandidate = {
+    /** e.g. "ICS 31" */
+    courseId: string;
+    courseTitle: string;
+    /** Raw WebSOC instructor name, e.g. "PATTIS, R." */
+    professorId: string;
+    /** AntAlmanac term shortName, e.g. "Fall 2024" */
+    term: string;
```
KevinWu098, line 27: question: do we have a term type? if not... maybe we should have one in another PR...

apps/antalmanac/src/backend/routers/review.ts
```diff
@@ -0,0 +1,89 @@
...
+    'Mandatory Lecture',
+    'Mandatory Discussion',
+    'Curved',
+    'Extra Credit',
+    'Project Based',
+    'Test Heavy',
+]);
+
+const submitReviewInput = z.object({
+    userId: z.string(),
```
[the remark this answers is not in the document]
KevinWu098: agree

KevinWu098 (changes requested):
some initial comments. will do more when code is broken up a little bit more

apps/antalmanac/src/components/ReviewPrompt/ReviewPrompt.tsx
```diff
@@ -0,0 +1,259 @@
...
+                </Button>
+                <Button size="small" variant="contained" disabled={rating === 0} onClick={onSubmit} disableElevation>
+                    Submit
+                </Button>
+            </CardActions>
+        </>
+    );
+}
+
+function ratingLabel(rating: number): string {
```
KevinWu098: Uh... is this based on planner? i wouldn't (and i think most users also wouldn't) consider 3 good. moreso, "fair". happy to discuss this more though

KevinWu098:
Screenshot 2026-04-27 at 11 37 47 AM

Elevation is tricky!

Screenshot 2026-04-27 at 11 37 55 AM

Mm... it's kinda small. i know AA has small font, but I think we can do better here for readability.

also, are we not collecting written reviews? could we?

apps/antalmanac/src/components/ReviewPrompt/ReviewPrompt.tsx
AureliaSindhu: changed to snackbar!

KevinWu098 (changes requested):
Overall very good. I think it's just about ready to ship.

However, I'm going to ask that we stop using localStorage to persist dismissed state. We should be storing this in the database.

packages/db/src/schema/review/review.ts
```diff
@@ -0,0 +1,118 @@
...
+         * Nullable here; NOT NULL in Planner — left empty by quick-review UI.
+         */
+        difficulty: integer('difficulty'),
+
+        /** Letter grade received. Not collected in the quick-review UI. */
+        gradeReceived: text('grade_received'),
+
+        createdAt: timestamp('created_at', { withTimezone: true }).defaultNow().notNull(),
+
+        updatedAt: timestamp('updated_at', { withTimezone: true }),
```
KevinWu098: Oh also, append this to `updatedAt`

```
        .notNull()
        .$onUpdate(() => new Date()),
```

apps/antalmanac/src/stores/ReviewPromptStore.ts
```diff
@@ -119,12 +86,14 @@ export const useReviewPromptStore = create(
...
             try {
-                const reviewed = await trpc.review.getReviewedCombos.query();
+                const [dismissed, reviewed] = await Promise.all([
+                    trpc.review.getDismissedCombos.query(),
+                    trpc.review.getReviewedCombos.query(),
+                ]);
+                dismissedSet = new Set(dismissed.map((d) => `${d.courseId}::${d.professorId}`));
                 reviewedSet = new Set(reviewed.map((r) => `${r.courseId}::${r.professorId}`));
             } catch {
                 // Non-fatal — proceed without DB filter.
```
KevinWu098: issue: if we can't filter properly, isn't that actually significant? should we bail out at this point, if we have an error?

KevinWu098 (dismissed):

apps/antalmanac/src/stores/ReviewPromptStore.ts
```diff
@@ -0,0 +1,161 @@
...
+
+            if (pastTermNames.size === 0) return;
+
+            const allCourses = AppStore.schedule.getAllCourses();
+            const seen = new Set<string>();
+            const candidates: ReviewCandidate[] = [];
+
+            for (const course of allCourses) {
+                // Schedule stores terms as "2025-Fall"; termData.shortName uses "2025 Fall".
+                const normalizedTerm = course.term.replace('-', ' ');
```
KevinWu098: question: this is harmless... but as far as I can tell... we store terms in the database as "2025 Fall", and `termData` is also "2025 Fall". can you confirm?

KevinWu098:
@cubic-dev-ai review

alexespejo (approved):

PR #1684, opened 2026-05-02
Planner search followup

Choollol:
## Summary

Some followup changes to Search with Planner.

## Test Plan

- Search with planner works as before
- Course search results are fetched and rendered correctly
- Searching courses with many sections, like Bio 199 Spring 2026, doesn't cause any network errors (previously trpc was batching every single `aggregateGrades` request)

## Issues

Closes #


KevinWu098 (commented):
--- cleared

Choollol:
These seem more related to the planner store changes, so it might be better to revisit this after that is merged?

KevinWu098:
> These seem more related to the planner store changes, so it might be better to revisit this after that is merged?

lol whoops, i meant to edit my initial feedback with that message. sorry one sec

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearchTextFields.tsx
```diff
@@ -40,8 +50,10 @@ function getRoadmapMenuItems({ isLoggedIn, roadmaps }: RoadmapMenuItemsProps) {
 
     if (roadmaps.length === 0) {
         return [
-            <MenuItem key="create" value="" onClick={() => window.open('https://antalmanac.com/planner', '_blank')}>
-                Create a roadmap!
+            <MenuItem key="create" value="">
```
sicn4rf: question: since this select uses `displayEmpty`, making the only no-roadmaps item `value=""` means “Create a roadmap!” can appear as the closed field value. should we keep a normal placeholder/empty display and show the create action only inside the dropdown?

Image

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchWithPlanner.tsx
```diff
@@ -223,13 +237,9 @@ export const SearchWithPlanner = ({ labelProps }: SearchWithPlannerProps) => {
...
-                        <MenuItem
-                            component="a"
-                            href={PLANNER_LINK}
-                            sx={(theme) => ({ color: theme.palette.text.primary, paddingTop: 1.5, paddingBottom: 1.5 })}
-                        >
-                            Create a roadmap!
-                        </MenuItem>
+                        <Link href={PLANNER_LINK} target="_blank">
+                            <MenuItem sx={{ paddingTop: 1.5, paddingBottom: 1.5 }}>Create a roadmap!</MenuItem>
+                        </Link>
```
sicn4rf: question: this renders the no-options action as `<Link><MenuItem>...`, whereas MUI menu items are usually the interactive element themselves. would `MenuItem component={Link} href={PLANNER_LINK}` be a better fit here? ff the blue link styling is intentional, that can still be preserved explicitly.

Image

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
sicn4rf: blocking: this mutates objects inside `websocResp`, which is now React Query cached data. Same for `course.sectionTypes` below. can we clone the course/sections while flattening instead of writing derived fields onto cached query data?

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
```diff
@@ -300,19 +302,73 @@ const ErrorMessage = () => {
...
+            }
+        },
+    });
+
+    const courseData = useMemo(
+        () => (websocResp ? getFilteredCourses(flattenSOCObject(websocResp, courseColors)) : []),
+        [websocResp, courseColors]
+    );
+
+    const andCourseCount = useMemo(() => getFilteredAndCourseCount(courseData, sharedCourseKeys), [courseData]);
```
sicn4rf: issue: missing dependency: this memo reads `sharedCourseKeys`, so it should be included in the dependency array.

apps/antalmanac/src/lib/api/trpc.ts
```diff
@@ -13,6 +13,8 @@ const trpc = createTRPCClient<AppRouter>({
                     credentials: 'include', // Send cookies with requests
                 });
             },
+            // https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/cloudfront-limits.html#limits-general
+            maxURLLength: 8192,
```
sicn4rf, line 17: blocking: this only configures the imperative tRPC client. the failing `grades.aggregateGrades` calls use the React tRPC client from `apps/antalmanac/src/providers/Query.tsx`, so this does not fully fix the staging 494/oversized URL issue. can we apply the same `maxURLLength` there too, preferably via a shared constant?

Image

sicn4rf (changes requested):
thanks for the follow-up here. the `useQuery` refactor is a good direction, but i think there are a few correctness issues to address before merge.

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearchTextFields.tsx
Choollol: Do you mean like white text as before? It shows the same component in both places rn so that part of the behavior is the same

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
Choollol: Do you think it's worth installing Immer, or just `slice()` everything for now?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearchTextFields.tsx
sicn4rf: i see its doing it on prod too. in terms of the text color, did you intend for it to be blue? if so, thats fine we should just use a lighter blue on dark mode for readability. is "Create a roadmap!" a fine default for the closed value ? can we just leave it empty like this maybe:
image

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
sicn4rf: i would probably avoid adding Immer for this. i think the shape here is small enough that a targeted clone is clearer and keeps the dependency surface unchanged.

something like cloning each `course` and mapping its `sections` should be enough, e.g. derive `sectionsWithColors` and `sectionTypes`, then push `{ ...course, sections: sectionsWithColors, sectionTypes }` into the flattened array. i dont think you need to deep-clone the whole WebSOC response.

sicn4rf:
also remind me, are we going to use the "or" split between quick search and roadmap by default now? that's intended right?
image
for me, on prod right now, its split into two separate rows at 100% zoom

also question: the width of the left tab thing on the term selector is drastically different between quick search and manual search, is this intended as well? 
image
image

KevinWu098:
I think the reduced size label is chudded!

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearchTextFields.tsx
Choollol: Design wanted it to be blue and the blue that's there rn is just the `Link` style. We can change it cuz I do agree it is not very readable.

Do you think it's better to leave it empty or do something like this?
image

Choollol:
> I think the reduced size label is chudded!

If it's still too much for you I can reduce the label size less

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearchTextFields.tsx
sicn4rf: i can live with this but i think it looks a bit redundant. id understand for visibilitys sake tho

sicn4rf:
> If it's still too much for you I can reduce the label size less

oh i see, you're reducing it so that it can fit in 1 row. makes sense to me but yeah i do think the mismatch is quite odd.  have you tried making all of it use the reduced size instead? idk the size label is pretty big anyway idk if u need allat space

Choollol:
The General Education and Course Number(s) seem to take up all the space:
image

so I don't think we can shrink them?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearchTextFields.tsx
Choollol: We do have other fields that are empty initially, so I think this is fine actually, though it does change the design from before

Choollol:
I think it looks kind of weird right now on prod, there's a lot of space the labels are taking up that don't really add anything. tho this is my personal taste, so maybe a third opinion would be best

apps/antalmanac/src/stores/PlannerStore.ts
```diff
@@ -59,6 +59,7 @@ export const usePlannerStore = create<PlannerStore>((set, get) => {
         isPlannerLoading: false,
 
         loadPlannerRoadmaps: async () => {
+            return;
```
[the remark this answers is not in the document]
Choollol: oops

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
```diff
@@ -42,31 +43,25 @@ function getColors() {
...
-
-                course.sections.forEach((section) => {
-                    sectionTypesSet.add(section.sectionType);
+                accumulator.push({
+                    ...course,
+                    sections: course.sections.map((section) => ({
+                        ...section,
+                        color: courseColors[section.sectionCode],
+                    })),
+                    sectionTypes: Array.from(new Set((course as AACourse).sectionTypes)),
```
sicn4rf: blocking: gpt 5.5 got mad and said this now derives `sectionTypes` from `(course as AACourse).sectionTypes`, but the raw WebSOC course does not have `sectionTypes` yet. the previous code derived this from `course.sections.map(section => section.sectionType)`. as written, this will likely produce an empty array and break downstream behavior that relies on `courseDetails.sectionTypes` when adding courses / checking missing required section types.


can we derive it from the cloned sections instead?
`sectionTypes: Array.from(new Set(course.sections.map((section) => section.sectionType)))`

sicn4rf (changes requested):

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
Choollol: whoops sorry idk why im missing so many things

Choollol:
@cubic-dev-ai review

sicn4rf (approved):
fire LGTM. thanks for addressing the review comments. i also re-tested the staging 494 repro and no longer see the failing request, and the code-side blockers look resolved.

PR #1688, opened 2026-05-03
feat: union catalogue search index with websoc

KevinWu098:
## What

`get-search-data.ts` unions the **course catalogue** (`GET /v2/rest/courses`) with **one WebSoc schedule**: `GET /v2/rest/websoc` for **`termData[0]`** (latest SOC-available quarter). Courses on WebSoc that are not already in the catalogue get a synthetic autocomplete row with id `ws:{deptCode}:{courseNumber}` so search can reflect the schedule before the catalogue lists the class.

**Dedup (build-time only):** While merging, catalogue coverage is tracked in a `Set` using keys `trim(dept)::trim(courseNumber)`. That string is **not** written to `searchData.ts` or term JSON; persisted course ids remain the catalogue API `id` or the synthetic `ws:…` id above.

Types for the WebSoc payload: **`@packages/anteater-api-types`** (direct dependency on `apps/antalmanac`). **`pnpm-lock.yaml`** includes that workspace link.

## CI / `cache-generated-data`

The action restores **`apps/antalmanac/src/generated`** so old **`terms/*.json`** can be reused. **`tsx scripts/get-search-data.ts` always runs from the checked-out branch**, so script changes always apply. Union logs (`Fetching WebSoc REST for …`, `WebSoc union: …`) confirm that path ran.

## Limits

- Union uses **only** `termData[0]`, not every cached term.
- WebSoc does not expose the catalogue **`course.id`**; overlap is inferred from **dept + course number** (after trim), not from shared ids.

<div><a href="https://cursor.com/agents/bc-5adda475-868a-44dd-ad15-90946af71425"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-5adda475-868a-44dd-ad15-90946af71425"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/apps/antalmanac/scripts/get-search-data.ts
+++ b/apps/antalmanac/scripts/get-search-data.ts
@@ -98,6 +98,11 @@ async function main() {
     await mkdir(join(__dirname, '../src/generated/'), { recursive: true });
     await mkdir(join(__dirname, '../src/generated/terms/'), { recursive: true });
 
+    /*
+     * WebSOC may receive updates sooner than the course catalogue updates, notably for a
+     * new academic year (i.e. Fall term). Querying Websoc to build course data ensures all
+     * available courses are represented.
+     */
     const latestTerm = termData[0];
     const [latestYear, latestQuarter] = latestTerm.shortName.split(' ');
     console.log(`Fetching WebSoc REST for ${latestTerm.shortName} (union with catalogue)...`);
--- a/apps/antalmanac/scripts/get-search-data.ts
+++ b/apps/antalmanac/scripts/get-search-data.ts
@@ -8,6 +8,12 @@ import type {
     DepartmentSearchResult,
     CoursesFilteredAPIResult,
 } from '@packages/antalmanac-types';
+import type {
+    WebsocAPIResponse,
+    WebsocAPIResult,
+    WebsocCourse,
+    WebsocDepartment,
+} from '@packages/anteater-api-types';
 
 import { fetchAnteaterAPI, queryGraphQL } from '../src/backend/lib/helpers';
 import { parseSectionCodes, SectionCodesGraphQLResponse, termData } from '../src/backend/lib/term-section-codes';
@@ -26,8 +32,32 @@ const ALIASES: Record<string, string | undefined> = {
     IN4MATX: 'INF',
 };
 
+function catalogCourseKey(department: string, courseNumber: string) {
+    return `${department.trim()}\0${courseNumber.trim()}`;
+}
+
+function indexWebsocCoursesByCatalogKey(data: WebsocAPIResponse) {
+    const out = new Map<
+        string,
+        Pick<WebsocDepartment, 'deptCode' | 'deptName'> & Pick<WebsocCourse, 'courseNumber' | 'courseTitle'>
+    >();
+    for (const school of data.schools) {
+        for (const dept of school.departments) {
+            for (const course of dept.courses) {
+                const key = catalogCourseKey(dept.deptCode, course.courseNumber);
+                out.set(key, {
+                    deptCode: dept.deptCode,
+                    deptName: dept.deptName,
+                    courseNumber: course.courseNumber,
+                    courseTitle: course.courseTitle,
+                });
+            }
+        }
+    }
+    return out;
+}
+
 async function main() {
-    // Staging/preview: merge or deploy from a feature branch to exercise the full pipeline.
     console.log('Generating cache for fuzzy search.');
     console.log('Fetching courses from Anteater API...');
     const courses: Course[] = [];
@@ -41,6 +71,7 @@ async function main() {
     console.log(`Fetched ${courses.length} courses.`);
     const courseMap = new Map<string, CourseSearchResult & { id: string }>();
     const deptMap = new Map<string, DepartmentSearchResult & { id: string }>();
+    const catalogueKeys = new Set<string>();
     for (const course of courses) {
         courseMap.set(course.id, {
             id: course.id,
@@ -52,6 +83,7 @@ async function main() {
                 number: course.courseNumber,
             },
         });
+        catalogueKeys.add(catalogCourseKey(course.department, course.courseNumber));
         deptMap.set(course.department, {
             id: course.department,
             type: 'DEPARTMENT',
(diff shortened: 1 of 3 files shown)


Choollol:
Are there limitations that prevent AAPI from adding new websoc courses to the courses response? It makes more sense to me that requesting courses would return all courses, rather than having to also request another API.

KevinWu098:
> Are there limitations that prevent AAPI from adding new websoc courses to the courses response? It makes more sense to me that requesting courses would return all courses, rather than having to also request another API.

I would be suspect of asking AAPI to further amend / patch their response in order to serve business (application) logic. Especially because the application-side patch is minimal. 

AAPI is correct here, I think. WebSOC endpoint returns WebSOC and courses endpoint returns the course catalogue. would be curious what @laggycomputer thinks, but I'd imagine he'd feel the same

laggycomputer:
This assessment is correct. We cannot impute new year-scale course data from WebSoc, especially where it does not exist (e.g. repeatability) or conflicts (e.g. pre- and co-requisites).

Choollol:
Ah ok, thanks for clarifying!

apps/antalmanac/scripts/get-search-data.ts
```diff
@@ -26,6 +32,31 @@ const ALIASES: Record<string, string | undefined> = {
     IN4MATX: 'INF',
 };
 
+function catalogCourseKey(department: string, courseNumber: string) {
+    return `${department.trim()}::${courseNumber.trim()}`;
+}
+
+function indexWebsocCoursesByCatalogKey(data: WebsocAPIResponse) {
+    const out = new Map<
```
Choollol, line 40: Could this return just a `Set` instead?

apps/antalmanac/scripts/get-search-data.ts
```diff
@@ -40,6 +71,7 @@ async function main() {
     console.log(`Fetched ${courses.length} courses.`);
     const courseMap = new Map<string, CourseSearchResult & { id: string }>();
     const deptMap = new Map<string, DepartmentSearchResult & { id: string }>();
+    const catalogueKeys = new Set<string>();
```
Choollol, line 74: What's the reason for this? Could we just check course IDs against `courseMap`?

apps/antalmanac/scripts/get-search-data.ts
```diff
@@ -64,16 +97,63 @@ async function main() {
...
+    );
+    const fromWebsoc = indexWebsocCoursesByCatalogKey(websocRest.data);
+
+    let addedFromWebsoc = 0;
+    for (const [key, row] of fromWebsoc) {
+        if (catalogueKeys.has(key)) {
+            continue;
+        }
+        addedFromWebsoc += 1;
+        const id = `ws:${row.deptCode}:${row.courseNumber}`;
```
Choollol: Related to the catalogueKeys comment, why use IDs different from the usual courseId here?

apps/antalmanac/scripts/get-search-data.ts
```diff
@@ -64,16 +97,63 @@ async function main() {
...
+        `https://anteaterapi.com/v2/rest/websoc?${new URLSearchParams({
+            year: latestYear,
+            quarter: latestQuarter,
+        }).toString()}`,
+        { isApiKeyRequired: true }
+    );
+    const fromWebsoc = indexWebsocCoursesByCatalogKey(websocRest.data);
+
+    let addedFromWebsoc = 0;
+    for (const [key, row] of fromWebsoc) {
```
Choollol: Nitpick: Rename `row` to something more readable, like `course`

apps/antalmanac/scripts/get-search-data.ts
```diff
@@ -26,6 +32,31 @@ const ALIASES: Record<string, string | undefined> = {
     IN4MATX: 'INF',
 };
 
+function catalogCourseKey(department: string, courseNumber: string) {
+    return `${department.trim()}::${courseNumber.trim()}`;
+}
+
+function indexWebsocCoursesByCatalogKey(data: WebsocAPIResponse) {
```
Choollol: Might just be a me thing, but I think this function could have a clearer name, maybe `getWebsocCoursesFromResponse`

Choollol (changes requested):
Small things

apps/antalmanac/scripts/get-search-data.ts
KevinWu098: WebSOC does not return course id, so we must construct an intermediary

apps/antalmanac/scripts/get-search-data.ts
KevinWu098: https://anteaterapi.com/reference#tag/websoc/GET/v2/rest/websoc

apps/antalmanac/scripts/get-search-data.ts
KevinWu098: If you feel strongly, yes, this can be written with a Set. But since we do care about the key value (which could be a field on a map, yes), I'm happy with how it currently is

apps/antalmanac/scripts/get-search-data.ts
Choollol: Is it possible for a course id to not follow the format `${course.deptCode.replace(' ', '')}${course.courseNumber}`? If it is, then I think this is fine. If not, could we construct the course id with the info we have from websoc?

KevinWu098: I am... not sure and would not feel good saying it always follows some sort of format

Happy for @laggycomputer to chime in if he knows anything for sure

In any case, I'm far more confident that department and course number are consistent across Websoc and catalogue

Choollol (approved):

PR #1702, opened 2026-05-05
feat: remove (userId, scheduleName) constraint

KevinWu098:
## Summary

As part of #993, a constraint was added to our DB declaring (userId, scheduleName) must be unique. Functionally, neither I nor Cursor can identify any reason why this constraint exists. No frontend or backend caller uses schedule name as an identifier.

The constraint, and associated business logic supporting it (e.g. `mangleScheduleNames`) appear to be uneeded.

A manual migration was created and added to the Drizzle joural as the constraint is not declared in drizzle itself, but rather was inserted manually as part of #1582

## Test Plan

1. Create schedules of the same name
2. Save, load, import schedules of the same name (e.g. if you were signed out and had some unsaved eidts, then logged in)
3. Reorder them, rename them
4. Just generally stress test various features which reference the schedule's name
5. Confirm no problems (e.g. data consistency)

## Issues

Closes #


alexespejo (approved):
lgtrm

PR #1712, opened 2026-05-06
feat: additional reviews

### LINKED ISSUE #1677: Prompt user for additional reviews
After a user reviews, we should have an optional (default off) way for them to provide additional reviews.

victorialee13:
## Summary
Prompts an option for users to be able to submit additional reviews.
After you submit your first review
image
<img width="2000" height="1812" alt="image" src="https://github.com/user-attachments/assets/f26a2aa1-3d6a-4a57-95d7-c2244236cf9d" />

Prompted another class, from the random class list ( as collected from the first review compilation). Order doesn't randomize between reviews.
image

<img width="1332" height="720" alt="image" src="https://github.com/user-attachments/assets/58033c49-4664-40ce-8b4a-2faead32866c" />

Is the modal size for the Success/Additional review popup too small?

Closes #1677

EDIT:

Initial Trigger
- 15s after session is confirmed valid --> initPrompt() builds a shuffled list of eligible candidates from the user's past-term schedule (finals ended, primary sections only, non-STAFF, not already reviewed/dismissed, outside 7-day cooldown)

Review prompt( hidden --> enrollment-confirm
- "Did you take ... in term ... with ...?"
- "I did not" or X --> hidden (dismissal saved to db, cooldown starts)
- "Yes!" --> review                                                                                                                                                                 
                                                                                                                                                                                    
Review (enrollment-confirm --> review)
User fills out Rating, difficulty, text field, tags
- Skip" or X --> hidden (dismissal saved, cooldown starts)
- "Submit" --> success

Submitted (review --> success)
- "Done" or X --> hidden
- "Review another course" --> review directly, skipping enrollment-confirm for the next candidate

So, instead of skipping enrollment-confirm, we can reprompt ...


KevinWu098:
Maybe we can set a consistent width (say... 500) for all of the steps?
Just setting width: 500 should be responsive by default on smaller screens (similar to how popover and GPA are the same size)

apps/antalmanac/src/components/ReviewPrompt/ReviewPrompt.tsx
```diff
@@ -39,9 +40,10 @@ export function ReviewPrompt() {
 
     return (
         <Snackbar open={open} anchorOrigin={{ vertical: 'bottom', horizontal: 'right' }}>
-            <Paper>
+            <Paper sx={{ width: 500, maxWidth: '100vw' }}>
```
KevinWu098: issue: you shouldn't need the max width, I don't think

KevinWu098 (changes requested):
The positioning of the primary action is inconsistent between steps. Also, perhaps we should consider a state machine to handle these steps. 

Relatedly, I'm not a big fan of setting like a bajillion fields per step, but I guess I don't have a solution off the top of my head.

AureliaSindhu (commented):
looks good overall, some comments:

**Re: state machine** — also agree, a proper state machine (e.g. XState or even a simple reducer) would clean this up a lot. The main pain point I see is the repeated full-reset object in `dismiss`, `advanceToNext` fallback, and `finishReviewing`, if the state shape changes again every callsite needs updating. A shared `const RESET_STATE` constant at minimum, or a state machine at most, would fix this.

**Re: primary action positioning** — the `Review another course` button is `variant="contained"` which is good for hierarchy, but it sits to the left of Done, making Done feel like the primary. Typically the primary action goes on the far right. Swapping the order (Done first in DOM, Review another course last) would match the pattern in other steps.

apps/antalmanac/src/components/ReviewPrompt/SuccessStep.tsx
```diff
@@ -0,0 +1,59 @@
...
+    return (
+        <>
+            <CardHeader
+                title={
+                    <Typography variant="subtitle1" fontWeight={600}>
+                        Review submitted!
+                    </Typography>
+                }
+                action={
+                    <IconButton size="small" onClick={dismiss} aria-label="close">
```
AureliaSindhu: Bug: this calls `dismiss` instead of `finishReviewing`. When the user closes via X from the success step, the "Review submitted — thanks for helping other Anteaters!" snackbar never fires and `REVIEW_DONE_CLICKED` analytics never logs. Should be `onClick={finishReviewing}`.

apps/antalmanac/src/stores/ReviewPromptStore.ts
```diff
@@ -141,16 +143,26 @@ export const useReviewPromptStore = create(
                 return;
             }
 
-            const candidate = eligible[Math.floor(Math.random() * eligible.length)];
-            set({ step: 'enrollment-confirm', candidate, rating: 0, difficulty: 0, selectedTags: [], textReview: '' });
+            const shuffled = [...eligible].sort(() => Math.random() - 0.5);
```
AureliaSindhu: `.sort(() => Math.random() - 0.5)` produces a biased shuffle because `Array.sort` doesn't guarantee a consistent number of comparisons per element. For a small list it's unlikely to matter in practice, but a Fisher-Yates shuffle would be correct:
```ts
const shuffled = [...eligible];
for (let i = shuffled.length - 1; i > 0; i--) {
    const j = Math.floor(Math.random() * (i + 1));
    [shuffled[i], shuffled[j]] = [shuffled[j], shuffled[i]];
}
```

apps/antalmanac/src/stores/ReviewPromptStore.ts
```diff
@@ -241,10 +262,55 @@ export const useReviewPromptStore = create(
...
+        advanceToNext: () => {
+            const { eligibleCandidates, eligibleIndex } = get();
+            const nextIndex = eligibleIndex + 1;
+            const next = eligibleCandidates[nextIndex];
+            if (!next) {
+                set({ step: 'hidden', candidate: null, eligibleCandidates: [], eligibleIndex: 0 });
+                return;
+            }
+            set({
+                step: 'review',
```
AureliaSindhu: `advanceToNext` jumps straight to `'review'`, skipping `enrollment-confirm`. Is this intentional? The first candidate goes through the enrollment confirmation step, but subsequent ones don't. If eligible candidates are already pre-validated this is fine, but worth a comment clarifying the intent.

AureliaSindhu (changes requested):

alexespejo:
how's progress here @victorialee13

apps/antalmanac/src/components/ReviewPrompt/ReviewPrompt.tsx
```diff
@@ -39,9 +40,10 @@ export function ReviewPrompt() {
 
     return (
         <Snackbar open={open} anchorOrigin={{ vertical: 'bottom', horizontal: 'right' }}>
-            <Paper>
+            <Paper sx={{ width: 500 }}>
```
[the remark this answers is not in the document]
victorialee13: i just removed that do we need that for sure

AureliaSindhu, line 45: nit: will overflow on narrow screens. Use responsive value like { xs: '90vw', sm: 500 }

apps/antalmanac/src/components/ReviewPrompt/SuccessStep.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+    return (
+        <>
+            <CardHeader
+                title={
+                    <Typography variant="subtitle1" fontWeight={600}>
+                        Review submitted!
+                    </Typography>
+                }
+                action={
```
AureliaSindhu, line 26: `finishReviewing` is called twice for Close and Done button. Consider a separate `REVIEW_SUCCESS_DISMISSED` event for the X button to tell them apart in analytics.

AureliaSindhu: i think u add it in analytics alrdy, so pls match it

AureliaSindhu (changes requested):

apps/antalmanac/src/components/ReviewPrompt/ReviewPrompt.tsx
victorialee13: we decided to not have a max value cus its gonna be capped anyways

apps/antalmanac/src/components/ReviewPrompt/SuccessStep.tsx
victorialee13: Sounds good, I separated the analytics strings to REVIEW_DONE_CLICKED for the Done button and REVIEW_SUCCESS_DISMISSED for pressing X. thx for the comment!

AureliaSindhu (dismissed):
lgtm!

apps/antalmanac/src/stores/ReviewPromptStore.ts
```diff
@@ -38,219 +45,270 @@ const REVIEW_PROMPT_COOLDOWN_MS = 7 * 24 * 60 * 60 * 1000;
...
+                    });
+                } catch {
+                    openSnackbar('error', 'Failed to submit review. Please try again.');
+                }
+            },
+
+            // Intentionally skips enrollment-confirm for subsequent candidates: eligibleCandidates
+            // is already filtered to courses from the user's own schedule, and the user implicitly
+            // opted in to reviewing more by clicking "Review another course".
+            advanceToNext: () => {
```
KevinWu098, line 274: Sorry I'm gonna block this because I'm confused — why would we not confirm if they took this course? Can we update the PR description to be very specific about our state flows and what UI corresponds to what?

KevinWu098 (changes requested):

victorialee13:
Also, I was thinking we modify the No and Yes! Buttons to alternative words
image

image

Also, I feel like having the button a big bigger than the Nope! would push users to fill out reviews.. (?)

apps/antalmanac/src/stores/ReviewPromptStore.ts
victorialee13: Changed the PR description, I think we can have a prompt that asks as well for confirmation, that makes sense. I'll work on that right now so we can probably have a loop flow where we basically go back to the inital trigger popup. What do you think?

victorialee13: ~~Actually, now that I think about it, we can maybe use the Skip function into play and actually just have it go through our list maybe, if they didn't take the course it would just prompt the user the next class in the list. So it would be a continuous form~~

For the inital review I think we can change the button to Cancel instead of Skip as well, ill change some of the buttonsand push that for now

apps/antalmanac/src/components/ReviewPrompt/ReviewStep.tsx
```diff
@@ -91,8 +89,13 @@ export function ReviewStep() {
...
             });
-            resetReview();
-            openSnackbar('success', 'Review submitted — thanks for helping other Anteaters!');
+            useReviewPromptStore.setState({
+                step: 'success',
+                rating: 0,
+                difficulty: 0,
+                selectedTags: [],
+                textReview: '',
+            });
```
KevinWu098: Seems weird to call the store directly here, no?

apps/antalmanac/src/components/ReviewPrompt/EnrollmentConfirmStep.tsx
```diff
@@ -58,11 +58,11 @@ export function EnrollmentConfirmStep() {
...
             <CardActions sx={{ justifyContent: 'flex-end' }}>
                 <Button size="small" color="inherit" onClick={handleDismiss}>
-                    I did not
+                    Nope
                 </Button>
 
                 <Button size="small" variant="contained" onClick={confirm}>
-                    Yes!
+                    Yes, continue!
                 </Button>
```
KevinWu098, line 72: Seems too casual imo. Can we just say

"No, I did not."
"Yes, continue"

apps/antalmanac/src/components/ReviewPrompt/SuccessStep.tsx
```diff
@@ -0,0 +1,63 @@
...
+                </Box>
+            </CardContent>
+
+            <CardActions sx={{ justifyContent: 'flex-end', gap: 1 }}>
+                <Button size="small" color="inherit" onClick={() => finishReviewing()}>
+                    Done
+                </Button>
+
+                {hasMore && (
+                    <Button size="small" variant="contained" onClick={advanceToNext}>
```
KevinWu098: review another course should be the secondary action. not the primary.

apps/antalmanac/src/stores/ReviewPromptStore.ts
```diff
@@ -37,180 +45,272 @@ const REVIEW_PROMPT_COOLDOWN_MS = 7 * 24 * 60 * 60 * 1000;
...
 const initialState = {
     candidate: null as ReviewCandidate | null,
+    eligibleCandidates: [] as ReviewCandidate[],
+    eligibleIndex: 0,
     step: 'hidden' as Step,
     rating: 0,
     difficulty: 0,
     selectedTags: [] as ReviewTag[],
     textReview: '',
 };
```
KevinWu098: This typing is weird — why aren't we just declaring a type so we don't have to cast with `as`

apps/antalmanac/src/stores/ReviewPromptStore.ts
```diff
@@ -37,180 +45,272 @@ const REVIEW_PROMPT_COOLDOWN_MS = 7 * 24 * 60 * 60 * 1000;
...
-
-            for (const course of allCourses) {
-                const term = course.term;
-                const sectionType = course.section.sectionType;
-
-                if (!pastTermNames.has(term)) {
-                    continue;
+    combine(initialState, (set, get) => {
+        const toStep = (next: Step, updates: Partial<typeof initialState> = {}) => {
+            const { step } = get();
```
KevinWu098: this should be omitting step from the `updates` type, no?

apps/antalmanac/src/components/ReviewPrompt/EnrollmentConfirmStep.tsx
```diff
@@ -16,9 +16,9 @@ export function EnrollmentConfirmStep() {
...
 
     const handleDismiss = () => {
-        const candidate = dismiss();
-        if (candidate) {
-            dismissReview({ professorId: candidate.professorId, courseId: candidate.courseId, term: candidate.term });
+        dismiss();
+        if (professorId && courseId && term) {
+            dismissReview({ professorId, courseId, term });
         }
     };
```
KevinWu098, line 29: `dismiss` calls trpc.dismiss but you're also calling it here. pick one place.

apps/antalmanac/src/components/ReviewPrompt/ReviewStep.tsx
victorialee13: Agree, fixed that

AureliaSindhu (dismissed):
lgtm

AureliaSindhu (approved):
lgtm

PR #1715, opened 2026-05-07
fix: open tutorial on page load, not menu open

### LINKED ISSUE #1689: On first visit, tutorial begins when you click the hamburger menu instead of the tutorial button
Steps:
1. open antalmanac scheduler in new private window
2. click on the menu button

expected: nothing happens until I click tutorial button
actual: tutorial starts and the menu remains open

Image

dsnsgithub:
## Summary
Move the `useEffect` and `setIsOpen` to the root component so that the tutorial launches on page load rather than only when you can see the tutorial button in the menu.

## Test Plan
- Open an incognito window, visit AntAlmanac
- Tutorial opens on page load, doesn't automatically open when viewing menu

## Issues

Closes https://github.com/icssc/AntAlmanac/issues/1689

### DIFF
--- a/apps/antalmanac/src/App.tsx
+++ b/apps/antalmanac/src/App.tsx
@@ -1,7 +1,6 @@
 import './App.css';
 import { undoDelete, redoDelete } from '$actions/AppStoreActions';
 import { AutoSignIn } from '$components/AutoSignIn';
-import { TutorialInitializer } from '$components/TutorialInitializer';
 import PosthogPageviewTracker from '$lib/analytics/PostHogPageviewTracker';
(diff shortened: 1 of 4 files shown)


apps/antalmanac/src/App.tsx
```diff
@@ -144,6 +145,7 @@ export default function App() {
                                 }),
                             }}
                         >
+                            <TutorialInitializer />
```
Choollol: Maybe put this in `Home` in case weird things happen where the user first goes to a page that isn't the home page?

Choollol (changes requested):
Generally looks good!

Choollol (approved):
lgtm

PR #1742, opened 2026-05-10
feat: refetch search/WebSOC caches for all active quarters

KevinWu098:
## Summary

`get-search-data.ts` only refreshed WebSOC union data and per-term section caches for `termData[0]`. Generated calendar terms are sorted by **latest instruction start first** (`fetch-term-data.ts`), so during overlapping registration (e.g. Summer sessions plus Fall) only the chronologically latest quarter (Fall) was updated.

This change filters `termData` with the existing **`canTermEnrollmentChange`** (week 2 Friday for 10-week quarters, week 1 Friday for 5-week summer; includes quarters whose instruction has not started yet).

- WebSOC REST catalogue union merges all such quarters (with rate-limit spacing between requests).
- Per-term section JSON files are refreshed when `canTermEnrollmentChange(term.shortName)` is true.
- If no term passes that check, the script logs and **returns immediately** (no catalogue/WebSOC work).

## Test Plan

- `pnpm lint`
- Full vitest run requires generated `src/generated/termData.ts` (not present in this workspace snapshot).

## Issues

<div><a href="https://cursor.com/agents/bc-47cf43de-6274-4b11-a71c-ac4bfafd348e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-47cf43de-6274-4b11-a71c-ac4bfafd348e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/apps/antalmanac/scripts/get-search-data.ts
+++ b/apps/antalmanac/scripts/get-search-data.ts
@@ -57,10 +57,6 @@ async function main() {
     console.log('Generating cache for fuzzy search.');
 
     const activeTerms = termData.filter((t) => canTermEnrollmentChange(t.shortName));
-    if (activeTerms.length === 0) {
-        console.log('No terms in an active enrollment window; skipping search data generation.');
-        return;
-    }
 
     console.log('Fetching courses from Anteater API...');
     const courses: Course[] = [];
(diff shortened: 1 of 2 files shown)


apps/antalmanac/scripts/get-search-data.ts
```diff
@@ -59,6 +55,13 @@ function getWebsocCoursesFromResponse(data: WebsocAPIResponse) {
 
 async function main() {
     console.log('Generating cache for fuzzy search.');
+
+    const activeTerms = termData.filter((t) => canTermEnrollmentChange(t.shortName));
+    if (activeTerms.length === 0) {
+        console.log('No terms in an active enrollment window; skipping search data generation.');
+        return;
+    }
```
sicn4rf: question: when does this actually fire in prod? i would think `termData` always contains some upcoming term whose drop deadline is in the future so this case should effectively be unreachable. 

however, a bigger concern, if this ever does fire, wouldn't it skip the entire catalogue fetch and the `searchData.ts` generation entirely? big correct me if im wrong here, but at build time, are the `searchData.ts` and the per-term section JSON files (e.g. Fall_2026.json) generated fresh each time? if that's the case, then if this branch is reached, all those files will not be generated at build time.

sicn4rf (changes requested):
this LGTM with one concern mentioned. but to answer my own concern, if my belief is true that this branch is unreachable, then it is benign.

apps/antalmanac/scripts/get-search-data.ts
KevinWu098: yeah, there's usually always a term, but `.filter()` doesn't know this to always be true. in any case, you're right about the early return, so its scoped now just to the websoc block

apps/antalmanac/scripts/get-search-data.ts
```diff
@@ -98,66 +97,85 @@ async function main() {
...
-        }).toString()}`,
-        { isApiKeyRequired: true }
-    );
-    const fromWebsoc = getWebsocCoursesFromResponse(websocRest.data);
-
-    let addedFromWebsoc = 0;
-    for (const [key, course] of fromWebsoc) {
-        if (catalogueKeys.has(key)) {
-            continue;
+    if (activeTerms.length > 0) {
```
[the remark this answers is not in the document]
sicn4rf: i disagree. the early return was intentionally removed in the previous commit. the `src/generated/*` directory is in .gitignore and the files get regenerated on every build, so skipping the catalogue fetch + `searchData.ts` generation write would mean producing zero generated files in the edge case where `activeTerms.length === 0`. any downstream imports would fail at build time.

sicn4rf (approved):
LGTM. thank you for this fix and implementing the feedback!

PR #1744, opened 2026-05-10
feat: apple auth

KevinWu098:
## Summary

Adds Apple Sign In support via the `icssc/auth` OIDC server, required by Apple App Store policy 4.8 for apps offering third-party sign-in.

chrome-capture-2026-5-10

### What changed

- **New `APPLE` account type**: Added to the `accountType` enum with a corresponding Drizzle migration (`0016_sparkling_mauler.sql`). This allows a user to have both an `OIDC` (Google) and `APPLE` account row, preventing `providerAccountId` clobbering on multi-provider users.
- **`getAuthUrl` procedure** (renamed from `getGoogleAuthUrl`): Accepts a `provider` field (`'google' | 'apple'`, defaults to `'google'`) and forwards it to `icssc/auth` as a query param. Removed the unnecessary outer `.optional()` on the input schema.
- **`accountTypeFromSub` helper**: Derives the account type from the `sub` claim prefix (`google_*` -> `OIDC`, `apple_*` -> `APPLE`), used in the callback handler when calling `registerUserAccount`.
- **`registerUserAccount` updates** (`rds.ts`): Removed dead `google_` prefix normalization (confirmed all prod OIDC rows are already `google_*` prefixed). Accepts `'APPLE'` as a valid account type. Only fills in `name`, `email`, and `avatar` when the existing user has empty values, preventing richer Google profile data from being overwritten by a sparser Apple profile.
- **Removed Google token logging**: The callback no longer warns on missing `google_access_token`/`google_refresh_token`, since Apple users won't have these.
- **Provider selection UI**: Added "Sign in with Apple" buttons alongside Google in `SignInDialog`, `Signin` (header), and the legacy-ID fallback alert. Both buttons pass the selected provider through `loginUser`.
- **Silent SSO** (`AutoSignIn`): Updated to use `getAuthUrl` (no provider change needed; `prompt=none` is provider-agnostic on the `icssc/auth` side).

### Depends on

- [icssc/auth Apple Sign In support](https://github.com/icssc/auth) (server-side Apple callback, PKCE changes, discriminated union schemas)

## Test Plan

- [x] Sign in with Google -- verify existing flow is unchanged
- [x] Sign in with Apple -- verify new account is created with `APPLE` type
- [x] Sign in with Apple using the same email as an existing Google account -- verify accounts are linked to the same user, and Google profile data (name/avatar) is preserved
- [ ] Silent re-auth (`prompt=none`) continues to work for returning users
- [x] Verify the DB migration applies cleanly (`ALTER TYPE account_type ADD VALUE 'APPLE'`)


apps/antalmanac/src/components/dialogs/SignInDialog.tsx
```diff
@@ -53,18 +53,33 @@ export function SignInDialog(props: SignInDialogProps) {
...
                     >
                         Sign in with Google
                     </Button>
+                    <Button
+                        onClick={() => loginUser({ provider: 'apple', postHog })}
+                        startIcon={<AppleIcon />}
+                        variant="contained"
+                        size="large"
+                        sx={{
+                            backgroundColor: isDark ? '#fff' : '#000',
```
alexespejo: the styling is redundant in SignInDialog.tsx, Signin.tsx or Worth extracting as a shared appleButtonSx constant or a small component to avoid drift

apps/antalmanac/src/components/dialogs/SignInDialog.tsx
```diff
@@ -53,18 +53,33 @@ export function SignInDialog(props: SignInDialogProps) {
                 <Stack spacing={1}>
                     {props.feature === 'Save' && (
                         <Alert severity="info" variant={isDark ? 'outlined' : 'standard'} sx={{ fontSize: 'small' }}>
-                            All changes made will be saved to your Google account
+                            All changes made will be saved to your account
                         </Alert>
                     )}
                     <Button
-                        onClick={() => loginUser(postHog)}
+                        onClick={() => loginUser({ provider: 'google', postHog })}
```
alexespejo: suggestion: I was always mildly dissatisfied that the sign in with Google button didn't look like a sign in with google button 

Image

alexespejo (changes requested):
Smoke test seems chill...

apps/antalmanac/src/components/dialogs/SignInDialog.tsx
KevinWu098: I (cursor) was playing with more black and white versions of google sign in, but I like the blue

image

WDYT @alexespejo

KevinWu098: If you're still on the fence about Google, I'm gonna invoke "in another PR please" on you 👅

alexespejo: image
hb this one of these :p

KevinWu098: :squint: im still gonna invoke it, but i feel like our button is pretty close to that already? i would rather encourage people via our primary blue to use google

KevinWu098: Screenshot 2026-05-11 at 5 42 02 PM

e.g. chatgpt uses the colored icon, but then an on brand button

alexespejo (approved):
LGTM

PR #1754, opened 2026-05-12
White Text on Notifications

alexespejo:
## Summary
image

image

## Issues

Closes #


alexespejo:
wait wtf

apps/antalmanac/src/components/NotificationSnackbar.tsx
```diff
@@ -40,9 +40,9 @@ export const NotificationSnackbar = () => {
                 severity={severity}
                 variant="filled"
                 onClose={handleClose}
-                sx={(theme) => ({
+                sx={() => ({
                     width: '100%',
-                    color: theme.palette.text.primary,
+                    color: '#fff',
```
Choollol: maybe `color: theme.palette.common.white` instead

Choollol (changes requested):

apps/antalmanac/src/components/NotificationSnackbar.tsx
sicn4rf: nit: why not keep theme and then just do `theme.palette.common.white` pretty sure this ships as a default in MUI theme. doesnt change anything functionally but might be more explicit than the hex value

sicn4rf (dismissed):
address/dismiss nits and then merge lgtm

apps/antalmanac/src/components/NotificationSnackbar.tsx
sicn4rf: oh we said the same thing

Choollol: twin

sicn4rf (approved):
fire

Choollol (approved):

PR #1771, opened 2026-05-13
feat: friends backend

calebwongg:
## Summary
Takes all of the friends backend (friends schema, migrations, etc.) and ports it into this pr, this way high-visibility files like rds.ts aren't continuously getting refactored because of merge conflicts.

## Issues

Closes #1511


apps/antalmanac/src/backend/lib/rds.ts
```diff
@@ -75,10 +76,27 @@ export class RDS {
...
+     */
+    static async getGoogleIdByUserId(db: DatabaseOrTransaction, userId: string): Promise<string | null> {
+        return db
+            .select({ providerAccountId: accounts.providerAccountId })
+            .from(accounts)
+            .where(eq(accounts.userId, userId))
+            .limit(1)
+            .then((res) => (res.length > 0 ? res[0].providerAccountId : null));
+    }
+
```
brelieu05: This looks like dead code to me. I grepped for getGoogleIdByUserId and only found the definition. 

Alex says to delete 👍

apps/antalmanac/src/backend/lib/rds.ts
```diff
@@ -87,12 +105,14 @@ export class RDS {
         email?: string,
         avatar?: string
     ) {
-        if (accountType !== 'OIDC' && accountType !== 'APPLE') {
-            throw new Error('Invalid account type. Must be OIDC or APPLE.');
+        if (accountType !== 'OIDC') {
+            throw new Error('Invalid account type. Must be OIDC.');
```
brelieu05: I don't think we should be removing the APPLE accountType check since people are logging in with apple now

brelieu05 (changes requested):

apps/antalmanac/src/backend/lib/rds.ts
calebwongg: are people logging in with apple? because people are on ios but they still log in with google right.

calebwongg: nevamind

brelieu05 (approved):

PR #1773, opened 2026-05-13
Always anchor schedule select popover

Choollol:
## Summary

Allows schedule select to open correctly without being clicked on.

Originally part of friends but that got pushed back so I think it would be better to have this as a separate PR.

## Test Plan

Open schedule select and see that it's anchored to the button as before

## Issues

Closes #


KevinWu098 (commented):
need more context. why are we adding a useEffect (an extreme sin)? what is before/after behavior?

Choollol:
Before (currently), if the schedule select opens without the user clicking on it, the anchor element isn't set so it opens in the wrong position. With this change, the schedule select can open correctly without the user clicking on it.

Now that I'm thinking about it, this can probably stay in the friends PR. tbh idk what i was thinking this morning but i'll close this

PR #1780, opened 2026-05-14
feat: display warning for cancelled courses

### LINKED ISSUE #1775: Cancelled classes are unhandled
Several courses (such as FLM&MDA H80) have been cancelled and no longer show up on websoc, but are still shown as open on scheduler

alexespejo:
## Summary
Cancelled classes still appear in our search results as if they were offered. This PR explores how to show the user this discrepancy

Current iteration: 

### A cancelled class will still appear in the search results
image

### When every section is cancelled (i.e the whole class is cancelled) 
image

### When only specifc sections are cancelled 
image

The search results will still appear as if it was offered and the user can interact with it normally but a warning will appear
image
 
## Issues

Closes #1775


apps/antalmanac/src/lib/courseAvailability.ts
```diff
@@ -0,0 +1,23 @@
+export function isSectionCancelled(section: { isCancelled?: boolean }) {
```
calebwongg: nit: our AASection already guarantees that we will always have the isCancelled attribute, right? So since it'll always be true or false we don't need to guard against it being undefined. That way when we check
`const cancelledSections = sectionList.filter(isSectionCancelled);`

we can change this to `const cancelledSections = sectionList.filter(section => section.isCancelled);`, and then we can delete the whole isSectionCancelled function

apps/antalmanac/src/lib/courseAvailability.ts
```diff
@@ -0,0 +1,23 @@
...
+    return section.isCancelled === true;
+}
+
+export function getCourseCancellationWarning(
+    sections: Iterable<{
```
calebwongg: nit: we don't need to widen the type to iterable here, since the only caller passes ` which is already a `AASection[]`. 

If we accept an array directly we can delete the spread into sectionList and just use sections directly. So the function head would be `getCourseCancellationWarning(sections: AASection[])`

calebwongg (changes requested):
looks good :D. two small little nitpicks

calebwongg (approved):
👍👍👍

PR #1784, opened 2026-05-15
refactor: term and quarter type safety

### LINKED ISSUE #1768: Type terms with year and quarter
A little bit of string parsing while waiting on AAPI https://github.com/icssc/anteater-api/issues/400

KevinWu098:
## Summary

This change tightens how we represent **academic terms** and **quarters** across the monorepo: in-app state prefers structured types (notably **`AATerm`** and the Anteater **`Quarter`** union), while **wire formats, DB fields, and legacy strings** are converted at clear boundaries with **schema-backed parse/serialize** so bad data fails fast instead of spreading as untyped `string`.

### Shared types and WebSOC inputs

- **`@packages/antalmanac-types`**: extend calendar/term modeling (e.g. **`Quarter` / `QuarterSchema`**, **`AATerm`**) and strengthen **`WebsocSearchInput`** (Zod) so search and API layers agree on valid quarters and query shapes.
- **`@packages/anteater-api`**: re-export or align calendar-related types where the client talks to WebSOC.

### App & stores (Antalmanac)

- Central **term helpers** (new `term` module) replace ad-hoc string assembly; **`termData`** is slimmed so consumers resolve **`AATerm`** consistently.
- **Schedules, calendarization, downloads (ICS), planner, notifications, and section/course UI** now thread **`AATerm` / `Quarter`** where the domain is a real term, and use explicit serialization when persisting or calling APIs.
- **Right pane / added courses**: course rows keyed and merged by stable term identity (e.g. **`shortName`**) and **`SectionTable`** continues to receive a real **`AATerm`** for search and related actions.
- **Backend tRPC**: **search** and **websoc** routers validate inputs against the shared Zod shapes; **review** and related DB touchpoints keep **quarter/term** fields consistent with the stricter types.

### AANTS (notifications worker)

- **`Quarter`** is used end-to-end for WebSOC fetches and email props instead of loose **`string`**.
- Subscription grouping no longer relies on a **`quarter-year`** delimiter string and **`split`**: DB rows are grouped into **`TermGroup`**s with **`quarter: Quarter`** and **`year`**, with **`QuarterSchema.safeParse`** skipping invalid rows.

### Review prompt (related cleanup)

- Store flow supports a dedicated **success** step, **eligible candidate queue**, and **state-machine style** step transitions; **dismiss** can persist server-side via **`dismissReview`** from the store. Review candidates still carry **`term` as a string** (AntAlmanac **`shortName`**) for API compatibility—aligned with the “serialize at the edge” approach.

## Test plan

- [ ] **Search / WebSOC**: run queries across terms; confirm term selector and multi-course search still match router **Zod** input.
- [ ] **Schedule**: add/remove sections, switch schedules, import; confirm **added courses** list and **ICS download** behave as before.
- [ ] **Calendar / finals**: events and finals still align with **`AATerm`** dates from **term data**.
- [ ] **Notifications**: subscription create/update paths; optional smoke of **AANTS** batch path if you have a staging harness.
- [ ] **Review prompt**: enrollment confirm → review → submit → **success**; dismiss/skip at each step; verify **dismissReview** / cooldown behavior.
- [ ] **Unit tests**: run CI or `pnpm test` for touched packages—especially **term**, **download/ICS**, **calendarize**, and **schedule** tests updated in this branch.

## Issues

Closes #1768


apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearchTextFields.tsx
```diff
@@ -39,11 +40,7 @@ function getRoadmapMenuItems({ isLoggedIn, roadmaps }: RoadmapMenuItemsProps) {
     }
 
     if (roadmaps.length === 0) {
-        return [
-            <MenuItem key="create" value="" onClick={() => window.open('https://antalmanac.com/planner', '_blank')}>
-                Create a roadmap!
-            </MenuItem>,
-        ];
+        return <CreateRoadmapLinkItem verticalPadding={'6px'} />;
```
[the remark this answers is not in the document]
Choollol: this one's my fault i'll make a PR for it

apps/antalmanac/src/lib/term.ts
```diff
@@ -0,0 +1,114 @@
...
+    const [y, month, day] = dateStr.split('-').map(Number);
+    return new Date(y, month - 1, day);
+}
+
+const termSchema = z
+    .object({
+        year: z.string(),
+        quarter: QuarterSchema,
+        shortName: z.string().refine(
+            (s): s is `${string} ${(typeof QUARTERS)[number]}` => {
```
Choollol: Could we use `AATerm['shortName']` instead of `${string} ${(typeof QUARTERS)[number]}` here?

packages/anteater-api/src/types/calendar.ts
Choollol, line 1: It might be worth exporting the `Year` type too. It doesn't change anything, since it's still just a string, but it'd be more readable and easier to change in case it ever does change down the line.

packages/types/src/websoc.ts
```diff
@@ -36,3 +44,28 @@ type AACourseExtendedProperties = {
...
+    division: z.string().optional(),
+    sectionType: z.string().optional(),
+    fullCourses: z.string().optional(),
+    cancelledCourses: z.string().optional(),
+    units: z.string().optional(),
+    startTime: z.string().optional(),
+    endTime: z.string().optional(),
+    excludeRestrictionCodes: z.string().optional(),
+    includeRelatedCourses: z.string().nullable().optional(),
+}) satisfies z.ZodType<{ [K in keyof WebsocQueryParams]: unknown }>;
```
Choollol: We should prob have a safer type here and check the values too
Maybe something like this?
```ts
[K in keyof WebsocQueryParams]: NonNullable<WebsocQueryParams[K]> extends string
        ? string | WebsocQueryParams[K]
        : WebsocQueryParams[K];
```

apps/antalmanac/src/backend/routers/websoc.ts
```diff
@@ -1,71 +1,61 @@
...
     }
-    if ('courseNumber' in p) {
-        p.courseNumber = p.courseNumber.toUpperCase();
+    if (courseNumber) {
+        sanitized.courseNumber = courseNumber.toUpperCase();
     }
-    for (const [key, value] of Object.entries(p)) {
-        if (value === '') delete p[key];
+    for (const [key, value] of Object.entries(sanitized)) {
+        if (value === '' || value == null) delete sanitized[key];
```
Choollol: Maybe `if (!value)` would be more concise?

apps/antalmanac/src/backend/routers/websoc.ts
```diff
@@ -1,71 +1,61 @@
...
-    }
-    if ('department' in p) {
-        if (p.department.toUpperCase() === 'ALL') {
-            delete p.department;
-        } else {
-            p.department = p.department.toUpperCase();
-        }
+function sanitizeWebsocParams(params: WebsocSearchInput): WebsocQueryParams {
+    const { department, courseNumber, ...rest } = params;
+    const sanitized: Record<string, string | null | undefined> = { ...rest };
```
Choollol: This loses a bit of type safety, could we do something like `const sanitized: typeof params`? Then maybe either a `delete sanitized[key as keyof typeof sanitized]` below or use a type-widening technique similar to my other comment

apps/antalmanac/scripts/update-terms.ts
```diff
@@ -18,8 +19,13 @@ interface DeployedTermsData {
...
-    const [year, quarter] = term.shortName.split(' ');
-    if (!year || !quarter) throw new Error(`Invalid term format: ${term.shortName}`);
+    const aaTerm = getTermByShortName(term.shortName);
+
+    if (!aaTerm) {
+        throw new Error(`Unknown term: ${term.shortName}`);
+    }
+
+    const { year, quarter } = aaTerm;
     console.log(`Checking section count for ${year} ${quarter}...`);
```
Choollol: Could this just be ```console.log(`Checking section count for ${aaTerm.shortName}...`);```?

apps/antalmanac/src/components/RightPane/RightPaneStore.ts
```diff
@@ -99,25 +95,27 @@ class RightPaneStore extends EventEmitter {
     }
 
     updateFormDataFromURL = (search: URLSearchParams) => {
-        const formFields = Object.keys(defaultFormValues) as CourseSearchParamKey[];
+        const paramTerm = search.get('term') || search.get('TERM');
```
Choollol, line 98: Do you know when it's possible for the param key to be in all caps? I saw that the previous logic checks for all caps too

apps/antalmanac/src/components/RightPane/RightPaneStore.ts
```diff
@@ -27,18 +20,21 @@ const defaultAdvancedSearchValues: Record<AdvancedSearchParam, string> = {
     days: '',
 };
 
-const defaultFormValues: Record<ManualSearchParam, string> = {
+export interface CourseSearchParams extends Record<Exclude<ManualSearchParam, 'term'>, string> {
+    term: AATerm;
+}
+
+type StringSearchParamKey = Exclude<keyof CourseSearchParams, 'term'>;
```
Choollol: Maybe just `SearchParamKey`? The `String` feels a little repetitive

apps/antalmanac/src/lib/helpers.ts
```diff
@@ -1,12 +1,14 @@
 import { openSnackbar } from '$stores/SnackbarStore';
 import { Theme } from '@mui/material/styles';
 import { SxProps, SystemStyleObject } from '@mui/system';
+import type { AATerm } from '@packages/antalmanac-types';
 import { MouseEvent } from 'react';
 
-export const warnMultipleTerms = (terms: Set<string>) => {
+export const warnMultipleTerms = (terms: Set<AATerm>) => {
+    const names = [...terms].map((t) => t.shortName).sort();
```
Choollol, line 8: Is this spread still necessary?

apps/antalmanac/src/lib/notifications.ts
```diff
@@ -2,17 +2,32 @@ import trpc from '$lib/api/trpc';
...
+        quarter: notification.term.quarter,
+        sectionCode: notification.sectionCode,
+        courseTitle: notification.courseTitle,
+        sectionType: notification.sectionType,
+        lastUpdatedStatus: notification.lastUpdatedStatus,
+        lastCodes: notification.lastCodes,
+        notifyOn: notification.notifyOn,
+    };
+}
+
```
Choollol, line 14: If the only difference here between input and output is term, we can probably use destructuring/spreading to reduce work needed when `Notification` changes. Maybe something like 
```ts
const {term, ...rest} = notification;
return {
  ...rest,
  year: term.year,
  quarter: term.quarter,
}
```

apps/antalmanac/src/lib/term.ts
Choollol: Possibly also leave a todo comment to convert this to [`z.literal.template()`](https://github.com/colinhacks/zod/pull/1786) when we upgrade to zod v4?

apps/antalmanac/src/stores/NotificationStore.ts
```diff
@@ -166,43 +161,53 @@ export const useNotificationStore = create<NotificationStore>((set) => {
...
+                    const key = term.shortName;
+                    const group = termGroups.get(key);
+                    if (group) {
+                        group.sectionCodes.add(sectionCode.toString());
                     } else {
-                        courseDict[term] = new Set([sectionCode.toString()]);
+                        termGroups.set(key, {
+                            term,
+                            sectionCodes: new Set([sectionCode.toString()]),
+                        });
```
Choollol: Optional: Could deduplicate a tiny bit of work here by checking if key doesn't exist first, then adding empty value

apps/antalmanac/src/stores/NotificationStore.ts
```diff
@@ -166,43 +161,53 @@ export const useNotificationStore = create<NotificationStore>((set) => {
...
+                            sectionCodes: Array.from(sectionCodes).join(','),
+                        });
+                        courseInfoDict.set(key, { term, courseInfo });
+                    })
+                );
 
                 const notifications: Partial<Record<string, Notification>> = {};
 
-                for (const [term, courseInfo] of courseInfoDict.entries()) {
+                for (const [, { term, courseInfo }] of courseInfoDict) {
```
Choollol: Could this iterate over `courseInfoDict.values()` instead?

apps/antalmanac/src/routes/UnsubscribePage.tsx
```diff
@@ -8,7 +9,7 @@ export const Unsubscribe = () => {
     const [searchParams] = useSearchParams();
 
     const sectionCode = searchParams.get('sectionCode');
-    const quarter = searchParams.get('quarter');
+    const _quarter = searchParams.get('quarter');
```
Choollol: Maybe `searchParamQuarter` instead? The underscore might make it seem like it's unused at first

apps/antalmanac/src/lib/term.ts
```diff
@@ -0,0 +1,114 @@
...
+        }
+    }
+    return termData[defaultTermIndex];
+}
+
+export function getTermByShortName(termShortName: string): AATerm | undefined {
+    return termData.find((t) => t.shortName === termShortName);
+}
+
+export function parseQuarter(rawQuarter: unknown) {
```
Choollol, line 78: Was there a particular reason this is `unknown` rather than a form of nullable `string`?

apps/antalmanac/src/lib/term.ts
```diff
@@ -0,0 +1,114 @@
...
+    const quarter = QuarterSchema.safeParse(rawQuarter);
+    return quarter.success ? quarter.data : undefined;
```
Choollol, line 80: Was there a particular reason for this returning `undefined` rather than `null`? It probably won't make a difference but I think usually `null` is used for an intentional absence of value

Choollol (changes requested):
Glad this problem is getting addressed, mostly small things and nitpicks

One more thing to consider:
Renaming db columns to fit these changes, like `quarter` and `term` that actually mean term short names. Probably a separate PR thing though

apps/antalmanac/src/backend/routers/websoc.ts
KevinWu098: I believe this carve out is because technically values such as "0" or "false" could be valid and end up filtered out. I've made this more explicit by adding undefined to the conditional (which `== null` was covering)

apps/antalmanac/src/components/RightPane/RightPaneStore.ts
KevinWu098: /shrug no idea — doesn't hurt i suppose. i have a NUQS PR brewing to refactor rightpanestore and clean up a lot of annoying bugs in advanced search, so I'll deal with it then

apps/antalmanac/src/lib/helpers.ts
KevinWu098: Turns a set into an array which can be sorted

apps/antalmanac/src/lib/term.ts
KevinWu098: Probably because Zod will validate anything, so it doesn't really matter. Saves us the trouble of updating this util if something somewhere thinks rawQuarter could be undefined, i guess

apps/antalmanac/src/lib/term.ts
KevinWu098: Probably because its neighboring utils use `.find()` which returns `undefined`

apps/antalmanac/src/backend/routers/websoc.ts
Choollol: Gotcha, thanks for the explanation

apps/antalmanac/src/lib/helpers.ts
Choollol: Ah right, missed that

packages/types/src/calendar.ts
```diff
@@ -0,0 +1,26 @@
...
+    'Spring',
+    'Summer1',
+    'Summer10wk',
+    'Summer2',
+] as const satisfies readonly Quarter[];
+
+export const QuarterSchema = z.enum(QUARTERS);
+
+export type AATerm = {
+    year: string;
```
Choollol: See main comment

apps/antalmanac/src/backend/lib/rds.ts
```diff
@@ -1131,7 +1132,8 @@ export class RDS {
         db: DatabaseOrTransaction,
         userId: string,
         sectionCode: string,
-        term: string,
+        year: string,
```
Choollol: See main comment

packages/types/src/calendar.ts
```diff
@@ -0,0 +1,26 @@
...
+    quarter: Quarter;
+    shortName: `${string} ${Quarter}`;
```
Choollol: See main comment

Choollol (changes requested):
Overall LGTM, besides a few places missing the `Year` type

Choollol (approved):
awesome

PR #1786, opened 2026-05-15
Use local mergeSx utility

Choollol:
## Summary

Import local mergeSx function instead of using mui internals

## Issues

Closes #

### DIFF
--- a/apps/antalmanac/src/lib/helpers.ts
+++ b/apps/antalmanac/src/lib/helpers.ts
@@ -17,6 +17,11 @@ export async function clickToCopy(event: MouseEvent<HTMLElement>, sectionCode: s
     openSnackbar('success', 'WebsocSection code copied to clipboard');
 }
 
+/**
+ * Merges MUI sx props. Later styles override earlier ones.
+ *
+ * Taken from [MUI internals](https://github.com/mui/mui-x/blob/master/packages/x-date-pickers/src/internals/utils/utils.ts)
+ */
 export function mergeSx(
     ...sxProps: (SxProps<Theme> | undefined)[]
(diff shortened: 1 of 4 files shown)


apps/antalmanac/src/lib/helpers.ts
```diff
@@ -15,6 +17,20 @@ export async function clickToCopy(event: MouseEvent<HTMLElement>, sectionCode: s
...
+): ReadonlyArray<boolean | SystemStyleObject<Theme> | ((theme: Theme) => SystemStyleObject<Theme>)> {
+    return sxProps.reduce((acc, sxProp) => {
+        if (Array.isArray(sxProp)) {
+            acc.push(...sxProp);
+        } else if (sxProp != null) {
+            acc.push(sxProp);
+        }
+
+        return acc;
+    }, [] as any);
```
KevinWu098, line 36: kind of sus of this any? what's going on here? what's the internal MUI implementation of mergeSx?

KevinWu098 (changes requested):

sicn4rf (dismissed):
LGTM, thanks for pulling this into a local helper. this removes the dependency on MUI internals and makes the follow-up drag-and-drop PR cleaner too.

apps/antalmanac/src/lib/helpers.ts
Choollol: This is copy pasted from https://github.com/mui/mui-x/blob/d9c7b7a3ee93fb8cbc290c34da2bc9e6f7ef33a4/packages/x-date-pickers/src/internals/utils/utils.ts#L91

KevinWu098: throw a comment on top of it and its good by me

KevinWu098 (approved):

PR #1788, opened 2026-05-15
fix: Pass value to CreateRoadmapLinkItem

Choollol:
## Test Plan

No errors/warnings about controlled selects or values

## Issues

Closes #


sicn4rf (approved):
LGTM, thanks for catching this bug. fix looks clean to me.

PR #1794, opened 2026-05-16
fix: extract parseTermShortName helper for update-terms script

sicn4rf:
`update-terms.ts` is the failure point of the `check-new-quarter` github actions workflow to check for new quarters/section changes to rebuild the app.

`update-terms.ts` imported `getTermByShortName` from `src/lib/term.ts`, which in turn imports `$generated/termData.json` and `$components/...` via path aliases that tsx cannot resolve at runtime. WebsocTerm only carries `shortName`/`longName` (not year/quarter), so we parse those out of `shortName` instead.

Extracted into a new `src/lib/termHelpers.ts` so both `term.ts` (zod refine) and `update-terms.ts` share one `parseTermShortName` helper instead of duplicating `split(' ')` logic. Per review, follow-up PR can move more termData-independent helpers out of `term.ts`.

thanks ai
## Summary

- new `src/lib/termHelpers.ts` with `parseTermShortName(shortName) -> { year, quarter } | undefined`
- `term.ts` zod refine uses it
- `update-terms.ts` uses it (no more transitive `$generated`/`$components` imports)

## Test Plan

- [x] `tsc --noEmit` clean for `scripts/update-terms.ts`
- [x] staging deploy goes green

## Issues

Closes #

### DIFF
--- a/apps/antalmanac/src/lib/term.ts
+++ b/apps/antalmanac/src/lib/term.ts
@@ -1,9 +1,11 @@
 import type { CourseEvent, CustomEvent } from '$components/Calendar/CourseCalendarEvent';
 import termJson from '$generated/termData.json';
-import { QuarterSchema, QUARTERS, type AATerm } from '@packages/antalmanac-types';
+import { QuarterSchema, type AATerm } from '@packages/antalmanac-types';
 import { Year } from '@packages/anteater-api/types';
 import { addWeeks, differenceInWeeks, setDay } from 'date-fns';
 import { z } from 'zod';
+
+import { parseTermShortName } from './termHelpers';
 
 export type { AATerm } from '@packages/antalmanac-types';
 
@@ -21,13 +23,11 @@ const termSchema = z
         year: z.string(),
         quarter: QuarterSchema,
         // TODO(zod-v4): Replace refine with z.literal.template() once we upgrade to Zod v4.
-        shortName: z.string().refine(
-            (s): s is AATerm['shortName'] => {
-                const parts = s.split(' ');
-                return parts.length === 2 && (QUARTERS as readonly string[]).includes(parts[1]);
-            },
-            { message: 'shortName must be "<year> <quarter>"' }
-        ),
+        shortName: z
+            .string()
+            .refine((s): s is AATerm['shortName'] => parseTermShortName(s) !== undefined, {
+                message: 'shortName must be "<year> <quarter>"',
+            }),
         longName: z.string(),
         instructionStart: z.string().date(),
         instructionEnd: z.string().date(),
--- a/apps/antalmanac/src/lib/term.ts
+++ b/apps/antalmanac/src/lib/term.ts
@@ -1,11 +1,10 @@
 import type { CourseEvent, CustomEvent } from '$components/Calendar/CourseCalendarEvent';
 import termJson from '$generated/termData.json';
+import { parseTermShortName } from '$lib/termHelpers';
 import { QuarterSchema, type AATerm } from '@packages/antalmanac-types';
 import { Year } from '@packages/anteater-api/types';
 import { addWeeks, differenceInWeeks, setDay } from 'date-fns';
 import { z } from 'zod';
-
-import { parseTermShortName } from './termHelpers';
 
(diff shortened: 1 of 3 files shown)


sicn4rf:
frick

Choollol (changes requested):
Could we instead extract shared logic into maybe a `termHelpers.ts` and use that in `terms.ts` and `update-terms.ts`?

We could make a `parseTermShortName` or something that returns the year and quarter, then use that instead of `getTermByShortName` in `update-terms.ts` and the `split(' ')` in `term.ts`.

We could also move some functions that don't need to access `termData` from `term.ts` over in a new PR

sicn4rf:
> Could we instead extract shared logic into maybe a `termHelpers.ts` and use that in `terms.ts` and `update-terms.ts`?
> 
> 
> 
> We could make a `parseTermShortName` or something that returns the year and quarter, then use that instead of `getTermByShortName` in `update-terms.ts` and the `split(' ')` in `term.ts`.
> 
> 
> 
> We could also move some functions that don't need to access `termData` from `term.ts` over in a new PR


yea true this makes sense. I'm happy to work on this

sicn4rf:
i think this is chill now @Choollol, gonna put up a PR for decoupling functions from `term.ts` shortly too

apps/antalmanac/src/lib/termHelpers.ts
```diff
@@ -0,0 +1,13 @@
+import { QuarterSchema } from '@packages/antalmanac-types';
+import type { Quarter, Year } from '@packages/anteater-api/types';
+
+export type TermParts = { year: Year; quarter: Quarter };
+
+export function parseTermShortName(shortName: string): TermParts | undefined {
+    const parts = shortName.split(' ');
+    if (parts.length !== 2) return undefined;
```
Choollol: nit: I'd be more comfortable if we used brackets around if blocks, it's more readable to me.

apps/antalmanac/src/lib/termHelpers.ts
```diff
@@ -0,0 +1,13 @@
+import { QuarterSchema } from '@packages/antalmanac-types';
+import type { Quarter, Year } from '@packages/anteater-api/types';
+
+export type TermParts = { year: Year; quarter: Quarter };
```
Choollol: Is this type used anywhere outside this module?

apps/antalmanac/src/lib/termHelpers.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export function parseTermShortName(shortName: string): TermParts | undefined {
```
Choollol: nit: Consider returning null instead to purposefully indicate that something is missing

apps/antalmanac/src/lib/term.ts
```diff
@@ -1,10 +1,12 @@
 import type { CourseEvent, CustomEvent } from '$components/Calendar/CourseCalendarEvent';
 import termJson from '$generated/termData.json';
-import { QuarterSchema, QUARTERS, type AATerm } from '@packages/antalmanac-types';
+import { QuarterSchema, type AATerm } from '@packages/antalmanac-types';
 import { Year } from '@packages/anteater-api/types';
 import { addWeeks, differenceInWeeks, setDay } from 'date-fns';
 import { z } from 'zod';
 
+import { parseTermShortName } from './termHelpers';
```
Choollol: Semiblocking: Why the relative import here?

Choollol (changes requested):
Thanks for getting the new changes in, just some small things

apps/antalmanac/src/lib/termHelpers.ts
sicn4rf: i chose to return undefined here because similar getter functions in `term.ts` (`getTermByShortName`, `parseQuarter`, `getTermByYearAndQuarter`) also return undefined. changing this to `null` may introduce inconsistencies.

apps/antalmanac/src/lib/term.ts
sicn4rf: woops

Choollol (approved):
LGTM!

PR #1799, opened 2026-05-17
refactor: Order added course sections during db selection instead of manually sorting

Choollol:
## Summary

Create a centralized function used to get user schedule data. Then move course sorting by index from manual implementation to using `orderBy`.

## Test Plan

1. Schedules load and save properly
2. Course reordering saves and loads

## Issues

Closes #


KevinWu098:
Think some of this was handled in #1821, but i think the order by improvements are still valid?

KevinWu098 (approved):
hype I love removing code

PR #1800, opened 2026-05-17
refactor: ge multi search

KevinWu098:
### WebSOC API (tRPC + types)

- **`getManyOfField`** is restricted to **`ge`** via `WebsocSearchInputKeysSchema` and returns **`WebsocAPIResponse[]`**—one sorted response per comma-separated GE value—so the client can compute intersection vs rest instead of getting a pre-merged tree.
- **`combineWebsocResponses`** is renamed **`mergeWebsocUnion`**; **`getMultiple`** uses it for deep union (courses + section dedupe).

Kevin: As a thought, the explosion of queries if we allowed M x M queries is probably untenable. Querying multiple GEs is already rather unperformant because every course must then perform an additional query in order to get the associated non-lecture sections (which do not have a GE assigned to them, see GEDataFetchProvider). Imagine if we allowed further queries that were `OR`

### Shared utils (`@packages/anteater-api`)

- **`websocCourseKey`** centralizes stable dept+course identity for set logic.
- **`splitWebsocIntersectAndRest`** turns N WebSOC responses into **`{ intersect, rest }`**: courses in every response (first response’s layout) vs courses only in some (deduped), for multi-GE AND/OR presentation.
- **`flattenSectionsWithCourse`** reuses **`flattenCourses`** (less duplicated walking).

Kevin: It... is weird that we were combining the intersection and the non-intersected courses, then having to figure out where they were being combined. I... frankly think it was the wrong decision to allow returning "OR" and it really complicates our queries. I would like to reduce the functionality of this feature in another PR.

### Course pane / manual search

- Search **`useQuery`** returns a small discriminant: **`{ kind: 'single', response }`** (planner multi-course path) or **`{ kind: 'split', intersect, rest }`** (manual search with `getManyOfField` + split).
- Results render as **intersect list → optional “at least one GE” alert → rest list**; removed **`sharedCourseKeys`**, banner index scanning, and **`queryManualSearchCourses`** indirection.
- **`getQueryParams`** is hoisted ahead of the query; manual branch inlines **`getSelectedGEs`**, **`getManyOfField`**, and **`splitWebsocIntersectAndRest`**.

Kevin: this is again something that I'd like to simplify by simply not supporting `OR` queries, but for now is refactored for readability

### GE helpers (`multiGeSearch.ts`)

- File is now **GE string parsing only** (`getSelectedGEs`, `normalizeGeSelection`); merge/query/banner helpers and dead exports removed.

### Misc

- Advanced search units placeholder: **"ex. 4 or VAR"** (drops comma-separated example).

Kevin: `getManyOfField` was written to support `OR` queries. I... don't know anyone who uses this functionality. And WebSOC doesn't support it either.

## Test plan

- [ ] Manual search: single GE, ANY GE, **multi-GE** (intersection block, rest block, warnings when empty intersection).
- [ ] Planner **multi-course** search still merges via **`getMultiple`**.
- [ ] **`getManyOfField`**: empty/invalid GE string falls back to single `queryWebsoc` (server).
- [ ] Spot-check: section tables still correct for GE-filtered courses (discussions/labs vs lectures).

## Issues

Closes #


packages/types/src/websoc.ts
```diff
@@ -45,6 +45,29 @@ type AACourseExtendedProperties = {
 
 export type AACourse = Omit<WebsocCourse, 'sections'> & AACourseExtendedProperties;
 
+export const WebsocSearchInputKeysSchema = z.enum([
```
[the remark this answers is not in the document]
KevinWu098: It doesn't.

apps/antalmanac/src/backend/routers/websoc.ts
```diff
@@ -44,14 +44,25 @@ const websocRouter = router({
...
-            );
+        .input(
+            z.object({
+                params: WebsocSearchInputSchema,
+                fieldName: z.enum([WebsocSearchInputKeysSchema.enum.ge]),
+            })
+        )
+        .query(async ({ input }): Promise<WebsocAPIResponse[]> => {
+            const { fieldName, params } = input;
+            const fieldValue = params[fieldName]
```
Choollol: Is this more similar to `fields` from before? If so, then we should prob keep the name to signal that this is the array of fields, not just a single value.

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
```diff
@@ -473,29 +468,49 @@ export default function CourseRenderPane(props: { id?: number }) {
...
-                                                },
-                                            }}
-                                        >
-                                            The courses below satisfy at least one of the selected GEs.
-                                        </Alert>
-                                    )}
+                                <LazyLoad once key={`rest-${index}`} overflow height={heightEstimate} offset={1000}>
                                     {SectionTableWrapped(index, {
-                                        courseData: courseData,
+                                        courseData: restCourseData,
```
Choollol: Is there a way we can reduce the amount of duplication we have to maintain here?

Off the top of my head, we could lump the alerts and courseDatas into one array and distinguish them in a `.map` with union type narrowing.

Choollol: As a side note, `WarningAlert` is the same as the `Alert` components here, so we should probably use that too.

packages/types/src/websoc.ts
Choollol: Could we use `Object.keys(WebsocSearchInputSchema.shape)` here?

packages/anteater-api/src/utils/index.ts
```diff
@@ -81,23 +94,116 @@ export function combineWebsocResponses(responses: WebsocAPIResponse[]): WebsocAP
...
+ * - {@link intersect}: courses that appear in **every** response (layout from the **first**
+ *   response). Sections are **not** merged across responses.
+ * - {@link rest}: courses that appear in **some** but not all responses, deduped by
+ *   {@link websocCourseKey} (first occurrence wins).
+ *
+ * Each value is a full {@link WebsocAPIResponse} (same non-`schools` fields as the first
+ * response) so callers can flatten/render independently. For a single merged tree, use
+ * {@link mergeWebsocUnion}; for one concatenated list (legacy), combine `schools` yourself.
+ */
+export function splitWebsocIntersectAndRest(responses: WebsocAPIResponse[]): {
```
Choollol, line 109: Pretty nice! I'm a little confused by the "same non-schools fields as the first response" in the doc comment though, doesn't each response only have a schools field?

Choollol: Optional: Could consider breaking up the chunks of logic into smaller functions for possible reusability but also readability. Probably a skill issue on my part but this function took a while to understand.

packages/anteater-api/src/utils/index.ts
```diff
@@ -37,11 +43,19 @@ export function sortWebsocResponse(response: WebsocAPIResponse): WebsocAPIRespon
     return response;
 }
 
+/** Stable identity for a course across WebSOC responses (dept + number, whitespace stripped). */
+export function websocCourseKey(deptCode: string, courseNumber: string): string {
+    return `${deptCode}::${courseNumber}`.replace(/\s+/g, '');
```
Choollol, line 48: Just out of curiosity, what does the "::" do?

Choollol (changes requested):
Much cleaner! I mostly looked at the bigger stuff, I'll do another pass with finer-grain detail afterward

packages/anteater-api/src/utils/index.ts
KevinWu098: Just a delimited (it's also used this way in the review code. But I believe this specific choice predates both (and was in the original)

packages/anteater-api/src/utils/index.ts
KevinWu098: It originally was many smaller funcs, but I got dizzy myself just staring at it a bit — do you want some time to stew on it? im... not against breaking it up, but i think this is just one of those methods that is just big (but will hopefully become smaller once we nix the OR)

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
KevinWu098: I had cursor do a map — not to beat a dead horse, but I'll be happy to see it go when we get rid of the OR

packages/anteater-api/src/utils/index.ts
```diff
@@ -37,11 +43,19 @@ export function sortWebsocResponse(response: WebsocAPIResponse): WebsocAPIRespon
...
- * Merges multiple WebSOC responses into one, deduplicating schools and departments
- * by name/code, merging courses by course number, and merging sections by section code.
+ * Fold several WebSOC responses into a single tree.
+ *
+ * - Schools and departments match on name / code and get merged.
+ * - Each course appears at most once per department; if it shows up in more than one
+ *   response, section lists are combined and deduped by section code.
  */
-export function combineWebsocResponses(responses: WebsocAPIResponse[]): WebsocAPIResponse {
+export function mergeWebsocUnion(responses: WebsocAPIResponse[]): WebsocAPIResponse {
```
Choollol: In the name, a bit of information was lost (the "response" part), and there's some repeated information too ("merge" and "union"). Maybe `unionWebsocResponses` instead?

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
```diff
@@ -73,6 +69,23 @@ function isCourseEntry(item: WebsocSchool | WebsocDepartment | AACourse): item i
...
+          kind: 'lazy-course';
+          key: string;
+          courseData: (WebsocSchool | WebsocDepartment | AACourse)[];
+          index: number;
+      };
+
+type CoursePaneSearchData =
+    | { kind: 'single'; response: WebsocAPIResponse }
+    | { kind: 'split'; intersect: WebsocAPIResponse; rest: WebsocAPIResponse };
+
```
Choollol: Would prefer the types at the top of the file so they aren't mixed in with the logic

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
```diff
@@ -354,43 +381,27 @@ export default function CourseRenderPane(props: { id?: number }) {
...
-    const courseData = useMemo(
-        () => (websocResp ? getFilteredCourses(flattenSOCObject(websocResp, courseColors)) : []),
-        [websocResp, courseColors]
-    );
-
-    const andCourseCount = useMemo(
-        () => getFilteredAndCourseCount(courseData, sharedCourseKeys),
-        [courseData, sharedCourseKeys]
-    );
+    const { intersectCourseData, restCourseData } = useMemo(() => {
```
Choollol, line 380: Optional: The "rest" part is kind of vague. A quick search said the not-intersection is the symmetric difference, so maybe something like `symDiffCourseData`? This might be more confusing for some tho so up to you

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
```diff
@@ -354,43 +381,27 @@ export default function CourseRenderPane(props: { id?: number }) {
...
-            fullCourses: searchData.coursesFull,
-            building: searchData.building,
-            room: searchData.room,
-            division: searchData.division,
-            excludeRestrictionCodes: searchData.excludeRestrictionCodes.split('').join(','),
-            days: searchData.days.split(/(?=[A-Z])/).join(','),
-        }),
-        []
-    );
+    const andCourseCount = useMemo(() => intersectCourseData.filter(isCourseEntry).length, [intersectCourseData]);
```
Choollol: To be consistent with the changes, maybe `intersectCourseCount`?

packages/anteater-api/src/utils/index.ts
```diff
@@ -81,23 +94,116 @@ export function combineWebsocResponses(responses: WebsocAPIResponse[]): WebsocAP
...
+        const nextKeys = perResponseKeys[i]!;
+        const [smaller, larger] =
+            intersectionCourseKeys.size <= nextKeys.size
+                ? [intersectionCourseKeys, nextKeys]
+                : [nextKeys, intersectionCourseKeys];
+        intersectionCourseKeys = new Set([...smaller].filter((k) => larger.has(k)));
+        if (intersectionCourseKeys.size === 0) {
+            break;
+        }
+    }
```
Choollol, line 142: What do you think about doing this in a single pass instead of two?

packages/anteater-api/src/utils/index.ts
```diff
@@ -81,23 +94,116 @@ export function combineWebsocResponses(responses: WebsocAPIResponse[]): WebsocAP
...
+ * Each value is a full {@link WebsocAPIResponse} (same non-`schools` fields as the first
+ * response) so callers can flatten/render independently. For a single merged tree, use
+ * {@link mergeWebsocUnion}; for one concatenated list (legacy), combine `schools` yourself.
+ */
+export function splitWebsocIntersectAndRest(responses: WebsocAPIResponse[]): {
+    intersect: WebsocAPIResponse;
+    rest: WebsocAPIResponse;
+} {
+    const first = responses[0];
+    if (!first) {
```
Choollol: Consider checking `reponses.length` instead for clearer intention

packages/anteater-api/src/utils/index.ts
```diff
@@ -81,23 +94,116 @@ export function combineWebsocResponses(responses: WebsocAPIResponse[]): WebsocAP
...
+
+                    targetDept.courses.push(course);
+                }
+            }
+        }
+    }
+
+    return {
+        intersect: { ...first, schools: intersectSchools },
+        rest: { ...first, schools: restSchools },
```
Choollol: How come `first` is being spread here? Shouldn't these only contain `intersectSchools` and `restSchools`?

Also don't the second `schools` property assignments overwrite the `first` spread anyway?

Choollol (changes requested):
Looks good other than the comments I left + merge conflicts

packages/anteater-api/src/utils/index.ts
KevinWu098: uhhh just let me get away with this one yeah — i don't want to look at this longer than I have to

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
KevinWu098: i want to nix this soon anyways, just uhhhh, let's leave it for now...

packages/anteater-api/src/utils/index.ts
KevinWu098: tbh i dont know and im just gonna let cursor deal with this

packages/anteater-api/src/utils/index.ts
Choollol, line 138: Thoughts on using `set.intersection()` instead?

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
Choollol, line 244: Why the un-shallow here?

Choollol (changes requested):
Thanks for addressing comments, just a couple more things i swear haha

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
KevinWu098: Probably because cursor (composer) wrote it

packages/anteater-api/src/utils/index.ts
KevinWu098: No let me out I'm nuking the OR anyways 

Minion laugh gif

Choollol (approved):
i will remember this

PR #1806, opened 2026-05-18
Fix legacy term data path

Choollol:
## Summary

Delete the generated `termData.ts` instead of the actual source file, which git already knows about.

## Test Plan

fetch-term-data.ts works

## Issues

Closes #

### DIFF
--- a/apps/antalmanac/scripts/fetch-term-data.ts
+++ b/apps/antalmanac/scripts/fetch-term-data.ts
@@ -1,11 +1,11 @@
 import 'dotenv/config';
-import { mkdir, unlink, writeFile } from 'node:fs/promises';
+import { mkdir, writeFile } from 'node:fs/promises';
 
 import { AATerm } from '@packages/antalmanac-types';
 import { createClient } from '@packages/anteater-api/client';
 import type { CalendarTerm, Quarter, Year } from '@packages/anteater-api/types';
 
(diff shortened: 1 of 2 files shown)


KevinWu098 (commented):
Can we just get rid of the whole thing? Isn't term data dead dead

Choollol:
> Can we just get rid of the whole thing? Isn't term data dead dead

Not everywhere I don't think, it's still there on older PRs, like the better auth one

KevinWu098:
> > Can we just get rid of the whole thing? Isn't term data dead dead
> 
> Not everywhere I don't think, it's still there on older PRs, like the better auth one

Well like, if those PRs haven't merged the branch, no problem. And if they have, also no problem?

Choollol:
My better auth branch merged main and deploy was failing because of the leftover generated file though. I'll try again with the commit reverted

Choollol:
@KevinWu098 Looks like it's happening on new PRs too for some reason
https://github.com/icssc/AntAlmanac/actions/runs/26075790187/job/76666426328

KevinWu098:
I went in and am just nuking all of our old term caches (we're actually at our 10gb limit anyways), so this PR may be redundant / could just be to remove the code? im out and ive forgotten exactly whats going on here, but prod deploy should work in a jif fingers crossed

dsnsgithub:
@KevinWu098 I can purge all caches if you want, it should fix everything.

Choollol:
Done by https://github.com/icssc/AntAlmanac/pull/1816

PR #1807, opened 2026-05-18
feat: add GitHub Codespaces dev container

dsnsgithub:
## Summary
Create an easy way to use Github Codespaces, which will make it easier for new developers to start working on AntAlmanac (and fix any machine-specific environment problems).

image
image

## Test Plan
- Create a new GitHub Codespaces (instructions in [README.md](https://github.com/icssc/AntAlmanac/tree/claude/setup-codespaces-dev-env-2GVjz#:~:text=Quickest%20Start%3A%20GitHub%20Codespaces) of this branch)
- Make sure that this branch is selected under "Dev container configuration" (**only needed until this PR is merged**)
- Wait a few minutes for the development environment to be created
- Run `pnpm dev` in the terminal.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,38 @@
+{
+    "name": "AntAlmanac",
+    "image": "mcr.microsoft.com/devcontainers/typescript-node:1-22-bookworm",
+    "features": {
+        "ghcr.io/devcontainers/features/docker-in-docker:2": {
+            "version": "latest",
+            "moby": true
+        },
+        "ghcr.io/devcontainers/features/github-cli:1": {}
+    },
+    "forwardPorts": [3000, 5432],
+    "portsAttributes": {
+        "3000": {
+            "label": "AntAlmanac Web",
+            "onAutoForward": "openPreview"
+        },
+        "5432": {
+            "label": "PostgreSQL",
+            "onAutoForward": "silent"
+        }
+    },
+    "postCreateCommand": "bash .devcontainer/post-create.sh",
+    "remoteUser": "node",
+    "customizations": {
+        "vscode": {
+            "extensions": [
+                "oxc.oxc-vscode",
+                "ms-azuretools.vscode-docker",
+                "bradlc.vscode-tailwindcss",
+                "mikestead.dotenv"
+            ],
+            "settings": {
+                "terminal.integrated.defaultProfile.linux": "bash",
+                "editor.formatOnSave": false
+            }
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,38 @@
...
+        }
+    },
+    "postCreateCommand": "bash .devcontainer/post-create.sh",
+    "remoteUser": "node",
+    "customizations": {
+        "vscode": {
+            "extensions": [
+                "oxc.oxc-vscode",
+                "ms-azuretools.vscode-docker",
+                "bradlc.vscode-tailwindcss",
```
calebwongg: nit: we never actually use tailwind for any of our styling, the only reason why claude thought we need it is because it is a dependency OF the aants dependency @react-email/components. I think the only instance we would ever write tailwind in our current codebase is if someone was editing the react-email component in node modules (which nobody ever does). so it should be safe to delete it here.

calebwongg (changes requested):
not the most versed in shell script but everything looks good to me after we delete the tailwind extension. lowkey my codespaces takes like 10 minutes to start up but it should be just because codespaces itself is heavy. hopefully this motivates more open source contributers🤞🤞

.devcontainer/devcontainer.json
dsnsgithub: resolved https://github.com/icssc/AntAlmanac/pull/1807/changes/9e344c9276dfb050016743e739f6dacc81c2299a

dsnsgithub:
> not the most versed in shell script but everything looks good to me after we delete the tailwind extension. lowkey my codespaces takes like 10 minutes to start up but it should be just because codespaces itself is heavy. hopefully this motivates more open source contributers🤞🤞

In my testing, most of the startup time is spent waiting for `pnpm --filter antalmanac get-data` to finish. 

The dev container can also be used locally as long as you have Docker installed if GitHub codespaces doesn't offer enough performance.

If https://github.com/icssc/AntAlmanac/pull/1848 is merged, it might be able to speed up codespace creation.
We could also [prebuild the codespace](https://docs.github.com/en/codespaces/prebuilding-your-codespaces/about-github-codespaces-prebuilds) with GitHub Actions, but that would be more involved.

calebwongg (approved):
It just occurred to me that users are still going to have to generate their own aapi key and their own better auth secret. So they still will have to do slight config for their environment, which I suppose takes away a little from the point of codespaces. I guess theres nothing we can really do about it because we don't want to globalize a key. Aside from that though codespaces does resolve other onboarding conflicts. So lgtm

dsnsgithub:
I think GitHub codespaces are a bit too slow for this use case, especially since the developer still needs to provide their own API keys. We can come back to this later.

PR #1817, opened 2026-05-19
feat: nuqs

KevinWu098:
## Summary

Course search state moves from `CoursePaneStore` to the URL via [nuqs](https://nuqs.dev). Shareable links, refresh, and back/forward all read the same query string.

**Preview:** https://scheduler-1817.antalmanac.com

## URL params

| Param | Role |
|--------|------|
| Form fields | `deptValue`, `ge`, `term`, `sectionCode`, `courseNumber`, `instructor`, advanced filters |
| `search` | `manual` or `quick` |
| `view` | `search` (form) or `results` |
| `importRoadmap` | Planner deep link |

Helpers (`isValidSearch`, `shouldShowSearchForm`, `hasManualParams`, `hasAdvancedParams`) compare against `DEFAULT_FORM_DATA`, not string literals.

## Code layout

**`SearchParams/`** — `constants`, `parsers`, `loaders`, `types`, `index`
- `useCourseSearchUrl()` — form data, mode/view, submit, reset, batch setters
- `useCourseSearchParam(field)` — per-field reads/writes for memoized form leaves
- `readCourseSearchParams()` — one-shot reads (mount/tab logic, no subscription)

**Zustand (non-URL only)**
- `SavedSearchStore` — stash `location.search` when leaving Search tab; restore on return. `savedManualSearch` preserves manual fields when switching to quick search.
- `TabStore` — tab index + shared tab config (`mobileOnly` on calendar)
- `RightPaneStore` — planner batch `multiSearchData` only

**Search form** — Advanced search split into `AdvancedSearchFields/*` (each field uses `useCourseSearchParam`). Mode toggle + reset use `SearchFormModeToggle` / `ManualSearchResetButton`.

**Schedule** — `ScheduleManagement` picks initial tab from URL + session; tab clicks navigate and replay saved search query.

## Other changes

- `nuqs` + `NuqsAdapter` (react-router v6)
- Remove `CoursePaneStore` and `advancedSearchEnabled` store flag
- `GEDataFetchProvider` reads advanced params from URL parsers
- `websoc` types: optional `instructor` on search requests

## Test plan

- [ ] Manual: dept / GE / section / instructor → Search → results; refresh keeps state
- [ ] `search=manual` vs quick toggle; manual fields restore after quick → manual
- [ ] Reset clears form (including advanced time fields)
- [ ] Advanced fields round-trip in URL; panel expands on load when advanced params present
- [ ] Shared link opens search tab with correct form/results
- [ ] Planner: roadmap picker and `?importRoadmap=` batch search
- [ ] Search → Added/Map → Search restores previous query string
- [ ] Exclude taken courses dropdown opens (signed in)


apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchForm.tsx
```diff
@@ -84,7 +81,9 @@ export const SearchForm = ({ toggleSearch }: SearchFormProps) => {
                                     action: analyticsEnum.classSearch.actions.MANUAL_SEARCH,
                                 });
                             }}
-                            onReset={RightPaneStore.resetFormValues}
+                            onReset={() => {
+                                void resetAll();
```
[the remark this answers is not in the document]
KevinWu098: This is intended

Choollol (changes requested):
Haven't looked at the code yet, but ran into a couple bugs from usage:

1. Just inputting into some manual search fields, like selecting department or GE or typing in section code, triggers the search flow. To reproduce: From https://scheduler-1817.antalmanac.com/?search=manual, select any department in manual search and see that it triggers search.
2. Pressing the back button on the results page takes users to quick search even if they searched from manual search. It was happening consistently at first but I can't reproduce anymore, so might've been a problem on my end.
3. Editing one manual search field resets all the other fields, including term. It was happening consistently at first but I can't reproduce anymore, so might've been a problem on my end.

also as a side note, the url in the PR description should have "staging" replaced with "scheduler"

Choollol:
More bugs:

4. Switching between quick search and manual search wipes the manual search input fields. To reproduce: Change anything besides term, e.g. course number, then switch to quick search then back to manual search, and see that form fields are reset.
5. Maybe not a bug exactly, but going from search results to added tab then pressing the browser back button doesn't change the UI at all, just the URL

KevinWu098:
> Haven't looked at the code yet, but ran into a couple bugs from usage:
> 
> 1. Just inputting into some manual search fields, like selecting department or GE or typing in section code, triggers the search flow. To reproduce: From [scheduler-1817.antalmanac.com?search=manual](https://scheduler-1817.antalmanac.com/?search=manual), select any department in manual search and see that it triggers search.
> 2. Pressing the back button on the results page takes users to quick search even if they searched from manual search. It was happening consistently at first but I can't reproduce anymore, so might've been a problem on my end.
> 3. Editing one manual search field resets all the other fields, including term. It was happening consistently at first but I can't reproduce anymore, so might've been a problem on my end.
> 
> also as a side note, the url in the PR description should have "staging" replaced with "scheduler"

Fixed 1. Can't reproduce 2 or 3, but perhaps the fix to 1 is helping there. Fixed 4. 5 is... technically a bug on prod as well (our tabs as routes is a little messy).

KevinWu098:
> More bugs:
> 
> 4. Switching between quick search and manual search wipes the manual search input fields. To reproduce: Change anything besides term, e.g. course number, then switch to quick search then back to manual search, and see that form fields are reset.
> 5. Maybe not a bug exactly, but going from search results to added tab then pressing the browser back button doesn't change the UI at all, just the URL

After having looked into it quite a bit, the problem here stems from how TabStore is not really synced to the actual page route. And it's not entirely easy to sync them via, say, an effect because both calendar and search share the base route (on mobile), so its difficult to disambiguate.

I think the shape of 5 is best handled in a follow up.

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchParams/hooks.ts
Choollol, line 1: Could you split this file up into multiple? A lot of things seem like they should be separated, like the hook. It'd also probably fix the circular dependency cubic brought up

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/searchParams.ts
```diff
@@ -0,0 +1,231 @@
...
+    division: string;
+    excludeRoadmapCourses: string;
+    excludeRestrictionCodes: string;
+    days: string;
+}
+
+export type CourseSearchField = keyof CourseSearchParams;
+
+export type CourseSearchMode = 'quick' | 'manual';
+export type CourseSearchView = 'search' | 'results';
```
Choollol: Can `search` be renamed to something more specific? The results page is also part of searching, so it's not entirely clear that this represents the search form display.

Choollol: They should also probably become constants, I see a lot of repeated "search" and "results" below. Same with "manual" and "quick"

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/searchParams.ts
```diff
@@ -0,0 +1,231 @@
...
+            return formData.term.shortName !== defaultFormData.term.shortName;
+        }
+        return formData[key] !== defaultFormData[key];
+    });
+}
+
+/** Enough to run a WebSOC search (dept, GE, section, or instructor). */
+export function isValidSearch(formData: CourseSearchParams) {
+    const { ge, deptValue, sectionCode, instructor } = formData;
+    return ge !== 'ANY' || deptValue !== 'ALL' || sectionCode !== '' || instructor !== '';
```
Choollol: Can we check against defaults instead of duplicated literals? Same with shouldShowSearchForm

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/searchParams.ts
```diff
@@ -0,0 +1,231 @@
...
+    eq: (a: AATerm, b: AATerm) => a.shortName === b.shortName,
+}).withDefault(defaultTerm);
+
+const parseAsNormalizedGe = createParser<string>({
+    parse: (value: string) => normalizeGeSelection(value),
+    serialize: (value: string) => normalizeGeSelection(value),
+    eq: (a: string, b: string) => normalizeGeSelection(a) === normalizeGeSelection(b),
+}).withDefault(defaultFormData.ge);
+
+export const courseSearchParamParsers = {
```
Choollol: Is there a way to lessen or reuse the amount of code for this?

Choollol: Also, is this a SCREAMING_SNAKE_CASE type of constant?

apps/antalmanac/src/stores/TabStore.ts
```diff
@@ -1,36 +1,61 @@
...
+export type TabInfo = {
+    name: TabName;
+    label: string;
+    href: string;
+    icon: SvgIconComponent;
+    id?: string;
+    mobile?: true;
 };
 
+export const Tabs: TabInfo[] = [
```
Choollol: Should this be in `TABS`?

apps/antalmanac/src/stores/TabStore.ts
```diff
@@ -1,36 +1,61 @@
...
-    search: 1,
-    added: 2,
-    map: 3,
+export type TabInfo = {
+    name: TabName;
+    label: string;
+    href: string;
+    icon: SvgIconComponent;
+    id?: string;
+    mobile?: true;
```
Choollol: Maybe `mobileOnly`?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/searchParams.ts
```diff
@@ -0,0 +1,231 @@
...
+    }, [formData.term, setFormData]);
+
+    const setSearchMode = useCallback(
+        (mode: CourseSearchMode) => {
+            return setSearchModeParam(mode);
+        },
+        [setSearchModeParam]
+    );
+
+    /** Navigate to the results pane. */
```
Choollol: Are these comments worth keeping?

apps/antalmanac/src/stores/TabStore.ts
Choollol, line 1: nice refactor

apps/antalmanac/src/stores/SavedSearchStore.ts
```diff
@@ -0,0 +1,34 @@
...
+    saveSearch: () => set({ savedSearch: location.search }),
+    popSavedSearch: () => {
+        const { savedSearch } = get();
+        set({ savedSearch: null });
+        return savedSearch;
+    },
+
+    savedManualSearch: null,
+    saveManualSearch: (params) => set({ savedManualSearch: { ...params, term: params.term } }),
+    getManualSearch: () => get().savedManualSearch,
```
Choollol: Is this get wrapper necessary? I think usually zustand getting is done by directly using the variable

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchForm.tsx
```diff
@@ -2,43 +2,59 @@ import { Footer } from '$components/RightPane/CoursePane/SearchForm/Footer';
...
 
     const toggleSearchMode = (event: React.MouseEvent<HTMLElement>, value: string) => {
         event.preventDefault();
         if (!value) return;
-        toggleManualSearch();
+
+        switch (value) {
+            case 'manual':
+                void setSearchMode('manual');
+                const savedManualSearch = useSavedSearchStore.getState().getManualSearch();
```
Choollol: Should these `useSavedSearchStore` uses be done with the hook call style instead?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchForm.tsx
```diff
@@ -2,43 +2,59 @@ import { Footer } from '$components/RightPane/CoursePane/SearchForm/Footer';
...
         (event: FormEvent<HTMLFormElement>) => {
             event.preventDefault();
-            toggleSearch();
+            submitSearch(formData);
         },
-        [toggleSearch]
+        [formData, manualSearchEnabled, submitSearch]
     );
 
     const toggleSearchMode = (event: React.MouseEvent<HTMLElement>, value: string) => {
```
Choollol: Since we have the types now, could this be more safe by typing `value` as a specific type?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchForm.tsx
Choollol: Can we use the new types/constants here?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearch.tsx
```diff
@@ -1,52 +1,19 @@
...
-        return () => {
-            RightPaneStore.off('formReset', resetField);
-        };
-    }, [resetField]);
-
     return (
         <>
             <Button
-                onClick={handleExpand}
+                onClick={() => setExpanded((value) => !value)}
```
Choollol: Should this be in the deleted `handleExpand`? So it isn't buried in all the rendering logic

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearchTextFields.tsx
```diff
@@ -159,63 +86,42 @@ export function AdvancedSearchTextFields() {
 
         switch (name) {
```
Choollol: Can this entire switch statement just be replaced by a `setField` call? This would also fix the problem where `startTime` and `endTime` don't actually change anything anymore.

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/searchParams.ts
```diff
@@ -0,0 +1,231 @@
...
+
+    const resetAll = useCallback(() => {
+        clearMultiSearchData();
+        return setFormData(defaultFormData);
+    }, [setFormData]);
+
+    const resetAllPreservingTerm = useCallback(() => {
+        clearMultiSearchData();
+        return setFormData({ ...defaultFormData, term: formData.term });
+    }, [formData.term, setFormData]);
```
Choollol: Optional: Should these two be combined so shared logic, which isn't a lot now but could grow in the future, only needs to be updated once? Off the top of my head, maybe it could take an options function with a `preserveTerm` bool or something

Choollol (changes requested):
Overall looking much better than before, main logic looks pretty solid

1. It looks like every component that uses `formData` is rerendering whenever any field is changed, even if they don't need that field. Is there a way to only rerender the relevant components when fields change?
2. Is `void`ing all the function calls necessary?
3. If you could double check things like the casing of constants and organization of files (e.g. grouping types together at the top), that'd be great

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/searchParams.ts
KevinWu098: I mean, yeah, but the view is definitely not search when you're looking at results! I can be more verbose and say "searchForm", but i dont wanna cuz it looks ugly :p

you can press me on this though, im not particularly attached

KevinWu098: > They should also probably become constants, I see a lot of repeated "search" and "results" below. Same with "manual" and "quick"

done

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchParams/hooks.ts
KevinWu098: Split it into a few files — might make recollecting your review comments a little annoying, but I've done my best to resolve them all and respond in thread with context if appropraite

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/searchParams.ts
KevinWu098: done

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/searchParams.ts
KevinWu098: eh... i don't really see it as a "constant" per se. No more than a useState would be a constant? As far as code reuse, I'd rather us not here since some of these fields should be enums and whatnot. i find overabstracting to be actually more of a headache when each thing is clearly its own... "thing" (?). 

In the same way that we don't reduce the number of `useState` we define in a file by turning it into an object useState, if each hook is obviously managing is own thing (if you follow my train of thought)

apps/antalmanac/src/stores/TabStore.ts
KevinWu098: done

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchForm.tsx
KevinWu098: done

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchForm.tsx
KevinWu098: done

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearchTextFields.tsx
KevinWu098: done

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/searchParams.ts
KevinWu098: done

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchParams/hooks.ts
```diff
@@ -0,0 +1,231 @@
+import {
+    ADVANCED_SEARCH_PARAMS,
+    MANUAL_SEARCH_PARAMS,
+    PLANNER_SEARCH_PARAM,
+    normalizeGeSelection,
+} from '$components/RightPane/CoursePane/SearchForm/constants';
+import RightPaneStore from '$components/RightPane/RightPaneStore';
```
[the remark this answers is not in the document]
KevinWu098: is it fixed now

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/AdvancedSearch/AdvancedSearch.tsx
KevinWu098, line 1: At risk of further blowing up the diff, I've opted not to reorganize the rest of SearchForm (e.g. quick and manual search)

KevinWu098:
@Choollol Sorry that the diff is rather big (bigger than before, at least). Hopefully it's still straightforward what's going on. 

I've resolved all of your comments. The one remaining one is about rerenders. Having burned some midnight oil on it... I've somewhat optimized it but I'm mostly leaning on `memo` right now which, from light testing via `react-scan`, appears to fix the perf issues. 

The tricky thing is that `formData` really is not great for reactivity... but I'm not really sure how to engineer around that... i was mulling over individual fields, but that uhhhhhh i want to say runs into the same issues anyways? input is appreciated here.

also happy to open an issue to track this, although i imagine it might sit open for quite some time...

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchParams/index.ts
```diff
@@ -0,0 +1,179 @@
...
+        formData.courseNumber !== DEFAULT_FORM_DATA.courseNumber ||
+        formData.ge !== DEFAULT_FORM_DATA.ge ||
+        formData.deptValue !== DEFAULT_FORM_DATA.deptValue ||
+        formData.instructor !== DEFAULT_FORM_DATA.instructor;
+
+    return !hasPrimarySearchInput || !isValidSearch(formData);
+}
+
+/** Drop planner batch search so URL-driven queries take precedence. */
+export function clearMultiSearchData() {
```
Choollol: What's the reason for this wrapper?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchParams/hooks.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    return !hasPrimarySearchInput || !isValidSearch(formData);
+}
+
+/** Drop planner batch search so URL-driven queries take precedence. */
+export function clearMultiSearchData() {
+    RightPaneStore.clearMultiSearchData();
+}
+
+export function useCourseSearchParam<K extends keyof CourseSearchParams>(
```
Choollol, line 23: Maybe more of a preference thing, but I think usually hooks have their own files

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchParams/index.ts
```diff
@@ -0,0 +1,179 @@
...
+        },
+        [setSearchModeParam]
+    );
+
+    const submitSearch = useCallback(
+        (data?: CourseSearchParams) => {
+            const payload = data ?? readCourseSearchParams();
+            if (isValidSearch(payload)) {
+                clearMultiSearchData();
+                void setViewParam(COURSE_SEARCH_VIEW.RESULTS);
```
Choollol: Is this somewhere `showResults` can be used?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchParams/index.ts
```diff
@@ -0,0 +1,179 @@
...
+            clearMultiSearchData();
+            void setValueRaw(next);
+        },
+        [setValueRaw]
+    );
+
+    return [value, setValue];
+}
+
+export function useCourseSearchUrl() {
```
Choollol: Do you think it would make sense to split this hook into multiple smaller ones? It'd be more code on the callers' side, but it'd also help with the unnecessary rerendering. idk just something off the top of my head

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchWithPlanner.tsx
```diff
@@ -40,10 +31,16 @@ function getDefaultTermRoadmapGrouping(): TermRoadmapGrouping {
     };
 }
 
-export const SearchWithPlanner = ({ labelProps }: SearchWithPlannerProps) => {
+export const SearchWithPlanner = () => {
+    const { formData, showResults } = useCourseSearchUrl();
```
Choollol: This looks like it can just use `term` and doesn't need the entire `formData`?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchWithPlanner.tsx
```diff
@@ -40,10 +31,16 @@ function getDefaultTermRoadmapGrouping(): TermRoadmapGrouping {
...
+    const [plannerSearchParam, setPlannerSearchParam] = useQueryState(
+        COURSE_SEARCH_PLANNER_KEY,
+        parseAsString.withOptions({ history: 'replace' })
+    );
     const [termRoadmapGrouping, setTermRoadmapGrouping] = useState<TermRoadmapGrouping>(getDefaultTermRoadmapGrouping);
     const [isLoadingSearch, setIsLoadingSearch] = useState(false);
     const [openSignInDialog, setOpenSignInDialog] = useState(false);
+    const hasSearchedWithUrlParamsRef = useRef(false);
```
Choollol, line 44: Since the `importRoadmap` param gets wiped when showing results, is this still necessary?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/FuzzySearch.tsx
```diff
@@ -168,11 +176,11 @@ const FuzzySearch = ({ toggleSearch, postHog, labelProps }: FuzzySearchProps) =>
                     console.error(e);
                 });
         },
-        [requestIsCurrent]
+        [term, requestIsCurrent]
     );
 
     const handleFormDataChange = useCallback(() => {
```
Choollol, line 189: Can this code just be moved directly into the `useEffect` it's called in?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchParams/parsers.ts
```diff
@@ -0,0 +1,75 @@
...
+import { type AATerm } from '@packages/antalmanac-types';
+import { createParser, createSerializer, parseAsString, parseAsStringLiteral, type SingleParserBuilder } from 'nuqs';
+
+export type CourseSearchParamParser<K extends keyof CourseSearchParams> = SingleParserBuilder<CourseSearchParams[K]> & {
+    readonly defaultValue: CourseSearchParams[K];
+};
+
+export type CourseSearchParamParserMap = {
+    [K in keyof CourseSearchParams]: CourseSearchParamParser<K>;
+};
```
Choollol: Are these types used in other files?

apps/antalmanac/src/components/ScheduleManagement/ScheduleManagement.tsx
```diff
@@ -40,50 +51,68 @@ export function ScheduleManagement() {
...
-                case 'added':
-                    setActiveTab('added');
-                    break;
-                case 'map':
-                    setActiveTab('map');
-                    break;
+    const handleTabChange = useCallback(
+        (nextTab: number) => {
+            if (activeTab === TAB_INDEX.search && nextTab !== TAB_INDEX.search) {
+                useSavedSearchStore.getState().saveSearch();
```
Choollol: Since this is inside a component, should `useSavedSearchStore` be called as a hook?

apps/antalmanac/src/components/ScheduleManagement/ScheduleManagement.tsx
```diff
@@ -30,6 +40,7 @@ export function ScheduleManagement() {
      * Ref to the scrollable container with all of the tabs-content within it.
      */
     const ref = useRef<HTMLDivElement>(null);
+    const hasAppliedInitialTab = useRef(false);
```
Choollol: Should this have a 'Ref' suffix?

apps/antalmanac/src/components/ScheduleManagement/ScheduleManagement.tsx
```diff
@@ -40,50 +51,68 @@ export function ScheduleManagement() {
...
+            hasAppliedInitialTab.current = true;
+            setActiveTab('added');
             return;
         }
-
-        if (shouldSearchPlannerFromParams()) {
-            setActiveTab('search');
+        if (tab === 'map') {
+            hasAppliedInitialTab.current = true;
+            setActiveTab('map');
```
Choollol: Could the literal strings in the `setActiveTab` calls pass `tab` instead? This way, we get type safety on `tab` too

Choollol (changes requested):
Thanks for making those changes, looks a lot better (especially the advanced search fields). I think once these new comments are addressed it'll be good to go

1. Does `clearMultisearchData` need to be called anywhere outside of `showSearchForm`?
2. Might have missed your answer to this, but why all the `void`ing with the function calls?
3. There's a lot of non-component related files under the components folder, which I think is actually better because related code is closer together. Should we start moving toward a feature-based file organization rather than grouping by file type? Would be a separate issue/PR thing ofc

apps/antalmanac/src/components/ScheduleManagement/ScheduleManagement.tsx
KevinWu098: done

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchParams/loaders.ts
```diff
@@ -0,0 +1,30 @@
...
+export function readAdvancedSearchParams(): AdvancedSearchParams {
+    return loadAdvancedSearchParams(globalThis.location?.search ?? '');
+}
+
+export function readCourseSearchParams(): CourseSearchParams {
+    return loadCourseSearchParams(globalThis.location?.search ?? '');
+}
+
+export function readSearchMode(): CourseSearchMode {
+    return loadSearchMode(globalThis.location?.search ?? '')[COURSE_SEARCH_MODE_KEY];
```
Choollol, line 29: What are the `globalThis`s for?

KevinWu098: Just more convenient for reading `window.location` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchParams/index.ts
KevinWu098: Just keeping things in one place, but eh we can inline until it becomes not a one liner to clear (but i dont expect it not to be)

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchParams/hooks.ts
KevinWu098: I am open to moving the other methods to smth like helpers.ts?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchParams/index.ts
KevinWu098: Uhhhh, I think so yes, but I think it's marginal with `memo`. I did experiment with breaking up the hooks a bit, but I would much rather do this as a follow up

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchWithPlanner.tsx
KevinWu098: eh, it'll rerender if it needs `showResults`. in any case, im going to refactor this whole file so let it slide?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchWithPlanner.tsx
KevinWu098: I want to say yes, but again I'm not super interested in perfect correctness here since I'm gonna clean this up

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/FuzzySearch.tsx
KevinWu098: yes, but stylistically i kinda like it outside /shrug its whatever

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchParams/parsers.ts
KevinWu098: doesn't look like it. fixed

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchParams/index.ts
KevinWu098: ill take a further look. but in the meantime ill rerequest you since i dealt with the rest

KevinWu098:
> Thanks for making those changes, looks a lot better (especially the advanced search fields). I think once these new comments are addressed it'll be good to go
> 
> 1. Does `clearMultisearchData` need to be called anywhere outside of `showSearchForm`?
> 2. Might have missed your answer to this, but why all the `void`ing with the function calls?
> 3. There's a lot of non-component related files under the components folder, which I think is actually better because related code is closer together. Should we start moving toward a feature-based file organization rather than grouping by file type? Would be a separate issue/PR thing ofc

1. Yes, but ive made it just call rightpanestore direct cuz whatever
2. just a correctness thing, but not important (nuqs url updates are async)
3. yeah, definitely future work

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchParams/hooks.ts
Choollol: I'm good with that too! I just like hooks in their own files cuz it makes them slightly easier to ctrl + p search

Choollol:
>     1. Yes, but ive made it just call rightpanestore direct cuz whatever

Could you explain an example flow of multisearchData not being cleared if it's only called in showSearchForm?

edit: nvm i just saw the diff ur good
edit 2: nvm i just looked closer, question still stands

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/SearchParams/loaders.ts
Choollol: Could this code be called outside the browser? Maybe I'm just misunderstanding the purpose of globalThis, but it seems like it's equivalent to window, which would be consistent with the rest of the codebase

KevinWu098: uh the point is that i don't have to do if window is defined yada yada (ssr vs spa nonsense), its why there's an optional chain

KevinWu098:
> > ```
> > 1. Yes, but ive made it just call rightpanestore direct cuz whatever
> > ```
> 
> Could you explain an example flow of multisearchData not being cleared if it's only called in showSearchForm?
> 
> edit: nvm i just saw the diff ur good edit 2: nvm i just looked closer, question still stands

sorry original reply was bad. here's a cursor reply that i did read and do agree with. but this issue will be fixed in follow ups as i consolidate planner stuff into the URL

```
Here’s a reply you can give Choollol, with a concrete flow:

multiSearchData is in-memory batch search state (planner roadmap → list of dept/course pairs). It is not in the URL. CourseRenderPane checks it first and, if non-empty, ignores the URL search params and runs the batch WebSOC path instead:


CourseRenderPane.tsx
Lines 341-358
                if (multiSearchData.length > 0) {
                    // ... uses multiSearchData, NOT formData from URL
                    response = await trpc.websoc.getMultiple.query({ params: offeredCourses });
                } else {
                    const websocQueryParams = getQueryParams(formData);
                    // ... normal single URL-driven search
                }
So if we only cleared in showSearchForm, stale planner data would keep winning whenever the user starts a new search without explicitly opening the search form.

Example flow (planner → quick search from calendar)
User searches with Planner → setMultiSearchData([...roadmap courses]) + showResults().
Results page shows the roadmap batch (not driven by URL params).
User clicks a course on the calendar → useQuickSearch() navigates to /?dept=COMPSCI&courseNumber=122&...&view=results.
showSearchForm is never called — they go straight to results with new URL params.
Without clearing elsewhere, multiSearchData is still populated.
CourseRenderPane sees multiSearchData.length > 0 and still runs the planner batch query → user asked for COMPSCI 122 but still sees the roadmap batch.
That’s why useQuickSearch clears it, and why param setters (useCourseSearchParam, setField, submitSearch, etc.) clear it too — any path that makes URL params authoritative needs to drop the batch.

Another example (tab switch)
User does a planner search (batch results).
Switches to Added tab — URL is stashed, but multiSearchData stays in RightPaneStore.
Switches back to Search — URL is restored, but batch data is still in memory.
Again, batch path wins over the restored URL unless something cleared it on the way out or on param change.
Short version: showSearchForm only runs when explicitly returning to the form UI. Many flows update search without that — quick search links, fuzzy submit, individual param edits, form reset, mode toggle. Clearing only in showSearchForm would leave stale planner batch data overriding URL-driven searches.
```

Choollol:
Ahh gotcha, didn't think about searching not from the search form

KevinWu098:
> Ahh gotcha, didn't think about searching not from the search form

ended up breaking them into individual hooks — i think perf win is marginal, but yeah its nice. little bit suspicious that the edges on the hooks are kind of blurred, but uhhhh whatever its fine for now i guess

KevinWu098:
@cubic

Choollol (approved):
noice

PR #1820, opened 2026-05-19
Soft delete DB rows

### LINKED ISSUE #1698: Soft delete DB rows
Continuing from the improvements in #1640 and #1582, which made insertions update via `ON CONFLICT` (rather than delete then reinsert), our deletes should be soft (meaning rather than removing the row, we mark it as "deleted" (or "archived")). 

This makes recovery much easier and any features (e.g. autosave) much safer.

xgraceyan:
## Issues

Closes #1698


packages/db/migrations/0019_schedule_soft_delete.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE "schedules" ADD COLUMN "archived_at" timestamp with time zone;--> statement-breakpoint
+ALTER TABLE "schedules" ALTER COLUMN "index" DROP NOT NULL;--> statement-breakpoint
+ALTER TABLE "schedules" ADD CONSTRAINT "schedules_archive_index_consistency" CHECK (
```
KevinWu098: this index consistency crap (which i know isn't your introduction) is absolutely hellish.

xgraceyan: What would you recommend then? Maybe we can make the unique index only apply to active rows then that would remove the need for the consistency check

KevinWu098: dunno, i need to think/research it more — i just think that any time we do a constraint which isn't enforced by drizzle, we're really playing with fire

PR #1821, opened 2026-05-19
refactor: RDS

### LINKED ISSUE #1158: Clean up RDS post auth

victorialee13:
## Summary
Refactored rds.ts to modular rds classes

Closes #1158


KevinWu098 (changes requested):
This is not what I had in mind for a refactor of rds — ideally, we should improve the shape of the code in addition to splitting it up.

1. The files are in the rds directory! They don't need a suffix.
2. Why are we bothering to create new classes that just get their methods reexported by rds.ts? This additional layer of indirection is unnecessary and makes reasoning about the code paths (somewhat) harder
3. Why are they classes? Does RDS being a class do anything for us? (this is a historical artifact, but since we're refactoring, we can change this)

The ideal shape would be having `rds/accounts`, `rds/foo`, `rds/bar`, each of which exports their own methods so that consumers can just import { fooMethod } from $rds/foo or something. 

Also open to a barrel file, but bit more ambivalent to that since we try not to barrel file _within_ an `app` (its mostly for stuff like the packages which can be more generic)

victorialee13:
@KevinWu098 Noted! I like the sound of a single barrel file but in that sense it would just keep everything monolithic once again as it doesn't tell sep our schedules, sessions, etc. . I like the sound of this "The ideal shape would be having rds/accounts, rds/foo, rds/bar, each of which exports their own methods so that consumers can just import { fooMethod } from $rds/foo or something."
will implement that now

apps/antalmanac/src/backend/lib/rds/friendships.ts
```diff
@@ -0,0 +1,225 @@
+import type { db } from '@packages/db';
+import type * as schema from '@packages/db/src/schema';
+import { friendships, users } from '@packages/db/src/schema';
+import { and, eq, ExtractTablesWithRelations, ne, or } from 'drizzle-orm';
+import type { PgTransaction, PgQueryResultHKT } from 'drizzle-orm/pg-core';
+
+type Transaction = PgTransaction<PgQueryResultHKT, typeof schema, ExtractTablesWithRelations<typeof schema>>;
+type DatabaseOrTransaction = Omit<typeof db, '$client'> | Transaction;
+
```
KevinWu098: looks like a lot of code duplication. can you take a pass and make a rds/types.ts file or something

KevinWu098 (changes requested):

PR #1822, opened 2026-05-19
feat: de-support OR in multi-GE search

### LINKED ISSUE #1814: De-support OR in multi-GE
Due to technical limitations in allowing large GE results (due to `GEDataFetchProvider`), we would like to de-support "OR" queries in the multi-GE search field.

Users also, generally, are familiar with checkbox filters which apply AND results. Further, we will make it more clear in the input field by changing the delimiter to `and` such that it is clear the query is an intersection, not an intersection.

This will also massively simplify the code handling, in addition to the migration to NUQS (being tracked in #1800 and #1186, respectively)

KevinWu098:
## Summary

Multi-GE search now only returns courses that satisfy **all** selected GEs (AND semantics). OR fallback results, split intersect/rest rendering, and related UI banners are removed.

- Deleted `multiGeSearch.ts`; GE parsing lives in `SearchForm/constants.ts` with an ` and ` delimiter (legacy comma URLs still parse).
- Replaced `websoc.getManyOfField` with `websoc.getGeIntersection` and removed `splitWebsocIntersectAndRest` in favor of `intersectWebsocResponses`.
- Simplified `CourseRenderPane` to a single WebSOC response and one lazy-loaded course list.

Net: **−125 lines** across the touched files.

## Test Plan

- [ ] Select two or more GEs in Advanced Search; confirm results only include courses fulfilling every selected GE.
- [ ] Confirm the GE field shows selections joined with ` and ` (e.g. `GE-1A and GE-2`).
- [ ] When no course satisfies all selected GEs, confirm the warning appears and no OR-only courses are listed below.
- [ ] Open a shared link with comma-separated `ge=` params; confirm it still normalizes and searches correctly.

## Issues

Closes #1814

<div><a href="https://cursor.com/agents/bc-a2b55307-780b-4ecd-be7d-fa7b6e5f0360"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-a2b55307-780b-4ecd-be7d-fa7b6e5f0360"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


packages/anteater-api/src/utils/index.ts
```diff
@@ -125,23 +106,22 @@ export function splitWebsocIntersectAndRest(responses: WebsocAPIResponse[]): {
...
+        } else {
+            intersectionCourseKeys = intersectionCourseKeys.intersection(keys);
+            if (intersectionCourseKeys.size === 0) {
+                return { schools: [] };
+            }
         }
     }
 
-    const intersectSchools = first.schools
+    if (intersectionCourseKeys === null) {
```
Choollol, line 120: Unnecessary null check?

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/constants.ts
```diff
@@ -46,3 +46,21 @@ export const GE_LIST = [
...
+
+    return validGEs.length === 0 ? [] : [...new Set(validGEs)];
+};
+
+export const getSelectedGEs = (ge: string) => parseSelectedGEs(ge);
+
+export const normalizeGeSelection = (ge: string) => {
+    const selectedGEs = parseSelectedGEs(ge);
+    return selectedGEs.length > 0 ? selectedGEs.join(',') : ANY_GE;
+};
```
Choollol: These seem more helpers/utils than constants. Maybe they could be moved to lib/searchHelpers or something? I also just noticed that this `constants.ts` is in the components folder, but that's probably something for another day

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
```diff
@@ -494,24 +439,20 @@ export default function CourseRenderPane(props: { id?: number }) {
...
-                                    })}
-                                </LazyLoad>
-                            )
-                        )}
+                        {courseData.map((_, index) => (
+                            <LazyLoad
+                                once
+                                key={index}
+                                overflow
+                                height={estimateCoursePaneLazyHeight(courseData[index])}
```
Choollol: Can we use the map parameter here instead? (Currently ignored as `_`)

packages/anteater-api/src/utils/index.ts
```diff
@@ -94,29 +88,16 @@ export function unionWebsocResponses(responses: WebsocAPIResponse[]): WebsocAPIR
...
- * flatten/render independently. For a single merged tree, use {@link unionWebsocResponses};
- * for one concatenated list (legacy), combine `schools` yourself.
- */
-export function splitWebsocIntersectAndRest(responses: WebsocAPIResponse[]): {
-    intersect: WebsocAPIResponse;
-    rest: WebsocAPIResponse;
-} {
-    if (responses.length === 0) {
-        return { intersect: { schools: [] }, rest: { schools: [] } };
+/** Courses that appear in every response (layout from the first). Sections are not merged. */
```
Choollol: Is the "(layout from the first)" necessary? I think everything that needs to be conveyed is done already by the words before, and this extra bit could get a little confusing.

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
```diff
@@ -143,7 +130,6 @@ const RecruitmentBanner = () => {
     const [bannerVisibility, setBannerVisibility] = useState(true);
     const theme = useTheme();
 
-    // Display recruitment banner if more than 11 weeks (in ms) has passed since last dismissal
```
Choollol, line 146: What was the reason behind deleting this? I think this comment was pretty harmless, possibly helpful to first-time readers

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/constants.ts
```diff
@@ -46,3 +46,21 @@ export const GE_LIST = [
...
+const parseSelectedGEs = (ge: string) => {
+    const validGEs = ge
+        .split(',')
+        .map((value) => value.trim().toUpperCase())
+        .filter((value) => VALID_GES.has(value));
+
+    return validGEs.length === 0 ? [] : [...new Set(validGEs)];
+};
+
+export const getSelectedGEs = (ge: string) => parseSelectedGEs(ge);
```
Choollol: Is this different from just exporting/using `parseSelectedGEs`?

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane.tsx
```diff
@@ -348,27 +334,30 @@ export default function CourseRenderPane(props: { id?: number }) {
...
+                const response =
+                    selectedGEs.length > 1
+                        ? intersectWebsocResponses(
+                              await trpc.websoc.getManyOfField.query({
+                                  params: { ...websocQueryParams, ge: selectedGEs.join(',') },
+                                  fieldName: 'ge',
+                              })
+                          )
+                        : await trpc.websoc.getOne.query(websocQueryParams);
                 setSearchedTerm(RightPaneStore.getFormData().term.longName);
```
Choollol, line 361: Could we deduplicate the shared logic between both branches by switching from early-return to using an else-branch? I know it's really minor right now since it's just one line, but it could grow in the future.

packages/anteater-api/src/utils/index.ts
```diff
@@ -94,29 +88,16 @@ export function unionWebsocResponses(responses: WebsocAPIResponse[]): WebsocAPIR
...
+export function intersectWebsocResponses(responses: WebsocAPIResponse[]): WebsocAPIResponse {
+    const first = responses[0];
+    if (first === undefined) {
+        return { schools: [] };
```
Choollol, line 95: Why the revert from `responses.length` check back to `first` check?

packages/anteater-api/src/utils/index.ts
```diff
@@ -94,29 +88,16 @@ export function unionWebsocResponses(responses: WebsocAPIResponse[]): WebsocAPIR
...
- * for one concatenated list (legacy), combine `schools` yourself.
- */
-export function splitWebsocIntersectAndRest(responses: WebsocAPIResponse[]): {
-    intersect: WebsocAPIResponse;
-    rest: WebsocAPIResponse;
-} {
-    if (responses.length === 0) {
-        return { intersect: { schools: [] }, rest: { schools: [] } };
+/** Courses that appear in every response (layout from the first). Sections are not merged. */
+export function intersectWebsocResponses(responses: WebsocAPIResponse[]): WebsocAPIResponse {
```
Choollol, line 92: nice, much cleaner

Choollol (changes requested):
Thanks for remembering, overall nice changes

Maybe something for design, but could we render the 'and' differently from the GEs? Maybe darker or semi-transparent or something. Right now the GEs and the 'and's kind of blend together:

Image

As you can also see from the screenshot, the field's UI breaks when the GEs overflow. This is behavior from before though, so not something that needs to be fixed in this PR.

Also super minor and maybe something for design, but since the GE selector is only intersection now, the "All" option reads a little differently from its previous meaning and could seem like it's trying to find courses that satisfy all GEs, which doesn't make sense but might be confusing to some.

apps/antalmanac/src/components/RightPane/CoursePane/SearchForm/constants.ts
KevinWu098: I'm gonna nuke these soon via NUQS, so id rather just let them be

packages/anteater-api/src/utils/index.ts
KevinWu098: Eh needed `first` down below so its like one line of code less

packages/anteater-api/src/utils/index.ts
KevinWu098: mm, the type still says null on it. harmless i think

KevinWu098:
> Thanks for remembering, overall nice changes
> 
> Maybe something for design, but could we render the 'and' differently from the GEs? Maybe darker or semi-transparent or something. Right now the GEs and the 'and's kind of blend together:
> 
> Image
> As you can also see from the screenshot, the field's UI breaks when the GEs overflow. This is behavior from before though, so not something that needs to be fixed in this PR.
> 
> Also super minor and maybe something for design, but since the GE selector is only intersection now, the "All" option reads a little differently from its previous meaning and could seem like it's trying to find courses that satisfy all GEs, which doesn't make sense but might be confusing to some.

We can explore the UI things in another PR — I don't think we can render the delimiter any different, but we could choose to just use commas again!

Choollol:
> We can explore the UI things in another PR — I don't think we can render the delimiter any different, but we could choose to just use commas again!

`renderValue` can return any `ReactNode`, so we could do something like this:
```ts
return values
  .map<ReactNode>((value) => getShortLabel(value))
  .reduce((prev, cur) => [
      prev,
      <Typography sx={{ opacity: 0.7, display: 'inline' }}> and </Typography>,
      cur,
  ]);
```
though we'll probably want to double check with design.

But again, true, something for another PR

packages/anteater-api/src/utils/index.ts
Choollol: darn, i guess typescript isn't smart enough for that yet

Choollol (approved):
lgtm!

packages/anteater-api/src/utils/index.ts
Choollol: optionally could add a comment saying that it isn't actually possible for it to be null and it's just to satisfy typescript

PR #1854, opened 2026-05-24
fix: return URL validation to fix encoding bug

### LINKED ISSUE #1851: Auth return URL isn't encoded
To reproduce:
1. From signed out, set manual search department to I&C SCI
2. Sign in
3. After being redirected to antalmanac.com, see that null page is returned because the "&" isn't encoded

Choollol:
## Summary

Refactors `getSafeAuthRedirectPath` to not decode the given URL. Handles open redirect attacks by checking against a pre-supplied allowed origin.

I think checking the URL's origin is enough to protect against attacks, but if there are other cases that should be considered, lmk

## Test Plan

1. Run `pnpm test` and ensure tests pass
2. Auth return URL still works. Testable by setting manual search field that sets a URL query param -> Sign in -> See that you are redirected to the same URL with the same query params

## Issues

Closes #1851

### DIFF
--- a/apps/antalmanac/src/lib/auth/auth.ts
+++ b/apps/antalmanac/src/lib/auth/auth.ts
@@ -50,8 +50,8 @@ export const auth = betterAuth({
                     if (additionalData.returnUrl) {
                         const returnUrl = getSafeAuthRedirectPath(
                             additionalData.returnUrl,
-                            ctx.request?.url ?? '',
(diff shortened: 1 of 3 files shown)


KevinWu098 (approved):
Uh yeah the candidates code looks weird anyways, such i trust this but i didn't really think about if this regresses anything

PR #1887, opened 2026-06-01
Revert "fix: gate AddedCoursesList skeleton when schedule is empty"

dsnsgithub:
The added course skeleton broke after #1881 (at least on prod with my schedule), as it displays the "no courses added" before skipping to my loaded courses.

Reverts icssc/AntAlmanac#1881


KevinWu098 (approved):
Hm cursor had some reason for doing it but I can't remember why... your problem is clear/obvious tho so

dsnsgithub:
Cubic took feedback about "no skeleton on empty first-load" when I was creating the loading skeleton, then hallucinated a problem when you made some changes to it. Cursor probably read the Cubic comment.

no skeleton on empty first-load != move empty state above loading skeleton

Can't trust Cubic for everything :pensive:.

PR #1888, opened 2026-06-01
refactor: replace GeDataFetchProvider with includeRelatedCourses on GE searches

KevinWu098:
## Summary

Removes `GeDataFetchProvider`, which refetched each course with `ge: ANY` while keeping advanced filters. When the user selects at least one GE category, the initial WebSOC query now passes `includeRelatedCourses=true` so Anteater API returns all sections for courses that match the search.

## Behavior

| Search | Before | After |
|--------|--------|-------|
| GE only | Extra AAPI call per course row; same sections | One query; same sections |
| GE + building/days/etc. | Section table often still filter-truncated | Full course sections when any section matches |
| Building only (no GE) | Unchanged | Unchanged — flag not sent |

## Implementation

- `getQueryParams` sets `includeRelatedCourses: 'true'` when `getSelectedGEs(ge).length > 0`
- `CourseListItem` always renders `SectionTable` from the list search payload
- Deleted `GEDataFetchProvider.tsx`

<div><a href="https://cursor.com/agents/bc-d56e4c4e-002a-4b77-889b-07f4788529a2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-d56e4c4e-002a-4b77-889b-07f4788529a2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


KevinWu098:
Although initially promising, `includeRelatedCourses` is too broad for AA's specific use case. Specifically, we "enrich" WebSOC requests with a specified GE, which only return the lectures (bit of a simplification), with the entire course (this is the lecture AND the discussions etc). 

`includeRelatedCourses` enables queries to return the entire course so long as at least one section matches the given parameters. This is, however, too broad.

Take for instance `SOCIOL 1`, which is GE 3 and has a lecture held in BS3 1200. Were we to use `includeRelatedCourses` with GE 3 AND a filter on location for BS3 1200, we would receive back every SOCIOL 1 course, including lectures which are NOT held in BS3 1200. 

Now, the semantics of what filtering should mean are a little hazy (e.g. if you filter by building, you may very likely receive course results which only contain a discussion section, which is not useful on its own). But rather than introduce new behavior, I'd rather us keep our existing behavior which is more specific, if more code heavy.

PR #1926, opened 2026-06-09
fix: canTermEnrollmentChange

### LINKED ISSUE #1818: canTermEnrollmentChange closes enrollment at Friday 00:00 instead of end of day
## Summary

`canTermEnrollmentChange` in `apps/antalmanac/src/lib/termHelpers.ts` (line ~74) closes the enrollment window at the start of Friday (00:00:00 local) instead of at the end.

## Repro

`term.instructionStart` is built by `parseLocalDate` to local midnight. `addWeeks` and date-fns `setDay` both preserve time-of-day from the input, so the right-hand side of the comparison resolves to Friday 00:00:00. Anyone checking after Friday 12:00:01 AM sees enrollment as closed, when UCI's drop policy treats Friday as a valid drop day.

## Root cause

```ts
return new Date() <= setDay(addWeeks(term.instructionStart, weeksUntilDropDeadline), 5);
```

Introduced in #1784. Caught by cubic on #1816 and tracked separately so the refactor PR stays focused.

## Suggested fix

```ts
const dropDeadline = setDay(addWeeks(term.instructionStart, weeksUntilDropDeadline), 5);
dropDeadline.setHours(23, 59, 59, 999);
return new Date() <= dropDeadline;
```

But — **before committing, confirm against UCI's actual deadline:**
- https://www.reg.uci.edu/enrollment/adc/adcpolicy.html (full quarters)
- https://summer.uci.edu/faq (summer)

The policy might say 5 PM Friday (registrar business close), not midnight. If so, use `setHours(17, 0, 0, 0)` instead.

## Acceptance criteria

- [ ] Verify the correct deadline time from UCI's policy
- [ ] Apply the matching `setHours` / `endOfDay` call
- [ ] Add a unit test pinning down the boundary with `vi.useFakeTimers()` (Thursday 11:59 PM → true; configured time on Friday → true; later Friday → false; Saturday → false)
- [ ] PR title prefix: `fix:`

jenniyan:
## Summary
- Changed `canTermEnrollmentChange` to close enrollment at Friday 5:00PM
- Sources:
  - https://uu.uci.edu/current-students/policies-and-procedures/enrollment-registration/
  - https://www.humanities.uci.edu/undergrad/add-drop-policy
- Wrote unit test for canTermEnrollmentChange in new file `termHelpers.test.ts`

## Test Plan
- `pnpm test` should succeed

## Issues
Closes #1818


apps/antalmanac/src/lib/termHelpers.ts
```diff
@@ -83,5 +83,7 @@ export function canTermEnrollmentChange(term: AATerm) {
         weeksUntilDropDeadline++;
     }
 
-    return new Date() <= setDay(addWeeks(term.instructionStart, weeksUntilDropDeadline), 5);
+    const dropDeadline = setDay(addWeeks(term.instructionStart, weeksUntilDropDeadline), 5);
+    dropDeadline.setHours(17, 0, 0, 0);
```
KevinWu098: We generally do expect the user to be in PST, but just curious if we declare that the application is always in PST? 

can you look into this please.

KevinWu098 (changes requested):
Looks solid. Two questions on correctness that would be good for you to confirm

1. see below
2. is drop deadline also 5 PM on friday for all summer sessions?

jenniyan:
> Looks solid. Two questions on correctness that would be good for you to confirm
> 
> 1. see below
> 2. is drop deadline also 5 PM on friday for all summer sessions?

After looking at the summer sessions FAQ, there is no specific indicator that the drop deadline is 5PM, so for now I adjusted it so that for all summer closing enrollment times are EOD. I also adjusted it so that for the shorter terms (summer sessions 1 and 2), the drop deadline is Week 1 Friday.

Additionally, the app isn’t globally set to PST. Some features do explicitly use America/Los_Angeles(for example the mailer that sends notifications for changes in course enrollment), but the main date logic still follows the runtime’s local timezone. So for any user in EST, a regular quarter's cutoff would look like Friday 5 PM EST.

apps/antalmanac/src/lib/termHelpers.ts
```diff
@@ -63,6 +63,24 @@ export function parseQuarter(rawQuarter: unknown) {
...
+    const isTermShort = differenceInWeeks(term.finalsStart, term.instructionStart) < 9;
+    const hasWeekZero = term.instructionStart.getDay() !== 1;
+    let weeksUntilDropDeadline = isTermShort ? 0 : 1;
+    if (hasWeekZero) {
+        weeksUntilDropDeadline++;
+    }
+
+    const dropDeadline = setDay(addWeeks(term.instructionStart, weeksUntilDropDeadline), 5);
+    if (term.isSummerTerm) {
+        return endOfDay(dropDeadline);
```
KevinWu098: A (brief) comment would be good here

KevinWu098 (commented):
Thanks for looking into it. For summer, EOD seems like a reasonable choice. For timezones, however, what I meant was that it would be... incorrect for someone on, say, the east coast to no longer be able to subscribe to notifications at 5 PM local time, which is only 2 PM PST.

Hence, it probably makes sense to figure out how to pin it to 5 PM PST! (and, as a good completeness thing, scan if there are any related areas which would benefit from also being pinned to PST without regressing behavior)

jenniyan:
Sorry for the delay on this... I pushed a fix that pins the drop deadline to 5 PM Pacific explicitly (via Intl.DateTimeFormat against America/Los_Angeles), so it now correctly resolves to 5 PM PT regardless of the viewer's local timezone or DST.

I also checked the rest of the codebase for anywhere else that might have the same bug. The only place that checks this deadline is the enrollment/notification code (canTermEnrollmentChange). Other places that use hours, like rendering the calendar or exporting to ICS, are showing actual class meeting times, which should stay in the user's local time, so those didn't need changing.

KevinWu098:
> Sorry for the delay on this... I pushed a fix that pins the drop deadline to 5 PM Pacific explicitly (via Intl.DateTimeFormat against America/Los_Angeles), so it now correctly resolves to 5 PM PT regardless of the viewer's local timezone or DST.
> 
> 
> 
> I also checked the rest of the codebase for anywhere else that might have the same bug. The only place that checks this deadline is the enrollment/notification code (canTermEnrollmentChange). Other places that use hours, like rendering the calendar or exporting to ICS, are showing actual class meeting times, which should stay in the user's local time, so those didn't need changing.

I'm squinting at this on mobile, so take what I'm saying with a grain of salt but uhhh the diff seems kinda big?

Is there really no concise way to express "5 PM PST" in our date helped library?

The diff in the test file also seems rather overly involved for the same reasons as above (I suspect you need far fewer LOC).

To be clear, if this is idiomatic, then I'm happy to be corrected here

jenniyan:
> > Sorry for the delay on this... I pushed a fix that pins the drop deadline to 5 PM Pacific explicitly (via Intl.DateTimeFormat against America/Los_Angeles), so it now correctly resolves to 5 PM PT regardless of the viewer's local timezone or DST.
> > I also checked the rest of the codebase for anywhere else that might have the same bug. The only place that checks this deadline is the enrollment/notification code (canTermEnrollmentChange). Other places that use hours, like rendering the calendar or exporting to ICS, are showing actual class meeting times, which should stay in the user's local time, so those didn't need changing.
> 
> I'm squinting at this on mobile, so take what I'm saying with a grain of salt but uhhh the diff seems kinda big?
> 
> Is there really no concise way to express "5 PM PST" in our date helped library?
> 
> The diff in the test file also seems rather overly involved for the same reasons as above (I suspect you need far fewer LOC).
> 
> To be clear, if this is idiomatic, then I'm happy to be corrected here

I'll take another look and see if I can trim it down. There's a decent chance I reached for Intl.DateTimeFormat boilerplate where the library already has a cleaner solution for having a fixed time.. I'll also simplify the test diff at the same time😅

jenniyan:
The plain package date-fns doesn't actually handle different timezones. I added date-fns-tz, which is the official add-on for date-fns that handles this, and its fromZonedTime function.

apps/antalmanac/src/lib/termHelpers.ts
```diff
@@ -63,6 +66,22 @@ export function parseQuarter(rawQuarter: unknown) {
...
+    const isTermShort = differenceInWeeks(term.finalsStart, term.instructionStart) < 9;
+    const hasWeekZero = term.instructionStart.getDay() !== 1;
+    let weeksUntilDropDeadline = isTermShort ? 0 : 1;
+    if (hasWeekZero) {
+        weeksUntilDropDeadline++;
+    }
+
+    const dropDeadline = setDay(addWeeks(term.instructionStart, weeksUntilDropDeadline), 5);
+    const closingTime = term.isSummerTerm
+        ? { hours: 23, minutes: 59, seconds: 59, milliseconds: 999 }
```
KevinWu098, line 79: i wonder if its 5 PM for summer too...

KevinWu098 (approved):
lgtm, nice work

apps/antalmanac/src/lib/termHelpers.ts
KevinWu098: (ik we already had this discussion, but i suspect 5 pm could be more likely because its more consistent... not that university systems are renowned for consistency...)

jenniyan: I'll change it to 5pm for summer then as well LOL

KevinWu098: 😸

jenniyan: actually I will just call the office when they are open tomorrow so this as accurate as it can be..

jenniyan: i called and the representative said its midnight for summer sessions so the current latest commit should be right 😸

KevinWu098: Nice. Good initiative! Will merge as is then

PR #1927, opened 2026-06-09
feat: reset split pane to default sizes on gutter double-click

KevinWu098:
## Summary

- Double-clicking the gutter between the calendar and schedule management panes now resets the split back to the default 42.5/57.5 ratio
- Extracts the hardcoded sizes into a `DEFAULT_SPLIT_SIZES` constant
- Wraps `<Split>` in a `display: contents` div with an `onDoubleClick` handler that guards on the `.gutter` class and calls `setSizes()` on the underlying Split.js instance via ref

#### Test plan
- [ ] Drag the gutter to resize the panes
- [ ] Double-click the gutter — panes should snap back to default sizes

Generated with [Devin](https://cli.devin.ai/docs)


KevinWu098:
@cubic

PR #1963, opened 2026-08-29
feat: add weighted review prompt selection by review count

### LINKED ISSUE #1706: Weight reviews by total reviews
If a professor, courseId pair has been reviewed a lot, we should negatively weight its likelihood of being selected for a review. Similarly, lesser reviewed courses should get more requests.

PranavKocharlakota:
## Summary
Review prompt selection previously did a uniform shuffle over eligible
candidates, so heavily-reviewed professor/course pairs surfaced for
review requests just as often as pairs with zero reviews.

This adds a `getReviewCounts` query (grouped by courseId + professorId,
filtered on the indexed courseId column) and replaces the uniform
shuffle in `initPrompt` with a weighted full-ordering permutation, using
weight = 1 / (1 + log(reviewCount + 1)). The sort key is `log(random) / weight`
rather than `random^(1/weight)`. Counts are global across all
users, not per-user.

This softens the correction significantly at low-to-mid review counts
(e.g. a 50-review pair now gets picked ~17% of the time against an
unreviewed pair, vs ~2% before) in exchange for eliminating starvation
at the high end.

Weighting the full ordering (not just the first pick) means "Review
another" stays weighted too, since `eligibleCandidates` keeps its
existing contract and `advanceToNext` needed no changes.

Counts are batched: the router caps input at
`REVIEW_COUNT_BATCH_SIZE` (200), so the store dedupes course IDs, chunks
them, and issues one query per batch in parallel (new shared module
`lib/reviewPrompt.ts` holds `chunk` and the batch size). Without this, a
user with more than 200 distinct courses would trip Zod validation and
lose weighting entirely. The batches run under one `Promise.all`, so a
single failed batch currently discards all counts, not just that
batch's.

The count fetch happens after the eligible-candidate filter, rather
than joining the existing `Promise.all` of eligibility queries, since
the cooldown early-return fires on most app loads. The fetch is wrapped
in its own try/catch that leaves the count map empty on failure, which
naturally degrades to uniform weighting rather than suppressing the
prompt entirely.

`reviewCount` and `reviewCountsLoaded` are logged on `PROMPT_SHOWN`
the flag matters because an empty map makes every candidate report
`reviewCount: 0`, indistinguishable from a genuinely-zero-review pair.
## Test Plan
- Manually verified against a seeded local DB: seeded a schedule with
  three real 2026 Spring COMPSCI Lec sections and gave the pairs distinct
  review counts. Confirmed end-to-end that `getReviewCounts` fires, that
  its `courseId::professorId` keys match the candidate keys, that the
  zero-count pair (which the `GROUP BY` returns no row for) resolves to 0
  via the `?? 0` fallback, and that the resulting weights are exactly
  `1 / (1 + ln(n+1))`. The full ordering is produced, not just the
  first pick, so `advanceToNext` inherits the weighting.
- Confirmed the fetch-placement decision: on a cooldown-suppressed load,
  `initPrompt` returns before the count query, so `getReviewCounts` never
  fires. It only appears on loads that actually surface a prompt.
- Incidentally exercised the dismissal path: dismissing wrote a
  `reviewDismissals` row, the 7-day cooldown blocked subsequent prompts,
  and the combo was excluded from `eligible` thereafter.
## Issues

Closes #1706 

<a href="https://cubic.dev/pr/icssc/AntAlmanac/pull/1963?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/antalmanac/src/backend/lib/rds/reviews.ts
+++ b/apps/antalmanac/src/backend/lib/rds/reviews.ts
@@ -1,6 +1,6 @@
 import type { DatabaseOrTransaction } from '$backend/lib/rds/types';
 import { type NewInstructorReview, instructorReviews, reviewDismissals } from '@packages/db/src/schema';
-import { eq, max } from 'drizzle-orm';
+import { count, eq, inArray, max } from 'drizzle-orm';
 
 export async function insertInstructorReview(
     db: DatabaseOrTransaction,
@@ -38,6 +38,22 @@ export async function getReviewedCombos(db: DatabaseOrTransaction, userId: strin
         })
         .from(instructorReviews)
         .where(eq(instructorReviews.userId, userId));
+}
+
+export async function getReviewCounts(db: DatabaseOrTransaction, courseIds: string[]) {
+    if (courseIds.length === 0) {
+        return [];
+    }
+
+    return db
+        .select({
+            professorId: instructorReviews.professorId,
+            courseId: instructorReviews.courseId,
+            reviewCount: count(),
+        })
+        .from(instructorReviews)
+        .where(inArray(instructorReviews.courseId, courseIds))
+        .groupBy(instructorReviews.courseId, instructorReviews.professorId);
 }
 
 export async function insertReviewDismissal(
--- a/apps/antalmanac/src/backend/routers/review.ts
+++ b/apps/antalmanac/src/backend/routers/review.ts
@@ -7,6 +7,7 @@ import {
     insertReviewDismissal,
 } from '$backend/lib/rds/reviews';
 import { protectedProcedure, router } from '$backend/trpc';
+import { REVIEW_COUNT_BATCH_SIZE } from '$lib/reviewPrompt';
 import { db } from '@packages/db';
 import { TRPCError } from '@trpc/server';
 import { z } from 'zod';
@@ -86,7 +87,7 @@ const reviewRouter = router({
      * Review counts per (professorId, courseId) across all users
      */
     getReviewCounts: protectedProcedure
-        .input(z.object({ courseIds: z.array(z.string()).min(1).max(200) }))
+        .input(z.object({ courseIds: z.array(z.string()).min(1).max(REVIEW_COUNT_BATCH_SIZE) }))
         .query(async ({ input }) => {
             return getReviewCounts(db, input.courseIds);
         }),
--- a/apps/antalmanac/src/backend/routers/review.ts
+++ b/apps/antalmanac/src/backend/routers/review.ts
@@ -1,5 +1,6 @@
 import {
     getDismissedCombos,
+    getReviewCounts,
     getReviewPromptLastInteractionAt,
     getReviewedCombos,
     insertInstructorReview,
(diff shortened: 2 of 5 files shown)


sicn4rf (changes requested):
hey pranav, thanks for taking this on, and i apologize that this took a while to get back to you on.

the implementation here is solid. the weighted-permutation algorithm makes sense, correct approach, review-another staying weighted is a good call, tests are thorough. 

i'm not rejecting the work here but i had to go through it with @KevinWu098 who originally wrote the issue, we want to scope this down before merging. two problems with the current approach that we don't think get fixed by a quick follow up commit:

1. **starvation**: `1 / (reviewCount + 1)` is unbounded, so a heavily-reviewed required course can end up 100-1000x underweighted relative to a unreviewed one. it doesn't just get deprioritized, it can effectively stop getting asked for a very long time, with no signal for recency. ideally we should care about how recent the reviews were not just the count alone.
2. **enrollment size**: a 10-person TA section with 0 reviews gets the same max weight as a 400-person lecture with 0 reviews. nothing here accounts for how many people could have even reviewed it. this is a matter of enrollment size of the course which is also not accounted for currently

both are real and hard problems, but they're their own scoping work (recency needs a "last meaningfully reviewed" signal that we don't track yet, enrollment-size needs enrollment data wired in). this is genuinely outside what the original issue asked for, that one's on us for being vague, not on you. also my bad for not consulting kevin before assigning the issue to you.

as the next best step, what we'd like you to look at first is to replace `1 / (reviewCount + 1)` with something bounded instead, so the gap between a 0-review course and a heavily-reviewed one doesn't grow without limit. we did some rough modeling on our end as a sanity check, but a log-compressed version like `1 / (1 + ln(reviewCount + 1))` drops that gap from about 1000x down to roughly 8x at 1000 reviews, and it barely moves even out past 5000. so the target we're going for is something in that range, noticeably deprioritized but never crushed toward zero, not a specific formula we're locking you into. if you land on something that gets a similar spread through a different shape, that's fine.

rest of the PR (the weighted-permutation ordering, review-another staying weighted, the batching) can stay as-is, this is really just about the shape of the weight curve. you'll probably want to update the expected values in `review-prompt-weighting.test.ts` once you land on a formula.

if you try this and the numbers don't cooperate, plan B is dropping the continuous formula for an explicit bounded rule instead, small fixed boost for courses under 3 reviews, small fixed penalty for courses with over 20, neutral for anything in between. figured i'd mention it now so you're not blocked if the first approach turns out to not work cleanly.

let me know if you want to sync on this before diving in, otherwise take a pass and we'll review from there.

KevinWu098:
I'm old news now, but please FFT ping me on discord if you want to run anything by me too! :)

PranavKocharlakota:
I switched the weighting to the log-compressed formula mentioned above. That should fix the starvation issue. I modified the PR description and tests accordingly. Let me know if you need anything else!

sicn4rf (approved):
lgtm! log-compressed formula matches what we modeled with kevin, math checks out exactly, tests updated correctly, lint is clean

noting again that recency and enrollment-size normalization stay out of scope here, tracking those separately, this PR still closes #1706 as originally asked.

thanks for sticking with and implementing the rescoping, we appreciate it

i'm going to merge this for you as well

PR #1965, opened 2026-09-09
feat: Update No courses screen text

### LINKED ISSUE #1264: Better "No Courses" Screen
## Description
- The graphic with "No Classes" and the notice to go to PP don't really go well together.
- There should be a cohesive message. Something along the lines of, "No sections found. See class information on PeterPortal".

rocdx:
## Summary
Changes the no courses screen text to match the image

## Issues

Closes #1264

<a href="https://cubic.dev/pr/icssc/AntAlmanac/pull/1965?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane/PlannerCourseLinkBanner.tsx
+++ b/apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane/PlannerCourseLinkBanner.tsx
@@ -42,7 +42,8 @@ export function PlannerCourseLinkBanner({ deptValue, courseNumber }: PlannerCour
                 }}
             >
                 <span>
-                    Search for <span style={{ textDecoration: 'underline' }}>{courseLabel}</span> on AntAlmanac Planner!
+                    No sections for <span style={{ textDecoration: 'underline' }}>{courseLabel}</span> in this session.
+                    Please see class information on AntAlmanac Planner!
                 </span>
             </Alert>
         </Link>


apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane/PlannerCourseLinkBanner.tsx
```diff
@@ -42,7 +42,8 @@ export function PlannerCourseLinkBanner({ deptValue, courseNumber }: PlannerCour
                 }}
             >
                 <span>
-                    Search for <span style={{ textDecoration: 'underline' }}>{courseLabel}</span> on AntAlmanac Planner!
+                    No sections for <span style={{ textDecoration: 'underline' }}>{courseLabel}</span> in this session.
```
sicn4rf, line 45: suggestion: we don't use `session` as a label anywhere in the app we use term/quarter, and it's vaguer than it needs to be here since we already know which term they searched.

can we name it directly maybe?

something like
> No sections for ICS 33 in Fall 2026.

the term the user picked should be reachable from this component, worth poking around the search param code. also worth checking how it reads for summer terms.

apps/antalmanac/src/components/RightPane/CoursePane/CourseRenderPane/PlannerCourseLinkBanner.tsx
```diff
@@ -42,7 +42,8 @@ export function PlannerCourseLinkBanner({ deptValue, courseNumber }: PlannerCour
...
+                    Please see class information on AntAlmanac Planner!
```
sicn4rf, line 46: suggestion: since we're rewriting the copy anyway, good chance to say what they'd actually find over there. "class information" is vague and it's also the thing they just failed to find, so it reads a bit like a dead end

the planner course page has a **Recent Offerings** grid (academic year x quarter) plus grade distributions and a prereq tree, which is exactly what someone who searched a term with no sections is after. something like:

`No sections for ICS 33 in Fall 2026. See recent offerings and course info on AntAlmanac Planner`

sicn4rf (changes requested):
copy direction is right, thanks for taking this one. two small suggestions inline, both verbiage related.

one thing that's out of scope here but i want to highlight, Issue #1264's actual complaint was that the mascot graphic and the planner notice don't read as one cohesive message. this PR fixes the tooltip half, but the graphic underneath still says "sorry! i couldn't find any classes for that search", so the screen ends up saying the same thing twice

we can merge this once the inline comments are addressed, and i'll open a separate issue scoped just to the graphic so anyone who wants to contribute the redesign has something to pick up.

rocdx:
replaced "session" with the actual term the user originally searched for and reworded the latter half of the sentence to specify recent offerings/class info in the planner.

PR #1966, opened 2026-09-12
fix: re-enable mobile schedule selector in Added Courses pane

ronaldw07:
## Summary
Re-enables the mobile schedule selector in the Added Courses pane, originally
shipped in #1428 and disabled directly on main without review in 66cd80f
over layout crowding concerns.

Addresses the crowding by giving the selector its own row instead of
inlining it next to the title, and removes the redundant schedule-name
display (the selector button already shows the name).

Follow-up per review: moved the copy/clear/columns/notifications buttons
off the floating overlay and onto the right side of the title row, for
both mobile and desktop, matching the layout Caleb suggested.

Screenshots for before and after:

OLD (CURRENT DESIGN):
Screenshot 2026-09-14 at 10 37 32 PM

NEW DESIGN:
Buttons are moved to the right, schedule selector stays on the left / the same
Screenshot 2026-09-14 at 10 37 22 PM

## Test Plan
1. Open AntAlmanac at a mobile viewport (<600px) on the Added tab
2. Confirm the schedule selector renders on its own row below the icon toolbar
3. Confirm no duplicate schedule name text
4. Add a second schedule and switch between them via the selector
5. Confirm the action buttons render right-aligned next to the title on both
   mobile and desktop, not floating on top of it

## Issues
Re-addresses #1177 (originally closed by #1428)

<a href="https://cubic.dev/pr/icssc/AntAlmanac/pull/1966?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/antalmanac/src/components/RightPane/AddedCourses/AddedCourses.tsx
+++ b/apps/antalmanac/src/components/RightPane/AddedCourses/AddedCourses.tsx
@@ -1,10 +1,12 @@
 import { ClearScheduleButton } from '$components/buttons/Clear';
 import { CopyScheduleButton } from '$components/buttons/Copy';
+import { SelectSchedulePopover } from '$components/Calendar/Toolbar/ScheduleSelect/ScheduleSelect';
 import { AddedCoursesList } from '$components/RightPane/AddedCourses/AddedCoursesList';
 import { CustomEventsBox } from '$components/RightPane/AddedCourses/CustomEventsBox';
 import { NotificationsDialog } from '$components/RightPane/AddedCourses/Notifications/NotificationsDialog';
 import { ScheduleNoteBox } from '$components/RightPane/AddedCourses/ScheduleNoteBox';
 import { ColumnToggleDropdown } from '$components/RightPane/CoursePane/CoursePaneButtonRow';
+import { useIsMobile } from '$hooks/useIsMobile';
 import analyticsEnum from '$lib/analytics/analytics';
 import {
     removeLocalStorageAddedCoursesSkeletonBlueprint,
@@ -66,6 +68,7 @@ export function AddedCourses() {
     const [courses, setCourses] = useState(getCurrentCourses);
     const [scheduleNames, setScheduleNames] = useState(() => AppStore.getScheduleNames());
     const [scheduleIndex, setScheduleIndex] = useState(() => AppStore.getCurrentScheduleIndex());
+    const isMobile = useIsMobile();
 
     const handleCourseOrderChange = useCallback(
         (updatedCourses: AACourseWithTerm[], _activeIndex: number, overIndex: number) => {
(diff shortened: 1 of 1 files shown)


calebwongg (commented):
thoughts:

good PR we will probably have to review these with the uiux team before we can implement this / add this back. i think kevin removed this because it looks kinda weird and i agree it does look weird, so we might have to experiment with different layouts and talk about whether or not the schedule popover is useful at all in the mobile context.

maybe we can see how it looks and if we have space to move it to the right? -> 

image

this might stay latent until school starts back up so lowkey you can just wait and see what the big francis says

ronaldw07:
moved the icon buttons off the floating overlay and onto the right side of the title row, like you suggested - 
should be same treatment for both mobile and desktop now. lmk how it looks and if that works better or not!
i think this fits the design better and doesn't have any awkward spacing.

sicn4rf:
hey ronald,

thanks for catching this and picking this back up! thanks as well for turning around caleb's suggestion really quickly.

a few notes i have:

i'm agreeing with caleb here, this one needs discussion with our UI/UX designers before we merge this. it's not really a code question we can easily answer. whether the schedule name on mobile should be a heading or the `SelectSchedulePopover` control (which also brings the rename, copy, delete, reorder functionalities into the added tab) is a much larger question. that's not something i want to approve on my own, so we'll bring it to the design team once fall quarter starts and everyone's back.

in the meantime, a few things that would make that review a lot easier:
- adding screenshots or a short recording to your PR desc for frontend work is preferred! the staging deploy that caleb got running is still from your first commit, so the inline button row isn't up anywhere yet. i can deploy it so we can see what that looks like
- take a quick look at cubic's overflow comment, since that's likely the first thing design will check

no need to rework the approach until we've talked it through. i'll post here once there's a direction

ronaldw07:
i added some screenshots to the original pr description, and pushed a fix for the overflow comment — buttons stay pinned right and can't get pushed off/overlap now regardless of schedule name length. also widened the selector for readability and kept the units text at its original size. should be good to go, lmk if anything still looks off!

> hey ronald,
> 
> thanks for catching this and picking this back up! thanks as well for turning around caleb's suggestion really quickly.
> 
> a few notes i have:
> 
> i'm agreeing with caleb here, this one needs discussion with our UI/UX designers before we merge this. it's not really a code question we can easily answer. whether the schedule name on mobile should be a heading or the `SelectSchedulePopover` control (which also brings the rename, copy, delete, reorder functionalities into the added tab) is a much larger question. that's not something i want to approve on my own, so we'll bring it to the design team once fall quarter starts and everyone's back.
> 
> in the meantime, a few things that would make that review a lot easier:
> 
> * adding screenshots or a short recording to your PR desc for frontend work is preferred! the staging deploy that caleb got running is still from your first commit, so the inline button row isn't up anywhere yet. i can deploy it so we can see what that looks like
> * take a quick look at cubic's overflow comment, since that's likely the first thing design will check
> 
> no need to rework the approach until we've talked it through. i'll post here once there's a direction

